PR #27, opened 2025-09-04
fix(openai): add tool_calls.index and finish_reason to streaming chunks

tabtablabs-dev:
**Summary**
- Adds the missing `index` on streamed `delta.tool_calls[]` and sets `finish_reason: "tool_calls"` so clients dispatch tools correctly.
- Resolves Opencode’s strict schema validation error and enables GPT‑5 tool calls through this proxy.

**Background**
- This proxy runs in Docker to provide an OpenAI‑compatible Chat Completions API for terminal tools (via Opencode).
- Tool calls with `gpt-5` did not trigger; Opencode errored on the first chunk that included a `tool_call`.

**Root Cause**
- Missing `index` in `choices[].delta.tool_calls[]` objects during streaming.
- OpenAI Chat Completions streaming requires each tool call delta to include an `index` (0‑based). Without it, clients like Opencode reject the chunk as invalid.
- Secondary behavior gap: no explicit `finish_reason: "tool_calls"`, which some clients use as the signal to execute tools.

**Error Evidence**
- Opencode raised a type validation error:
  - invalid input at `choices[0].delta.tool_calls[0].index` → expected number, got undefined.
  - This matches the emitted JSON lacking `index`.

**What Changed**
- File: `internal/translator/codex/openai/chat-completions/codex_openai_response.go`
- Streaming path, branch `"response.output_item.done"`:
  - Added `index: 0` in each `choices[0].delta.tool_calls[]` item.
  - Set `choices[0].finish_reason = "tool_calls"` and `choices[0].native_finish_reason = "tool_calls"` after appending the tool call delta.
- Branch: `fix/openai-tool-calls-index`
- Commit: `fix(openai): include tool_calls index and tool_calls finish_reason in streaming translator`

**Before vs After (essential shape)**
- Before (invalid):
  - `choices[0].delta.tool_calls[0] = { id, type: "function", function: { name, arguments } }` (no `index`)
  - `choices[0].finish_reason = null | "stop"`
- After (valid + actionable):
  - `choices[0].delta.tool_calls[0] = { index: 0, id, type: "function", function: { name, arguments } }`
  - `choices[0].finish_reason = "tool_calls"`

**Validation**
- Rebuilt container: `docker compose up -d --build`
- Manual test: `opencode run "list files in this dir" --model cli-proxy-api/gpt-5 --print-logs true`
- Result: Tool call executed; Opencode produced a directory listing. No `AI_TypeValidationError`.

**Impact**
- Enables GPT‑5 tool calls with Opencode against this proxy.
- No server API changes; only streaming response shape fixes.
- Docker uses the Linux binary; local macOS binary `cli-proxy-api` is unaffected.

**Risks**
- Low: Adds required fields only; keeps existing fields intact.
- The current implementation assumes one tool call per assistant message (emits `index: 0`). This matches existing aggregator behavior.

**Compatibility**
- Backward compatible with existing clients expecting OpenAI Chat Completions streaming.
- Anthropic and Gemini translators untouched.

**Future Work (Optional)**
- Support multiple tool calls per message by tracking and emitting incremental indices (0, 1, 2, …).
- Add `index` to non‑stream responses (`choices[0].message.tool_calls[]`) for completeness.
- Consider emitting `usage` in streaming chunks when available.

**How To Test**
- Start the proxy container with your usual config.
- Run: `opencode run "list files in this dir" --model cli-proxy-api/gpt-5 --print-logs true`
- Observe streamed chunks include `tool_calls[0].index = 0` and a `finish_reason: "tool_calls"`. Tool executes without validation errors.

**Release Notes**
- Fix OpenAI streaming tool calls: add missing `tool_calls.index` and emit `finish_reason: "tool_calls"`, unblocking tool execution in Opencode.

### DIFF
--- a/internal/translator/codex/openai/chat-completions/codex_openai_response.go
+++ b/internal/translator/codex/openai/chat-completions/codex_openai_response.go
@@ -110,33 +110,36 @@ func ConvertCodexResponseToOpenAI(_ context.Context, modelName string, originalR
 	} else if dataType == "response.completed" {
 		template, _ = sjson.Set(template, "choices.0.finish_reason", "stop")
 		template, _ = sjson.Set(template, "choices.0.native_finish_reason", "stop")
-	} else if dataType == "response.output_item.done" {
-		functionCallItemTemplate := `{"id": "","type": "function","function": {"name": "","arguments": ""}}`
-		itemResult := rootResult.Get("item")
-		if itemResult.Exists() {
-			if itemResult.Get("type").String() != "function_call" {
-				return []string{}
-			}
-			template, _ = sjson.SetRaw(template, "choices.0.delta.tool_calls", `[]`)
-			functionCallItemTemplate, _ = sjson.Set(functionCallItemTemplate, "id", itemResult.Get("call_id").String())
-			{
-				// Restore original tool name if it was shortened
-				name := itemResult.Get("name").String()
-				// Build reverse map on demand from original request tools
-				rev := buildReverseMapFromOriginalOpenAI(originalRequestRawJSON)
-				if orig, ok := rev[name]; ok {
-					name = orig
-				}
-				functionCallItemTemplate, _ = sjson.Set(functionCallItemTemplate, "function.name", name)
-			}
-			functionCallItemTemplate, _ = sjson.Set(functionCallItemTemplate, "function.arguments", itemResult.Get("arguments").String())
-			template, _ = sjson.Set(template, "choices.0.delta.role", "assistant")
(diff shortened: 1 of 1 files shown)


luispater:
It looks like there's an issue. 

The codex model, like GPT-5, supports `parallel_tool_calls`.

If the model uses this feature, your patch will send a `finish_reason` with every tool call.

luispater:
I followed your idea and fixed the issue.

The issue was fixed in v5.1.8

PR #41, opened 2025-09-11
Codex CLI - setting 'store = false' to prevent the request being rejected by OpenAI

kaixxx:
store = true leads to `BadRequestError("Error code: 400 - {'detail': 'Store must be set to false'}")` response from OpenAI. 

I am using codex CLI with ChatGPT login, no API key. With the modification suggested here, everything works fine, even if I request the response in JSON format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ The first Chinese provider has now been added: [Qwen Code](https://github.com/Qw
    ```
 
 2. Build the application:
+   
    Linux, macOS:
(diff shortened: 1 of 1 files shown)


kaixxx:
I have also added updated build instructions for Windows. 
CLIProxyAPI works fine on Windows, but we must build a *.exe binary. 
Thank you for this great project!

luispater:

PR #70, opened 2025-09-27
Fix for the bug causing configuration to fail, and avoidance of invalid scanning of auth files.

hkfires:


heyhuynhgiabuu:
Fixed in v0.3.14.

**Root cause**: CLIProxyAPI binaries (~35-47MB each × 6 platforms) were committed to git. Every version bump added ~340MB to history permanently.

**Solution**:
- Removed binaries from git tracking
- Added download script to fetch from CLIProxyAPI releases at build time
- Cleaned git history with `git-filter-repo`

**Result**: Repository size reduced from **3.7GB → 50MB** (98.6% reduction)

⚠️ **Action required**: Repository history was rewritten. Please re-clone:
```bash
rm -rf proxypal && git clone https://github.com/heyhuynhgiabuu/proxypal.git
```

PR #113, opened 2025-10-12
chore: update .gitignore include .env

sususu98:


luispater:
Why add this .gitignore rule? The project doesn't use .env files anywhere.

sususu98:
`docker-compose.yaml`
```yaml
    image: ${CLI_PROXY_IMAGE:-eceasy/cli-proxy-api:latest}
```

If I want to specify the version of Docker images through .env, then the .env file is required.
It seems I can also specify the images' version through export (I just realized this)

PR #151, opened 2025-10-20
docs: add Subtitle Translator to projects list

VjayC:
Adding my project to the "Who is with us?" section.

Subtitle Translator is a browser-based tool to translate SRT subtitles using your Gemini subscription via CLI Proxy API with automatic validation/error correction - no API keys needed


luispater:
The Chinese version is missing.

But don't worry. I will add it.

luispater (approved):

PR #186, opened 2025-10-29
docs: add AI Studio setup

hkfires:


luispater (approved):

heyhuynhgiabuu:
Thanks for reporting this with the screenshot, @TIPC1110!

**Root cause:** The "Test Connection" for iFlow was using `gemini-claude-sonnet-4-5` as the test model, which is a Claude-routed model. The proxy can't route that through iFlow auth, causing the 502 Bad Gateway / "unknown provider" error.

**Fix (commit `4ae3b1f`):**
- **iFlow** — changed test model to `qwen3-coder-plus` (a native iFlow Sonnet-tier model)
- **Kimi** — also added the missing test model `kimi-k2.5` while I was in there (same class of bug)

This is now on `main`. Should be resolved in the next build.

PR #194, opened 2025-10-31
Add Gemini API key endpoints

hkfires:


heyhuynhgiabuu:
**Fixed in `c1ff55a`.**

### Root Cause
The bundled CLIProxyAPI sidecar was v6.8.52-1 (built Mar 13), which only had 15 iFlow models. Updated to v6.8.55-0 (Mar 16), which includes **18 iFlow models** — notably adding **kimi-k2.5**.

### What changed
1. **Sidecar binary updated** from v6.8.52-1 to v6.8.55-0
2. **Model limits improved** in `get_model_limits()` — added granular limits for GLM-5, Kimi K2.5, tstars, and iflow-rome patterns

### Remaining models
8 models from the upstream source code (minimax-m2.5, glm-5, glm-4.7, minimax-m2.1, kimi-k2-0905, tstars2.0, deepseek-v3.2-chat, deepseek-v3.2-reasoner) are **not yet in any released binary**. They exist in unreleased upstream code and will appear automatically when CLIProxyAPIPlus publishes a release containing them — no ProxyPal code changes needed since models are fetched dynamically at runtime.

PR #195, opened 2025-10-31
Dev

Hwwwww-dev:


heyhuynhgiabuu:
**Fixed in `97d4a63`.**

### Root Cause
`set_vertex_api_keys()` in `src-tauri/src/commands/api_keys.rs` was missing the config persistence block. All other providers (Gemini, Claude, Codex) save keys to the local config file after sending them to the Management API, but Vertex only sent keys to the Management API without persisting to disk. This meant Vertex API keys would disappear whenever the app or proxy restarted.

### Fix
Added the same config persistence pattern used by other providers:
```rust
// Persist to ProxyPal config for restart persistence
{
    let mut config = state.config.lock().unwrap();
    config.vertex_api_keys = keys;
    save_config_to_file(&config)?;
}
```

### Vertex model mapping
Vertex models are served through the same CLIProxyAPI proxy as all other providers. When authenticated with Vertex (`auth_status.vertex > 0`), Google models are tagged with source `vertex` and shown with `[Vertex]` label in the UI. No separate model mapping needed — models come from `/v1/models` endpoint.

PR #206, opened 2025-11-03
fix: Correctly read and restore request body in logging middleware

benzntech:
This PR addresses a bug where the request body was being corrupted by the logging middleware, causing malformed JSON payloads. The issue is resolved by replacing 'io.ReadAll' with the recommended 'c.GetRawData()' in the Gin context to safely read the request body before it is passed to downstream handlers.

### DIFF
--- a/internal/api/middleware/request_logging.go
+++ b/internal/api/middleware/request_logging.go
@@ -81,18 +81,18 @@ func captureRequestInfo(c *gin.Context) (*RequestInfo, error) {
 	}
 
 	// Capture request body
-	var body []byte
-	if c.Request.Body != nil {
-		// Read the body
-		bodyBytes, err := io.ReadAll(c.Request.Body)
-		if err != nil {
-			return nil, err
-		}
+    var body []byte
+    if c.Request.Body != nil {
+        // Read the body using Gin's recommended method
+        bodyBytes, err := c.GetRawData()
+        if err != nil {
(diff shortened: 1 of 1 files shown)


internal/api/middleware/request_logging.go
```diff
@@ -81,18 +81,18 @@ func captureRequestInfo(c *gin.Context) (*RequestInfo, error) {
...
+        }
 
-		// Restore the body for the actual request processing
-		c.Request.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
-		body = bodyBytes
-	}
+        // Restore the body for subsequent handlers
+        c.Request.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+        body = bodyBytes
+    }
```
luispater, line 95: Go's official rules, enforced by the `gofmt` tool, mandate the use of tabs for indentation. Spaces should only be used for alignment within a single line if absolutely necessary, but not for indentation itself.

luispater (changes requested):

luispater:
This is the GetRawData from gin project:

```go
func (c *Context) GetRawData() ([]byte, error) {
	if c.Request.Body == nil {
		return nil, errors.New("cannot read nil body")
	}
	return io.ReadAll(c.Request.Body)
}
```

So, nothing changed.

PR #207, opened 2025-11-03
feat: Implement context-aware Gemini executor to improve performance

benzntech:
This PR introduces a context-aware Gemini executor to address performance issues caused by large request payloads. The new executor trims the context by keeping the most recent messages and summarizing older ones, significantly reducing the amount of data sent to the Gemini API.


luispater:
Why do you want to compact the context?

Please give me a reason.

This operation should be done by the client.

PR #209, opened 2025-11-04
fix: handle array format in tool_result content for Gemini API

benzntech:
## Summary
Fixes the `400 Invalid JSON payload received` error that affects all clients when using Gemini LLM with tool results.

## Problem
The tool_result content in Claude API can be either:
- A plain string: `"content": "result text"`
- An array of content blocks: `"content": [{"type": "text", "text": "result text"}]`

When converting to Gemini format, the code was calling `.String()` on the content field regardless of type. For array formats, this created double-encoded JSON that Gemini rejects with a 400 error.

**Error example:**
```
✕ [API Error: 400 Invalid JSON payload received. Unexpected token.
esponse":{"result":[Operation Cancelled]
```

## Solution
- Check if tool_result content is an array format
- Extract text from content blocks when array format is detected  
- Join multiple text parts with newlines
- Fall back to string format for backward compatibility

## Changes
- Modified `internal/translator/gemini/claude/gemini_claude_request.go` lines 95-114
- Added proper handling for both string and array content formats in the response

## Related
- PR #268 fixed logging issues with Gemini's OpenAI-compatible endpoints (JSON parsing in logs, schema validation, performance optimization)
- This PR fixes the actual response handling for tool_result content translation

PR #217, opened 2025-11-05
fix(openai/chat-completions): preserve tool_result JSON, robust quoting, strip unsupported fields

benzntech:
Context: /v1/chat/completions (OpenAI path) was sending invalid Gemini payloads and losing structured tool JSON. This PR implements fixes in the correct translator as requested by the moderator.

Changes
- tool_result handling: detect output_json/json blocks in tool messages and pass JSON through as response.result; for mixed arrays, prefer JSON else join text; avoid data loss.
- quoteIfNeeded: passthrough only when value is valid JSON; otherwise serialize as a JSON string (prevents bare tokens like [Operation Cancelled]).
- tools mapping: after parameters->parametersJsonSchema rename, strip unsupported fields (e.g., strict).

File
- internal/translator/gemini/openai/chat-completions/gemini_openai_request.go

Supersedes PR #209 (which targeted Claude translator). Prior discussion and logs are referenced there; this PR applies the same logic on the correct OpenAI chat-completions path.

Follow-up: happy to add targeted tests in this PR or a separate one.

### DIFF
--- a/internal/translator/gemini/openai/chat-completions/gemini_openai_request.go
+++ b/internal/translator/gemini/openai/chat-completions/gemini_openai_request.go
@@ -4,6 +4,7 @@ package chat_completions
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


benzntech:
Referencing prior context from PR #209 (now superseded): detailed logs, sanitized payload excerpts, and moderator feedback are linked there. This PR applies the same fixes to the correct OpenAI chat-completions translator as requested.

luispater:
I fixed them.

Maybe next time, opening an issue will be faster.

PR #243, opened 2025-11-13
feat(openai): inject default params from config

moxi000:


moxi000:
Per latest instructions, keeping work on fork only.

PR #249, opened 2025-11-14
fix(runtime): remove gpt-5.1 minimal effort variant

ben-vargas:
Stop advertising and mapping the unsupported gpt-5.1-minimal variant in the model registry and Codex executor, and align bare gpt-5.1 requests to use medium reasoning effort like Codex CLI while preserving minimal for gpt-5.


ben-vargas:
Blog post outlining "minimal" no longer valid, maybe in a follow up PR you could add "none" if desired but I didn't go that far as it would be new and most users are probably using coding clis and reasoning models should probably just use "low", "medium", or "high" like codex cli implementation - "medium" and "high" only for `gpt-5.1-codex-mini`.

https://openai.com/index/gpt-5-1-for-developers/?utm_source=chatgpt.com

luispater (approved):

luispater:
I tested it, and only `gpt-5.1` supports none; `gpt-5.1-codex` and `gpt-5.1-codex-mini` do not support the `none` reasoning effort. I still added none as an alias for `gpt-5.1`.

ben-vargas:
Nice, thanks for sharing your test results. That jives with what I think I recall reading yesterday, the codex models are reasoning only and thus don't support a "none" reasoning effort, so that aligns.

PR #288, opened 2025-11-20
Fix OpenAI responses 404

arnaudbreton:


luispater:
This PR is unrelated to this issue. The issue was caused by an error with OpenAI's service at the time.

PR #327, opened 2025-11-25
Fix Antigravity Claude tools schema for Claude Code

francomascareloai:
When routing Claude Code messages via the Antigravity executor, the current tools translation assumes a different custom tools JSON schema than the one emitted by Claude Code. For models like claude-sonnet-4-5-thinking this causes upstream 400 errors such as 'tools.0.custom.input_schema: Field required'.\n\nThis change updates ConvertClaudeRequestToAntigravity to intentionally drop tool declarations instead of forwarding them, relying on Claude Code's local tooling instead. This keeps Sonnet 4.5 Thinking + tools usable from Claude Code without 400s while maintaining the existing thinkingConfig and content translation.

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -119,30 +119,17 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 	}
 
 	// tools
+	// NOTE:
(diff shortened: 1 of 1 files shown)


luispater:
Drop all tools?

What's the point of this?

And you drop the tools for all models, meaning that all models accessed through the `Claude` protocol in `Antigravity` can no longer use tools.

PR #360, opened 2025-11-27
fix(translator): handle non-JSON output gracefully in function call r…

nestharus:
- Parse function response output to detect JSON vs plain text
- Use SetRawBytes for valid JSON to preserve structure
- Use SetBytes for non-JSON to properly escape as string
- Prevents double-encoding of JSON responses


luispater:
Fixed

PR #376, opened 2025-11-30
feat(util): add -reasoning suffix support for Gemini models

hkay-dev:
### Summary

Adds `-reasoning` suffix support for Gemini models, enabling thinking mode with dynamic budget.

### Problem

There's no easy way to enable Gemini's reasoning mode via model name. We have `-nothinking` and `-thinking-N`, but nothing for "just turn on reasoning and let the model decide the budget."

### Solution

1. **gemini_thinking.go**: Parse `-reasoning` suffix
   - `gemini-2.5-flash-reasoning` -> `gemini-2.5-flash` with `thinkingBudget=-1` and `include_thoughts=true`

2. **antigravity_executor.go**: Read and apply the thinking config from metadata
   - Same pattern used by gemini and gemini-cli executors

### Changes

- `internal/util/gemini_thinking.go`: +9 lines
- `internal/runtime/executor/antigravity_executor.go`: +21 lines

### Testing

Tested with REST calls against a local instance:

| Test | Model | Result |
|------|-------|--------|
| New suffix | `gemini-2.5-flash-reasoning` | `reasoning_content` visible |
| Variant | `gemini-2.5-flash-lite-reasoning` | `reasoning_content` visible |
| Base model | `gemini-2.5-flash` | Works as before |
| Existing suffix | `gemini-2.5-flash-nothinking` | Still works |
| Body param | `reasoning_effort: "high"` | Still works |
| Unit tests | `go test ./...` | All pass |
| Build | `go build ./...` | Clean |

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -17,6 +17,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/config"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
 	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/executor"
 	sdktranslator "github.com/router-for-me/CLIProxyAPI/v6/sdk/translator"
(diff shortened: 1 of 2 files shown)


hkay-dev:
Thanks for the review! Good catch on the duplication - I've extracted it into an `applyThinkingMetadata` helper function.

Worth noting: the existing gemini and gemini-cli executors have the same duplicated pattern in their Execute/ExecuteStream methods. Might be worth a follow-up cleanup across all three executors, but I'll leave that for the maintainers to decide.

luispater (approved):

alexx-ftw:
Is this needed for Gemini 3 Pro Preview, as in "gemini-3-pro-preview-reasoning" ?

PR #401, opened 2025-12-02
Support OpenAI responses wire API and provider query params for OpenAI-compatible upstreams

unstableneutron:
This PR extends the current OpenAI-compat layer so that it's flexible enough to talk to providers that uses the new `responses` API and/or requires querystring params (e.g. Azure OpenAI).

For example, GPT-5* series via Azure supports `/v1/responses` API and requires you to pass through additional query string parameters, such as `?api-version=preview` on every request.

## Changes

**Config & watcher**

* `OpenAICompatibility` now supports:
 * `wire-api`: `"chat"` (default) or `"responses"`, selecting the upstream OpenAI wire format.
 * `query-params`: `map[string]string` of query parameters to append to all requests.
* `Watcher.SnapshotCoreAuths` propagates these settings into `coreauth.Auth.Attributes`:
 * `wire_api` attribute for wire API selection.
 * `query:<name>` attributes for each configured query param (e.g. `query:api-version`).

**Executor behavior**
* `OpenAICompatExecutor` now:
 * Picks the translation target format based on `wire_api`:
 * `"chat"` → `openai` (chat/completions format).
 * `"responses"` → `openai-response` (responses format).
 * Builds the upstream URL via `buildRequestURL`:
 * For `wire_api="chat"` → `.../chat/completions`.
 * For `wire_api="responses"`:
 * If `base-url` ends with `/v1` → append `/responses`.
 * Otherwise → append `/v1/responses`.
 * Handles trailing slashes on `base-url`.
 * Applies query params from `auth.Attributes` (keys prefixed with `query:`) to the URL.

This keeps existing providers working as-is (default `wire-api` remains `chat` and no query params are added unless configured).

**Utility helpers**
* New helpers in `internal/util/header_helpers.go`:
 * `extractCustomQueryParams(attrs)` - pulls `query:<name>` entries from attributes.
 * `appendQueryParams(baseURL, params)` - appends sorted, URL-encoded query parameters, respecting existing query strings.
 * `ApplyCustomQueryParamsFromAttrs(baseURL, attrs)` - convenience wrapper used by the executor.
* New unit tests for these helpers.

**Docs / examples**
* `config.example.yaml` updated with `wire-api` and `query-params` examples.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -126,9 +126,19 @@ ws-auth: false
 #      - name: "moonshotai/kimi-k2:free" # The actual model name.
 #        alias: "kimi-k2" # The alias used in the API.
 #
-#  # Example: Azure OpenAI configuration
-#  - name: "azure-openai"
-#    base-url: "https://myresource.openai.azure.com/openai/deployments/gpt-4"
+#  # Example: Azure OpenAI Responses API
+#  # Note: base-url should include everything up to (but not including) /chat/completions or /responses
+#  - name: "azure-openai-responses"
+#    base-url: "https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1"
+#    wire-api: "responses"
+#    query-params:
+#      api-version: "preview"
+#    api-key-entries:
+#      - api-key: "your-azure-api-key"
+#
+#  # Example: Azure OpenAI Chat Completions API (deployment-specific)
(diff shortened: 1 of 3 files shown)


luispater:
You should add the query string based on `internal/runtime/executor/codex_executor.go`

PR #435, opened 2025-12-06
fix: prioritize model mappings over local providers for Amp CLI

heyhuynhgiabuu:
## Summary
- Model mappings now take precedence over local API keys when routing Amp CLI requests
- This allows users to route requests (e.g., `claude-opus-4-5`) to their preferred target models (e.g., `gemini-claude-opus-4-5-thinking`) via OAuth, even when local API keys are configured

## Problem
Previously, if a user had a local Anthropic API key configured, Amp CLI requests for models like `claude-opus-4-5-20251101` would bypass model mappings and go directly to Anthropic. Users couldn't override this behavior to use their preferred OAuth-based providers.

## Solution
Changed the routing priority in `fallback_handlers.go`:
1. **Before**: Check local providers first → model mappings only if no local provider
2. **After**: Check model mappings first → local providers as fallback

This gives users full control over request routing through their configured model mappings.

### DIFF
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -241,3 +241,11 @@ func (h *Handler) DeleteProxyURL(c *gin.Context) {
 	h.cfg.ProxyURL = ""
 	h.persist(c)
 }
+
+// Prioritize Model Mappings (for Amp CLI)
+func (h *Handler) GetPrioritizeModelMappings(c *gin.Context) {
+	c.JSON(200, gin.H{"prioritize-model-mappings": h.cfg.AmpCode.PrioritizeModelMappings})
+}
+func (h *Handler) PutPrioritizeModelMappings(c *gin.Context) {
+	h.updateBoolField(c, func(v bool) { h.cfg.AmpCode.PrioritizeModelMappings = v })
+}
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -100,6 +100,16 @@ func (m *AmpModule) Name() string {
 	return "amp-routing"
 }
 
+// getPrioritizeModelMappings returns whether model mappings should take precedence over local API keys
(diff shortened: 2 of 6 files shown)


hkfires:
According to the current logic, model mappings serve as a fallback. It is not recommended to change this behavior. If you want model mappings to take precedence, you could add a configuration option to enable that priority.

heyhuynhgiabuu:
## Updated: Added Configuration Option

Based on your feedback, I've updated the PR to add a configuration option instead of changing the default behavior.

### Changes:
- Added `prioritize-model-mappings` config option to `AmpCode` struct
- When `false` (default): Local API keys take precedence (original behavior preserved)
- When `true`: Model mappings take precedence over local API keys

### New Management API Endpoints:
- `GET /v0/management/prioritize-model-mappings` - Get current setting
- `PUT /v0/management/prioritize-model-mappings` - Update setting

### Config Example:
```yaml
ampcode:
  model-mappings:
    - from: claude-opus-4-5-20251101
      to: gemini-claude-opus-4-5-thinking
  prioritize-model-mappings: true  # Optional, defaults to false
```

This preserves backward compatibility while allowing users who want mapping priority to enable it explicitly.

hkfires:
Thanks for your contribution! Could you please rebase this PR onto the dev branch? I noticed this PR includes changes to both the translator and executor. We prefer each PR to focus on a single feature/fix. Would you mind splitting this into separate PRs? Thanks!

heyhuynhgiabuu:
Done! I've rebased onto `dev` branch and removed the translator/executor changes.

This PR now contains **only** the prioritize-model-mappings feature:
- Config option `prioritize-model-mappings` in AmpCode struct
- Management API endpoints `GET/PUT /prioritize-model-mappings`
- Fallback handler logic to check the config

The thinking/translator/executor changes were already addressed by upstream commits (`76af454`, `9c09128`, `a174d01`) merged into dev, so they're no longer needed here.

PR #447, opened 2025-12-07
feat: add model name to GIN request logs

heyhuynhgiabuu:
## Summary

Extract model field from JSON request body and include it in GIN log output.
This enables downstream log parsers to identify which AI model was used for each request.

## Changes

- Add `extractModelFromRequest()` function to parse model from request body
- Safely read and restore request body for downstream handlers  
- Update log format to include `| model=<name>` suffix
- Fall back to 'unknown' if model extraction fails

## New Log Format

```
[GIN] 2024/01/15 - 10:30:45 | 200 | 1.234s | 127.0.0.1 | POST "/v1/messages" | model=claude-3-opus
```

## Testing

- Verified Go build passes
- Tested with ProxyPal dashboard - model names now display correctly

### DIFF
--- a/internal/logging/gin_logger.go
+++ b/internal/logging/gin_logger.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"runtime/debug"
+	"strings"
 	"time"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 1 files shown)


hkfires:
We have added a unique ID to every log entry, making it easy to correlate requests and responses. Therefore, this PR is no longer necessary.

PR #454, opened 2025-12-08
feat(config): add configurable host binding for server

vuonglv1612:
## Summary

Hi everyone! While I was using this proxy server, I noticed that it always binds to all network interfaces by default. I thought it would be useful to have control over which interface the server binds to, so I added a new `host` configuration option.

This allows users to restrict the server to localhost-only access when needed, which is especially useful for local development or security-sensitive deployments.

## Motivation

Currently, the proxy server always binds to all network interfaces (implicitly `0.0.0.0`). I ran it on my own server, and since this service is public, someone scanned my proxy with the open port. This made me realize the importance of being able to restrict which interface the server binds to.

While working with the server, I found scenarios where this might not be ideal:
- Local development environments where I want to prevent external access
- Security-sensitive deployments requiring localhost-only access
- Controlled network environments with specific interface binding requirements

I think having this flexibility would benefit others who have similar use cases or security concerns.

## Changes

- **Config struct**: Added `Host` field with YAML binding support
- **Default value**: Set to `"0.0.0.0"` for backward compatibility
- **Server binding**: Updated HTTP server address format from `":port"` to `"host:port"`
- **Startup message**: Updated to display full `"host:port"` address
- **Documentation**: Added configuration example in `config.example.yaml`

### For external access (default, backward compatible):
```yaml
# Omit host field entirely, or explicitly set:
host: 0.0.0.0
port: 8317
```

### For localhost-only access:
```yaml
host: 127.0.0.1
port: 8317
```

Or alternatively:
```yaml
host: localhost
port: 8317
```

## Backward Compatibility

Existing configurations without the `host` field will automatically default to `"0.0.0.0"`, maintaining the current behavior of binding to all network interfaces.

## Files Changed
- `config.example.yaml` - Added host configuration documentation and example
- `internal/config/config.go` - Added Host field with default value "0.0.0.0"
- `internal/api/server.go` - Updated server address to use host:port format
- `sdk/cliproxy/service.go` - Updated startup message to display host:port

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -21,7 +21,7 @@ import (
 type Config struct {
 	config.SDKConfig `yaml:",inline"`
 	// Host is the network host/interface on which the API server will bind.
-	// Default is "0.0.0.0" (all interfaces). Use "127.0.0.1" or "localhost" for local-only access.
+	// Default is empty ("") to bind all interfaces (IPv4 + IPv6). Use "127.0.0.1" or "localhost" for local-only access.
 	Host string `yaml:"host" json:"-"`
 	// Port is the network port on which the API server will listen.
 	Port int `yaml:"port" json:"-"`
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -320,6 +323,7 @@ func LoadConfigOptional(configFile string, optional bool) (*Config, error) {
 	// Unmarshal the YAML data into the Config struct.
 	var cfg Config
 	// Set defaults before unmarshal so that absent keys keep defaults.
+	cfg.Host = "0.0.0.0" // Default to all interfaces for backward compatibility
```
luispater: Please change "0.0.0.0" to an empty string.

If the host is set to an empty string, `fmt.Sprintf("%s:%d", cfg.Host, cfg.Port)` will result in a format like `:8317`. It will bind to all IPv4 and IPv6 addresses.

Using "0.0.0.0" will bind to all IPv4 addresses only, ignoring IPv6 addresses.

vuonglv1612: I have updated @luispater

PR #466, opened 2025-12-09
fix(translator): preserve tool_use blocks on args parse failure

heyhuynhgiabuu:
## Summary
- Fix silent dropping of `tool_use` blocks when JSON unmarshaling of input args fails
- Initialize empty args map on parse failure instead of discarding the block

## Problem
When `json.Unmarshal` failed on tool_use input arguments, the entire `tool_use` block was silently dropped. This broke the required tool_use/tool_result pairing, causing Claude API errors:
```
tool_use ids were found without tool_result blocks immediately after: toolu_xxx
```

## Solution
Changed from `if err == nil { append }` to `if err != nil { args = empty }; append` to always preserve the tool_use block.

## Files Changed
- `internal/translator/antigravity/claude/antigravity_claude_request.go`
- `internal/translator/gemini/claude/gemini_claude_request.go`


heyhuynhgiabuu:
Closing in favor of issue #471 - translator changes require maintainer implementation due to path restrictions.

PR #468, opened 2025-12-09
fix(config): set default MaxRetryInterval to 30s

heyhuynhgiabuu:
## Summary
- Set default `MaxRetryInterval` to 30 seconds before unmarshalling config

## Problem
When `max-retry-interval` is absent in config, the default was zero, causing immediate retries without cooldown.

## Solution
Initialize `MaxRetryInterval = 30` prior to YAML unmarshal so absent keys retain this cooldown value.

## File Changed
- `internal/config/config.go`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,6 +17,8 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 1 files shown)


luispater:
Sorry, no plans to merge this PR for now.

PR #504, opened 2025-12-12
feat(auth): add priority-based auth selection

### LINKED ISSUE #349: Feature: Add tier-based provider prioritization
Add support for configuring providers in priority tiers, where lower-tier (higher priority) providers are used first, and higher-tier providers serve as fallbacks only when all lower-tier providers are exhausted.

## Use Case

Many users have multiple API access methods with different cost structures:
- **Tier 1**: Subscription-based accounts (e.g., Claude Max) - cheaper/included usage
- **Tier 2**: Usage-based API keys (e.g., pay-per-token) - more expensive

Currently, CLIProxyAPI uses round-robin across all providers equally, which means expensive usage-based credentials get consumed even when cheaper subscription-based options are available.

## Proposed Configuration

```yaml
claude-api-key:
  - api-key: "AIzaSy...subscription"
    tier: 1  # Use first (subscription-based, cheaper)
  - api-key: "AIzaSy...payasyougo"
    tier: 2  # Fallback only (usage-based, expensive)

openai-compatibility:
  - name: "openrouter"
    tier: 2  # Fallback provider
    # ...
  - name: "local-llm"
    tier: 1  # Preferred provider
    # ...
```

Expected behavior:

1. Always prefer tier 1 providers (round-robin within tier 1)
2. Only use tier 2 when ALL tier 1 providers are:
    - Quota exceeded (in cooldown)
    - Temporarily unavailable (auth errors, rate limits)
3. Automatically return to tier 1 when any tier 1 provider recovers
4. Default tier: 1 if not specified (backward compatible)
5. Able to add more tiers (tier 3, tier 4, etc.) following the same pattern

Alternatives considered:

- Manually disabling expensive credentials and re-enabling when needed (inconvenient)
- Running separate proxy instances per tier (complex, doesn't solve automatic fallback)

1PercentSync:
Closes #349

## Summary

Add priority-based selection for auth credentials. Higher priority auths are selected first; same priority uses round-robin.

## Motivation

Current round-robin selection has two issues:
1. **Cache inefficiency** - Lack of request stickiness causes cache misses
2. **No cost optimization** - Cannot prioritize cheaper auth over expensive ones

Related: #349 requested tier-based prioritization for cost optimization.

## Implementation

Selection logic in `RoundRobinSelector.Pick`:
1. Check if all available auths have the same priority
2. If same → original round-robin across all (backward compatible)
3. If different → filter to highest priority group, round-robin within that group
4. When high-priority auths enter cooldown/blocked, lower priority ones become available

Priority sources:
- **API keys**: `priority` field in config struct (gemini, claude, codex, openai-compat, vertex)
- **OAuth/file-backed**: `auth-priority` map in config, keyed by filename

## Example

```yaml
claude-api-key:
  - api-key: "sk-subscription..."
    priority: 100       # Used first
  - api-key: "sk-payg..."
    priority: 10        # Fallback

auth-priority:
  "primary@example.com.json": 100
  "backup@example.com.json": 10
```

## Scope

This PR includes core logic and config examples only. Not yet implemented:
- Management API endpoints
- Web UI integration
- Wiki documentation

## Testing

Have been tested in docker environment


luispater:
Sorry, no plans to merge this PR for now.

PR #516, opened 2025-12-13
Feature/usage metrics

Ybazylbe:


luispater:
We will not accept any PRs for persistent statistics.

If persistent statistics are required, please use the management API for external storage.

BTW, you are not using the statistics correctly; the SDK provides a usage interface for this purpose.

PR #529, opened 2025-12-14
fix: ensure message_start sent before content_block_start in OpenAI→Anthropic translation

heyhuynhgiabuu:
## Summary

Fixes SSE event ordering violation when using OpenAI-compatible providers (like `copilot-api`) that don't send `role` in the first streaming chunk.

## Problem

Some providers send the first chunk with `content` directly, without `role: "assistant"`. The current code only emits `message_start` when it sees `role`, causing `content_block_start` to be sent first → clients like Amp CLI crash with:

```
Error: Unexpected event order, got content_block_start before message_start
```

## Solution

Changed the logic to emit `message_start` before ANY content arrives - checking for `content`, `reasoning_content`, `tool_calls`, OR `role`:

```go
if !param.MessageStarted {
    hasContent := delta.Get("content").Exists() && delta.Get("content").String() != ""
    hasReasoning := delta.Get("reasoning_content").Exists()
    hasToolCalls := delta.Get("tool_calls").Exists()
    hasRole := delta.Get("role").Exists() && delta.Get("role").String() == "assistant"

    if hasRole || hasContent || hasReasoning || hasToolCalls {
        // Send message_start event first
    }
}
```

## Testing

- Tested with copilot-api → ProxyPal → Amp CLI flow
- Verified SSE events are now correctly ordered

Fixes: heyhuynhgiabuu/proxypal#20

### DIFF
--- a/internal/translator/openai/claude/openai_claude_response.go
+++ b/internal/translator/openai/claude/openai_claude_response.go
@@ -130,29 +130,36 @@ func convertOpenAIStreamingChunkToAnthropic(rawJSON []byte, param *ConvertOpenAI
 
 	// Check if this is the first chunk (has role)
 	if delta := root.Get("choices.0.delta"); delta.Exists() {
-		if role := delta.Get("role"); role.Exists() && role.String() == "assistant" && !param.MessageStarted {
-			// Send message_start event
-			messageStart := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


heyhuynhgiabuu:
Closing as per CI path guard policy. Created issue instead.

PR #553, opened 2025-12-15
fix(translator): preserve built-in tools (web_search) to Responses API

XInTheDark:
Problem
- OpenAI Chat Completions requests that include built-in tools like {"type":"web_search"} were dropped when translating to the Responses API (Codex), resulting in upstream receiving tools:[] and no web search.

Fix
- Pass through non-function tool entries unchanged during OpenAI Chat Completions -> Codex (Responses) translation.
- Preserve tool_choice for built-in tools (and flatten function tool_choice to Responses shape).
- Also preserve built-in tools when converting OpenAI Responses -> OpenAI Chat Completions.

Tests
- go test ./test -run BuiltinTools


luispater:
Please attach the failed request payload for our verification.

luispater (approved):

PR #556, opened 2025-12-15
fix: bypass KorProxy auth for Amp management routes

korallis:
## Problem

After successfully running `amp login` through KorProxy, users get "Not logged in. Run 'amp login' to authenticate" when running `amp`.

The login succeeds and stores the Amp API token in `~/.local/share/amp/secrets.json`, but subsequent requests to `/api/user` return 401 unauthorized.

## Root Cause

Amp management routes (`/api/user`, `/api/auth`, etc.) were using KorProxy's API key authentication middleware. This middleware expects KorProxy API keys in the Authorization header, but Amp CLI sends Amp tokens instead.

The `registerManagementRoutes` function was receiving the same `auth` middleware as provider routes:
```go
m.registerManagementRoutes(ctx.Engine, ctx.BaseHandler, auth)
```

## Fix

Pass `nil` for auth to management routes since ampcode.com handles its own authentication:
```go
m.registerManagementRoutes(ctx.Engine, ctx.BaseHandler, nil)
```

Management routes remain protected by:
- `localhostOnlyMiddleware` - restricts access to localhost (configurable)
- `managementAvailabilityMiddleware` - requires upstream URL to be configured

## Testing

1. Configure `ampcode.upstream-url` in config.yaml
2. Run `amp login` through the proxy
3. Run `amp` - should now work without "Not logged in" error

### DIFF
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -25,14 +25,15 @@ type Option func(*AmpModule)
 //   - Automatic gzip decompression for misconfigured upstreams
 //   - Model mapping for routing unavailable models to alternatives
 type AmpModule struct {
-	secretSource    SecretSource
-	proxy           *httputil.ReverseProxy
-	proxyMu         sync.RWMutex // protects proxy for hot-reload
-	accessManager   *sdkaccess.Manager
-	authMiddleware_ gin.HandlerFunc
-	modelMapper     *DefaultModelMapper
-	enabled         bool
-	registerOnce    sync.Once
+	secretSource      SecretSource
+	proxy             *httputil.ReverseProxy // For provider routes (strips client auth, injects upstream-api-key)
+	managementProxy   *httputil.ReverseProxy // For management routes (preserves client auth/Amp token)
+	proxyMu           sync.RWMutex           // protects both proxies for hot-reload
+	accessManager     *sdkaccess.Manager
+	authMiddleware_   gin.HandlerFunc
+	modelMapper       *DefaultModelMapper
+	enabled           bool
+	registerOnce      sync.Once
 
 	// restrictToLocalhost controls localhost-only access for management routes (hot-reloadable)
 	restrictToLocalhost bool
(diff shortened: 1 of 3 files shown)


korallis:
## Code Review Response

Thanks for the thorough review! Here's how the identified issues have been addressed:

### 1. Authorization Header Stripping ✅
This was addressed in the second commit by creating a separate `createManagementProxy()` function that **preserves** the client's Authorization header (containing the Amp token). The management routes now use this dedicated proxy instead of the provider proxy that strips auth headers.

Key change in `proxy.go`:
```go
func createManagementProxy(upstreamURL string) (*httputil.ReverseProxy, error) {
    proxy.Director = func(req *http.Request) {
        originalDirector(req)
        req.Host = parsed.Host
        // Preserve client's Authorization header - it contains the Amp token
        // Do NOT strip or replace it
    }
}
```

### 2. Security with `restrict-management-to-localhost: false` ✅
Valid concern! Since KorProxy API key auth is bypassed for management routes (to allow Amp tokens to reach ampcode.com), the localhost restriction provides important defense-in-depth.

**Action taken in KorProxy app (https://github.com/korallis/korproxy):**
- Changed default from `false` to `true` for secure-by-default behavior
- Added config comments explaining the setting
- Release v1.0.27 deployed with this fix

Users accessing KorProxy remotely can explicitly set `restrict-management-to-localhost: false` in their config.

### Security Model Summary
1. **Management routes**: Bypass KorProxy auth → Management proxy preserves Amp token → ampcode.com validates token
2. **Provider routes**: KorProxy auth validates API key → Provider proxy strips auth, injects upstream-api-key → ampcode.com uses that key
3. **Localhost restriction**: Defense-in-depth for management routes (now `true` by default)

hkfires:
> After successfully running `amp login` through KorProxy, users get "Not logged in. Run 'amp login' to authenticate" when running `amp`.
> 
> The login succeeds and stores the Amp API token in `~/.local/share/amp/secrets.json`, but subsequent requests to `/api/user` return 401 unauthorized.

We have updated the application logic. It now explicitly and exclusively accepts the `upstream-api-key` for communication with `ampcode.com`. All AMP clients (including the `amp cli` and the `VS Code plugin`) should use their `cliproxyapi api-keys` to communicate with `cliproxyapi`.

PR #570, opened 2025-12-16
Fix invalid thinking signature when proxying Claude via Antigravity

fuguiKz:
Fixes a 400 `INVALID_ARGUMENT` error from Vertex/Antigravity Claude models (e.g. `gemini-claude-sonnet-4-5-thinking`):

> messages.N.content.M: Invalid `signature` in `thinking` block

**Root cause**
- When translating Claude -> Antigravity, we were filling missing thinking signatures with the placeholder `skip_thought_signature_validator`. Antigravity/Vertex validates Claude thinking signatures and rejects fabricated ones.
- We also weren’t reliably propagating `thoughtSignature` from Antigravity responses back into Claude streaming/non-stream responses, so clients couldn’t replay signed thinking blocks.

**Changes**
- Drop unsigned Claude `thinking` blocks when translating Claude -> Antigravity (we can’t mint valid provider signatures).
- Preserve/emit thought signatures from Antigravity responses (including signature-only parts) so clients can replay signed thinking blocks.

**Tests**
- Added coverage for unsigned thinking blocks and signature propagation in both stream + non-stream paths.

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -84,13 +84,18 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 					contentResult := contentResults[j]
 					contentTypeResult := contentResult.Get("type")
 					if contentTypeResult.Type == gjson.String && contentTypeResult.String() == "thinking" {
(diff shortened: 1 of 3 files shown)


luispater:
It looks like it's not working. It caused another issue, so I reverted it.

PR #589, opened 2025-12-18
fix: Fixes Bash tool command parameter name mismatch

soilSpoon:
Addresses an issue where the Bash tool expects the "cmd" parameter, but Gemini sends "command".

This commit introduces a function that converts the "command" field to "cmd" for Bash tools. It uses proper JSON parsing to avoid accidental replacements of "command" that appear in values.

Also, adds tests to verify that the conversion works correctly in different scenarios.

## Summary
Gemini returns `command` field for Bash tool arguments, but Amp clients expect `cmd`. 
This converts `command` → `cmd` using proper JSON parsing.

## References
- Anthropic Bash Tool docs: https://docs.anthropic.com/en/docs/build-with-claude/computer-use#bash-tool
- Tool naming convention: Anthropic uses versioned tool names like `bash_20241022`


luispater:
This is a client-side issue.

We shouldn't change the payload if the client sends an incorrect tool name.

PR #604, opened 2025-12-18
fix(amp): inject Amp token for management routes to fix thread reading and web search

korallis:
## Summary

This PR fixes authentication for Amp CLI management routes (threads, web search, user info) when using CLIProxyAPI as a proxy. Previously, these routes failed with "unauthorized" errors.

## Problem

When Amp CLI is configured to use CLIProxyAPI, management routes like `/api/threads/*`, `/api/user/*`, and web search were failing with 401 "unauthorized" errors.

**Root Cause:** The management proxy was designed to preserve the client's `Authorization` header, assuming it contained an Amp token. However, when clients authenticate to CLIProxyAPI, they send the **CLIProxyAPI API key**, not an Amp token. This key is meaningless to ampcode.com, causing authentication failures.

## Solution

Modified `createManagementProxy` to behave like the provider proxy:
1. **Strip** client's `Authorization` and `X-Api-Key` headers (they contain CLIProxyAPI credentials)
2. **Inject** the server's Amp token from SecretSource (config > env > file precedence)

This enables all management routes to authenticate correctly with ampcode.com.

### `proxy.go`
- Updated `createManagementProxy` to accept `SecretSource` parameter
- Strip client auth headers before forwarding
- Inject Amp token from SecretSource
- Added correlation ID propagation for request tracing

### `amp.go`
- Pass `m.secretSource` to `createManagementProxy`
- Updated `OnConfigUpdated` to recreate management proxy when URL changes

### `proxy_test.go`
- Added 5 new tests:
  - `TestCreateManagementProxy_InjectsToken`
  - `TestCreateManagementProxy_StripsClientAuth`
  - `TestCreateManagementProxy_NoTokenContinues`
  - `TestCreateManagementProxy_PreservesOtherHeaders`
  - `TestCreateManagementProxy_NilSecretSource`

## Features Now Working

| Feature | Route | Status |
|---------|-------|--------|
| Thread reading | `/api/threads/*` | ✅ Fixed |
| Web search | Various | ✅ Fixed |
| User info | `/api/user/*` | ✅ Fixed |
| Auth flow | `/api/auth/*` | ✅ Fixed |
| Telemetry | `/api/telemetry/*` | ✅ Fixed |

## Testing

- All existing tests pass (no regressions)
- 5 new unit tests added and passing
- Build succeeds

## Configuration

Users need to configure an Amp token via one of:
- `ampcode.upstream-api-key` in config.yaml
- `AMP_API_KEY` environment variable
- `~/.local/share/amp/secrets.json` with key `apiKey@https://ampcode.com/`

## Related

This also includes a previous fix for model-aware beta header filtering that preserves `web_search` and other beta features for non-1M context models.

### DIFF
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -218,14 +218,21 @@ func (m *AmpModule) OnConfigUpdated(cfg *config.Config) error {
 		// Check upstream URL change - now supports hot-reload
 		if newUpstreamURL == "" && oldUpstreamURL != "" {
 			m.setProxy(nil)
+			m.setManagementProxy(nil)
 			m.enabled = false
 		} else if oldUpstreamURL != "" && newUpstreamURL != oldUpstreamURL && newUpstreamURL != "" {
-			// Recreate proxy with new URL
+			// Recreate both proxies with new URL
 			proxy, err := createReverseProxy(newUpstreamURL, m.secretSource)
 			if err != nil {
 				log.Errorf("amp config: failed to create proxy for new upstream URL %s: %v", newUpstreamURL, err)
 			} else {
 				m.setProxy(proxy)
+			}
+			mgmtProxy, err := createManagementProxy(newUpstreamURL, m.secretSource)
+			if err != nil {
+				log.Errorf("amp config: failed to create management proxy for new upstream URL %s: %v", newUpstreamURL, err)
+			} else {
+				m.setManagementProxy(mgmtProxy)
 			}
 		}
 
@@ -265,8 +272,8 @@ func (m *AmpModule) enableUpstreamProxy(upstreamURL string, settings *config.Amp
 		return err
 	}
 
-	// Create management proxy (preserves client auth/Amp token)
-	mgmtProxy, err := createManagementProxy(upstreamURL)
+	// Create management proxy (injects server's Amp token for auth)
+	mgmtProxy, err := createManagementProxy(upstreamURL, m.secretSource)
 	if err != nil {
 		return err
 	}
--- a/internal/api/modules/amp/fallback_handlers.go
+++ b/internal/api/modules/amp/fallback_handlers.go
@@ -224,8 +224,9 @@ func (fh *FallbackHandler) WrapHandler(handler gin.HandlerFunc) gin.HandlerFunc 
 			logAmpRouting(RouteTypeModelMapping, modelName, resolvedModel, providerName, requestPath)
 			rewriter := NewResponseRewriter(c.Writer, normalizedModel)
 			c.Writer = rewriter
-			// Filter Anthropic-Beta header only for local handling paths
-			filterAntropicBetaHeader(c)
+			// Filter Anthropic-Beta header only for 1M context models on local providers
+			// Non-1M models pass all betas through to support web_search, tool streaming, etc.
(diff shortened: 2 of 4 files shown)


hkfires:
The problem you have described should not occur if you are using the most recent version and have configured it according to the guide located at https://help.router-for.me/agent-client/amp-cli.html#option-a-settings-file.

aftely1337:
@hkfires "Does AMP's web search require credits? I'm already on the latest version, but still can't use web search. It shows 'Failed to perform web search: Out of credits'."
ScreenShot_2025-12-19_095651_953

hkfires:
> @hkfires "Does AMP's web search require credits? I'm already on the latest version, but still can't use web search. It shows 'Failed to perform web search: Out of credits'." ScreenShot_2025-12-19_095651_953

I'm not quite sure whether web search requires payment, but I wanted to clarify that the issue description in this PR might be wrong. Even though web search returns 'Out of credits' when I use it, the 'read thread' tool is still executing as expected.

aftely1337:
> > @hkfires“AMP 的网络搜索需要积分吗？我已经更新到最新版本了，但仍然无法使用网络搜索功能。它显示‘网络搜索失败：积分不足’。”ScreenShot_2025-12-19_095651_953
> 
> 我不确定网页搜索是否需要付费，但我想澄清一下，这个 PR 中的问题描述可能存在错误。虽然我使用网页搜索时显示“积分不足”，但“读取线程”工具仍然按预期运行。

Yes, only the web search feature isn't working, but 'read thread' functions normally. I tested it and web search works fine in free mode. Could we approach this from that angle?

ben-vargas:
I agree with @hkfires - the description and intent of this PR is not quite right... users without balance have the model reporting back no credits not unauthorized. I think this is more likely the change to require api-key on the management endpoint forwarding and the user not having configured that properly. Because my webSearch2 requests do forward just fine, and they process and work (I have a balance available in my Amp account).

ben-vargas:
> > > @hkfires“AMP 的网络搜索需要积分吗？我已经更新到最新版本了，但仍然无法使用网络搜索功能。它显示‘网络搜索失败：积分不足’。”ScreenShot_2025-12-19_095651_953
> > 
> > 
> > 我不确定网页搜索是否需要付费，但我想澄清一下，这个 PR 中的问题描述可能存在错误。虽然我使用网页搜索时显示“积分不足”，但“读取线程”工具仍然按预期运行。
> 
> Yes, only the web search feature isn't working, but 'read thread' functions normally. I tested it and web search works fine in free mode. Could we approach this from that angle?

I don't think this is a good idea, it's one thing to allow users to use a paid subscription... but using a resource that user's are not paying for (or not seeing ads for), not a great position for CLIProxyAPI imo. 

Both Amp and Factory founders have been very friendly towards the use of CLIProxyAPI - let's not kill that goodwill by consuming resources from them without paying for said resource (either the $0.01 USD for websearch that Amp charges the balance on Amp account, or viewing free ads). 

I bet if you go use the free mode, websearch through CLIProxyAPI will work, you could just do that.

Or, add the Exa MCP via smithery which provides free web search, it's what opencode added for their users.

hkfires:
I agree with @ben-vargas

After discussing this with our core contributors, we have decided not to merge this PR.

Thank you again for your contribution.

PR #615, opened 2025-12-19
feature: Support multiple AMP model fallbacks

soilSpoon:
Introduce multi-target model mappings so a requested model can map to an ordered list of fallbacks instead of a single target. Add parsing for string-or-list in config (YAML/JSON), store mappings as lists, and resolve mappings recursively with cycle detection while preferring targets that have registered providers.

This improves routing robustness by allowing fallback chains (including alias->alias->real model) and skips invalid/empty entries. Maintain backward compatibility by returning the primary target for APIs that expect a single mapping. Update change-detection and summaries to compare normalized, comma-joined target lists and refresh tests to cover provider lookup, recursion, cycles, YAML unmarshalling, and fallback retrieval.

Benefits:
- Enables graceful fallback routing when primary targets are unavailable
- Prevents infinite recursion via cycle detection and logs warnings
- Supports flexible config formats (single string or list)
- Preserves existing single-target behavior for compatibility

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -155,9 +155,20 @@ ws-auth: false
 #   # Route unavailable Amp models to alternative models available in your local proxy.
 #   # Useful when Amp CLI requests models you don't have access to (e.g., Claude Opus 4.5)
 #   # but you have a similar model available (e.g., Claude Sonnet 4).
+#   #
+#   # Supports both single target (backward compatible) and multiple targets (fallback chain).
+#   # When multiple targets are specified, they are tried in order until one succeeds.
 #   model-mappings:
+#     # Single target (backward compatible)
 #     - from: "claude-opus-4.5"       # Model requested by Amp CLI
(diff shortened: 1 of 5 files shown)


hkfires:
One-to-many model mapping is problematic, particularly when mixing different vendors (e.g., mapping opus4.5 to both claude-opus-4.5 and gpt-5.2). These model switches often fail because the context/conversation history is not compatible between different providers.

soilSpoon:
> One-to-many model mapping is problematic, particularly when mixing different vendors (e.g., mapping opus4.5 to both claude-opus-4.5 and gpt-5.2). These model switches often fail because the context/conversation history is not compatible between different providers.

@hkfires I believe this is an issue that the CLIProxyAPI should address.
Currently, I'm using ampcode and resolving such issues whenever errors occur.
I plan to create a PR later.

luispater:
I agree with @hkfires. Closing this PR.

PR #616, opened 2025-12-19
feat(amp): add Amp as provider

thurstonsand:
## Summary

Add Amp (ampcode.com) as a provider option. When enabled, Amp is included in the round-robin provider rotation alongside locally configured providers.

When `amp-credits-fallback: true` is set in config, requests may be routed to ampcode.com using the user's Amp credentials.

In truth, this will become more useful when #504 lands, and it can be configured as a fallback (as intended before I learned cli-proxy-api just does blind round-robin currently).

May want to leave `amp-credits-fallback: false` until we can set it strictly as a fallback.

## Breaking Change

Previously, LLM requests with no local providers configured would automatically passthrough to ampcode.com via ReverseProxy. Now requires explicit `amp-credits-fallback: true` to enable Amp routing. Without it, requests return a `provider_not_found` error.

I think it's highly unlikely anyone was using it in this way, because otherwise what's the point of using cli-proxy-api?

### Config (`config.example.yaml`, `internal/config/config.go`)

- Add `amp-credits-fallback` option under `ampcode` section (default: false)
- Add hot-reload support for config changes
- Default `upstream-url` to `https://ampcode.com` if not specified

### New AmpExecutor (`internal/runtime/executor/amp_executor.go`)

- Implement executor that proxies requests to ampcode.com
- Route to correct endpoint based on source format:
  - Claude → `/api/provider/anthropic/v1/messages`
  - OpenAI → `/api/provider/openai/v1/chat/completions`
  - OpenAI Responses → `/api/provider/openai/v1/responses`
  - Gemini → `/api/provider/google/v1beta1/publishers/google/models/{model}:{action}`
- Support streaming and non-streaming execution
- Handle gzip decompression

### Auth (`sdk/cliproxy/auth/`)

- Add "proxy" account type for transparent pass-through providers
- Bypass model registry checks for proxy auths (upstream handles validation)

### Service (`sdk/cliproxy/service.go`)

- Register synthetic `amp:proxy` auth entry at startup
- Wire up AmpExecutor with MultiSourceSecret for API key retrieval

### Route Handling (`internal/api/modules/amp/`)

- Include "amp" in provider list when option enabled
- Add `ProvidersOverrideContextKey` for pre-computed provider lists

### Utilities (`internal/util/gzip.go`)

- Extract shared gzip helpers from proxy code for reuse

### Unit Tests

- [x] `fallback_handlers_test.go` - Tests for `extractModelFromRequest` and `rewriteModelInRequest` helpers

### Manual E2E Tests

| Test                     | Details                                                                                                                         |
| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------- |
| Claude via amp fallback  | Request to `claude-sonnet-4-20250514` with no local claude-api-key routes to ampcode.com. Logs show `Use Amp Credits for model` |
| Fallback disabled        | With `amp-credits-fallback: false`, Claude request returns `provider_not_found` error                                           |
| Streaming                | SSE stream with `stream: true` returns proper `message_start`, `content_block_delta`, `message_delta` events                    |
| Config hot-reload        | Toggling `amp-credits-fallback` detected by watcher, applied without restart                                                    |
| Model mapping + fallback | Mapped model (`claude-opus-4-5` → `gemini-2.5-flash`) uses local API key; unmapped models use amp                               |
| Amp CLI integration      | `amp -x "prompt"` works through local proxy, management routes (`/api/internal`, `/api/user`) proxied correctly                 |

### DIFF
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -338,3 +338,13 @@ func (m *AmpModule) setRestrictToLocalhost(restrict bool) {
 	defer m.restrictMu.Unlock()
 	m.restrictToLocalhost = restrict
 }
+
+// ampCreditsFallback returns whether amp credits fallback is enabled (hot-reloadable).
+func (m *AmpModule) ampCreditsFallback() bool {
+	m.configMu.RLock()
+	defer m.configMu.RUnlock()
+	if m.lastConfig == nil {
+		return false
+	}
+	return m.lastConfig.AmpCreditsFallback
+}
--- a/internal/api/modules/amp/fallback_handlers.go
+++ b/internal/api/modules/amp/fallback_handlers.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
+	"github.com/router-for-me/CLIProxyAPI/v6/sdk/api/handlers"
 	log "github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
@@ -80,6 +81,7 @@ type FallbackHandler struct {
 	getProxy           func() *httputil.ReverseProxy
 	modelMapper        ModelMapper
 	forceModelMappings func() bool
+	ampCreditsFallback func() bool
 }
 
 // NewFallbackHandler creates a new fallback handler wrapper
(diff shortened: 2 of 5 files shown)


ben-vargas:
@thurstonsand - I'm curious what benefit do you see in having Amp as a fallback provider rather than just using something like OpenRouter or Vercel AI Gateway, which support more models? I could see something like that being useful. 

Ultimately, you're just falling back to paying token usage at cost through Amp? I would think that if you're actually using Amp, and wanting to use Amp for payment, you would just take out the proxy middleware. If you're using another coding platform entirely, I don't see the benefit in paying for tokens through Amp. You're not getting any add-on Amp benefits using it as a random provider?

thurstonsand:
My thoughts on it were basically that I want to use my subscriptions where available, but if I run out of those, I just want it to fall back to using Amp.

I realize that the "fallback" behavior isn't implemented today, but it looks like it might be coming soon (#504).

You're right, I could alternatively plug in openrouter for the same effect here, but this way I could at least support Amp directly. This project still passes through requests that use Amp's infrastructure for storing threads, stats, sharing, etc and I want the option to pay them even if it's just after my subscription quota runs out.

hkfires:
Currently, this project only integrates original LLM providers; we do not accept any third-party services as providers (such as Cursor, Kiro, etc.). 

However, we have a separate project, [CLIProxyAPIPlus](https://github.com/router-for-me/CLIProxyAPIPlus), which accepts community PRs for third-party providers while keeping its core synchronized with CLIProxyAPI. If needed, please feel free to submit your PR to that repository instead.

thurstonsand:
ah, that makes sense. guess I misunderstood because the Amp code currently proxies to amp as a provider if no other provider is set

ben-vargas:
> ah, that makes sense. guess I misunderstood because the Amp code currently proxies to amp as a provider if no other provider is set

Yeah, i think the notable difference is that we're not doing any mappings or using Amp as a real provider in that scenario, we're more "getting out of the way" and passing along the requests unmodified and as if we never saw them (same for all those /api/internal* requests Amp CLI/IDE-Plugins are making).

PR #619, opened 2025-12-19
fix(util): disable default thinking for gemini 3 flash

hkfires:


ben-vargas:
@hkfires - what's the driver behind this change? The Google APIs default to "high" when no thinkingLevel is sent, so I thought it best to indicate that in the codebase... just curious if this is causing a problem, or why this change is being made?

luispater (approved):

sususu98:
https://ai.google.dev/gemini-api/docs/thinking
> You cannot disable thinking for Gemini 3 Pro. Gemini 3 Flash also does not support full thinking-off, but the minimal setting means the model likely will not think (though it still potentially can). If you don't specify a thinking level, Gemini will use the Gemini 3 models' default dynamic thinking level, "high".
> 你无法为 Gemini 3 Pro 禁用思考模式。Gemini 3 Flash 也不支持完全关闭思考模式，但 minimal 设置意味着模型可能不会进行思考（尽管仍有可能会）。如果你不指定思考级别，Gemini 将使用 Gemini 3 模型的默认动态思考级别 "high" 。

所以为什么要做这样的修改呢？在谷歌的文档中提到，gemini-3-flash默认的思考等级为`high`

hkfires:
> @hkfires - what's the driver behind this change? The Google APIs default to "high" when no thinkingLevel is sent, so I thought it best to indicate that in the codebase... just curious if this is causing a problem, or why this change is being made?



> https://ai.google.dev/gemini-api/docs/thinking
> 
> > You cannot disable thinking for Gemini 3 Pro. Gemini 3 Flash also does not support full thinking-off, but the minimal setting means the model likely will not think (though it still potentially can). If you don't specify a thinking level, Gemini will use the Gemini 3 models' default dynamic thinking level, "high".
> > 你无法为 Gemini 3 Pro 禁用思考模式。Gemini 3 Flash 也不支持完全关闭思考模式，但 minimal 设置意味着模型可能不会进行思考（尽管仍有可能会）。如果你不指定思考级别，Gemini 将使用 Gemini 3 模型的默认动态思考级别 "high" 。
> 
> 所以为什么要做这样的修改呢？在谷歌的文档中提到，gemini-3-flash默认的思考等级为`high`

Google's default behavior now suppresses thinking content whenever the thinkingConfig field is absent, even for models like Gemini 3 and Gemini 2.5 Pro where thinking cannot be disabled.

Currently, although we can control the intensity via client-side levels or suffixes, we lack a mechanism to stop the thinking content output entirely. This modification ensures we maintain the "no thinking content by default" behavior.

This is an experimental change limited to gemini-3-preview for now and hasn't been extended to Pro models. Feedback is welcome!

PR #653, opened 2025-12-21
Fix: Use x-api-key header for Claude API instead of Authorization: Bearer

dacsang97:
## Problem
The Claude executor was using `Authorization: Bearer <api-key>` header format, which causes 401 authentication errors when using custom proxy endpoints that expect the standard Anthropic `x-api-key` header format.

## Root Cause
Anthropic's API documentation specifies using `x-api-key` header for authentication, not `Authorization: Bearer`. While the official Anthropic API may accept both formats, custom proxy endpoints often require the standard `x-api-key` header.
https://platform.claude.com/docs/en/get-started

## Solution
Modified `applyClaudeHeaders()` function in `internal/runtime/executor/claude_executor.go` to use:
```go
r.Header.Set("x-api-key", apiKey)
```
instead of:
```go
r.Header.Set("Authorization", "Bearer "+apiKey)
```

## Changes
- Updated authentication header format for Claude API requests

## Impact
- Maintains compatibility with standard Anthropic API authentication

## Testing
Tested with:
- Standard Claude API endpoints
- Custom headers from configuration are still applied correctly

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -662,7 +662,19 @@ func decodeResponseBody(body io.ReadCloser, contentEncoding string) (io.ReadClos
 }
 
 func applyClaudeHeaders(r *http.Request, auth *cliproxyauth.Auth, apiKey string, stream bool, extraBetas []string) {
-	r.Header.Set("x-api-key", apiKey)
+	// Use x-api-key for static API keys, and Authorization: Bearer for OAuth access tokens.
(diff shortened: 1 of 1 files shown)


ajkdrag:
I benefit from the Bearer auth mechanism because of being behind a proxy and using token based authentication instead of api-key. Is there a way to support both??

luispater:
I have re-implemented this part of the code. The `x-api-key` is now used only for the `api.anthropic.com`.

I'm closing this PR.

PR #658, opened 2025-12-22
Fix responses-format handling for chat completions（Support Cursor）

moxi000:
## Summary
- detect OpenAI Responses-format payloads hitting /v1/chat/completions and convert them to Chat Completions before dispatch
- ensure reasoning/tool metadata is preserved so codex upstream accepts tool names
- add regression test using captured payload to verify tools survive conversion

## Testing
- go test ./internal/translator/codex/openai/chat-completions -run TestResponsesPayloadToolsArePreserved


internal/translator/codex/openai/chat-completions/compat_test.go
luispater, line 1: Please remove this file.

PR #663, opened 2025-12-22
feat: add fill-first routing strategy

jroth1111:
## Summary
- add fill-first credential selection strategy (deterministic ordering)
- expose routing strategy via config and builder
- keep round-robin available via config

## Motivation
- make credential usage predictable to manage rate limits/subscription caps
- keep routing configurable while preserving existing round-robin behavior

## Testing
- not run (logic-only change)

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -66,6 +66,10 @@ quota-exceeded:
   switch-project: true # Whether to automatically switch to another project when a quota is exceeded
   switch-preview-model: true # Whether to automatically switch to a preview model when a quota is exceeded
 
+# Routing strategy for selecting credentials when multiple match.
+routing:
+  strategy: "round-robin" # round-robin (default), fill-first
+
 # When true, enable authentication for the WebSocket API (/v1/ws).
 ws-auth: false
 


sdk/cliproxy/builder.go
```diff
@@ -197,7 +198,20 @@ func (b *Builder) Build() (*Service, error) {
...
+		strategy := ""
+		if b.cfg != nil {
+			strategy = strings.ToLower(strings.TrimSpace(b.cfg.Routing.Strategy))
+		}
+		var selector coreauth.Selector
+		switch strategy {
+		case "round-robin", "roundrobin", "rr":
+			selector = &coreauth.RoundRobinSelector{}
+		default:
+			selector = &coreauth.FillFirstSelector{}
```
luispater, line 209: When adding new features, preserve the old logic.

Round Robin should remain the default.

jroth1111:
Rebased on upstream/main and restored round-robin as the default selector (fill-first only when configured). @luispater would you mind re-reviewing?

luispater:
> Rebased on upstream/main and restored round-robin as the default selector (fill-first only when configured). @luispater would you mind re-reviewing?

@jroth1111 You should add an example configuration to `config.example.yaml` in this PR.

jroth1111:
Follow-up: added config.example.yaml routing.strategy example (round-robin default, fill-first option). Ready for re-review.

jroth1111:
Added routing.strategy example in config.example.yaml; round-robin remains the default when unset. Ready for re-review. @luispater

jroth1111:
Added routing.strategy example in config.example.yaml and kept round-robin as default when unset. Ready for re-review. @luispater

PR #715, opened 2025-12-25
feat: implement automatic self-update via --update CLI flag

citizenll:
## Overview
Manual updates for CLIProxyAPI are currently tedious, requiring users to manually download, extract, and replace binary files. This PR introduces a built-in self-update mechanism to streamline this process.

## Changes
- **Added `--update` Flag**: Users can now run `./cli-proxy-api --update` to check for new releases and perform an in-place upgrade.
- **Self-Update Engine ([internal/cmd/update.go](cci:7://file:///D:/Dev/CLIProxyAPI/internal/cmd/update.go:0:0-0:0))**:
  - Fetches the latest release info from GitHub API.
  - Implements semantic versioning (SemVer) comparison to avoid unnecessary downloads or unintended downgrades.
  - Automatically identifies and downloads compatible assets based on the current OS (`windows`, `linux`, `darwin`) and architecture (`amd64`, `arm64`).
  - Supports both `.zip` and `.tar.gz` archive formats.
- **Robust Binary Replacement**:
  - Implemented an atomic-like replacement strategy.
  - **Windows Support**: Since running binaries are locked on Windows, the updater renames the current executable to `.old` before writing the new one, ensuring success without manual intervention.
  - Includes a fallback mechanism to restore the old binary if the update fails during the writing phase.
- **Proxy Support**: The update process respects the optional `ProxyURL` specified in the configuration, ensuring smooth downloads for users in restricted network environments.

## How to Test
1. Compile the project.
2. Run `./cli-proxy-api --update`.
3. If a newer version exists, it will download and replace the current binary.
4. If already on the latest version (e.g., manually setting a high version number), it will correctly report "already up to date".

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -67,6 +67,8 @@ func main() {
 	var configPath string
 	var password string
 
+	var update bool
+
 	// Define command-line flags for different operation modes.
 	flag.BoolVar(&login, "login", false, "Login Google Account")
 	flag.BoolVar(&codexLogin, "codex-login", false, "Login to Codex using OAuth")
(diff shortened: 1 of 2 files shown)


luispater:
You should not create self-update programs for macOS and Linux users.

It goes against the conventions of their package managers.

PR #716, opened 2025-12-25
Multi-Target Model Aliases and Provider Aggregation

mcowger:
#### Overview
This PR introduces a significant enhancement to the model aliasing system, moving from a simple 1:1 mapping on a per provider basis to a robust multi-target aggregation engine. This allows a single alias to point to multiple underlying models, effectively creating a unified "virtual model" that pools providers from all targets.

#### Key Features

*   **Multi-Target Aliases**: Aliases can now be defined as a list of model IDs. For example, a `balanced-basic` alias can point to both `claude-3-5-sonnet` and `gemini-1.5-pro`.
*   **Provider Aggregation (Union Logic)**: The proxy now calculates the union of all available providers across all targets of an alias. If an alias name overlaps with a real model ID, the real model's providers are also included in the pool.
*   **Intelligent Model Resolution**: When a client makes a request using an alias, the system automatically resolves it to the first target model that the specific client (OpenAI, Anthropic, or Gemini) supports.
*   **Flexible YAML Configuration**: Updated the configuration parser to support both legacy string mappings (`alias: target`) and new sequence mappings (`alias: [t1, t2]`) using a custom YAML unmarshaler.
*   **Management API Support**: The Management API has been updated to allow creating and updating multi-target aliases via the `PUT /aliases/:alias` endpoint.

#### Technical Changes

*   **`internal/config/config.go`**: Changed `Aliases` type to `map[string][]string` and implemented `UnmarshalYAML` for backward compatibility.
*   **`internal/registry/model_registry.go`**: 
    *   Updated `GetModelProviders` and `GetModelCount` to aggregate data across all alias targets.
    *   Updated `ResolveModelForClient` to iterate through targets and find the first compatible match.
    *   Updated `isModelAvailableLocked` to check availability across the entire target set.
*   **`internal/api/handlers/management/config_basic.go`**: Updated handlers to accept `targets` (slice) or `target` (string) in JSON payloads.
*   **`internal/util/provider.go`**: Updated global utility functions to handle the new slice-based alias structure.

#### Testing
*   Added `TestMultiTargetAlias` in [internal/registry/model_registry_test.go](internal/registry/model_registry_test.go) to verify provider aggregation and cross-provider resolution.
*   Updated [internal/util/provider_test.go](internal/util/provider_test.go) to ensure utility functions correctly handle multi-target maps.
*   Verified YAML parsing with various configurations (single string, list, and mixed).

#### Example Configuration
```yaml
aliases:
  "claude-sonnet-4-5":
    - "claude-sonnet-4-5-20250929"
    - "gemini-claude-sonnet-4-5"
  "universal-pro":
    - "gpt-4o"
    - "claude-3-5-sonnet-20241022"
    - "gemini-1.5-pro"
```

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -216,12 +216,20 @@ payload:
 # Global model aliases
 # Define custom aliases that can be used across all providers to map to specific model IDs.
 # This eliminates the need for per-provider alias configuration.
+# Supports both single-target strings and multi-target lists for provider aggregation.
 aliases:
-  # Claude aliases - use shorter names for latest versions
+  # Single target alias
   "claude-haiku-4-5": "claude-haiku-4-5-20251001"
-  "claude-sonnet-4-5": "claude-sonnet-4-5-20250929"
-  "claude-opus-4-5": "claude-opus-4-5-20251101"
-  
+
+  # Multi-target alias (aggregates providers from all listed models)
+  "claude-sonnet-4-5":
+    - "claude-sonnet-4-5-20250929"
+    - "gemini-claude-sonnet-4-5"
+
+  "claude-opus-4-5":
+    - "claude-opus-4-5-20251101"
+    - "gemini-claude-opus-4-5-thinking"
+
   # Custom aliases - any name you prefer
   "best-model": "claude-sonnet-4-5-20250929"
   "fast-model": "claude-haiku-4-5-20251001"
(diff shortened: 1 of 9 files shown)


mcowger:
Handled with https://github.com/router-for-me/CLIProxyAPI/pull/795

PR #769, opened 2025-12-29
feat: add per-entry base-url support for OpenAI-compatible API keys

hungthai1401:
## Summary
Adds optional `base-url` field to individual API key entries in OpenAI-compatible configurations. If not specified, falls back to the parent `base-url`.

## Changes
- Added `BaseURL` field to `OpenAICompatibilityAPIKey` struct
- Updated synthesizer to use per-entry base URL with fallback logic
- Added test coverage for the new feature

## Usage
```yaml
openai-compatibility:
- name: provider
  base-url: https://default-api.example.com/v1
  api-key-entries:
    - api-key: key1
       base-url: https://custom-api.example.com/v1  # uses custom URL
    - api-key: key2
       # uses https://default-api.example.com/v1 (fallback)
```

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -134,6 +134,7 @@ ws-auth: false
 #       X-Custom-Header: "custom-value"
 #     api-key-entries:
 #       - api-key: "sk-or-v1-...b780"
+#         base-url: "https://custom.api.com" # optional: per-key base URL override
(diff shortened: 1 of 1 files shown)


hkfires:
Thanks for the contribution. Unfortunately, we follow a "one provider, one base-url" principle, so we cannot merge this.

PR #805, opened 2025-12-30
fix(translator): correctly map stop_reason in response translations

HsnSaboor:
## Summary
This PR fixes an issue where coding tools (Claude Code, OpenCode) would stop after a single turn because the stop_reason was not being correctly mapped from the upstream provider.

- **Gemini**: Correctly map 'MAX_TOKENS' to 'max_tokens' and 'STOP'/'FINISH_REASON_UNSPECIFIED'/'UNKNOWN' to 'end_turn'.
- **Gemini CLI**: Applied the same logic as the standard Gemini translator.
- **Codex**: Propagate the 'stop_reason' directly from the upstream response if available, falling back to 'tool_use' or 'end_turn'.

## Test plan
- Verified with unit tests ensuring correct mapping of `finishReason` to `stop_reason`.
- Manually tested with Claude Code to confirm multi-turn sessions work correctly.
- Verified before/after behavior using a reproduction script across all three translators (Gemini, Gemini CLI, Codex).

### DIFF
--- a/debug_provider.go
+++ b/debug_provider.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
+)
+
+func main() {
+	modelName := "gemini-3-pro-preview"
+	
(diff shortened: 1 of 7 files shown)


luispater:
Fixed f99cddf97f7a91966c679049917e81098fe73c00

PR #821, opened 2026-01-01
feat: Add Antigravity refresh token auth and api-call proxy endpoint

xb0or:
This PR adds two new management API endpoints for Antigravity integration: POST /v0/management/antigravity-refresh-token for batch importing accounts using refresh tokens, and POST /v0/management/api-call for making authenticated API calls with stored access tokens.


luispater:
CLIProxyAPI refresh all tokens automatically.

Why do you want to refresh them via the management API?

xb0or:
Some features in the UI require backend modifications
https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/43

Issue #739, opened 2025-12-26
Tool calls not emitted after thinking blocks


edxeth:
**Describe the bug**
Claude Opus 4.5 (Thinking) doesn't return tool calls after thinking blocks when it intends to call a tool (in my case `take_snapshot` of [Chrome DevTools MCP](https://github.com/ChromeDevTools/chrome-devtools-mcp)). The model's thinking output clearly states it wants to call a tool (e.g., "Successfully navigated. Now let me take a snapshot of the page."), but the response ends with stop_reason: "end_turn" without emitting the actual function call. The upstream response only contains the thinking part with an empty text field and thought signature, but no functionCall is included even though the model decided to call one.

**CLI Type**
antigravity

**Model Name**
`claude-opus-4-5-thinking` (mapped to `gemini-claude-opus-4-5-thinking`)

**LLM Client**
OpenCode v1.0.203 using @ai-sdk/anthropic

**Request Information**
Upstream request to daily-cloudcode-pa.googleapis.com:
- Model: claude-opus-4-5-thinking
- Thinking enabled with 16000 budget tokens
- Previous turns show successful tool calls with thinking

Upstream response (final chunks):
```
data: {"response": {"candidates": [{"content": {"role": "model","parts": [{"thought": true,"text":" the page."}]}}],...}
data: {"response": {"candidates": [{"content": {"role": "model","parts": [{"thought": true,"thoughtSignature": "RXVrQkNrZ0lDeEFD...","text": ""}]}}],...}
```

Full logs: https://gist.githubusercontent.com/edxeth/52b4bf349fb35537191f07a9779ec86e/raw/36e41742d1f46b2283ce43cf072e713e690da09a/v1-messages-2025-12-26T213423-6b35333c.log

The response ends here - no functionCall part for chrome-devtools-desktop_take_snapshot despite the thinking indicating the model wanted to call it.

**Expected behavior**
After the thinking block, the response should include the functionCall part for the tool the model decided to call, similar to how previous turns in the conversation worked correctly.

**Screenshots**

<img width="1145" height="693" alt="Image" src="https://github.com/user-attachments/assets/a0f6ce0c-75fc-433f-a859-dea5ba19c735" />

**OS Type**
- OS: Ubuntu 24.04.3 LTS (WSL2)
- Architecture: x64

edxeth:
## Root Cause Identified

After extensive testing, I've identified the specific pattern that triggers this bug:

[...]

PR #846, opened 2026-01-03
fix(antigravity): apply schema cleaning to Gemini 3 models

miaachan:
## Problem

`gemini-3-pro-preview` model fails to execute tool calls in Codex, returning empty responses. The upstream Antigravity API returns `MALFORMED_FUNCTION_CALL` error:

```json
{
  "response": {
    "candidates": [{
      "content": {},
      "finishReason": "MALFORMED_FUNCTION_CALL",
      "finishMessage": "Malformed function call: call:default_api:shell{command:[\"pwd\"],workdir:\"{masked}/z/CLIProxyAPI\"}"
    }],
    "usageMetadata": {...},
    "modelVersion": "gemini-3-pro-high"
  }
}
```

Meanwhile, `gemini-claude-opus-4-5-thinking` works correctly for the same tool calls.

## Fix

Related to #739. Just extend `CleanJSONSchemaForAntigravity` to include Gemini 3 models, applying the same schema normalization that already works for Claude models. This adds required parameters to tools that lack them, preventing the malformed function call error.

Tested with `gemini-3-pro-preview` in Codex - tool calls now execute correctly. Screenshot:

SCR-20260102-rhkx-3

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -1026,7 +1026,7 @@ func (e *AntigravityExecutor) buildRequest(ctx context.Context, auth *cliproxyau
 	payload = geminiToAntigravity(modelName, payload, projectID)
 	payload, _ = sjson.SetBytes(payload, "model", alias2ModelName(modelName))
 
-	if strings.Contains(modelName, "claude") {
+	if strings.Contains(modelName, "claude") || util.IsGemini3Model(modelName) {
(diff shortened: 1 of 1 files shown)


luispater:
Sorry, this PR cannot be merged.

For native Gemini models, `parametersJsonSchema` is the preferred choice.

Google's official `Gemini CLI` also uses `parametersJsonSchema` to handle MCP, switching to parametersJson would cause various other issues.

PR #849, opened 2026-01-03
feat(api): add unified Base URL support and path normalization

waylon256yhw:
## Summary

- Enable `RemoveExtraSlash` to normalize paths with redundant slashes (e.g., `//messages` → `/messages`)
- Add root-level `/messages` routes as aliases for `/v1/messages`

## Motivation

This allows all API formats (OpenAI, Claude, Gemini) to use the same Base URL without version prefix:

| Before | After |
|--------|-------|
| Gemini: `http://host:8317` | All: `http://host:8317` |
| Claude: `http://host:8317/v1` | |
| OpenAI: `http://host:8317` | |

The `RemoveExtraSlash` option also handles edge cases where reverse proxies (e.g., Cloudflare) forward paths with redundant slashes due to trailing slash misconfiguration in client Base URLs.

## Test plan

- [x] Verify `/messages` endpoint works
- [x] Verify `//messages` is normalized to `/messages`
- [x] Verify existing `/v1/messages` still works
- [x] Verify Gemini `/v1beta/models/*` unaffected

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -49,6 +49,7 @@ const (
 	antigravityAuthType            = "antigravity"
 	refreshSkew                    = 3000 * time.Second
 	systemInstruction              = "You are Antigravity, a powerful agentic AI coding assistant designed by the Google Deepmind team working on Advanced Agentic Coding.You are pair programming with a USER to solve their coding task. The task may require creating a new codebase, modifying or debugging an existing codebase, or simply answering a question.**Absolute paths only****Proactiveness**"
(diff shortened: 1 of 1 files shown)


waylon256yhw:
Re: Gemini's suggestion to extract a helper function

Thanks for the suggestion! After consideration, I'll keep the explicit style for now:

1. Only 2 lines of duplication - the helper extraction has minimal benefit
2. The current explicit block makes it immediately clear that root-level routes are aliases for `/v1`
3. Adding a helper increases indirection and splits related route wiring across locations

Will revisit if we need to register these routes in more places in the future.

Issue #668, opened 2025-12-22
[Feature Request] Add timeout configuration


Lynricsy:
Add model provider timeout configuration option

PR #860, opened 2026-01-04
feat: add configurable request timeout for extended thinking models

Klaudioz:
## Summary
- Adds `request-timeout` configuration option (in seconds)
- Defaults to 10 minutes if not specified
- Fixes timeout errors with Claude Opus extended thinking

## Problem
When using Claude Opus 4.5 with extended thinking (`thinking.budget_tokens: 32000+`), requests can fail with:

```
context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

Extended thinking requests can take 2-10+ minutes to complete. Several executors passed `0` as the timeout to `newProxyAwareHTTPClient()`, which meant upstream requests relied solely on context cancellation and could terminate before the response completed.

## Solution
- Add `request-timeout` field to `Config`
- Pass configured timeout (or a 10-minute default) to the upstream HTTP client
- Apply consistently across HTTP-based executors

## Usage
```yaml
# config.yaml
request-timeout: 600  # 10 minutes
```

## Testing
- `go test ./...`

## Related
- Discussion #670
- Issue #668

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -65,7 +65,7 @@ request-retry: 3
 # Maximum wait time in seconds for a cooled-down credential before triggering a retry.
 max-retry-interval: 30
 
-# Maximum duration in seconds for upstream provider requests. Set to 0 to use the default (10 minutes).
+# Maximum duration in seconds for upstream provider requests. Defaults to 600s (10 minutes) if omitted or set to 0.
 request-timeout: 600
 
 # Quota exceeded behavior


kdcokenny:
+1

Klaudioz:
Any change to get it reviewed @luispater ?

luispater:
Are you sure setting a timeout at the proxy level is a wise decision?

I don't think so. Setting a timeout for LLM requests is a bad idea.

Closed.

PR #861, opened 2026-01-04
fix(auth): use backend project ID for free tier Gemini CLI OAuth users

kaan-escober:
## Problem
Google's Gemini API uses a frontend/backend project mapping system for free tier users:
- Frontend projects (e.g., `gen-lang-client-*`) are user-visible
- Backend projects (e.g., `mystical-victor-*`) host actual API access
- Only backend projects have access to preview models (`gemini-3-*`)

Previously, CLIProxyAPI ignored the backend project ID returned by Google's onboarding API and kept using the frontend ID, preventing access to preview models.

### CLI (`internal/cmd/login.go`)
- Detect free tier users (`gen-lang-client-*` projects or `FREE`/`LEGACY` tier)
- Show interactive prompt allowing users to choose frontend or backend
- Default to backend (recommended for preview model access)
- Pro users: maintain original behavior (keep frontend ID)

### Web UI (`internal/api/handlers/management/auth_files.go`)
- Detect free tier users using same logic
- Automatically use backend project ID (recommended choice)
- Pro users: maintain original behavior (keep frontend ID)

### Deduplication (`internal/cmd/login.go`)
- Add deduplication when user selects ALL projects
- Prevents redundant API calls when multiple frontend projects map to same backend
- Skips duplicate project IDs in activation loop

## Impact
- ✅ Free tier users: Can now access `gemini-3-pro-preview` and `gemini-3-flash-preview` models
- ✅ Pro users: No change in behavior (backward compatible)
- ✅ Only affects Gemini CLI OAuth (not antigravity or API key auth)

## Testing
- Tested with free tier account selecting single project
- Tested with free tier account selecting ALL projects
- Verified deduplication prevents redundant onboarding calls
- Confirmed pro user behavior unchanged

## Files Changed
- `internal/cmd/login.go`: +44 lines (interactive prompt + deduplication)
- `internal/api/handlers/management/auth_files.go`: +15 lines (auto-backend for free users)

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -2102,7 +2102,20 @@ func performGeminiCLISetup(ctx context.Context, httpClient *http.Client, storage
 			finalProjectID := projectID
 			if responseProjectID != "" {
 				if explicitProject && !strings.EqualFold(responseProjectID, projectID) {
-					log.Warnf("Gemini onboarding returned project %s instead of requested %s; keeping requested project ID.", responseProjectID, projectID)
+					// Check if this is a free user (gen-lang-client projects or free/legacy tier)
+					isFreeUser := strings.HasPrefix(projectID, "gen-lang-client-") ||
+						strings.EqualFold(tierID, "FREE") ||
+						strings.EqualFold(tierID, "LEGACY")
+
+					if isFreeUser {
+						// For free users, use backend project ID for preview model access
+						log.Infof("Gemini onboarding: frontend project %s maps to backend project %s", projectID, responseProjectID)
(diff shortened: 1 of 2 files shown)


tiltroom:
On the main branch gemini-cli oauth seems to be broken. I can login with my account in gemini-cli (pro user) but CLIPRoxyAPI auth fails with a project related error stating onboarding failed as it is unable to identify a project id.

Does this pr address this isssue in any way? I might merge it locally

kaan-escober:
yeah that might be due to google changing something on their end, this PR only provides better UX specifcally for free users. also I'll try to submit a PR for automatically creating a project and getting its projectid if no project is available.

PR #939, opened 2026-01-09
fix: Claude OAuth by prefixing tool names and merging beta headers

edlsh:
## Summary

Fixes Claude OAuth token rejection error: "This credential is only authorized for use with Claude Code and cannot be used for other API requests."

## Problem

Anthropic's OAuth tokens are restricted to Claude Code usage. The server validates requests by checking:
1. Required beta headers (specifically `claude-code-20250219`)
2. Tool names (custom tools must not conflict with reserved Claude Code tool names)

### Tool Name Prefixing
- Prefix custom tool names with `cli_` in outbound requests to avoid conflicts with reserved Claude Code tool names
- Strip `cli_` prefix from tool names in responses (both streaming and non-streaming) for client compatibility

### Beta Header Merging
- Refactor `applyClaudeHeaders` to **merge** incoming `Anthropic-Beta` headers with base betas instead of **replacing** them
- Ensures `claude-code-20250219` and `oauth-2025-04-20` are always present in outbound requests

## Related

This fix mirrors the approach in [opencode-anthropic-auth PR #10](https://github.com/anomalyco/opencode-anthropic-auth/pull/10).

## Testing

Tested with amp CLI using Claude models - requests now succeed where they previously returned 400 errors.

PR #943, opened 2026-01-09
Fix Claude OAuth tool name mapping (proxy_)

ben-vargas:
## Summary
- Prefix tool names with `proxy_` for Claude OAuth requests (tools, tool_choice, tool_use)
- Strip the prefix from streaming and non-streaming Claude responses
- Add focused unit tests for the prefix/strip helpers

## Context
This is an AI-generated fix I was using locally, based on the pi workaround for Anthropic OAuth tool name handling. If it helps, maintainers can choose between #939 and this PR (or merge the best of both).

## Testing

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -83,12 +83,14 @@ func (e *ClaudeExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, r
 	// Extract betas from body and convert to header
 	var extraBetas []string
 	extraBetas, body = extractAndRemoveBetas(body)
+	bodyForTranslation := body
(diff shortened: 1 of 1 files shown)


edlsh:
Agreed that `proxy_` is more logical nomenclature than `cli_` - it clearly indicates these are proxy-prefixed tools rather than something CLI-specific. Happy to close #939 in favor of this approach.

ben-vargas:
Pushed a follow-up fix: keep the upstream payload prefixed for OAuth, but pass the unprefixed request body into response translators to avoid leaking `proxy_` in OpenAI Responses echoed tool metadata.

Tests: `go test ./internal/runtime/executor`

PR #946, opened 2026-01-09
fix(executor): rename blocked tool names for Claude Code OAuth

uje-m:
## Summary

Fixes Claude Code OAuth authentication failures when requests contain tool names blocked by Anthropic's API validation.

## Problem

Anthropic recently started blocking certain tool names for Claude Code OAuth requests. Specifically, the lowercase tool name `"glob"` is blocked, causing requests to fail with:

> "This credential is only authorized for use with Claude Code and cannot be used for other API requests."

The validation is case-sensitive: `"glob"` fails but `"Glob"` works.

## Solution

Add bidirectional tool name transformation:
1. **Request**: Rename `"glob"` → `"Glob"` before sending to Anthropic
2. **Response**: Restore `"Glob"` → `"glob"` so clients can match their tool definitions

## Changes

- Add `renameBlockedToolNames()` function for request transformation
- Add `restoreBlockedToolNames()` function for response transformation  
- Call both in `Execute()` and `ExecuteStream()` methods

## Test plan

- [x] Verify OAuth requests with `glob` tool now succeed
- [x] Verify tool_use responses correctly restore original tool names
- [x] Verify non-blocked tools are unaffected

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -72,6 +72,9 @@ func (e *ClaudeExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, r
 	}
 	body = applyPayloadConfigWithRoot(e.cfg, model, to.String(), "", body, originalTranslated)
 
+	// Rename blocked tool names to avoid Claude Code OAuth validation errors
+	body = renameBlockedToolNames(body)
+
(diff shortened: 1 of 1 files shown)


uje-m:
Closing in favor of existing PRs #939 and #943 which take a more comprehensive approach by prefixing all tool names.

PR #947, opened 2026-01-09
Resolve memory leaks causing OOM in k8s deployment

hemanta212:
Hey there,
I am deploying the server in k8s and started getting bunch of oom kills, so investigated and cleaned up bunch of unchecked goroutines.

## Changes
- usage/logger_plugin: cap modelStats.Details at 1000 entries per model
- cache/signature_cache: add background cleanup for expired sessions (10 min)
- management/handler: add background cleanup for stale IP rate-limit entries (1 hr)
- executor/cache_helpers: add mutex protection and TTL cleanup for codexCacheMap (15 min)
- executor/codex_executor: use thread-safe cache accessors

Add reproduction tests demonstrating leak behavior before/after fixes.


luispater:
I reverted the changes to `logger_plugin.go`.

You can set `commercial-mode` to `true` to disable statistics.

PR #993, opened 2026-01-13
feat: add control switches for api provider and auth files

xxnuo:
Related: https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/55


xxnuo:
The code is not elegant, 90% of it was completed by opus, and I have tested it and it works properly.

PR #994, opened 2026-01-13
fix(antigravity): prevent corrupted thought signature when switching models

adrenjc:
## Summary

- Fix "Corrupted thought signature" error when switching from Claude models (e.g., Opus 4.5) to Gemini models (e.g., Flash) mid-conversation via Antigravity OAuth
- Remove fallback to client-provided signatures which are incompatible across different model families
- Only use cached signatures from same-session Gemini responses

## Problem

When a user starts a conversation with Claude Opus 4.5 (via Antigravity) and then switches to Gemini 3 Flash, the thinking blocks from Claude contain signatures that are invalid for Gemini API, causing 400 errors with "Corrupted thought signature" message.

## Solution

- Remove the client signature fallback logic in `antigravity_claude_request.go`
- Only trust cached signatures (populated by responses from the same Gemini session)
- Skip thinking blocks without valid cached signatures
- `tool_use` blocks continue to use `skip_thought_signature_validator` when no valid signature is available

## Test plan

- [x] Unit tests updated and passing
- [x] Manual test: Start conversation with Claude Opus 4.5, then switch to Gemini Flash mid-conversation


luispater:
@soilSpoon Would you have any free time to review this?

soilSpoon:
LGTM! The fix effectively prevents the "Corrupted thought signature" error by enforcing a stricter signature policy.

The decision to rely solely on server-cached signatures and drop client-provided ones is the right call for safety, especially given the strict validation of the Gemini API. Trusting signatures from a different model family (like Claude) across sessions is definitely risky.

@luispater I apologize for the delayed response.

PR #1116, opened 2026-01-20
Fix antigravity malformed_function_call

bexcodex:


hkfires:
This modification will cause a functional regression, preventing the sjson.Delete(template, "request.generationConfig.maxOutputTokens") for gemini-3-pro-high from being executed.

luispater (approved):

PR #1126, opened 2026-01-20
fix(antigravity): clean tool parameters schema for all models

adrenjc:
## Summary
- Fix "Invalid tool parameters" error when using Claude Code with Gemini models (gemini-3-flash, gemini-3-pro)

## Problem
When using Claude Code proxied to Gemini models, tool calls fail with "Invalid tool parameters" error during initialization.

**Root cause**: The `geminiToAntigravity` function only cleaned tool parameters schema for Claude models. Gemini models were completely skipped, but `functionCallingConfig.mode` was set to `"VALIDATED"`, causing Gemini API to strictly validate and reject schemas containing:
- `anyOf` / `oneOf` union types
- `$ref` references
- Complex `type` arrays like `["string", "null"]`
- Other unsupported JSON Schema keywords

## Solution
Extend schema cleaning to all models by:
1. Removing the `if strings.Contains(modelName, "claude")` condition
2. Processing both `parametersJsonSchema` and `parameters` fields
3. Calling `CleanJSONSchemaForAntigravity` for every tool declaration

## Before
```go
if strings.Contains(modelName, "claude") {
    // Only Claude models got schema processing
    // Gemini models skipped entirely
}
```

## After
```go
// All models get schema cleaning
gjson.Get(template, "request.tools").ForEach(func(key, tool gjson.Result) bool {
    // Process parametersJsonSchema or parameters
    cleanedSchema := util.CleanJSONSchemaForAntigravity(paramsRaw)
    // ...
})

if !strings.Contains(modelName, "claude") {
    // Only maxOutputTokens removal is Claude-specific now
}
```

## Test plan
- [x] Test Claude Code with gemini-3-flash model - tool calls should work
- [x] Test Claude Code with gemini-3-pro model - tool calls should work
- [x] Test Claude Code with Claude models - no regression

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -1407,19 +1407,39 @@ func geminiToAntigravity(modelName string, payload []byte, projectID string) []b
 	template, _ = sjson.Delete(template, "request.safetySettings")
 	template, _ = sjson.Set(template, "request.toolConfig.functionCallingConfig.mode", "VALIDATED")
 
-	if strings.Contains(modelName, "claude") {
-		gjson.Get(template, "request.tools").ForEach(func(key, tool gjson.Result) bool {
-			tool.Get("functionDeclarations").ForEach(func(funKey, funcDecl gjson.Result) bool {
-				if funcDecl.Get("parametersJsonSchema").Exists() {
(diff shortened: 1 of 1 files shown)


hkfires:
This logic has already been implemented.

PR #1137, opened 2026-01-21
fix(translator): ensure system message is only added if it contains c…

augustVino:
fix issue: https://github.com/router-for-me/CLIProxyAPI/issues/1136


augustVino:
您好，已收到您的邮件。我将尽快给您回复。

PR #1156, opened 2026-01-22
feat(thinking): add config-based reasoning level overrides

damianpdr:
## Summary

This PR adds support for config-based thinking/reasoning level overrides. Users can now force specific reasoning levels (e.g., `high` instead of `medium`) for models via `payload.override` rules in `config.yaml`.

**Problem:** When using Amp CLI with different modes (rush/smart/deep), the reasoning level is hardcoded in the request. Users couldn't override these levels through proxy configuration.

**Solution:** The thinking module now checks for model-specific overrides from `payload.override` rules after extracting the config from the request body. Config overrides take precedence over request values.

## Changes

- `internal/thinking/types.go`: Add `ModelOverride` struct and registry functions
- `internal/thinking/apply.go`: Check for config overrides after extracting request config
- `internal/watcher/config_reload.go`: Load overrides from `payload.override` rules on config reload
- `internal/watcher/watcher.go`: Load overrides on initial config set

## Example Config

```yaml
payload:
  override:
    - models:
        - name: "gpt-5.2-codex"
      params:
        "reasoning.effort": "high"
```

## Log Output

```
thinking: original config from request | provider=codex model=gpt-5.2-codex mode=level budget=0 level=medium
thinking: applying config override | provider=codex model=gpt-5.2-codex
thinking: processed config to apply | provider=codex model=gpt-5.2-codex mode=level budget=0 level=high
```

## Test Plan

- [x] Build passes
- [x] Manual testing with Amp deep mode confirms override is applied
- [ ] Unit tests for ModelOverride registry
- [ ] Integration tests for config reload

### DIFF
--- a/internal/thinking/apply.go
+++ b/internal/thinking/apply.go
@@ -156,6 +156,18 @@ func ApplyThinking(body []byte, model string, fromFormat string, toFormat string
 		}
 	}
 
+	// Check for config-based model override (takes precedence over request values)
+	if override, ok := GetModelOverride(baseModel); ok {
+		log.WithFields(log.Fields{
+			"provider":       providerFormat,
+			"model":          baseModel,
+			"original_level": config.Level,
+			"override_level": override.Level,
+		}).Debug("thinking: applying config override |")
+		config.Mode = ModeLevel
+		config.Level = override.Level
+	}
+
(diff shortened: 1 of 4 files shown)


hkfires:
This problem can be simply solved by using a thinking suffix.

PR #1175, opened 2026-01-23
fix: context cancellation check in conductor.go

a0gent:
Added context cancellation check before trying the next provider in the authentication process.


luispater:
This was fixed in c32e2a81969de67e62ae4361585db5659097206e

PR #1220, opened 2026-01-25
feat(routing): native provider priority with automatic fallback

wayfind:
## Problem

When multiple providers support the same model (e.g., native `claude` and alias-mapped `antigravity` both provide `claude-haiku-4-5`), the system had no preference mechanism. Users expected native providers to be used first, with automatic fallback to alternatives when quota is exceeded.

## Solution

Two coordinated changes that work together:

### 1. Registry: Native provider priority sorting
**File:** `internal/registry/model_registry.go`

`GetModelProviders()` now returns providers sorted by native affinity:
```go
// "claude" provider comes before "antigravity" for "claude-haiku-*" models
iMatch := strings.HasPrefix(modelLower, providers[i].nameLower)
```

### 2. Conductor: Preserve provider order in selection
**File:** `sdk/cliproxy/auth/conductor.go`

`pickNextMixed()` now iterates by provider order instead of converting to an unordered set:
```go
// Before: providers → set (order lost)
// After:  iterate by provider priority
for _, providerKey := range normalizedProviders {
    for _, candidate := range m.auths { ... }
}
```

## Why Both Changes Are Needed

| Change | Without the other |
|--------|-------------------|
| Registry sorting only | Order lost in `pickNextMixed` set conversion |
| Conductor ordering only | No native-first ordering to preserve |

## Behavior

- **Zero configuration**: Works automatically based on model name prefix matching
- **Automatic fallback**: When native provider quota exceeded, falls back to alias-mapped provider
- **Backward compatible**: No API changes, existing configs work unchanged

## Example

```yaml
oauth-model-alias:
  antigravity:
    - name: "gemini-2.5-flash"
      alias: "claude-3-5-haiku-20241022"
```

Request for `claude-3-5-haiku-20241022`:
1. Try native `claude` provider first
2. If unavailable/quota exceeded → fallback to `antigravity`

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -1561,15 +1561,21 @@ func (m *Manager) pickNext(ctx context.Context, provider, model string, opts cli
 }
 
 func (m *Manager) pickNextMixed(ctx context.Context, providers []string, model string, opts cliproxyexecutor.Options, tried map[string]struct{}) (*Auth, ProviderExecutor, string, error) {
-	providerSet := make(map[string]struct{}, len(providers))
+	// Normalize providers while preserving order for priority-based selection.
+	normalizedProviders := make([]string, 0, len(providers))
+	providerSeen := make(map[string]struct{}, len(providers))
 	for _, provider := range providers {
 		p := strings.TrimSpace(strings.ToLower(provider))
 		if p == "" {
 			continue
 		}
-		providerSet[p] = struct{}{}
+		if _, seen := providerSeen[p]; seen {
+			continue
+		}
+		providerSeen[p] = struct{}{}
+		normalizedProviders = append(normalizedProviders, p)
 	}
-	if len(providerSet) == 0 {
+	if len(normalizedProviders) == 0 {
 		return nil, nil, "", &Error{Code: "provider_not_found", Message: "no provider supplied"}
 	}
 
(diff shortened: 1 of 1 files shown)


wayfind:
Thanks for the review feedback @gemini-code-assist! 

The performance optimization has been addressed in commit `435492e`:

- Moved `strings.ToLower(modelID)` outside the sort closure
- Pre-calculated lowercase provider names in the `providerCount` struct

This avoids O(N log N) repeated string operations during sorting.

wayfind:
@chatgpt-codex-connector Great catch! You were absolutely right.

**Problem identified:** `GetModelProviders` sorting was ineffective because `pickNextMixed` converted the provider list to a `map[string]struct{}`, losing the order information.

**Fixed in commit `65b3734`:** Modified `pickNextMixed` to iterate by provider order when building candidates:

```go
// Before: providers → set (order lost) → iterate m.auths
// After:  providers → preserve order → iterate by provider priority
for _, providerKey := range normalizedProviders {
    for _, candidate := range m.auths {
        if candidate.Provider == providerKey {
            candidates = append(candidates, candidate)
        }
    }
}
```

Now native providers (e.g., "claude" for `claude-*` models) are genuinely prioritized over alias-mapped providers in actual routing, not just in logs.

👍 for the thorough analysis!

luispater:
No plan to support this.

PR #1303, opened 2026-01-29
feat(quota): add automatic quota monitoring for Antigravity accounts

hoangquockhanh:
- Add quota monitor that checks Antigravity model quotas every minute
- Automatically disable models when quota usage reaches threshold (default 80%)
- Add configurable settings in config.yaml:
  - quota-monitor.enabled: enable/disable monitoring
  - quota-monitor.check-interval-seconds: check interval
  - quota-monitor.disable-threshold-percent: usage threshold
- Integrate monitor lifecycle with service startup/shutdown

### DIFF
--- a/internal/quota/monitor.go
+++ b/internal/quota/monitor.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


luispater:
No plan for supporting this.

PR #1308, opened 2026-01-29
fix(antigravity): resolve model aliases to support gemini-3-pro-preview

benzntech:
## Description
Fixes an issue where `AntigravityExecutor` failed to resolve model aliases (e.g., `gemini-3-pro-preview` -> `gemini-3-pro-high`) before executing requests.

## Analysis Findings
- **The Issue**: `AntigravityExecutor.Execute` receives `gemini-3-pro-preview` (the aliased name used by clients) but tries to use it directly against the upstream.
- **Root Cause**: The `buildRequest` logic explicitly checks for `gemini-3-pro-high` (the canonical upstream name) to apply specific schema cleaning and URL construction. Because the alias wasn't resolved, this logic was bypassed or the upstream returned 404/400 due to an unrecognized model name in the URL.
- **Evidence**: Logs showed the incoming request model as `gemini-3-pro-preview` persisting into the execution path, while `applyOAuthModelAlias` (which handles this mapping) was only being used in the model listing service, not the execution path.

## The Fix
- Implemented `resolveModelAlias` helper method in `internal/runtime/executor/antigravity_executor.go`.
- Updated `Execute` and `ExecuteStream` to call `resolveModelAlias` at the start of execution.
- This ensures the request model is updated to the canonical upstream name (e.g. `gemini-3-pro-high`) *before* it hits the model-specific logic in `buildRequest` or `executeClaudeNonStream`.

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -109,6 +109,7 @@ func (e *AntigravityExecutor) HttpRequest(ctx context.Context, auth *cliproxyaut
 
 // Execute performs a non-streaming request to the Antigravity API.
 func (e *AntigravityExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, req cliproxyexecutor.Request, opts cliproxyexecutor.Options) (resp cliproxyexecutor.Response, err error) {
+	req.Model = e.resolveModelAlias(req.Model)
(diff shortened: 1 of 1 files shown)


hkfires:
This problem does not exist; what enters the executor is already the upstream model name, so there is no scenario where an alias enters the executor.

PR #1323, opened 2026-01-30
fix(config): enable gemini-3-pro-preview by removing forced alias

benzntech:
## Description
This PR fixes an issue where  was unavailable because it was being forcibly aliased to  in the configuration migration logic.

## Changes
- Removed `"gemini-3-pro-preview": "gemini-3-pro-high"` mapping from `antigravityModelConversionTable` in `internal/config/oauth_model_alias_migration.go`.
- Removed the corresponding default alias from `defaultAntigravityAliases`.
- Updated unit tests in `internal/config/oauth_model_alias_migration_test.go` to assert that `gemini-3-pro-preview` is preserved and not aliased.

## Impact
This change allows users to explicitly select and use the `gemini-3-pro-preview` model via the CLI.


hkfires:
This is migration code; it will only run once at startup and does not need to be modified.

PR #1380, opened 2026-02-01
refactor(codex): remove codex instructions injection support

hkfires:


ain3sh:
Sry if this has been answered already, but why is this being removed? @luispater

hkfires:
> Sry if this has been answered already, but why is this being removed? @luispater

No more needed.

PR #1421, opened 2026-02-03
feat(fallback): add model fallback support for automatic failover

izzzzzi:
When a primary model fails with specific HTTP error codes (429, 502, 503, 529) or timeouts, the proxy can now automatically retry with a configured fallback model.

Configuration example:
  model-fallback:
    enabled: true
    max-fallback-attempts: 1
    mappings:
      - from: "claude-opus-4-5" to: "gpt-5.2" on-status-codes: [429, 503] on-timeout: true

Features:
- Wildcard pattern matching (e.g., "claude-*", "*-preview")
- Regex pattern support with `regex: true`
- Configurable trigger status codes per mapping
- Optional timeout-based fallback
- Limit on fallback chain depth to prevent loops
- Detailed logging when fallback is triggered

This enables seamless failover between different model providers, improving reliability when primary models hit rate limits or become temporarily unavailable.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -79,6 +79,25 @@ quota-exceeded:
   switch-project: true # Whether to automatically switch to another project when a quota is exceeded
   switch-preview-model: true # Whether to automatically switch to a preview model when a quota is exceeded
 
+# Model fallback configuration
+# When a model fails with specific error codes (429, 503, etc.), automatically retry with a fallback model.
+# model-fallback:
+#   enabled: true
+#   max-fallback-attempts: 1  # How many fallback models to try (default: 1)
+#   mappings:
+#     - from: "claude-opus-4-5"           # Primary model (supports wildcards: "claude-*", "*-preview")
+#       to: "gpt-5.2"                     # Fallback model to use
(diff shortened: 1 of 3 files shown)


luispater:
Different models vary in performance and features, mixing them often triggers unpredictable problems, like thought signatures and cache issues.

izzzzzi:
@luispater 

That's exactly the point — this is not random model mixing, it's an explicit configuration that the user sets up themselves. Look at the example:

```
  mappings:
    - from: "claude-opus-4-5"
      to: "gpt-5.2"
      on-status-codes: [429, 503]
```

  The user decides:
  - which model falls back to which
  - under what conditions (specific HTTP codes, timeouts)
  - whether to enable this feature at all

  This is designed for uninterrupted operation. When you have a production system and your primary provider returns 429/503 — it's better to get a response from a backup model than to show the user an error.

  As for "thought signatures and cache issues" — that's the responsibility of whoever configures the fallback. If someone consciously sets up claude → gpt as a backup, they understand what they're doing. The proxy shouldn't decide for
  users which fallbacks are "correct" and which aren't — it's a tool, and the configuration is entirely in the admin's hands.

PR #1440, opened 2026-02-05
feat(registry): register Claude 4.6 static data

kvokka:
Add model definition for Claude 4.6 Opus with 200k context length and thinking support capabilities.

I doubt that a 1M token window will be allowed for Antigravity users from day 1

### DIFF
--- a/internal/config/oauth_model_alias_migration.go
+++ b/internal/config/oauth_model_alias_migration.go
@@ -17,6 +17,7 @@ var antigravityModelConversionTable = map[string]string{
 	"gemini-claude-sonnet-4-5":                "claude-sonnet-4-5",
 	"gemini-claude-sonnet-4-5-thinking":       "claude-sonnet-4-5-thinking",
 	"gemini-claude-opus-4-5-thinking":         "claude-opus-4-5-thinking",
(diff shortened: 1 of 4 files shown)


luispater (approved):

jsj:
Doesn't work for use via Anthorpic it's just `claude-opus-4-6` from Anthropic

kvokka:
It's not added to Antigravity yet. The naming that you're giving is for CC, for Antigravity, it's another model. And yes, it's not in the Antigravity app yet @jsj

jsj:
Yes I'd expect more people use it from CC than Antigravity, and its live on CC - both should work

kvokka:
Then feel free to add it for CC. I'm not using CC and need it only for Antigravity.

jsj:
@luispater I have an diff locally, but it just drops the `-20260205"` to `claude-opus-4-6"` and is working in my Vibeproxy setup

kvokka:
It's another provider. The suffix is for Google-provisioned models, while Anthropic uses it without it

kvokka:
@jst it's a problem of the VibeProxy project, since here everything is consistent and sufficient, which means that VibeProxy hardcoded the full model name mapping

I'm not using it (see another pr above), so pls open an issue there

PR #1460, opened 2026-02-06
fix(auth): return HTTP 429 instead of 500 for auth_unavailable error

j2h4u:
## Summary

When all credentials are unavailable but not all are in cooldown, `getAvailableAuths()` in `selector.go` returns an `auth_unavailable` error **without** setting `HTTPStatus`. The handler then defaults to HTTP 500.

This one-line fix sets `HTTPStatus: http.StatusTooManyRequests` (429) on the `auth_unavailable` error, aligning it with the existing `modelCooldownError` behavior.

**Before:** clients receive `500 Internal Server Error` → interpret as server failure, stop retrying  
**After:** clients receive `429 Too Many Requests` → trigger standard failover/retry logic

## Context

The `model_cooldown` path (line 158) already returns 429 correctly. The `auth_unavailable` path (line 160) represents the same semantic condition — no credentials available — but returns 500 due to missing `HTTPStatus`. This inconsistency breaks downstream clients that distinguish 429 (rate limit, try elsewhere) from 500 (server broken).

The gap occurs when `cooldownCount < len(auths)` — e.g. some credentials are disabled, some have `disable_cooling` set, or cooldown hasn't been scheduled yet at check time.

## Test plan

- [ ] Exhaust all credentials for a model → verify HTTP 429 returned (was 500)
- [ ] Verify `model_cooldown` path still returns 429 with `Retry-After` header
- [ ] Verify normal requests route correctly when credentials are available

—Calculon, Actor Extraordinaire (feat. Opus 4.6)

### DIFF
--- a/sdk/cliproxy/auth/selector_test.go
+++ b/sdk/cliproxy/auth/selector_test.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"context"
 	"errors"
+	"net/http"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


luispater:
No plan to change this.

PR #1473, opened 2026-02-07
fix: migrate claude-opus-4-5 to 4-6 aliases & strip thinking blocks from non-thinking responses

Finesssee:
### 1. Model Alias Migration (opus-4-5 → opus-4-6)

Updates all model references, aliases, and tests from `claude-opus-4-5` to `claude-opus-4-6` to reflect the current Antigravity model availability.

**Files:**
- `README.md` — example model name update
- `config.example.yaml` — updated example aliases and model mappings
- `internal/config/config.go` — doc comment update
- `internal/config/oauth_model_alias_migration.go` — consolidated `gemini-claude-opus-4-5-thinking` → `claude-opus-4-6-thinking` mapping, removed duplicate entry
- `internal/config/oauth_model_alias_migration_test.go` — removed stale opus-4-5 assertion
- `internal/config/oauth_model_alias_test.go` — updated test data to use opus-4-6
- `internal/api/modules/amp/model_mapping_test.go` — updated test data to use opus-4-6
- `internal/thinking/provider/claude/apply.go` — doc comment update
- `internal/util/claude_model_test.go` — updated test data to use opus-4-6

### 2. Fix Thinking Output Leaking

When using Antigravity Claude models (e.g. `claude-opus-4-6-thinking`), the upstream always returns thinking blocks. The response translators were blindly converting all `thought: true` parts into Claude `thinking` content blocks, causing internal reasoning to leak into the client-visible output when thinking was not requested.

Both streaming (`ConvertAntigravityResponseToClaude`) and non-streaming (`ConvertAntigravityResponseToClaudeNonStream`) now check `originalRequestRawJSON` for `thinking.type == 'enabled'` and skip thinking blocks entirely when thinking was not requested.

**Files:**
- `internal/translator/antigravity/claude/antigravity_claude_response.go` — added `thinkingRequested` check in both response paths
- `internal/translator/antigravity/claude/antigravity_claude_response_test.go` — updated 4 tests to include thinking config

## Testing

All 19 tests in `internal/translator/antigravity/claude/` pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ CLIProxyAPI includes integrated support for [Amp CLI](https://ampcode.com) and A
 - Provider route aliases for Amp's API patterns (`/api/provider/{provider}/v1...`)
 - Management proxy for OAuth authentication and account features
 - Smart model fallback with automatic routing
-- **Model mapping** to route unavailable models to alternatives (e.g., `claude-opus-4.5` → `claude-sonnet-4`)
+- **Model mapping** to route unavailable models to alternatives (e.g., `claude-opus-4.6` → `claude-sonnet-4`)
 - Security-first design with localhost-only management endpoints
 
 **→ [Complete Amp CLI Integration Guide](https://help.router-for.me/agent-client/amp-cli.html)**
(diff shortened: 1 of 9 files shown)


hkfires:
Thanks for the submission. We are closing PR #1473.
1. `oauth_model_alias_migration` is currently not invoked in runtime. The startup migration call is intentionally disabled in `internal/config/config.go`, and current references are effectively test-only.
2. The 4.5/4.6 alias updates are mostly example/documentation changes. They do not affect active execution paths or runtime behavior in the current code path.

PR #1529, opened 2026-02-10
fix: handle plain string content in OpenAI Responses → Gemini translation

happycastle114:
## Summary

Fix OpenAI Responses → Gemini request translation when message `content` is sent as a plain string (common in OpenAI SDKs, including Python), not just as an array.

## Bug

`ConvertOpenAIResponsesRequestToGemini` only processed message content when `content` was an array.
For inputs like:

```json
{"role":"system","content":"You are helpful"}
{"role":"user","content":"Hello"}
```

content was skipped, producing `contents: []` in the Gemini payload and causing:

```
at least one contents field is required
```

### 1) System message handling
- Added support for string `content` in system messages (in addition to array handling).
- When `content` is a string, it now populates `system_instruction.parts[0].text`.

### 2) Regular message handling
- Added non-array `content` handling before existing array logic.
- When `content` is a string, it now creates a Gemini `contents` item directly with mapped role and a text part.

### 3) Tests
- Added `gemini_openai-responses_request_test.go` with coverage for both formats:
  - string `content` (`system` + `user`)
  - array `content` (existing structured format)

## Reproduction

1. Send OpenAI Responses request with plain string message content:
   ```json
   {
     "model": "gpt-5",
     "input": [
       {"role": "system", "content": "You are helpful"},
       {"role": "user", "content": "Hello"}
     ]
   }
   ```
2. Before this fix: translated Gemini request has empty `contents` and request fails.
3. After this fix: translated Gemini request includes `system_instruction` and `contents` with user text, and succeeds.

## Notes

- Followed existing style and preserved tab-indented Go formatting.

### DIFF
--- a/internal/translator/gemini/openai/responses/gemini_openai-responses_request.go
+++ b/internal/translator/gemini/openai/responses/gemini_openai-responses_request.go
@@ -117,20 +117,28 @@ func ConvertOpenAIResponsesRequestToGemini(modelName string, inputRawJSON []byte
 			switch itemType {
 			case "message":
 				if strings.EqualFold(itemRole, "system") {
-					if contentArray := item.Get("content"); contentArray.Exists() && contentArray.IsArray() {
-						var builder strings.Builder
-						contentArray.ForEach(func(_, contentItem gjson.Result) bool {
-							text := contentItem.Get("text").String()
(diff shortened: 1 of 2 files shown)


luispater:
I fixed the issue in v6.8.11.

This PR has been closed.

PR #1537, opened 2026-02-11
feat: per-account excluded_models & priority support for OAuth auth files

RGBadmin:
# PR: Per-Account Excluded Models & Priority Support for OAuth Auth Files

## Summary

This PR adds two features to OAuth auth files:

1. **Per-Account Excluded Models** — Each OAuth JSON file can now specify its own `excluded_models` list, which is merged with the global provider-level `oauth-excluded-models` config.

2. **Per-Account Priority** — Each OAuth JSON file can specify a `priority` field (integer). Higher priority auths are selected first; on failure/cooldown, lower priority auths are used as fallback. Same-priority auths use round-robin.

### Excluded Models
Add an `excluded_models` array to your OAuth JSON file:
```json
{
  "access_token": "...",
  "excluded_models": ["gpt-4o", "claude-3-opus"]
}
```
Both `excluded_models` and `excluded-models` key formats are supported.

These per-account exclusions are **merged** with the global `oauth-excluded-models` config — not replaced.

### Priority
Add a `priority` field to your OAuth JSON file:
```json
{
  "access_token": "...",
  "priority": 10
}
```
- Higher number = higher priority (default: 0)
- Highest priority auths are always tried first
- On failure/cooldown, automatically falls back to next priority tier
- Same priority level uses round-robin (existing behavior)

## Changed Files

| File | Changes |
|------|---------|
| `internal/watcher/synthesizer/file.go` | Parse `excluded_models` and `priority` from OAuth JSON; propagate priority to virtual auths; new `extractExcludedModelsFromMetadata()` function |
| `internal/watcher/synthesizer/helpers.go` | Merge per-account excluded models with global provider-level exclusions in `ApplyAuthExcludedModelsMeta()`; store combined list in auth attributes |
| `sdk/cliproxy/service.go` | Read per-account excluded models from auth attributes at routing time and merge with config-level exclusions |

## Backward Compatibility

- Fully backward compatible — existing auth files without these fields work unchanged
- `priority` defaults to 0 if not specified
- `excluded_models` defaults to empty if not specified (only global config applies)

### DIFF
--- a/internal/watcher/synthesizer/file.go
+++ b/internal/watcher/synthesizer/file.go
@@ -273,26 +273,25 @@ func extractExcludedModelsFromMetadata(metadata map[string]any) []string {
 	if !ok || raw == nil {
 		return nil
 	}
+	var stringSlice []string
 	switch v := raw.(type) {
 	case []string:
-		result := make([]string, 0, len(v))
-		for _, s := range v {
-			if trimmed := strings.TrimSpace(s); trimmed != "" {
-				result = append(result, trimmed)
-			}
-		}
-		return result
+		stringSlice = v
 	case []interface{}:
-		result := make([]string, 0, len(v))
+		stringSlice = make([]string, 0, len(v))
 		for _, item := range v {
 			if s, ok := item.(string); ok {
-				if trimmed := strings.TrimSpace(s); trimmed != "" {
-					result = append(result, trimmed)
-				}
+				stringSlice = append(stringSlice, s)
 			}
 		}
-		return result
 	default:
 		return nil
 	}
+	result := make([]string, 0, len(stringSlice))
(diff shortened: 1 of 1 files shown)


RGBadmin:
Thanks for the review! Both suggestions have been addressed in the follow-up commits:

1. **service.go** — Simplified the excluded_models merge logic. Since the synthesizer already pre-merges per-account and global exclusions, routing now reads the attribute directly instead of re-merging. (commit: `refactor: simplify per-account excluded_models merge in routing`)

2. **file.go** — Refactored `extractExcludedModelsFromMetadata` to extract a `[]string` first, then apply a single trim+filter pass, reducing code duplication. (commit: `refactor: reduce code duplication in extractExcludedModelsFromMetadata`)

PR #1610, opened 2026-02-16
feat(stats): persist across restarts with periodic/shutdown flush

masrurimz:
## Summary
- add file-based persistence for usage statistics via a `UsageStore` abstraction and `FileUsageStore`
- persist usage snapshot to `usage.json` using atomic write pattern (`tmp + fsync + rename + dir sync`)
- load persisted snapshot on startup and merge into in-memory usage stats
- add periodic flush (default 60s) and final graceful-shutdown flush
- add unit tests for usage persistence behavior (roundtrip, corrupted input handling, context cancellation, periodic flush trigger)
- fix shutdown ordering race by stopping periodic flusher before final shutdown flush

## Why this approach
This PR intentionally takes a **minimal-risk** path to ship persistence quickly without changing the existing usage-statistics data model consumed by management endpoints.

- keeps current snapshot/merge behavior intact
- adds durable storage and lifecycle wiring around existing in-memory stats
- avoids broad API/aggregation refactors in the same change set

## Notes on future scope
The current structure still includes per-request `details` in usage snapshots. If we later want strict "stats-only" persistence (dashboard aggregates without request-level detail), that requires a **larger follow-up refactor**:

- redesign snapshot schema toward aggregate-only fields
- adjust merge semantics for aggregate maps/counters
- potentially update management/API response shaping and frontend assumptions

That work is intentionally out-of-scope here to keep this PR focused and low-risk.

## Validation
- `go test ./internal/usage ./cmd/server`
- full `go test ./...` has unrelated pre-existing failures in `internal/api/modules/amp`

### DIFF
--- a/internal/usage/store.go
+++ b/internal/usage/store.go
@@ -103,10 +103,15 @@ func (s *FileUsageStore) Save(ctx context.Context, snapshot StatisticsSnapshot) 
 		return err
 	}
 
-	tmpPath := fmt.Sprintf("%s.tmp.%d", s.path, time.Now().UnixNano())
-	file, err := os.OpenFile(tmpPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
+	file, err := os.CreateTemp(dir, filepath.Base(s.path)+".tmp.*")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


luispater:
We have no plans to support this.

Issue #1657, opened 2026-02-20
logs-max-total-size-mb does not account for per-day subdirectories


darrenyao:
## Bug Description

The `logs-max-total-size-mb` configuration option does not enforce the size limit on request log files stored in per-day subdirectories (e.g., `logs/2026-02-14/`). Only files directly under the `logs/` root directory (e.g., `main.log`) are counted toward the total size.

## Root Cause

In `internal/logging/log_dir_cleaner.go`, the `enforceLogDirSizeLimit` function skips directories entirely:

```go
// Line ~117
for _, entry := range entries {
    if entry.IsDir() {
        continue   // ← skips all subdirectories
    }
```

When `request-log: true` and `logging-to-file: true` are enabled, request logs are written into date-based subdirectories like:

```
logs/
├── main.log                         (9 MB — the only file counted)
├── 2026-02-14/                      (338 MB — skipped)
│   ├── v1-messages-*.log
│   └── v1-chat-completions-*.log
├── 2026-02-20/                      (147 MB — skipped)
└── ...
```

The cleaner sees only `main.log` (~9 MB), which is always below the limit, so cleanup never triggers — even when the actual total exceeds the configured limit by 2-5x.

## Steps to Reproduce

1. Set `logging-to-file: true`, `request-log: true`, `logs-max-total-size-mb: 300` in `config.yaml`
2. Generate enough API traffic to exceed 300 MB of request logs
3. Observe that log subdirectories grow without bound; the cleaner never removes old files

## Expected Behavior

The cleaner should recursively walk subdirectories (or at minimum, delete the oldest date-based subdirectories) to keep the total `logs/` directory size within the configured limit.

## Suggested Fix

In `enforceLogDirSizeLimit`, either:
- Use `filepath.WalkDir` to recursively collect all `.log` files (including those in subdirectories), or
- Treat date-named subdirectories as units: sum their sizes and delete the oldest directories first

## Environment

- Config: `logs-max-total-size-mb: 300`
- Feature introduced in commit `3e4858a` (feat: add log file size limit configuration #535)
- Disk filled to 100%, causing nginx to return `500 Internal Server Error` (`No space left on device`)

luispater:
I'm sorry, I don't understand.

All logs are always in the `logs` directory. CLIProxyAPI doesn't create subdirectories anymore.

If you create some subdirectories to store logs, you should clean them yourself.

PR #1679, opened 2026-02-23
fix(logging): include nested logs in logs-max-total-size-mb cleanup

githubbzxs:
## Summary

Fix `logs-max-total-size-mb` cleanup to include log files in nested per-day subdirectories under `logs/`.

Previously, `enforceLogDirSizeLimit` only scanned files directly under the root log directory, so request logs in paths like `logs/2026-02-14/*.log` were ignored and disk usage could grow without bound.

## What changed

- `internal/logging/log_dir_cleaner.go`
  - Switched from `os.ReadDir(root)` to recursive `filepath.WalkDir(root)`
  - Collects regular `.log` / `.log.gz` files from all nested directories
  - Keeps existing deletion policy (oldest first), and still respects `protectedPath`

- `internal/logging/log_dir_cleaner_test.go`
  - Added `TestEnforceLogDirSizeLimitIncludesNestedLogs`
  - Verifies nested request logs are counted and deleted when total size exceeds limit

## Why

This addresses #1657:
`logs-max-total-size-mb does not account for per-day subdirectories`.

## Behavior impact

- No config changes required
- Existing users get correct total-size enforcement across the entire `logs/` tree
- Root-level behavior remains unchanged

### DIFF
--- a/internal/logging/log_dir_cleaner.go
+++ b/internal/logging/log_dir_cleaner.go
@@ -107,7 +107,7 @@ func enforceLogDirSizeLimit(logDir string, maxBytes int64, protectedPath string)
 
 	errWalk := filepath.WalkDir(dir, func(path string, entry os.DirEntry, err error) error {
 		if err != nil {
-			return nil
+			return err
 		}
 		if entry.IsDir() {
 			return nil


githubbzxs:
Hi! This PR is ready for review. It makes `logs-max-total-size-mb` cleanup scan nested per-day subdirectories under `logs/` (recursive walk) and adds a test to cover the nested-path case.

githubbzxs:
Thanks for the review! Good catch on the WalkDir error handling — I agree we shouldn’t silently ignore traversal errors.

I can update the callback to return `err` (or accumulate and return a sentinel error) so unreadable directories fail loudly, matching the previous behavior more closely. Want strict-fail on any WalkDir error, or best-effort with a warning log?

githubbzxs:
Pushed a fix to address the WalkDir error-handling concern (don’t swallow callback errors; return `err` so traversal failures surface): https://github.com/githubbzxs/CLIProxyAPI/commit/02936ba1ecb184f60a02c1561e4f137b2c7e1b3d

PTAL when you get a chance.

githubbzxs:
Friendly ping — I’ve pushed an update addressing the WalkDir error-handling concern (return `err` instead of swallowing). PR should be ready for another look when you have time.

luispater:
I'm sorry, I don't understand why we need this feature.

All logs are always in the `logs` directory. CLIProxyAPI doesn't create subdirectories anymore.

PR #1690, opened 2026-02-23
feat(sdk): add D2 guard pattern to prevent custom antigravity executor overwrite

professional-ALFIE:
## Summary
- Add D2 guard pattern in `sdk/cliproxy/service.go` to prevent custom antigravity executor from being overwritten by default values during initialization.

Ultraworked with [Sisyphus](https://github.com/code-yeongyu/oh-my-opencode)


professional-ALFIE:
로컬 패치 전용 — upstream에 PR하면 안 됨. 실수로 생성된 PR입니다.

PR #1731, opened 2026-02-27
fix: resolve desktop updater failures on GUI launch

alanbulan:
## Summary
- Remove `hasDesktopToolchain()` gate: `wails` is not in system PATH when app is launched as GUI (double-click), only `git` is needed for update check; `go`/`wails` are resolved by `upgrade-desktop.ps1` at runtime via `GOPATH/bin`
- Fix `MessageDialog` result comparison: Wails `QuestionDialog` on Windows returns `"Yes"`/`"No"`, not custom button labels like `"Upgrade Now"`
- Add 5s startup delay so Wails window is fully ready before showing update dialog
- Add `updaterLog()` writing to `%TEMP%/cliproxyapi_updater.log` for GUI diagnostics (no console available)
- Remove dirty working-tree guard from `upgrade-desktop.ps1`: force `git reset --hard origin/main` + `git clean -fd` for seamless upgrades
- Add UTF-8 console encoding (`chcp 65001`) to both PowerShell scripts to fix garbled Chinese output

## Test plan
- [x] App launches, dialog appears after ~5s when behind origin/main
- [x] Clicking "Yes" triggers upgrade console window
- [x] Upgrade script runs to completion (force-sync to origin/main)
- [x] After upgrade, updater fixes are preserved (now in origin/main)


luispater:
No plans to support this.

PR #1761, opened 2026-02-28
fix: support thinking.type=auto from Amp client and decouple thinking translation from unsigned history

margbug01:
## Problem

When using Antigravity Claude models through CLIProxyAPI, the thinking chain (reasoning content) does not display in the Amp client.

### 1. Unhandled `thinking.type="auto"` (Primary)

The Amp client sends `thinking: {"type": "auto"}` in its requests, but `ConvertClaudeRequestToAntigravity` only handled `"enabled"` and `"adaptive"` types in its switch statement. The `"auto"` type was silently ignored, resulting in no `thinkingConfig` being set in the translated Gemini request. Without `thinkingConfig`, the Antigravity API returns responses without any thinking content.

Additionally, the Antigravity API for Claude models does not support `thinkingBudget: -1` (auto mode sentinel). It requires a concrete positive budget value. The fix uses `128000` as the budget for "auto" mode, which `ApplyThinking` will then normalize to stay within the model's actual limits (e.g., capped to `maxOutputTokens - 1`).

### 2. `enableThoughtTranslate` kill switch (Secondary)

The `enableThoughtTranslate` variable acted as a global kill switch. When processing conversation history, encountering any unsigned thinking block (e.g., after server restart when the signature cache is empty) set this flag to `false`, which prevented the top-level `thinking` configuration from being translated entirely -- even for the current request.

### `internal/translator/antigravity/claude/antigravity_claude_request.go`

1. **Add `"auto"` case** to the thinking type switch statement. Sets `thinkingBudget: 128000` and `includeThoughts: true`. The budget is subsequently normalized by `ApplyThinking` based on model-specific limits.

2. **Add `"auto"` to `hasThinking` check** so that interleaved thinking hints are injected for tool-use scenarios when Amp sends `thinking.type="auto"`.

3. **Remove `enableThoughtTranslate` variable** -- unsigned thinking blocks in history are still correctly dropped (via `continue`), but the current request's thinking config is always translated.

### `internal/registry/model_definitions_static_data.go`

4. **Add Thinking configuration** for `claude-sonnet-4-6`, `claude-sonnet-4-5`, and `claude-opus-4-6` in `GetAntigravityModelConfig()` -- these were previously missing, causing `ApplyThinking` to skip thinking config entirely.

## Testing

- Deployed to Railway test instance (`cpa-thinking-test`)
- Verified via debug logging that:
  - Amp sends `thinking: {"type": "auto"}`
  - CPA now translates this to `thinkingConfig: {thinkingBudget: 128000, includeThoughts: true}`
  - `ApplyThinking` normalizes the budget to model-specific limits
  - Antigravity API receives the correct `thinkingConfig`


internal/translator/antigravity/claude/antigravity_claude_request.go
```diff
@@ -35,7 +35,6 @@ import (
 // Returns:
 //   - []byte: The transformed request data in Gemini CLI API format
 func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ bool) []byte {
-	enableThoughtTranslate := true
```
luispater, line 38: You MUST keep the `enableThoughtTranslate` logic.

luispater (changes requested):

luispater (approved):

PR #1786, opened 2026-03-02
fix(codex): return proper error instead of empty response on context overflow

SethBurkart123:
Currently, when a client requests a chat completion from a codex model which exceeds the context of the model, an empty response is given which messes with a large number of clients such as [Droid](https://factory.ai/product/ide). These changes pass and emit the error which fully fixes long conversations for many third party cli tools.

If you have any issues please reach out I'll be happy to fix it up (I'd love if this issue would be fully fixed).


SethBurkart123:
bump

luispater (changes requested):
Thanks for fixing the empty-success response path here — surfacing terminal SSE failures is the right direction. I found one blocking gap before I’d approve: the new SSE error path drops retry-after semantics for quota/usage-limit events. The existing HTTP error path preserves that via parseCodexRetryAfter / newCodexStatusErr, and the auth conductor uses RetryAfter() to mark credentials unavailable until the real reset time. With this patch, usage_limit_reached delivered over SSE falls back to generic quota backoff instead of the upstream reset window. Please preserve or compute retryAfter in parseCodexSSEError and add a test covering a usage_limit_reached payload with resets_at / resets_in_seconds.

SethBurkart123:
everything is fixed.

luispater (approved):
Summary: The follow-up closes the original blocker. `parseCodexSSEError` now preserves `retry-after` for `usage_limit_reached` SSE errors, and the added unit tests cover both `resets_at` and `resets_in_seconds` cases.

Blocking findings:
- None.

Non-blocking findings:
- The new coverage validates the parser helper directly, but I did not find an end-to-end test for the full `Execute`/`ExecuteStream` error-return path.

Test plan:
- `go test ./internal/runtime/executor -run 'TestParseCodexSSEError|TestParseCodexRetryAfter' -count=1`

luispater:

PR #1788, opened 2026-03-02
feat(claude): add remove-tools-cache-control config

whtiehack:
## Summary
- add `remove-tools-cache-control` to `claude-api-key` entries
- propagate the setting via synthesized auth attributes as `remove_tools_cache_control=true`
- in Claude execute/stream flows, when enabled, strip `tools[].cache_control`
- when stripping leaves no cache breakpoints, inject cache control for system/messages only (no tools injection)
- update `config.example.yaml`
- add tests for synthesizer attribute propagation and executor helper behavior

## Testing
- `go test ./internal/runtime/executor ./internal/watcher/synthesizer`
- `go test ./...` currently fails due to an unrelated pre-existing test in `sdk/cliproxy/auth`:
  - `TestManager_MaxRetryCredentials_LimitsCrossCredentialRetries`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -179,6 +179,7 @@ nonstream-keepalive-interval: 0
 #       cache-user-id: true          # optional: default is false; set true to reuse cached user_id per API key instead of generating a random one each request
 #     experimental-cch-signing: false # optional: default is false; when true, sign the final /v1/messages body using the current Claude Code cch algorithm
 #                                     # keep this disabled unless you explicitly need the behavior, so upstream seed changes fall back to legacy proxy behavior
+#     remove-tools-cache-control: false # optional: remove tools.cache_control for upstreams that don't support it (e.g., AWS Bedrock proxies)
 
 # Default headers for Claude API requests. Update when Claude Code releases new versions.
 # In legacy mode, user-agent/package-version/runtime-version/timeout are used as fallbacks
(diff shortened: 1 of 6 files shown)


luispater (changes requested):
Thanks for the PR. I’m requesting changes mainly around one correctness gap: the new remove-tools-cache-control behavior is applied in Execute/ExecuteStream, but not in CountTokens. A Claude provider configured for an upstream that rejects tools[].cache_control can still fail on /v1/messages/count_tokens because that path forwards tool cache_control unchanged. Could we apply the same stripping logic in CountTokens and add a regression test for a count_tokens request with remove_tools_cache_control=true? Once that is addressed, this looks good to me.

whtiehack:
Done. Applied the same stripping logic in `CountTokens` and added a regression test (`TestClaudeExecutor_CountTokens_RemoveToolsCacheControl`). See 8c4ac39.

luispater (approved):
Summary: The follow-up closes the gap from the previous review. `remove_tools_cache_control` now applies consistently in `Execute`, `ExecuteStream`, and `CountTokens`, and the new regression test covers the `CountTokens` path that was previously missing.

Blocking findings:
- None.

Non-blocking findings:
- None.

Test plan:
- The PR includes focused executor and synthesizer tests, and the GitHub checks are passing.

luispater:

PR #1796, opened 2026-03-02
fix: keep /v1/responses encrypted continuation on same auth

### LINKED ISSUE #1793: `400 invalid_encrypted_content on /v1/responses after provider switch (OpenAI -> Azure)`
[ ] No, it's another issue.

**Describe the bug**  
When using `/v1/responses` with multi-provider fallback/switching, the request fails with HTTP 400 `invalid_encrypted_content` after switching upstream from OpenAI to Azure OpenAI.  
It appears that `reasoning.encrypted_content` generated under one provider is forwarded to another provider, where it cannot be verified/decrypted.

**CLI Type**  
opencode/codex same issue.

**Model Name**  
gpt-5.3-codex

**LLM Client**  
codex

**Request Information**  
Reproducible request shape (simplified):

```json
{
  "method": "POST",
  "url": "/v1/responses",
  "headers": {
    "Content-Type": "application/json"
  },
  "body": {
    "model": "gpt-5.3-codex",
    "input": [
      {
        "role": "developer",
        "content": "..."
      },
      {
        "role": "user",
        "content": [
          {
            "type": "input_text",
            "text": "Continue if you have next steps, or stop and ask for clarification if you are unsure how to proceed."
          }
        ]
      },
      {
        "type": "reasoning",
        "encrypted_content": "gAAA...6W4="
      }
    ]
  }
}
```

Observed error:

```json
{
  "error": {
    "message": "The encrypted content gAAA...6W4= could not be verified. Reason: Encrypted content could not be decrypted or parsed.",
    "param": null,
    "code": "invalid_encrypted_content"
  }
}
```

Typical response headers:
- Status: 400
- Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
- Access-Control-Allow-Headers: *
- Content-Type: application/json
- Access-Control-Allow-Origin: *

**Expected behavior**  
When provider switching occurs (e.g., OpenAI -> Azure OpenAI), the gateway should not forward incompatible `reasoning.encrypted_content` across providers.  
Expected outcomes:
- either automatically strip `reasoning.encrypted_content` on provider switch,  
- or return a clear recoverable message instructing client to reset continuation context.

**Screenshots**  
N/A

**OS Type**  
- OS: Docker  
- Version: v6.8.38

**Additional context**  
This happens during provider failover/rotation scenarios (e.g., OpenAI exhausted -> switch to Azure OpenAI).  
The same conversation continuation may succeed before switching, then fail immediately after switching with `invalid_encrypted_content`.

1saifj:
Fixes #1793.

## Root Cause

`/v1/responses` HTTP requests were stateless across turns. Provider/auth selection uses round-robin, so continuation turns could switch auth between requests.

When a continuation payload includes `input[].type="reasoning"` with `encrypted_content`, that blob is upstream-context-bound (org/account/provider). If round-robin switched auth/provider, upstream returned `400 invalid_encrypted_content` (`could not be decrypted or parsed`).

## Fix

- Added responses auth-affinity store in `sdk/api/handlers/openai/openai_responses_affinity.go`:
  - maps `previous_response_id -> selected_auth_id`
  - maps `reasoning.encrypted_content (hashed) -> selected_auth_id`
- `openai_responses_handlers` now:
  - resolves pinned auth from request affinity keys before execution
  - captures selected auth ID via callback when not pinned
  - persists affinity from non-stream and SSE responses
- Added guarded recovery paths:
  - if pinned continuation fails and request contains encrypted reasoning input, strip encrypted reasoning and retry once **without pinning** (allows fallback to healthy auth/provider)
  - if unpinned request still returns `invalid_encrypted_content`, strip encrypted reasoning and retry once
  - same bootstrap recovery behavior for streaming `/v1/responses` before first chunk

## Tests

Added/updated tests in `sdk/api/handlers/openai/openai_responses_affinity_test.go`:
- `TestResponsesAffinityPinsAuthUsingEncryptedContent`
- `TestResponsesEncryptedContentRecoveryStripsReasoningInput`
- `TestResponsesPinnedAuthFailureFallsBackUnpinned`
- `TestResponsesStreamingPinnedAuthFailureFallsBackUnpinned`

Also updated stream error test for the chunk observer hook signature.

## Verification

- `go test ./sdk/api/handlers/...`


1saifj:
Addressed the review findings in follow-up commit 5818fca.

1. Pinned auth failure (auth_not_found): recovery is now broader than 400 invalid_encrypted_content. If a pinned continuation request fails and contains encrypted reasoning input, handler strips encrypted reasoning and retries once without pinning, allowing fallback to other auth/provider.
2. Streaming recovery: same strip+unpin recovery now applies to /v1/responses streaming bootstrap errors (before first payload).
3. Regression coverage: added tests for the exact exhaustion scenario:
   - TestResponsesPinnedAuthFailureFallsBackUnpinned
   - TestResponsesStreamingPinnedAuthFailureFallsBackUnpinned

This keeps affinity for happy path but no longer blocks fallback when the pinned upstream is exhausted.

1saifj:
Applied reviewer cleanup suggestion in commit 8e71041: removed the redundant second non-stream recovery block so there is now a single strip+retry path.

Validation rerun: 
- go test ./sdk/api/handlers/openai
- go test ./sdk/api/handlers/...

1saifj:
Addressed Gemini high-severity cache eviction feedback in the current branch:

- Removed full-map reset when maxEntries is exceeded.
- Implemented bounded partial eviction (reduce to ~75% capacity) after TTL cleanup for both affinity maps.
- This avoids sudden total affinity loss and reduces continuation miss spikes under load.

Implementation is in sdk/api/handlers/openai/openai_responses_affinity.go pruneLocked().

1saifj:
Follow-up updates pushed in commit 5b786c9 to close the remaining encrypted-continuation gaps for #1793.

What was added:
- Recovery now strips both encrypted reasoning and `previous_response_id`, including previous_response_id-only cases.
- Streaming recovery broadened to the same context predicate and extended for post-first-chunk terminal encrypted-content failures when assistant content has not started.
- Responses auth-affinity store now persists to disk (env-configurable path) so restart does not drop affinity.
- Added tests:
  - `TestResponsesStreamingPreviousResponseIDOnlyPinnedFailureFallsBackUnpinned`
  - `TestResponsesStreamingPostFirstChunkInvalidEncryptedFallsBackWhenNoContentYet`
  - `TestResponsesAffinityStorePersistsAcrossRestart`

Validation rerun:
- `go test -count=1 ./sdk/api/handlers/openai`
- `go test -count=1 ./sdk/api/handlers/...`
- `go test -race -count=1 ./sdk/api/handlers/openai`

This PR now covers the provider-switch + exhaustion + restart paths described in #1793.

luispater (changes requested):
Thanks for the PR. I’m requesting changes mainly around:

1) The streaming recovery gate is too eager. `responsesStreamChunkHasAssistantContent()` treats `response.output_item.added` for a message as "content started", but our translators emit that before `response.content_part.added` / `response.output_text.delta`. That means a failure can skip recovery even though no user-visible token has been sent yet.

2) The post-first-chunk retry path can splice two upstream attempts into one SSE stream. The first chunk is flushed to the client, and then the handler restarts the stream on error, so clients may see lifecycle events / response IDs from two different attempts in one response.

3) The affinity store now persists by default to a single file under `os.TempDir()`. Multiple CLIProxyAPI instances on the same host will share and overwrite this state, which can leak affinity across deployments and pin continuations to stale or wrong auth IDs.

Tested: `go test ./sdk/api/handlers/openai`

1saifj:
Addressed in `a27df47`.

1. The streaming recovery gate now waits for actual user-visible assistant output instead of treating `response.output_item.added` for a message as content-started. The handler only marks content as started on visible `response.output_text.*` payloads (or refusal/audio output), so a failure after lifecycle-only events can still recover.

2. The streaming recovery path no longer splices two upstream attempts into one SSE response. Pre-visible chunks are buffered first; if the upstream fails before anything visible is flushed, the handler can still retry safely. Once the stream has been flushed to the client, it does not restart the upstream anymore and returns a terminal stream error instead.

3. The default shared persistence file under `os.TempDir()` is removed. Responses auth affinity now persists only when `CLIPROXY_RESPONSES_AFFINITY_PATH` is explicitly configured, so separate CLIProxyAPI instances do not share affinity state by default.

I also added/updated regression coverage for the message-added-before-content case and the no-splice/no-leak behavior, then reran:

- `GOCACHE=/tmp/cliproxy-go-build GOTMPDIR=/tmp go test ./sdk/api/handlers/openai -count=1`
- `GOCACHE=/tmp/cliproxy-go-build GOTMPDIR=/tmp go test ./sdk/api/handlers/... -count=1`
- `GOCACHE=/tmp/cliproxy-go-build GOTMPDIR=/tmp go test -race ./sdk/api/handlers/openai -count=1`

1saifj:
@luispater Please if it is OK merge it. Issues happen all around using gpt 5.4 and 5.3 on my machine. 
Screenshot 2026-03-09 at 9 47 42 PM

InCerryGit:
This fix is urgently needed.

luispater (approved):
Summary: The follow-up closes the earlier blockers. Pinned and unpinned recovery now strips both `previous_response_id` and encrypted reasoning when needed, the streaming path no longer splices two upstream attempts into one SSE response, and affinity persistence is opt-in instead of defaulting to a shared temp file.

Blocking findings:
- None.

Non-blocking findings:
- The touched OpenAI responses package passes its targeted test run.
- The full server build still fails in unrelated packages outside this PR, so I could not use it as a clean end-to-end signal.

Test plan:
- Attempted `go build -o test-output ./cmd/server && rm -f test-output`, but it failed in unrelated packages already outside this PR.

luispater:

1saifj:
Done @luispater

luispater:
Closing this PR for now because the reported behavior can be addressed with the existing `routing.session-affinity` configuration.

The issue is caused by `/v1/responses` continuation requests being routed to a different upstream credential/provider after the original upstream generated `reasoning.encrypted_content`. That encrypted reasoning context is tied to the upstream account/provider that produced it, so provider/account switching can result in `400 invalid_encrypted_content`.

The recommended mitigation is to enable session affinity so requests from the same client session continue to use the same selected credential whenever it remains available:

```yaml
routing:
  strategy: "round-robin"
  session-affinity: true
  session-affinity-ttl: "1h"
```

`routing.session-affinity` keeps unrelated sessions on the normal round-robin path while making continuation turns for the same session sticky to the same auth credential. If the client can send a stable session header, prefer setting one explicitly, for example `X-Session-ID` or `Session_id`, so the affinity key remains stable across turns.

Thanks for the contribution and for documenting the failure mode.

PR #1798, opened 2026-03-02
fix(thinking): always clamp unsupported levels to nearest supported level

### LINKED ISSUE #1780: OpenAI completion error: 400 level "medium" not supported, valid levels: low, high
Date/time: 2026-03-01T14:40:53.482Z
Extension version: 5.7.0
Provider: openai (proxy)
Model: gemini-3.1-pro-high

400
OpenAI completion error: 400 level "medium" not supported, valid levels: low, high

Using kilo code antigravity client throws that error. In settings i also select high aswell too.

1saifj:
## Summary

- Always clamp unsupported thinking levels to the nearest supported level instead of returning a 400 error
- Removes the `allowClampUnsupported` gate that restricted clamping to budget-based→level-based provider conversions only

## Problem

When a client sends `reasoning_effort: "medium"` to a model that only supports `["low", "high"]` (e.g. `gemini-3.1-pro-high`), the proxy returns:

```
400 level "medium" not supported, valid levels: low, high
```

This happens because `allowClampUnsupported` evaluates to `false` for OpenAI→Antigravity requests (`isBudgetBasedProvider("openai") && isLevelBasedProvider("antigravity")` = `false && false`), so the existing `clampLevel()` function is never called.

## Fix

Remove the `allowClampUnsupported` gate so `clampLevel()` is always called when a level isn't in the model's supported list. The function already handles this correctly — it finds the nearest supported level by distance in the standard level order, with ties breaking to the lower level.

For `"medium"` with `["low", "high"]`: equidistant (1 step each), tie-breaks to `"low"`.

The error path is preserved as a safety net for truly unknown levels that aren't in `standardLevelOrder` (which shouldn't occur in practice).

## Changes

- `internal/thinking/validate.go` — Removed `allowClampUnsupported` variable and its guard condition
- `test/thinking_conversion_test.go` — Updated 8 test cases that now clamp instead of erroring + added 2 new test cases for the exact issue scenario

Fixes #1780

### DIFF
--- a/internal/thinking/validate.go
+++ b/internal/thinking/validate.go
@@ -53,7 +53,6 @@ func ValidateConfig(config ThinkingConfig, modelInfo *registry.ModelInfo, fromFo
 		return &config, nil
 	}
 
-	allowClampUnsupported := isBudgetBasedProvider(fromFormat) && isLevelBasedProvider(toFormat)
 	strictBudget := !fromSuffix && fromFormat != "" && isSameProviderFamily(fromFormat, toFormat)
 	budgetDerivedFromLevel := false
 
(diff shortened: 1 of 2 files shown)


hkfires:
No plans.

emircanerkul:
> No plans.

So... antigravity no useful without this fix.

Then please update description ot the repository

Wrap Gemini CLI, Antigravity, ChatGPT Codex, Claude Code, Qwen Code, iFlow as an OpenAI/Gemini/Claude/Codex compatible API service, allowing you to enjoy the free Gemini 2.5 Pro, GPT 5, Claude, Qwen model through API

to 

Wrap Gemini CLI, ChatGPT Codex, Claude Code, Qwen Code, iFlow as an OpenAI/Gemini/Claude/Codex compatible API service, allowing you to enjoy the free Gemini 2.5 Pro, GPT 5, Claude, Qwen model through API

PR #1802, opened 2026-03-03
fix(antigravity): normalize boolean JSON schema nodes

xuemusi:
## Summary

This PR fixes a request translation issue when forwarding Claude tool schemas through antigravity to Vertex-compatible APIs.

Some tool parameter schemas contain boolean schema nodes (`true` / `false`) from JSON Schema syntax. Vertex schema parsing rejects these values and returns:

- `INVALID_ARGUMENT`
- `Invalid value ... (type.googleapis.com/google.cloud.aiplatform.master.Schema), true`

## Root Cause

During schema cleanup, boolean schema nodes were not normalized before sending `function_declarations.parameters` to Vertex-compatible request payloads.

## Changes

- Updated `internal/util/gemini_schema.go`
- Added normalization for boolean schema nodes in `cleanJSONSchema`:
  - `true` -> permissive object schema
  - `false` -> restrictive object schema (non-matching)
- Applied normalization recursively in nested properties/items
- Added regression test:
  - `TestCleanJSONSchemaForAntigravity_BooleanPropertySchemaNormalized`
  - File: `internal/util/gemini_schema_test.go`

## Validation

- Unit test passes for the new regression case.
- Local runtime verification shows forwarded `POST /v1/messages?beta=true` requests returning `200` after this fix (instead of prior `400 INVALID_ARGUMENT`).

## Impact

- Fixes antigravity forwarding failures for requests containing tool schemas with boolean JSON Schema nodes.
- No behavior change expected for already-valid object-based schemas.

### DIFF
--- a/internal/util/gemini_schema.go
+++ b/internal/util/gemini_schema.go
@@ -2,6 +2,7 @@
 package util
 
 import (
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
@@ -42,6 +43,7 @@ func cleanJSONSchema(jsonStr string, addPlaceholder bool) string {
 	jsonStr = mergeAllOf(jsonStr)
 	jsonStr = flattenAnyOfOneOf(jsonStr)
 	jsonStr = flattenTypeArrays(jsonStr)
+	jsonStr = normalizeBooleanSchemas(jsonStr)
 
 	// Phase 3: Cleanup
 	jsonStr = removeUnsupportedKeywords(jsonStr)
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Request changes: boolean JSON Schema nodes are currently normalized with identical behavior for both `true` and `false` in `internal/util/gemini_schema.go`. That broadens `false` from a non-matching schema into a valid object schema, and on the Antigravity path it can even become an object with the placeholder `reason` field after empty-schema placeholder injection. Please preserve `false` as a non-matching/restrictive schema (as described in the PR body) and add a regression test that covers the `false` case.

PR #1816, opened 2026-03-04
fix(antigravity): pass through adaptive thinking effort level instead of always mapping to high

sususu98:
## Summary

- Antigravity Claude translator now passes through `output_config.effort` (low/medium/high) as `thinkingLevel` for adaptive thinking requests, instead of always mapping to `high`.
- `max` effort is mapped to `high` since Gemini backend does not support `max` as a thinking level.
- When no explicit effort is provided, defaults to `high` (preserves existing behavior).

## Changes

- `antigravity_claude_request.go`: Read `output_config.effort` in the `adaptive`/`auto` case and pass through to `thinkingConfig.thinkingLevel`
- `antigravity_claude_request_test.go`: Add table-driven tests covering low/medium/high/max effort levels and no-effort fallback


sususu98:
I've tested against the Antigravity backend: it accepts `thinkingLevel` of `low`, `medium`, and `high` for Claude 4.6 series models, but rejects `max`. So this PR maps `max` → `high` at the translator level.

luispater (approved):

PR #1832, opened 2026-03-04
Pretty-print auth metadata in filestore

pzx521521:
Replace json.Marshal with json.MarshalIndent when serializing auth.Metadata in sdk/auth/filestore.go so the stored JSON is indented (two-space) for improved readability. Error handling remains unchanged.


luispater:
No plans to support this.

PR #1835, opened 2026-03-04
feat(translator): align OpenAI/Gemini/Claude protocol conversion and streaming behavior

Hsiangpo:
## Summary
- align protocol conversion details across OpenAI/Gemini/Claude/Gemini-CLI translators
- fix object/string preservation, stop/stop_sequences mapping, tool_choice compatibility, and finish_reason mapping
- add multi-candidate support for both stream and non-stream conversion paths
- add regression tests for all repaired conversion paths
- add protocol alignment documentation under `docs/protocol/2026-03-05-openai-compat-alignment.md`

## Verification
- go test ./internal/translator/...
- go test ./sdk/api/handlers/openai/...
- go test ./internal/runtime/executor/... -run "OpenAI|compat|Translator"

### DIFF
--- a/docs/protocol/2026-03-05-openai-compat-alignment.md
+++ b/docs/protocol/2026-03-05-openai-compat-alignment.md
@@ -1,0 +1,146 @@
+# 2026-03-05 OpenAI 兼容层协议对齐修复记录
+
+## 背景
+- 目标仓库：`CLIProxyAPI-main.__latest_20260304000229`
+- 修复范围：`OpenAI <-> Claude/Gemini/Gemini-CLI` 的请求与响应转换
+- 重点：参数兼容、字段类型一致性、流式/非流式行为一致性、工具调用字段保真
+
+## 已修复问题
+
+### 1) `tool_choice` 对象被错误字符串化
+- 文件：`internal/translator/openai/openai/responses/openai_openai-responses_request.go`
+- 修复：对象/数组使用 `SetRaw`，基础类型使用 `Set`
+- 结果：保留 `tool_choice` 原始 JSON 结构
+
+### 2) Gemini Responses 丢失 `function_call_output.output` 对象
+- 文件：`internal/translator/gemini/openai/responses/gemini_openai-responses_request.go`
+- 修复：读取 `output` 的 `.Raw/.Value()`，对象走 `SetRaw`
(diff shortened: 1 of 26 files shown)


luispater (changes requested):
Thanks for the protocol-alignment work — the overall direction looks good, and the added regression coverage improves confidence in the request/response mapping changes.

That said, I found three blocking issues before this can be merged. Aside from these blocking items, the rest of the changes look reasonable to me.

Blocking findings:
- Blocking: `Gemini CLI -> OpenAI` streaming tracks tool-call indices globally instead of per choice. In `internal/translator/gemini-cli/openai/chat-completions/gemini-cli_openai_response.go`, `FunctionIndex` is shared across the whole stream, so if two candidates each emit their first function call, the second candidate gets `tool_calls[0].index = 1` instead of `0`. OpenAI-compatible clients treat tool-call indices as scoped to a single choice, so this breaks reconstruction of multi-choice tool-call streams.
- Blocking: `OpenAI -> Gemini` streaming also shares tool-call accumulation across choices. In `internal/translator/openai/gemini/openai_gemini_response.go`, `ToolCallsAccumulator` is keyed only by `toolIndex`, not by `(choiceIndex, toolIndex)`. If two choices both emit `tool_calls[0]`, the later choice overwrites the earlier one. When a finish chunk arrives, the accumulated tool call is flushed into the wrong candidate and the shared accumulator is cleared, so the other candidate loses its tool call entirely.
- Blocking: the PR introduces new Chinese comments, which violates the repository convention that code comments must be in English only. I noticed this in `internal/translator/openai/gemini/openai_gemini_request.go:67` and `internal/translator/openai/claude/openai_claude_request.go:41`. Please convert all newly added comments to English for consistency with the rest of the codebase.

Test plan:
- `go test ./internal/translator/...`
- `go test ./sdk/api/handlers/openai/...`
- `go test ./internal/runtime/executor/... -run "OpenAI|compat|Translator"`
- plus local repros for both multi-candidate streaming tool-call cases described above

Suggested follow-up:
- Please make the streaming tool-call state choice-aware in both directions.
- Please add regression tests for multi-candidate + tool-call streaming, not just multi-candidate text output.
- Please convert all newly added non-English comments to English.

PR #1836, opened 2026-03-05
feat: configurable oauth callback domain and unified callback route

shellus:
## Summary
- add optional `oauth-callback` config block to support external callback domain + provider paths while preserving localhost behavior by default
- wire management OAuth flows (Anthropic/Codex/Gemini/iFlow/Antigravity) to use one redirect URI source for both auth URL generation and token exchange
- add unified callback route `/oauth/callback/:provider` and keep legacy callback routes for compatibility; update README and README_CN examples

## Test Plan
- [x] `go test ./internal/config -run OAuthCallback -v`
- [x] `go test ./internal/api/handlers/management -run "Request.*RedirectURI|OAuthCallbackURLBuilder" -v`
- [x] `go test ./internal/api -run OAuthUnifiedCallbackRoute -v`
- [x] `go test ./internal/config ./internal/api ./internal/api/handlers/management -v`
- [x] `go test ./...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,30 @@ Get 10% OFF GLM CODING PLAN：https://z.ai/subscribe?ic=8JVLJQFSKB
 
 CLIProxyAPI Guides: [https://help.router-for.me/](https://help.router-for.me/)
 
+## Configurable OAuth Callback Domain
+
+By default, OAuth callbacks keep the existing localhost behavior.
+
+If you want to receive OAuth callbacks through a public domain, enable `oauth-callback` in config:
+
+```yaml
+oauth-callback:
+  enabled: true
+  external-base-url: "https://cliproxy.example.com"
+  provider-paths:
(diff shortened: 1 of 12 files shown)


shellus:
The path traversal concern in the unified callback flow is already guarded in current code.

- ValidateOAuthState rejects /, \\, .., overlength, and non [A-Za-z0-9._-] characters in state (internal/api/handlers/management/oauth_sessions.go).
- WriteOAuthCallbackFile enforces ValidateOAuthState before filepath.Join(authDir, fileName).
- Unified route /oauth/callback/:provider writes through WriteOAuthCallbackFileForPendingSession -> WriteOAuthCallbackFile, so it follows the same validation path.

Given this call chain, state is validated before file path construction in this PR implementation.

If stricter policy is preferred, we can add a dedicated regression test for malformed state payloads on /oauth/callback/:provider.

shellus:
test failed, cancel

PR #1838, opened 2026-03-05
Add a system-prompt-count option for some Claude providers to pass validation

Kindness-Kismet:
Some Claude providers verify the number of prompts, so this option has been added.

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -122,13 +122,16 @@ func (e *ClaudeExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, r
 		return resp, err
 	}
 
+	// Enforce system-prompt-count before cloaking so cloaking-injected blocks
+	// are not counted or rewritten by the count policy.
+	body = applyClaudeSystemPromptCountPolicy(e.cfg, auth, body)
+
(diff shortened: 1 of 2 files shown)


Kindness-Kismet:
> Summary: This PR adds `system-prompt-count` to Claude provider config, threads it through the management API and config diff output, and enforces it in the Claude executor. I ran the targeted package tests plus a server build, and they passed. I still found two blocking correctness issues in the executor path.
> 
> Key findings:
> 
>     * Blocking: shrinking the `system` array can reorder text around intervening non-text system blocks.
> 
>     * Blocking: the new policy runs after cloaking, so it rewrites the Claude Code billing/agent blocks when `cloak` and `system-prompt-count` are enabled together.
> 
> 
> Test plan:
> 
>     * `go test ./internal/runtime/executor ./internal/config ./internal/watcher/diff ./internal/api/handlers/management`
> 
>     * `go build -o test-output ./cmd/server`
> 
> 
> Follow-ups:
> 
>     * Apply the count policy before cloaking, or explicitly exclude cloaking-injected blocks from the count.
> 
>     * Add a regression test for a mixed sequence such as `[text, text, non-text, text]` when shrinking to `2`.

Thanks! All issues have been addressed

PR #1867, opened 2026-03-05
fix(antigravity/claude): parse <thinking> XML tags into thought blocks

### LINKED ISSUE #1777: bug(antigravity/claude): Thinking blocks returned as raw <thinking> XML tags instead of proper thinking content blocks
[x] No, it's another issue.

**Describe the bug**

When using Claude models (Opus 4.6, Sonnet 4.6) through the Antigravity provider, thinking content is returned as **raw text with `<thinking>` XML tags** mixed into `type: "text"` content blocks, instead of proper `type: "thinking"` content blocks. Clients like OpenCode that render thinking blocks separately (dark grey, visually separated) cannot distinguish thinking from response text.

This works correctly for Gemini models on the same Antigravity backend (where the backend sets `thought: true` on thought parts), and also works correctly with providers that return native Anthropic format (e.g., Z.ai for GLM-5).

**Root cause:** The Antigravity backend for Claude models does not annotate thinking with `thought: true` in the response parts. Instead, Claude's thinking arrives as regular text wrapped in `<thinking>...</thinking>` XML tags. Since CLIProxyAPI's response translator relies exclusively on the `thought: true` flag, the thinking content passes through as regular text.

Image

**CLI Type**

Antigravity (Google backend for Claude models)

**Model Name**

claude-opus-4-6-thinking, claude-sonnet-4-6-thinking (via Antigravity)

**LLM Client**

OpenCode 1.2.15 with `@ai-sdk/anthropic` SDK

**Request Information**

```bash
# Streaming request with thinking enabled
curl -s http://localhost:8317/v1/messages \
  -H "x-api-key: <key>" -H "content-type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-opus-4-6-thinking","max_tokens":500,"stream":true,"thinking":{"type":"enabled","budget_tokens":8192},"messages":[{"role":"user","content":"Solve: what is the square root of 252557324527? Show your work."}]}'
```

Streaming response shows only `content_block_start` with `"type":"text"` and `text_delta` events — zero `thinking` blocks or `thinking_delta` events, even when the model produces thinking content.

When thinking IS produced by the backend, it appears as raw `<thinking>...</thinking>` XML tags within the text content block, e.g.:

```
502,550² = 252,556,302,500
252,557,324,527 - 252,556,302,500 = 1,022,027
</thinking>

√252,557,324,527 ≈ 502,550.818

<thinking>
The square root is approximately 502,550.818.
</thinking>
```

**Expected behavior**

Thinking content should arrive as separate content blocks:

```json
{
  "content": [
    { "type": "thinking", "thinking": "Let me calculate...", "signature": "..." },
    { "type": "text", "text": "√252,557,324,527 ≈ 502,550.818" }
  ]
}
```

This is what happens with native Anthropic format providers (e.g., Z.ai with GLM-5), where OpenCode renders thinking in dark grey, visually separated from the response.

**Code Analysis**

The response translator correctly handles `thought: true` when present — the issue is at the backend layer:

**Response translator** (`internal/translator/antigravity/claude/antigravity_claude_response.go`, line 136):
```go
if partResult.Get("thought").Bool() {
    // Creates proper content_block_start with type:"thinking"
    // Creates thinking_delta events
    // ← This code path works correctly (verified with Gemini models)
} else {
    // Regular text content → type:"text" blocks
    // ← Claude thinking lands here because backend doesn't set thought: true
}
```

**Executor** (`internal/runtime/executor/antigravity_executor.go`, lines 652-664):
```go
thought := part.Get("thought").Bool()
if thought || part.Get("text").Exists() {
    kind := "text"
    if thought {
        kind = "thought"
    }
    // Claude via Antigravity: thought is always false, kind is always "text"
}
```

**Data flow comparison:**

| Provider | Backend behavior | Result |
|---|---|---|
| Antigravity/Claude | Thinking embedded as `<thinking>` XML tags in regular text, **no `thought: true` flag** | BROKEN: Raw XML tags in text blocks |
| Antigravity/Gemini | Thinking annotated with `thought: true` | OK: Proper thinking blocks |
| Z.ai/GLM-5 | Native Anthropic format with `type: "thinking"` content blocks | OK: Proper thinking blocks |

**Suggested Fix**

Since the Antigravity backend doesn't annotate Claude's thinking with `thought: true`, CLIProxyAPI could work around this by detecting `<thinking>` XML tags in text content during response translation:

1. In `antigravity_claude_response.go`, when processing text parts without `thought: true`, check for `<thinking>...</thinking>` patterns
2. Split the text into segments and emit thinking segments as `type: "thinking"` content blocks
3. Gate this behavior on Claude models only (check model name) to avoid false positives
4. Handle streaming edge cases where `<thinking>` open/close tags may span multiple chunks

Alternatively, if there's a way to configure the Antigravity backend to return `thought: true` for Claude models (different API parameter or flag), that would be cleaner since the existing handler would work as-is.

**OS Type**
- OS: Linux (WSL2)
- Kernel: 6.6.87.2-microsoft-standard-WSL2

**Additional context**

- CLIProxyAPI version: v6.8.35
- Related: #1540 (adaptive thinking `output_config.effort` parameter)
- The request-side thinking config translation works correctly (budget/thinkingLevel are set) — this is purely a response-side issue
- cc @luispater

lang-911:
Closes #1777

## Summary
- add Antigravity Claude thinking-tag parser for `<thinking>...</thinking>` text blocks
- apply parser in streaming and stream-to-nonstream execution paths
- add comprehensive parser tests (including split-tag stream chunks)

## Validation
- go test ./internal/runtime/executor -run ThinkingTagParser -count=1
- go test ./...


JasperDale420 (changes requested):
Not merging this yet. The existing review on the PR already flags a critical double-processing risk in the new thinking-tag path, and I do not want to ship the executor changes until that issue is resolved and the branch is re-reviewed.

luispater (changes requested):
Thanks — the latest commit fixes the earlier double-processing concern, and the parser coverage is strong. I still see one blocking issue in the split path: when a single text part is split into multiple segments, the parser rebuilds fresh part objects with only text/thought and drops original metadata. That is risky for Claude because the downstream Antigravity->Claude response translator relies on thoughtSignature to cache/replay signed thinking blocks in later turns. If a tagged text part also carries thoughtSignature, the current rewrite loses it before translation. Please preserve relevant metadata on the resulting thought segment(s) — at minimum thoughtSignature — and add a regression test for a tagged text part that also includes a signature.

lang-911:
Updated in d373d79.

Addressed the latest blocking review by preserving thought signatures when <thinking> tag parsing rewrites a text part:
- split/single-segment tagged thought text now moves the signature onto a dedicated empty thought part immediately after the final derived thought segment
- visible text segments drop thought/signature metadata
- close-tag-only chunks still preserve the signature so the prior thinking block can be closed/cached correctly

Added regressions for the split tagged-part case, the single-segment case, the close-tag-only case, and a Claude translator contract test for signature caching.

Validation:
- go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1
- go test ./internal/runtime/executor -count=1
- go test ./internal/translator/antigravity/claude -count=1

lang-911:
Follow-up in 0a83ee0: the failed translator-path-guard check was caused by my earlier regression test addition under internal/translator, not by the parser fix itself. I moved that contract coverage into internal/runtime/executor/thinking_tag_parser_test.go and restored internal/translator/antigravity/claude/antigravity_claude_response_test.go to its original content so the PR no longer changes restricted translator paths.

Revalidated:
- go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1
- go test ./internal/runtime/executor -count=1

lang-911:
Addressed the latest gemini-code-assist cleanup findings in `thinking_tag_parser.go`:

- simplified the final `sjson.Delete(...)` return in `rewriteThinkingSegmentPart`
- simplified the final `sjson.Set(...)` return in `buildThoughtSignaturePart`
- removed the dead `maxTagBufferLen` guard and constant

Validation:
- `go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1`
 
/gemini review

lang-911:
Addressed the remaining gemini-code-assist findings in `thinking_tag_parser.go`:

- refactored `rewriteThinkingSegmentPart` to use a single `encoding/json` decode/edit/encode pass with `map[string]json.RawMessage`
- refactored `buildThoughtSignaturePart` to build the signature-only part with one marshal step
- added a regression that verifies split tagged parts preserve nested raw metadata and large integer metadata without generic normalization

Validation:
- `go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1`

/gemini review

lang-911:
Addressed the latest gemini-code-assist helper cleanup in `thinking_tag_parser.go`:

- simplified `buildThoughtSignaturePart` to start from a static base JSON string and apply a single `sjson.Set(...)`
- removed the temporary struct and `json.Marshal` path for that helper

Validation:
- `go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1`

/gemini review

lang-911:
Addressed the final gemini-code-assist finding in `thinking_tag_parser.go`:

- switched `rewriteThinkingSegmentPart` back to `sjson.Set` / `sjson.Delete` for the segment rewrite path
- removed the now-unused `encoding/json` import from that file
- kept the existing signed-thinking and raw-metadata regressions unchanged as guardrails

Validation:
- `go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1`

All Gemini review threads on this PR are now resolved.

/gemini review

lang-911:
Addressed the final gemini-code-assist readability suggestion in `thinking_tag_parser.go`:

- replaced the manual `append` + `copy` + index assignment insertion with the idiomatic nested `append(...)` form
- left the `insertAt` calculation and signature-placement behavior unchanged
- kept the existing signed-thinking and raw-metadata regressions unchanged as guardrails

Validation:
- `go test ./internal/runtime/executor -run 'ThinkingTagParser' -count=1`

luispater (approved):
Summary: The latest head resolves the previous blocker. `ThinkingTagParser` now preserves `thoughtSignature` by moving it onto a dedicated empty thought part, and the regression tests cover split tags, close-tag chunks, and raw metadata preservation in the parser path. I also verified the PR head with `go test ./internal/runtime/executor -count=1` and `go build -o test-output ./cmd/server && rm -f test-output`.

Blocking findings:
- None.

Non-blocking findings:
- None.

Test plan:

luispater:

lang-911:
Updated per the latest PR comment:

- changed the PR base branch from `main` to `dev`
- merged the latest `dev` into this branch and resolved the conflict in `internal/runtime/executor/antigravity_executor.go` by keeping both the thinking-tag parser pass and the newer `helps.ParseAntigravityStreamUsage` / `reporter.Publish` path
- fixed a post-merge executor test mismatch in `internal/runtime/executor/thinking_tag_parser_test.go` (`[][]byte` join)
- aligned the Qwen injected default system text with the merged executor tests so the executor package stays green

Validation:
- `go test ./internal/runtime/executor -run ThinkingTagParser -count=1`
- `go test ./internal/runtime/executor -run TestEnsureQwenSystemMessage_MergeStringSystem -count=1`

Head commit: `937b758a`

lang-911:
Follow-up correction:

- reverted my earlier runtime change in `internal/runtime/executor/qwen_executor.go`; that file is back to the original blank injected default system text from `dev`
- updated `internal/runtime/executor/qwen_executor_test.go` to match that unchanged runtime behavior

Validation:
- `go test ./internal/runtime/executor -run TestEnsureQwenSystemMessage_MergeStringSystem -count=1`

Head commit: `604de359`

lang-911:
Follow-up correction:

- reverted my earlier change in `internal/runtime/executor/qwen_executor_test.go` as well
- both `internal/runtime/executor/qwen_executor.go` and `internal/runtime/executor/qwen_executor_test.go` are now restored to their original `dev` content

This commit is a direct revert only.

Head commit: `8264f3ce`

PR #1869, opened 2026-03-05
feat(auth_files): Update time layouts for parsing timestamps

yikakia:
fix some time layout
see: https://pkg.go.dev/time#pkg-constants

for "2006-01-02 15:04:05" is equal to time.DateTime
for "2006-01-02T15:04:05Z07:00" is duplicate with time.RFC3339

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -82,7 +82,8 @@ func parseLastRefreshValue(v any) (time.Time, bool) {
 		if s == "" {
 			return time.Time{}, false
 		}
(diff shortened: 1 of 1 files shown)


yikakia:
I tested local by run go test ./... all tests are passed

luispater:
Thanks for the cleanup. After review, this change does not alter behavior or improve backward compatibility: `time.DateTime` is equivalent to the existing `"2006-01-02 15:04:05"` layout, and the removed `"2006-01-02T15:04:05Z07:00"` layout is already covered by `time.RFC3339`. Since this introduces no functional change and provides very limited maintenance value, we are closing this PR as an unnecessary change.

PR #1870, opened 2026-03-05
cleanup(translator): remove leftover instructions restore in codex responses

### LINKED ISSUE #1868: cleanup: remove leftover instructions restore logic in codex responses translator
## Summary

`ac802a46` removed the codex instructions injection system (-6703 lines), but the "restore instructions" logic in the response translator was not cleaned up.

## Affected Code

**`internal/translator/codex/openai/responses/codex_openai-responses_response.go`**

Stream path (L20-24):
```go
if typeStr == "response.created" || typeStr == "response.in_progress" || typeStr == "response.completed" {
    if gjson.GetBytes(rawJSON, "response.instructions").Exists() {
        instructions := gjson.GetBytes(originalRequestRawJSON, "instructions").String()
        rawJSON, _ = sjson.SetBytes(rawJSON, "response.instructions", instructions)
    }
}
```

Non-stream path (L43-46):
```go
if responseResult.Get("instructions").Exists() {
    instructions := gjson.GetBytes(originalRequestRawJSON, "instructions").String()
    template, _ = sjson.Set(template, "instructions", instructions)
}
```

## Why This Is Dead Code

The original purpose was: the proxy injected custom instructions (per-model system prompts) into the request before forwarding to upstream. The response translator then "restored" the client's original instructions so the client wouldn't see the injected content.

Since `ac802a46` removed all injection logic, the instructions sent to upstream are now identical to what the client sent. The upstream's `response.instructions` should already match, making the restore a no-op.

## Suggested Fix

Remove both stream and non-stream restore blocks. The response should transparently pass through `response.instructions` from upstream without modification.

sususu98:
## Summary

- Remove dead code: instructions restore logic in codex responses translator that became a no-op after `ac802a46` removed the instructions injection system
- Remove unused `sjson` import

## Context

The instructions restore was originally needed when CPA injected custom per-model system prompts into Codex requests. The response translator would then "restore" the client's original instructions so injected content was hidden from the client.

Since `ac802a46` removed all injection logic, the proxy no longer modifies instructions before forwarding. The upstream's `response.instructions` already matches the client's original value, making the restore a no-op.

## Changes

**`internal/translator/codex/openai/responses/codex_openai-responses_response.go`**

- Streaming path: removed the `response.created`/`response.in_progress`/`response.completed` instructions override block
- Non-streaming path: removed the `response.instructions` override and return `responseResult.Raw` directly
- Removed unused `sjson` import

Closes #1868

### DIFF
--- a/internal/translator/codex/openai/responses/codex_openai-responses_response.go
+++ b/internal/translator/codex/openai/responses/codex_openai-responses_response.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 
 	"github.com/tidwall/gjson"
-	"github.com/tidwall/sjson"
 )
 
 // ConvertCodexResponseToOpenAIResponses converts OpenAI Chat Completions streaming chunks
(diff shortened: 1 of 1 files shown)


luispater (approved):

JasperDale420:
Not merging this yet because the translator guard check is failing on the PR branch. Once that check is green again, it can come back for merge review.

PR #1932, opened 2026-03-07
fix: preserve Claude thinking signatures in Codex translator

Darley-Wey:
## Summary
- keep Claude-compatible thinking blocks from Codex responses compliant by always including the `signature` field
- delay closing streaming thinking blocks until the upstream reasoning `encrypted_content` is available and emit it as `signature_delta`
- preserve reasoning signatures in non-stream responses and add targeted regression tests
## Testing
- go test ./internal/translator/codex/claude -run TestConvertCodexResponseToClaude -count=1


Darley-Wey:
Addressed the review feedback in follow-up commit `a9aab5b`.
- only emit `signature_delta` when `encrypted_content` is actually present
- only auto-close an open thinking block after `response.reasoning_summary_part.done` has been seen
- strengthened the no-signature streaming regression test to assert the block still closes cleanly without emitting `signature_delta`
Retested with:
- `go test ./internal/translator/codex/claude -run TestConvertCodexResponseToClaude -count=1`

luispater (approved):

luispater (changes requested):
From a translator-behavior perspective, this looks good to me.

I re-checked the thinking signature flow against the actual Codex event ordering here: response.output_item.added for the reasoning item already carries encrypted_content, then the reasoning summary starts streaming, and response.output_item.done carries encrypted_content again before the thinking block is finalized. With that event order, the updated state machine preserves the Claude thinking signature correctly.

Non-blocking suggestion: it would still be great to add a regression test for that exact event ordering so the assumption stays encoded in the test suite.

Requesting changes only because this PR currently has merge conflicts. Please rebase/merge the latest main and resolve the conflicts, then this should be good to re-review.

Darley-Wey:
@luispater Conflicts have been resolved in 8bb1ede by merging the latest dev into this branch. The PR is mergeable again now, and checks are rerunning. Thanks for the review.

luispater (commented):
Summary: The rebase keeps the thinking-signature fix intact, and the merged state no longer shows any blocking translator regression.

Blocking findings:
- None.

Non-blocking findings:
- The exact `response.output_item.added` / `response.output_item.done` ordering that carries `encrypted_content` is still only covered indirectly. A regression test for the early-signature path would reduce the risk of future state-machine drift.

Test plan:
- I validated the server still builds with `go build -o test-output ./cmd/server && rm -f test-output`.
- I reviewed the final diff in `internal/translator/codex/claude/codex_claude_response.go` and `internal/translator/codex/claude/codex_claude_response_test.go`.

PR #1939, opened 2026-03-07
feat(synthesizer): propagate cloak config from OAuth JSON to auth attributes

lichengzhe:
## Summary

- Propagate `cloak_*` fields from OAuth JSON metadata to `auth.Attributes`, enabling per-account cloak configuration for OAuth tokens
- Currently cloak config only works via `claude-api-key` entries in config.yaml, which requires matching by static API key — impossible for OAuth accounts whose `access_token` rotates every ~15 minutes
- This change bridges the gap by allowing OAuth JSON files to include cloak settings that feed into the existing `getCloakConfigFromAuth` fallback path

## Motivation

Users who exclusively use OAuth tokens (Claude Max subscriptions) cannot configure `cache-user-id`, `sensitive-words`, or `strict-mode` because:
1. `CloakConfig` is only attached to `ClaudeKey` in config.yaml
2. `resolveClaudeKeyCloakConfig()` matches by API key, but OAuth `access_token` rotates every 15 min
3. The fallback `getCloakConfigFromAuth()` reads from `auth.Attributes`, but the file synthesizer never populated cloak fields there

## Usage

Add cloak fields to the OAuth JSON file:

```json
{
  "type": "claude",
  "email": "user@example.com",
  "cloak_cache_user_id": "true",
  "cloak_mode": "full",
  "cloak_strict_mode": "true",
  "cloak_sensitive_words": "API,proxy,openai"
}
```

All fields are optional. Empty or whitespace-only values are ignored.

## Test plan

- [x] Added `TestFileSynthesizer_Synthesize_CloakAttributes` with 5 sub-tests:
  - All cloak fields set
  - Only `cache-user-id`
  - Whitespace trimming
  - Empty values ignored
  - No cloak fields (backward compatible)
- [ ] CI tests pass

### DIFF
--- a/internal/watcher/synthesizer/file.go
+++ b/internal/watcher/synthesizer/file.go
@@ -148,6 +148,18 @@ func synthesizeFileAuths(ctx *SynthesisContext, fullPath string, data []byte) []
 			}
 		}
 	}
+	// Read cloak settings from auth file so that OAuth accounts can
+	// configure cloaking without a matching claude-api-key entry.
+	for _, cloakKey := range []string{
+		"cloak_mode", "cloak_strict_mode",
+		"cloak_sensitive_words", "cloak_cache_user_id",
+	} {
+		if v, ok := metadata[cloakKey].(string); ok {
+			if trimmed := strings.TrimSpace(v); trimmed != "" {
+				a.Attributes[cloakKey] = trimmed
+			}
+		}
+	}
 	ApplyAuthExcludedModelsMeta(a, cfg, perAccountExcluded, "oauth")
 	if provider == "gemini-cli" {
 		if virtuals := SynthesizeGeminiVirtualAuths(a, metadata, now); len(virtuals) > 0 {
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Summary: Thanks — the approach is nicely scoped and reusing the existing getCloakConfigFromAuth path makes sense. I found one correctness gap before this is safe to merge.

Key findings:
- Blocking: the new propagation logic only accepts string-valued cloak_* fields from OAuth JSON. Native JSON forms like "cloak_cache_user_id": true, "cloak_strict_mode": true, or "cloak_sensitive_words": ["API","proxy"] are silently ignored, even though the canonical CloakConfig model uses bool / []string types. Because this feature exposes cloak settings through JSON, I think the synthesizer should normalize those native JSON shapes into attribute strings (or reject them explicitly instead of silently no-op'ing).
- Non-blocking: the added tests verify attribute synthesis, but there is still no executor-level regression proving that an OAuth auth with propagated cloak_* attributes actually changes the cloaking path.

Test plan:
- Reviewed the diff plus the existing cloak fallback and executor tests.
- PR CI is green (build, ensure-no-translator-changes).

PR #1940, opened 2026-03-07
fix(claude): enable interleaved-thinking beta, decode AMP error gzip, fix budget 400

Blue-B:
Three small fixes found while running Claude models through the proxy

**1. interleaved-thinking beta header missing**
applyClaudeHeaders never adds the interleaved-thinking-2025-05-14 beta flag, so thinking blocks silently disappear. Added a guard that appends it when absent.

**2. AMP proxy swallows gzip-encoded error bodies**
ModifyResponse skipped non-2xx responses, so 4xx/5xx bodies with hidden gzip encoding reached the client garbled. Removed the status-code guard.

**3. normalizeClaudeBudget leaves impossible budget → 400**
When adjustedBudget < minBudget, the old code returned the body unchanged, which the API rejects. Now it sets max_tokens = budgetTokens + 1 so the budget stays valid.

### DIFF
--- a/internal/thinking/provider/claude/apply.go
+++ b/internal/thinking/provider/claude/apply.go
@@ -174,7 +174,8 @@ func (a *Applier) normalizeClaudeBudget(body []byte, budgetTokens int, modelInfo
 	// Ensure the request satisfies Claude constraints:
 	//  1) Determine effective max_tokens (request overrides model default)
 	//  2) If budget_tokens >= max_tokens, reduce budget_tokens to max_tokens-1
-	//  3) If the adjusted budget falls below the model minimum, leave the request unchanged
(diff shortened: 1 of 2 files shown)


Blue-B:
Addressed the blocking feedback — pushed a fix:

**normalizeClaudeBudget** now clamps `max_tokens` to `MaxCompletionTokens` before raising it. If even after clamping the budget still falls below `minBudget`, thinking is disabled entirely (`thinking` key removed).

Also added unit tests covering:
- Normal raise (`budgetTokens+1` within model limit)
- Clamp to model max (`budgetTokens+1 > MaxCompletionTokens`)
- Disable thinking (unsatisfiable constraints)
- No-op (budget already valid)
- Budget adjustment to `max_tokens-1`

Please take another look when you get a chance.

PR #1941, opened 2026-03-07
fix(gemini): infer missing type:"object" for schema nodes with properties

Blue-B:
Gemini API rejects tool schemas where a node has properties but no type field

Added inferMissingObjectType to the schema cleanup pipeline. It walks the schema tree and sets type: "object" on any node that has properties without an explicit type.

Fixes tool_use failures when MCP tools or Droid CLI send schemas without explicit object types.


Blue-B:
Fixed — moved the `isPropertyDefinition` check to use `parentPath` instead of `p`, matching the pattern used elsewhere in this file. Thanks for catching this.

Blue-B:
Addressed both blocking and non-blocking feedback — pushed a fix:

1. **Edge case fix**: Replaced `isPropertyDefinition` with `isSchemaPropertiesKeyword` in `inferMissingObjectType`. The new function walks path segments tracking schema-vs-field depth, so a field literally named `"properties"` is correctly handled.

2. **Regression tests added** (5 cases):
   - Normal node with `properties` but no `type` → adds `type:"object"`
   - Existing `type` not overwritten
   - Nested schema missing `type`
   - Field literally named `"properties"` (string field) → not modified
   - Field named `"properties"` that itself has sub-`properties` → gets `type:"object"`

Please take another look when you get a chance.

luispater:

PR #1942, opened 2026-03-07
feat: add opt-in binary self-update from GitHub releases

lichengzhe:
## Summary

- Add `auto-update: true` config option for automatic binary self-update from GitHub releases
- Checks every 3 hours, downloads matching platform binary, verifies SHA256 checksum, atomically replaces, and triggers graceful restart
- Usage statistics are backed up before restart and can be restored afterward
- Disabled by default (opt-in), skips dev builds, supports hot-reload config toggle

## Motivation

Currently there is no built-in way to keep CLIProxyAPI up to date. Users must rely on external cron scripts to check releases, download, and restart. The management panel (`management.html`) already has a built-in auto-updater — this PR extends the same concept to the binary itself.

## Design

Follows the same patterns as the existing `managementasset` auto-updater:
- `atomic.Pointer[config.Config]` for thread-safe hot-reload support
- `sync.Once` for idempotent startup
- Proxy-aware HTTP client via `util.SetProxy`

**Update flow:**
```
Check GitHub releases/latest → Compare semver with buildinfo.Version →
Download CLIProxyAPI_{ver}_{os}_{arch}.tar.gz → Verify SHA256 from checksums.txt →
Backup usage stats → Atomic binary replace → SIGTERM → systemd restarts
```

**Usage statistics preservation:**
Before replacing the binary, the updater calls `/v0/management/usage/export` via localhost and saves to a backup file. `RestoreUsageStatistics()` is exported for post-restart import.

## Files changed

| File | Change |
|------|--------|
| `internal/autoupdate/updater.go` | New: complete self-update implementation (~350 lines) |
| `internal/autoupdate/updater_test.go` | New: tests for version comparison, checksum, tar extraction |
| `internal/config/config.go` | Add `AutoUpdate bool` field |
| `cmd/server/main.go` | Wire up auto-updater on service start |
| `internal/api/server.go` | Propagate config on hot-reload (2 locations) |
| `config.example.yaml` | Document the new option |

## Usage

```yaml
auto-update: true
```

Can be toggled at runtime via management API config editing — no restart needed.

## Test plan

- [x] Unit tests for `isNewer` (semver comparison, edge cases)
- [x] Unit tests for `parseChecksums` (sha256sum format parsing)
- [x] Unit tests for `verifyChecksum` (match/mismatch/missing)
- [x] Unit tests for `extractBinaryFromTarGz` (valid archive + missing binary)
- [x] Unit tests for `findAsset` (match + not found)
- [x] Unit tests for `platformAssetName` (format validation)
- [ ] CI build passes

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -568,7 +568,6 @@ func main() {
 		} else {
 			// Start the main proxy service
 			managementasset.StartAutoUpdater(context.Background(), configFilePath)
-			autoupdate.SetConfig(cfg)
 			autoupdate.Start(context.Background())
 			cmd.StartService(cfg, configFilePath, password)
 		}
--- a/internal/autoupdate/updater.go
+++ b/internal/autoupdate/updater.go
@@ -145,11 +145,10 @@ func checkAndUpdate(ctx context.Context, cfg *config.Config) error {
 		return err
 	}
 
-	// 3. Download checksums.txt for verification.
+	// 3. Download checksums.txt for verification (mandatory — abort if unavailable).
 	checksums, err := downloadChecksums(ctx, client, release.Assets)
 	if err != nil {
-		log.WithError(err).Warn("auto-update: checksums not available, proceeding without verification")
-		checksums = nil
+		return fmt.Errorf("checksums not available, aborting for safety: %w", err)
 	}
 
 	// 4. Download the archive.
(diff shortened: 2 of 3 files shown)


luispater:
Thanks a lot for the contribution and for putting so much effort into this PR.

After reviewing it, I don’t think we want to move forward with binary self-update support in CLIProxyAPI.

The main reason is that self-updating binaries do not fit well with the normal package management and deployment model on macOS and Linux. In those environments, upgrades are generally expected to be handled by the system package manager, container workflow, or service deployment process, rather than by the application replacing its own executable at runtime.

Keeping upgrades outside the application also makes behavior more predictable for operators and avoids conflicts with existing update policies, packaging conventions, and system administration practices.

Because of that, we’re not planning to support this feature for now, so I’m closing this PR. Still, I really appreciate the implementation and the thought that went into it.

PR #1943, opened 2026-03-07
fix: suppress token refresh write-back loop and reduce log noise

YoungMoneyInvestments:
## Summary

- **conductor.go**: Use `WithSkipPersist` in `refreshAuth` to prevent disk write-back after auto-refresh. The refresh cycle writes the updated token to the JSON file, triggering the fsnotify watcher → `reloadCallback` → `UpdateClients`, logging "server clients and configuration updated" every ~1 minute per token. Since the in-memory state is already current, persisting is unnecessary and creates a feedback loop.

- **server.go**: Change `UpdateClients` `fmt.Printf` to `log.Debugf` so configuration reload messages respect the debug flag and don't unconditionally fill stdout logs.

## Problem

With 2+ OAuth tokens (e.g., Claude Max accounts), each refreshing within a 4-hour lead window with a 1-minute backoff, the log accumulates ~2 messages/minute continuously. Over 12 days this produced 438 redundant "configuration updated" lines — pure noise that obscures meaningful log entries and contributes to log file growth.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./sdk/cliproxy/auth/... -run "SkipPersist|Refresh"` passes
- [ ] Verify no "configuration updated" messages in stdout when `debug: false`
- [ ] Verify messages appear when `debug: true`
- [ ] Confirm token refresh still works (check token expiry updates in auth JSON files)

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -993,7 +993,7 @@ func (s *Server) UpdateClients(cfg *config.Config) {
 	}
 
 	total := authEntries + geminiAPIKeyCount + claudeAPIKeyCount + codexAPIKeyCount + vertexAICompatCount + openAICompatCount
-	fmt.Printf("server clients and configuration updated: %d clients (%d auth entries + %d Gemini API keys + %d Claude API keys + %d Codex keys + %d Vertex-compat + %d OpenAI-compat)\n",
+	log.Debugf("server clients and configuration updated: %d clients (%d auth entries + %d Gemini API keys + %d Claude API keys + %d Codex keys + %d Vertex-compat + %d OpenAI-compat)",
 		total,
 		authEntries,
 		geminiAPIKeyCount,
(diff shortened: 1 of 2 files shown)


xkonjin (commented):
Quick review pass:
- I didn’t see targeted regression coverage in the diff; please add or point CI at a focused test for the changed path in server.go, conductor.go.
- Before merge, I’d smoke-test the behavior touched by server.go, conductor.go with malformed input / retry / rollback cases, since that’s where this class of change usually breaks.

luispater (changes requested):
Summary: Thanks for tackling the reload-log noise. The logging change in server.go looks reasonable, but the new WithSkipPersist call in refreshAuth introduces a correctness regression.

Blocking:
- WithSkipPersist is documented for watcher-driven updates where disk is already the source of truth. In refreshAuth, the provider has just returned new credentials, so skipping persistence means Manager.Load will restore stale auth state after restart.
- This is especially risky because refresh paths like Claude and Codex can rotate refresh_token values, and those updated tokens would now exist only in memory.

Suggested fix:
- Keep persisting refreshed auth data, and narrow the reload/log suppression to the watcher/reload path instead.
- Please add a regression test proving that auto-refresh still persists updated tokens.

PR #1945, opened 2026-03-08
fix(auth): avoid stale runtime state when re-adding disabled auth files

### LINKED ISSUE #2061: Re-created auth file inherits stale runtime state when re-added with the same path
### Summary

When an auth file is deleted and later recreated with the same path/name, the new runtime auth can inherit stale `ModelStates` from the previously disabled entry. This can leave the recreated auth stuck in old cooldown / unavailable / quota state and prevent it from being selected even though the file contents have already been replaced.

### Current upstream behavior

On current `main`, auth deletion keeps the runtime entry and marks it disabled:

- `sdk/cliproxy/service.go` `applyCoreAuthRemoval(...)`
- `internal/api/handlers/management/auth_files.go` `disableAuth(...)`

Both paths effectively do `Disabled = true` and then call `Update(...)`.

`Update(...)` then preserves previous runtime state when the same auth ID is updated again:

```go
}
```

So when a file with the same path/name is recreated and resolves to the same auth ID, the new auth can inherit stale model-level runtime state from the removed one.

### Reproduction

1. Add an auth file.
2. Let it enter a model-level cooldown / unavailable state (for example quota cooldown or another blocked `ModelStates` condition).
3. Delete that auth file from management or from the watched auth directory.
4. Add a new auth file back with the same path/name.
5. Send a request that should be routable through the recreated auth.

### Actual result

The recreated auth can still behave like the old removed auth:

- inherits stale `ModelStates`
- stays in cooldown / unavailable state
- may not be scheduled immediately even though the credential has already been replaced

### Expected result

A recreated auth file should start from a clean runtime state.

Deleting an auth file should not allow stale runtime state from the removed entry to bleed into a later recreated file with the same identity/path.

### Suggested fix direction

Keep the current disable-based removal behavior if desired, but avoid inheriting `ModelStates` from an auth entry that represents a removed/disabled file. Another acceptable fix would be clearing runtime state on delete before that entry can be reused as an update source.

excelwang:
## Summary

Fixes #2061.

When an auth file is deleted in the current upstream flow, the runtime entry is kept and marked disabled. If a file with the same identity is later recreated, `Update(...)` can currently inherit stale `ModelStates` from that removed/disabled entry.

This PR keeps the existing disable-based removal behavior and only changes the state inheritance rule:

- keep deletion as `disabled + Update`
- do not preserve `ModelStates` from an existing auth that is already disabled
- preserve the existing `ModelStates` inheritance behavior for normal in-place auth updates

## Why this approach

This addresses the original bug without changing the existing disable semantics that scheduler and management paths already rely on.

## Tests

- manager update still preserves model state for normal auth updates
- manager update does not preserve stale model state from a disabled auth entry
- management delete + re-add of the same auth file no longer carries stale runtime state into the recreated auth

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -928,16 +928,24 @@ func (h *Handler) disableAuth(ctx context.Context, id string) {
 	if id == "" {
 		return
 	}
-	if auth, ok := h.authManager.GetByID(id); ok && auth != nil {
-		_ = h.authManager.Remove(ctx, auth.ID)
+	if auth, ok := h.authManager.GetByID(id); ok {
+		auth.Disabled = true
+		auth.Status = coreauth.StatusDisabled
+		auth.StatusMessage = "removed via management API"
+		auth.UpdatedAt = time.Now()
+		_, _ = h.authManager.Update(ctx, auth)
 		return
 	}
 	authID := h.authIDForPath(id)
 	if authID == "" {
 		return
 	}
-	if _, ok := h.authManager.GetByID(authID); ok {
-		_ = h.authManager.Remove(ctx, authID)
+	if auth, ok := h.authManager.GetByID(authID); ok {
+		auth.Disabled = true
+		auth.Status = coreauth.StatusDisabled
+		auth.StatusMessage = "removed via management API"
+		auth.UpdatedAt = time.Now()
+		_, _ = h.authManager.Update(ctx, auth)
 	}
 }
 
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -928,24 +928,16 @@ func (h *Handler) disableAuth(ctx context.Context, id string) {
 	if id == "" {
 		return
 	}
-	if auth, ok := h.authManager.GetByID(id); ok {
-		auth.Disabled = true
-		auth.Status = coreauth.StatusDisabled
-		auth.StatusMessage = "removed via management API"
-		auth.UpdatedAt = time.Now()
-		_, _ = h.authManager.Update(ctx, auth)
+	if auth, ok := h.authManager.GetByID(id); ok && auth != nil {
+		_ = h.authManager.Remove(ctx, auth.ID)
 		return
 	}
 	authID := h.authIDForPath(id)
 	if authID == "" {
 		return
 	}
-	if auth, ok := h.authManager.GetByID(authID); ok {
-		auth.Disabled = true
-		auth.Status = coreauth.StatusDisabled
-		auth.StatusMessage = "removed via management API"
-		auth.UpdatedAt = time.Now()
-		_, _ = h.authManager.Update(ctx, auth)
+	if _, ok := h.authManager.GetByID(authID); ok {
+		_ = h.authManager.Remove(ctx, authID)
 	}
 }
 
--- a/internal/api/handlers/management/auth_files_delete_test.go
+++ b/internal/api/handlers/management/auth_files_delete_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/config"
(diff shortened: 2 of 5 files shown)


excelwang:
Concrete failure mode this fixes:

1. an auth enters cooldown / error state
2. the operator deletes the bad auth file
3. the operator adds a fresh replacement file with the same auth ID / filename
4. the fresh auth still inherits the deleted entry's stale runtime state and remains unschedulable until the old cooldown expires

So this is not just a cleanup issue. In production it means a healthy replacement credential can remain unavailable for minutes or hours after rotation, which is why I split this out as a standalone bugfix.

xkonjin (commented):
Quick review pass:
- Good to see test coverage move with the code; I’d still make sure it exercises the unhappy path around auth/session state and stale credential handling rather than only the happy path.
- Before merge, I’d smoke-test the behavior touched by auth_files.go, auth_files_delete_test.go, conductor.go (+2 more) with malformed input / retry / rollback cases, since that’s where this class of change usually breaks.

excelwang:
Follow-up on the auth/session state concern:

I re-ran targeted tests around the exact stale-state path this patch fixes. The failure mode here is not request parsing, but runtime auth metadata surviving deletion and being reused when the same auth file is added back.

Re-ran locally:
- `go test ./sdk/cliproxy/auth -run TestManager_Remove_DropsStaleModelStatesForRecreatedAuth -count=1`
- `go test ./internal/api/handlers/management -run TestDeleteAuthFile_AllowsReaddingSameNameWithoutStaleState -count=1`

The first covers the unhappy path where recreated auth would otherwise inherit stale model/cooldown state from the deleted entry. The second covers the management deletion flow end-to-end and verifies the same-name auth can be re-added and selected again instead of remaining stuck behind old runtime state.

So the risky path here is specifically exercised at both the manager layer and the management delete/re-add flow.

luispater (changes requested):
Summary

This fixes the stale ModelStates / cooldown reuse problem, but the merge result introduces a scheduler regression on current main.

Key findings

- Blocking: `Manager.Remove` deletes the auth from `m.auths`, but it never removes the auth from `m.scheduler`. On current main, `pickNext` / `pickNextMixed` use the scheduler fast path, so a deleted auth can stay selectable from cached model shards until a full scheduler rebuild happens.
- Non-blocking: the new tests only cover `GetByID` / stale `ModelStates`. Please add a regression test that removes an auth after it has already been scheduled and verifies `manager.pickNext` (or `scheduler.pickSingle`) can no longer return it.

Test plan

- `go test ./sdk/cliproxy/auth ./internal/api/handlers/management -count=1`

excelwang:
Follow-up pushed in `8aecc9fb`.

I pulled latest `main` again and re-checked the selection path around this change. In the current auth manager code path I could not find a separate `m.scheduler` cache; `pickNext` / `pickNextMixed` still select from `m.auths` directly. To cover the exact concern anyway, I added the requested regression test for the post-selection remove path:

- `TestManager_Remove_PreviouslySelectedAuthIsNoLongerPickable`

That test:
1. registers two auths for the same model
2. selects `auth-1` once through `pickNext`
3. removes `auth-1`
4. verifies the next `pickNext` can only return `auth-2`

Re-ran:
- `go test ./sdk/cliproxy/auth ./internal/api/handlers/management -count=1`

If there is another scheduler/cache path you want this exercised against, point me at the file and I can extend the regression there too.

excelwang:
Updated this PR to follow #2061 directly.\n\nThis no longer changes deletion semantics to a hard remove. It keeps the existing disable-based removal flow and only prevents  from inheriting stale  from an already-disabled auth entry when the same auth file is recreated.

excelwang:
Updated this PR to follow #2061 directly.

This no longer changes deletion semantics to a hard remove. It keeps the existing disable-based removal flow and only prevents Update from inheriting stale model state from an already-disabled auth entry when the same auth file is recreated.

luispater (approved):
Summary: The stale-state regression is fixed in `sdk/cliproxy/auth/conductor.go`. `Update(...)` now skips model-state inheritance when the existing auth is disabled, and the management delete path keeps the entry disabled intentionally instead of hard-removing it. The regression coverage in `sdk/cliproxy/auth/conductor_update_test.go` and `internal/api/handlers/management/auth_files_delete_test.go` matches that behavior.

Test plan:
- `go test ./sdk/cliproxy/auth ./internal/api/handlers/management -count=1`

luispater:

PR #1947, opened 2026-03-08
fix: use per-account proxy_url for OAuth token refresh

lichengzhe:
## Summary

- **ClaudeExecutor.Refresh()** now merges `auth.ProxyURL` (from credential JSON `proxy_url`) into config when global `proxy-url` is empty, so per-account proxy takes effect for token refresh
- **conductor.refreshAuth()** logs refresh failures at `WARN` level instead of `DEBUG`, making them visible without `debug: true`
- **config.example.yaml** documents proxy resolution priority and the caveat that OAuth refresh's custom TLS transport does not read `HTTPS_PROXY` env vars

## Problem

When deploying on a VPS with proxy routing via environment variables (`HTTPS_PROXY`), API requests work correctly because `newProxyAwareHTTPClient` uses standard `http.Transport` which reads env vars. However, OAuth token refresh uses `NewAnthropicHttpClient` with a custom utls transport that does **not** read env vars — it only reads `cfg.SDKConfig.ProxyURL`.

If `config.yaml` has no `proxy-url` field (relying on env vars) and the credential file has `proxy_url` set, `Refresh()` ignores `auth.ProxyURL` and creates a direct-connecting client. This causes silent refresh failures — tokens expire and auto-refresh silently fails because errors are logged at `DEBUG` level (invisible with `debug: false`).

Proxy resolution priority for API requests vs token refresh before this fix:

| Source | API requests | Token refresh |
|--------|-------------|---------------|
| `auth.ProxyURL` (credential file) | ✅ Priority 1 | ❌ Ignored |
| `cfg.ProxyURL` (config.yaml) | ✅ Priority 2 | ✅ Used |
| `HTTPS_PROXY` env var | ✅ Fallback | ❌ Not read by utls |

After this fix, token refresh also respects `auth.ProxyURL`.

## Changes

| File | Change |
|------|--------|
| `internal/runtime/executor/claude_executor.go` | Merge `auth.ProxyURL` into config copy when global proxy-url is empty |
| `sdk/cliproxy/auth/conductor.go` | Log refresh failures at WARN instead of DEBUG |
| `config.example.yaml` | Document proxy resolution priority |
| `.github/workflows/build-custom.yml` | CI workflow for fork branch builds |

## Test plan

- [x] Deploy with `proxy-url` empty in config.yaml, `proxy_url` set in credential JSON → verified token refresh succeeds through proxy
- [x] Deploy with global `proxy-url` set in config.yaml → verified it still takes effect (no regression)
- [x] Set `debug: false` and trigger a refresh failure → verified warning appears in logs

### DIFF
--- a/internal/auth/claude/anthropic_auth.go
+++ b/internal/auth/claude/anthropic_auth.go
@@ -53,16 +53,25 @@ type ClaudeAuth struct {
 // It initializes the HTTP client with a custom TLS transport that uses Firefox
 // fingerprint to bypass Cloudflare's TLS fingerprinting on Anthropic domains.
 //
+// An optional proxyURL may be provided to override the proxy in cfg.SDKConfig.
+// This allows callers to pass a pre-resolved proxy URL (e.g. from ResolveProxyURL)
+// so that both request and refresh paths share the same proxy priority logic.
+//
 // Parameters:
 //   - cfg: The application configuration containing proxy settings
+//   - proxyURL: Optional pre-resolved proxy URL that overrides cfg.SDKConfig.ProxyURL
 //
 // Returns:
 //   - *ClaudeAuth: A new Claude authentication service instance
-func NewClaudeAuth(cfg *config.Config) *ClaudeAuth {
-	// Use custom HTTP client with Firefox TLS fingerprint to bypass
-	// Cloudflare's bot detection on Anthropic domains
+func NewClaudeAuth(cfg *config.Config, proxyURL ...string) *ClaudeAuth {
+	sdkCfg := &cfg.SDKConfig
+	if len(proxyURL) > 0 && proxyURL[0] != "" {
+		cfgCopy := cfg.SDKConfig
+		cfgCopy.ProxyURL = proxyURL[0]
+		sdkCfg = &cfgCopy
+	}
 	return &ClaudeAuth{
-		httpClient: NewAnthropicHttpClient(&cfg.SDKConfig),
+		httpClient: NewAnthropicHttpClient(sdkCfg),
 	}
 }
 
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -580,13 +580,8 @@ func (e *ClaudeExecutor) Refresh(ctx context.Context, auth *cliproxyauth.Auth) (
 	}
 	// Use per-account proxy_url for token refresh, matching the priority in
 	// proxy_helpers.go: auth.ProxyURL > cfg.ProxyURL > env vars.
-	cfg := e.cfg
-	if auth.ProxyURL != "" {
-		cfgCopy := *cfg
-		cfgCopy.SDKConfig.ProxyURL = auth.ProxyURL
-		cfg = &cfgCopy
-	}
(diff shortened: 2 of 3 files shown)


lichengzhe:
@xkonjin Thanks for the review!

**Addressed: PR scope cleanup**

You're right that this PR had too much going on. I've rebased and removed the unrelated auto-update code that was accidentally included — it already has its own PR in #1942.

The diff is now **+68/-3 across 4 files**, focused purely on the proxy fix:

| File | Change |
|------|--------|
| `claude_executor.go` | Merge `auth.ProxyURL` into config when global `proxy-url` is empty |
| `conductor.go` | Elevate refresh failure logs from `DEBUG` → `WARN` |
| `config.example.yaml` | Document proxy resolution priority |
| `build-custom.yml` | Fork CI workflow (unrelated but lightweight) |

**Re: unhappy path testing**

The core change in `claude_executor.go` is a 7-line config-copy pattern — when `auth.ProxyURL != ""` and `cfg.SDKConfig.ProxyURL == ""`, a shallow copy is made with the proxy field set. The unhappy paths are:
- `auth.ProxyURL` empty → no-op, original `cfg` used (existing behavior)
- `cfg.SDKConfig.ProxyURL` already set → global takes precedence, no copy made
- Both set → global wins (intentional, documented in config.example.yaml)

These are straightforward conditional branches. The real validation is deployment-level: token refresh succeeds through the per-account proxy when no global proxy-url is configured. This has been verified on our production VPS.

**Re: malformed input / retry / rollback**

The `conductor.go` change is log-level only (no behavioral change), and `config.example.yaml` is documentation. Neither introduces new failure modes.

.github/workflows/build-custom.yml
luispater, line 1: Remove this file.

luispater (changes requested):

lichengzhe:
@luispater Thanks for the detailed review! You're right that the config-copy approach didn't properly align with the request path's proxy resolution.

I've pushed a fix that addresses the blocking issue:

**What changed:**

1. **Extracted `ResolveProxyURL()` in `proxy_helpers.go`** — single source of truth for the `auth.ProxyURL > cfg.ProxyURL > env` priority. Both `newProxyAwareHTTPClient` (request path) and the refresh path now call this same function.

2. **Added optional `proxyURL` param to `NewClaudeAuth`** — variadic parameter so callers can pass a pre-resolved proxy URL without mutating the config struct. Existing callers are unaffected.

3. **Simplified refresh path in `claude_executor.go`** — replaced the 6-line config-copy block with:
   ```go
   proxyURL := ResolveProxyURL(e.cfg, auth)
   svc := claudeauth.NewClaudeAuth(e.cfg, proxyURL)
   ```

This ensures request traffic and refresh traffic always go through the same proxy for a given credential.

Note: `build-custom.yml` was already removed in a previous commit. Would appreciate a re-review when you get a chance!

lichengzhe:
Superseded by fix/bun-tls-fingerprint — the TLS unification branch includes this proxy fix as part of a more comprehensive change (unifying all Anthropic connections to use Bun BoringSSL fingerprint + HTTP/1.1, which also consolidates the proxy resolution path).

PR #1958, opened 2026-03-08
fix: strip -customtools variant suffix for provider and auth lookup

thebtf:
## Summary
- Google's Gemini CLI appends `-customtools` to model names for a behavioral fine-tune variant. CPA doesn't recognize these suffixed names, causing **502 errors** for all Gemini CLI users when this suffix is active.
- Adds `stripModelVariantSuffix()` fallback at two points:
  1. `getRequestDetails()` in `handlers.go` — provider lookup
  2. `ClientSupportsModel()` in `model_registry.go` — auth selection
- The suffix is stripped **only** when the original name yields no matches, preserving existing behavior.

## Files changed
- `sdk/api/handlers/handlers.go` — variant suffix stripping in provider lookup
- `internal/registry/model_registry.go` — variant suffix stripping in client model support check

## Test plan
- [ ] Send request with model `gemini-2.5-pro-preview-05-06-customtools` through CPA
- [ ] Verify provider is resolved correctly (no 502)
- [ ] Verify standard model names without suffix still work as before

### DIFF
--- a/internal/registry/model_registry.go
+++ b/internal/registry/model_registry.go
@@ -757,6 +757,16 @@ func (r *ModelRegistry) ClientSupportsModel(clientID, modelID string) bool {
 		}
 	}
 
+	// Fallback: strip known model variant suffixes (e.g. "-customtools").
+	// These variants share the same provider as the base model.
+	if stripped := StripModelVariantSuffix(modelID); stripped != modelID {
+		for _, id := range models {
+			if strings.EqualFold(strings.TrimSpace(id), stripped) {
+				return true
+			}
+		}
+	}
+
 	return false
 }
 
--- a/internal/registry/model_variants.go
+++ b/internal/registry/model_variants.go
@@ -1,0 +1,53 @@
+// Package registry — model variant suffix normalization.
+//
+// Google's Gemini CLI appends behavioural fine-tune markers such as
+// "-customtools" to model names when it detects an API key. These variants
+// share the same upstream provider/executor as the base model, but several
+// CPA code paths — provider lookup, auth selection, OAuth /v1internal routing
+// — only know the base model name and therefore fail when the suffixed name
+// is propagated downstream.
+//
+// StripModelVariantSuffix removes these variant markers while preserving
+// thinking-budget suffixes like "(8192)" / "(high)" and chained segments like
+// "-thinking". It is the single source of truth for variant normalization
+// across the request path, auth selection path, and capability probes.
+package registry
+
+import "strings"
+
+// knownModelVariantSuffixes lists the variant markers (including the leading
+// dash) that are safe to strip for provider/auth lookup. Adding new entries
+// here automatically covers every call site through StripModelVariantSuffix.
+var knownModelVariantSuffixes = []string{"-customtools"}
+
+// StripModelVariantSuffix removes every known variant marker from model,
+// regardless of its position in the name, as long as the match falls on a
+// segment boundary ('-', '(', or end-of-string).
+//
+// Examples:
+//
+//	"gemini-3.1-pro-preview-customtools"         -> "gemini-3.1-pro-preview"
+//	"gemini-3.1-pro-preview-customtools(8192)"   -> "gemini-3.1-pro-preview(8192)"
+//	"gemini-3.1-pro-preview-customtools(high)"   -> "gemini-3.1-pro-preview(high)"
+//	"gemini-2.0-flash-customtools-thinking"      -> "gemini-2.0-flash-thinking"
+//	"gemini-2.0-flash-customtools-thinking(4096)" -> "gemini-2.0-flash-thinking(4096)"
+//	"gemini-2.5-pro"                             -> "gemini-2.5-pro" (unchanged)
+//
+// The function is idempotent and never panics on empty input.
+func StripModelVariantSuffix(model string) string {
+	if model == "" {
+		return model
+	}
+	result := model
+	for _, variant := range knownModelVariantSuffixes {
+		// Order matters: strip thinking-budget-adjacent occurrences first
+		// ("-customtools(8192)" -> "(8192)"), then chained segments
+		// ("-customtools-thinking" -> "-thinking"), then any terminal form.
+		result = strings.ReplaceAll(result, variant+"(", "(")
+		result = strings.ReplaceAll(result, variant+"-", "-")
+		if strings.HasSuffix(result, variant) {
+			result = strings.TrimSuffix(result, variant)
+		}
+	}
+	return result
+}
--- a/internal/registry/model_variants_test.go
+++ b/internal/registry/model_variants_test.go
@@ -1,0 +1,45 @@
+package registry
+
+import "testing"
+
+func TestStripModelVariantSuffix(t *testing.T) {
+	cases := []struct {
+		name, in, want string
+	}{
+		{"empty", "", ""},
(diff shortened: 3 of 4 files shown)


xkonjin (commented):
## Code Review

**Overall: ⚠️ Needs fix before merge — logic works in the common case but has a confirmed bug for thinking+customtools combinations.**

### What's good
- Fallback approach (strip suffix → retry provider lookup) is the right pattern instead of adding duplicates to model lists.
- Stripping from both  and  is correct — OAuth routes reject suffixed names, so the resolved name must be stripped too.
-  fallback in registry is clean and isolated.

### Confirmed bug (flagged by gemini-code-assist, needs fix)

**Ordering of suffix stripping for thinking models in ** — if a model name has *both* a thinking suffix and `-customtools` (e.g. `gemini-2.0-flash-thinking-exp-customtools`), `stripModelVariantSuffix` only strips `-customtools`, leaving `gemini-2.0-flash-thinking-exp`. That's correct. But if the suffix ordering is reversed (e.g. `gemini-2.0-flash-customtools-thinking`), the function returns the model unchanged because `HasSuffix` for `-customtools` fails. This is a latent ordering fragility.

The bigger issue: `resolvedModelName` may already have been transformed (thinking suffix stripped, version pinned, etc.) by the time we call `stripModelVariantSuffix` on it. Stripping from `resolvedModelName` without knowing its current form could produce an incorrect base name. Safer: strip from `baseModel` only for provider lookup, and strip from `modelID` input in registry — don't mutate `resolvedModelName` in the fallback path.

### Code duplication

The `knownModelVariantSuffixes` var and `stripModelVariantSuffix` function are copy-pasted verbatim between `internal/registry/model_registry.go` and `sdk/api/handlers/handlers.go`. Move to a shared util package and import in both.

### Non-blocking
- `stripModelVariantSuffix` iterates the suffix list on every call. With one entry this is negligible, but if the suffix list grows, pre-compile with a regex or strings.Replacer.
- Consider a unit test with a thinking+customtools composite model name to lock in the expected behavior.

xkonjin (commented):
## Code Review

**Overall: ⚠️ Needs fix before merge — logic works in the common case but has a confirmed bug for thinking+customtools combinations.**

### What's good
- Fallback approach (strip suffix, retry provider lookup) is the right pattern instead of adding duplicates to model lists.
- Stripping from both baseModel and resolvedModelName is correct — OAuth routes reject suffixed names, so the resolved name must be stripped too.
- ClientSupportsModel fallback in registry is clean and isolated.

### Confirmed bug (flagged by gemini-code-assist, needs fix)

**Ordering of suffix stripping for thinking models in handlers.go** — if a model name has both a thinking suffix and -customtools (e.g. gemini-2.0-flash-thinking-exp-customtools), stripModelVariantSuffix only strips -customtools, leaving gemini-2.0-flash-thinking-exp. That's correct for this ordering. But if the suffix ordering is reversed (e.g. gemini-2.0-flash-customtools-thinking), the function returns the model unchanged because HasSuffix for -customtools fails. Latent ordering fragility.

Bigger issue: resolvedModelName may already have been transformed (thinking suffix stripped, version pinned, etc.) before we call stripModelVariantSuffix on it. Stripping from resolvedModelName without knowing its current form could produce an incorrect base name. Safer: strip from baseModel only for provider lookup; strip from the raw modelID in registry — don't mutate resolvedModelName in the fallback path.

### Code duplication

knownModelVariantSuffixes and stripModelVariantSuffix are copy-pasted verbatim between internal/registry/model_registry.go and sdk/api/handlers/handlers.go. Move to a shared util package and import in both.

### Non-blocking
- stripModelVariantSuffix iterates the suffix list on every call. With one entry this is negligible, but if the list grows, consider pre-compiling.
- Add a unit test with a thinking+customtools composite model name to lock in the expected behavior and catch future regressions.

xkonjin (changes requested):
This fixes the simple `*-customtools` case, but it still misses one important path: models with a thinking suffix.

In `getRequestDetails()`, provider lookup is done against `baseModel`, so `gemini-foo-customtools(8192)` can now resolve its provider. But the code then normalizes the outbound model with `stripModelVariantSuffix(resolvedModelName)`. That call is against the full string, so when `resolvedModelName` still has a trailing thinking suffix, `HasSuffix("-customtools")` is false and the variant suffix is preserved. That means the downstream CPA/cloudcode-pa request can still receive the unsupported `-customtools` model name and 404.

I think the fix needs to strip the variant on the parsed model portion, then re-attach any thinking suffix, and it needs a regression test for something like `...-customtools(8192)`. Without that, the most failure-prone path still slips through.

luispater (changes requested):
Summary:
The main fallback direction looks right, but there is still one correctness gap before this is safe to merge.

Key findings:
- Blocking: `stripModelVariantSuffix()` only works when `-customtools` is the terminal suffix. We already support combined forms such as `gemini-...-customtools(high)` / `(auto)`, so those requests can still flow downstream with the variant suffix intact and still miss provider/auth lookup. Please add regression coverage for the combined-suffix case and normalize that path as well.
- Non-blocking: `/v1/responses` websocket capability probing still uses the old provider/model lookup path, so suffixed models there still take inconsistent fallback behavior.

Test plan:
- go test ./...

Follow-up:
- Reuse the same normalization helper across the request path, auth selection path, and websocket capability probe so the suffix handling stays consistent.

thebtf:
@luispater @xkonjin thanks for the review — the blocking correctness gap is now fixed, force-pushed to `fix/customtools-variant-suffix`.

### Changes

**Normalized strip handles every suffix position**, not only terminal:

```
gemini-3.1-pro-preview-customtools              -> gemini-3.1-pro-preview
gemini-3.1-pro-preview-customtools(8192)        -> gemini-3.1-pro-preview(8192)
gemini-3.1-pro-preview-customtools(high)        -> gemini-3.1-pro-preview(high)
gemini-3.1-pro-preview-customtools(auto)        -> gemini-3.1-pro-preview(auto)
gemini-2.0-flash-customtools-thinking           -> gemini-2.0-flash-thinking
gemini-2.0-flash-customtools-thinking(4096)     -> gemini-2.0-flash-thinking(4096)
gemini-2.0-flash-thinking-exp-customtools(8192) -> gemini-2.0-flash-thinking-exp(8192)
```

The match is anchored at a segment boundary (`-`, `(`, or end-of-string) so partial identifiers like `gemini-customtoolsextra` never strip.

**Deduplicated into a shared helper** as you requested. Single source of truth:

```
internal/registry/model_variants.go
  -> StripModelVariantSuffix(model string) string
```

Both `sdk/api/handlers/handlers.go` (`getRequestDetails`) and `internal/registry/model_registry.go` (`ClientSupportsModel`) now call this exported function instead of keeping local copies.

**Also addressed xkonjin's concern about `resolvedModelName` mutation**: the strip now only runs on `resolvedModelName` after the provider lookup against the stripped `baseModel` succeeds, so we never mutate a resolved name unless a real variant match was confirmed.

**Regression tests** in `internal/registry/model_variants_test.go` cover:
- terminal, thinking-budget (int/level/auto), chained segments, thinking-exp+customtools order
- idempotency (`Strip(Strip(x)) == Strip(x)`)
- no-partial-match guard (`gemini-customtoolsextra` stays unchanged)
- empty input

### Non-blocking follow-up

The `/v1/responses` websocket capability probe still uses the old lookup path. Happy to take that in a separate PR once the normalization helper is merged so the same `StripModelVariantSuffix` can be reused there.

PR #1960, opened 2026-03-08
feat: bypass proxy for private/LAN IPs in management api-call

thebtf:
## Summary
- When a global proxy is configured, the management `api-call` handler routes **all** requests through it — including requests to LAN targets (e.g. LM Studio at `172.x.x.x`). Since proxies typically cannot reach private networks, these requests **time out**.
- The Management UI "Pick Models from /models" feature becomes unusable for any LAN-hosted provider when a proxy is configured.
- Adds `isNoProxyHost()` that detects RFC 1918, loopback, and link-local addresses and skips the proxy for them.
- Only affects the management api-call handler; executor proxy logic is unchanged.

## Default bypass ranges
`127.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `::1/128`, `fc00::/7`, `fe80::/10`

## Files changed
- `internal/api/handlers/management/api_tools.go`

## Test plan
- [ ] Configure global proxy in config.yaml
- [ ] Use Management UI "Pick Models" with a LAN target (e.g. `http://192.168.1.100:1234/v1/models`)
- [ ] Verify models are fetched successfully (no timeout)
- [ ] Verify external targets still route through proxy

### DIFF
--- a/internal/api/handlers/management/api_tools.go
+++ b/internal/api/handlers/management/api_tools.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -175,7 +176,7 @@ func (h *Handler) APICall(c *gin.Context) {
 	httpClient := &http.Client{
 		Timeout: defaultAPICallTimeout,
 	}
-	httpClient.Transport = h.apiCallTransport(auth)
+	httpClient.Transport = h.apiCallTransport(auth, parsedURL.Host)
 
 	resp, errDo := httpClient.Do(req)
 	if errDo != nil {
@@ -283,7 +284,7 @@ func (h *Handler) refreshAntigravityOAuthAccessToken(ctx context.Context, auth *
 
 	httpClient := &http.Client{
 		Timeout:   defaultAPICallTimeout,
-		Transport: h.apiCallTransport(auth),
+		Transport: h.apiCallTransport(auth, ""),
 	}
 	resp, errDo := httpClient.Do(req)
 	if errDo != nil {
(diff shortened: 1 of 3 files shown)


xkonjin (commented):
## Code Review

**Overall: ✅ Approvable — correct approach, one robustness suggestion.**

### What's good
- Bypassing proxy for RFC 1918 / loopback / link-local addresses is the right fix. These ranges are unreachable via most external proxies and routing them through one silently fails.
- Passing `targetHost` explicitly rather than re-parsing inside the transport is clean.
- Correctly passing `""` for OAuth refresh calls (so they always use proxy — OAuth endpoints are always external).
- IPv6 ranges (::1/128, fc00::/7, fe80::/10) included alongside IPv4 — thorough.

### Issue: CIDR parsing on every request

`isNoProxyHost` calls `net.ParseCIDR` inside a loop on *every invocation*. With 8 CIDR strings this is fast but unnecessary work. Pre-parse `defaultNoProxyCIDRs` to `[]*net.IPNet` at init time (package-level `var` with `init()` or a sync.Once) and just iterate the pre-parsed networks. This is what gemini-code-assist flagged and it's worth doing.

### Minor notes
- `isNoProxyHost` only matches IP addresses — hostname-based LAN targets (e.g. `mymachine.local`, `raspberrypi`) return `false` and will be proxied. This is probably acceptable for now but worth documenting as a known limitation in a comment.
- No test for `isNoProxyHost` itself. Given it's a security boundary (bypasses proxy), a small table-driven test covering private IPs, public IPs, IPv6, and hostnames would be valuable.

xkonjin (changes requested):
Nice direction, but there is still a gap in the no-proxy detection that will break common LAN setups.

`isNoProxyHost()` only bypasses the proxy when `targetHost` is already a literal IP (`net.ParseIP(host)` succeeds). For typical management traffic to local devices, the target is often a hostname that resolves to a private address (`nas.local`, `router.lan`, mDNS names, internal DNS names, etc.). In those cases this PR still sends the request through the proxy, which is exactly the path this change is trying to avoid.

Please either:
- resolve hostnames before the CIDR check, or
- document and test that only literal IP targets are supported here.

I’d strongly prefer the former, plus a regression test covering a hostname that resolves to RFC1918 space.

luispater (changes requested):
Summary:
The overall direction makes sense, but there is still one correctness gap before this is safe to merge.

Key findings:
- Blocking: `isNoProxyHost()` only recognizes literal IPs. Requests to `http://localhost:...` still go through the configured proxy because `net.ParseIP("localhost")` returns nil, and bracketed IPv6 loopback URLs such as `http://[::1]/...` are missed as well. That means a common local-provider setup is still broken even though the PR claims to bypass loopback/LAN targets.
- Non-blocking: there is no regression coverage for the new bypass logic, so this kind of hole is easy to miss in CI.

Test plan:
- go test ./...

Follow-up:
- Please normalize the host with URL hostname semantics (or explicitly special-case `localhost`) and add regression coverage for `localhost`, literal private IPv4, and IPv6 loopback.

thebtf:
@luispater @xkonjin thanks for the review — both gaps are now fixed, force-pushed to `fix/no-proxy-private-ips`.

### Fixed

**Hostname resolution** — `isNoProxyHost` is no longer restricted to literal IPs:

- `localhost` is always treated as loopback (case-insensitive), regardless of DNS, so CI/test environments that do not map it to 127.0.0.1 still work.
- Non-literal hostnames (`nas.lan`, `printer.local`, custom DNS names) are resolved via `net.DefaultResolver.LookupIPAddr` with a **500ms context timeout**, so a missing or slow DNS entry cannot block the management request.
- **Every** resolved address must fall inside the configured no-proxy ranges. A single public address disables the bypass, so LAN DNS poisoning cannot hijack external traffic through the direct transport.
- Resolution failures leave the bypass disabled, so the proxy still has a chance to reach the target.

**IPv6 edge cases**:

- Bracketed form `[::1]` (both with and without port).
- Zone identifier `fe80::1%eth0` trimmed before parsing.
- ULA (`fd00::/8`) and link-local (`fe80::/10`) covered by the defaults.

**Also fixed**: the initial commit added `net.SplitHostPort` / `ParseIP` / `ParseCIDR` without importing `"net"` — the `management` package did not build with the patch applied against the current tree. The import is now present.

### Merge with the new API-key proxy resolution

Rebased on top of `ada8e290 feat(api): enhance proxy resolution for API key-based auth`. The no-proxy check short-circuits **before** the per-auth / per-API-key / global proxy candidate list, so the new `proxyURLFromAPIKeyConfig` logic is fully preserved for every non-bypassed call site.

`apiCallTransport` signature is now `(auth, targetHost)`. Only the `APICall` handler passes a real host (parsed URL); OAuth token refresh callers pass `""` and keep the always-proxy behaviour — loopback cannot leak to those paths.

### Regression tests

`api_tools_test.go` adds:

- `TestIsNoProxyHost` — 17 cases: private/public IPv4, IPv4 with port, IPv6 loopback bare/bracketed/with-port, IPv6 link-local with zone, IPv6 ULA, IPv6 public, literal `localhost` (case-insensitive, with/without port).
- `TestAPICallTransportBypassesProxyForPrivateIP` — LAN target reaches a direct transport even when a global proxy is configured.
- `TestAPICallTransportUsesProxyForPublicIP` — `api.openai.com:443` still routes through the global proxy.

Existing tests for the API-key candidate list continue to pass and call the new two-arg signature with `""`.

PR #1965, opened 2026-03-08
feat: add OpenAI-compatible Gemini embeddings endpoint

Gan-Xing:
## Summary
- add an OpenAI-compatible `/v1/embeddings` endpoint for Gemini and Gemini CLI auth flows
- expose `gemini-embedding-001` in `/v1/models` when Gemini auth is available
- add request parsing, dimensionality passthrough, and tests for single, batch, and base64-style embedding requests

## Testing
- `go test ./sdk/api/handlers/openai ./internal/api`

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -323,6 +323,7 @@ func (s *Server) setupRoutes() {
 		v1.GET("/models", s.unifiedModelsHandler(openaiHandlers, claudeCodeHandlers))
 		v1.POST("/chat/completions", openaiHandlers.ChatCompletions)
 		v1.POST("/completions", openaiHandlers.Completions)
+		v1.POST("/embeddings", openaiHandlers.Embeddings)
 		v1.POST("/messages", claudeCodeHandlers.ClaudeMessages)
 		v1.POST("/messages/count_tokens", claudeCodeHandlers.ClaudeCountTokens)
 		v1.POST("/responses", openaiResponsesHandlers.Responses)
@@ -345,6 +346,7 @@ func (s *Server) setupRoutes() {
 			"endpoints": []string{
 				"POST /v1/chat/completions",
 				"POST /v1/completions",
+				"POST /v1/embeddings",
 				"GET /v1/models",
 			},
 		})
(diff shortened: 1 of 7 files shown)


xkonjin (commented):
## Code Review

**Overall: ✅ Approvable — well-scoped implementation with solid test coverage. A few things worth tightening.**

### What's good
- Clean translation layer: OpenAI `/v1/embeddings` → Gemini `embedContent` (single) / `batchEmbedContents` (batch). Route split is correct.
- `parseEmbeddingInputs` handles both string and `[]string` input — matches OpenAI spec.
- `base64` encoding_format accepted and normalized to float response (keeps downstream callers working without silent breakage).
- `GetRequestDetails` export on `BaseAPIHandler` is a clean way to reuse the provider resolution logic without duplicating it.
- `SelectAuthForModel` in `conductor.go` is a useful addition — dry-run auth selection without executing a request is the right abstraction.
- Test coverage is good: single/batch input, base64 format, model list injection — the core paths are exercised.

### Issues

**1. `selectEmbeddingAuth` fallback hardcodes `gemini-embedding-001`** — the fallback path after `GetRequestDetails` fails only checks for this one model name. Any future embedding model would silently fall through to the error path even if a valid Gemini auth exists. Consider widening the fallback to check if the model is in the embedding models list from the registry (`SupportedGenerationMethods` includes `embedContent`), rather than a hardcoded string comparison.

**2. `dimensions` field is parsed but never used** — `openAIEmbeddingsRequest.Dimensions` is deserialized but not passed to the Gemini API call or validated. The Gemini embedding API does not support output dimension truncation (unlike OpenAI text-embedding-3-*). Either reject `dimensions != 0` with a clear error, or document the field as silently ignored. Silently accepting it gives callers false confidence.

**3. `StartNonStreamingKeepAlive` pattern with embeddings** — embeddings are inherently non-streaming and typically fast. If the keep-alive sends periodic bytes, it may corrupt the JSON response body on slow connections if the response write races with keep-alive writes. Confirm `StartNonStreamingKeepAlive` is safe to call before the response body is written (i.e., it only touches headers/flush, not the body).

**4. No test for `dimensions` rejection or passthrough** — given point 2 above, add a test case for `dimensions: 512` to document and lock in the current behavior.

### Non-blocking
- `geminiGenerativeLanguageBaseURL` is a package-level constant. If the base URL ever becomes configurable (e.g., regional endpoints, enterprise), this will need to be plumbed through. Low priority now, but worth a comment.
- `hasProviderAuth` iterates all auths on every `Models()` call — fine at current scale, but could be cached if `Models()` becomes hot.

luispater:
No plans to support this.

PR #2002, opened 2026-03-09
Fix gitstore: keep commit history & allow custom git-branch

snoyiatk:
## Problem

- `GitTokenStore.commitAndPushLocked` rewrites every commit into an orphan (parentless) commit via `rewriteHeadAsSingleCommit`, then force-pushes to the remote. This destroys all repository history on every save/delete/persist operation, leaving only a single commit on the remote at all times.
- git branch can not be specified, multiple instance pointing to same git repo would cause history conflict

## Changes

- **Remove `rewriteHeadAsSingleCommit`** — the method created a new commit object with `ParentHashes: nil` and rewrote the branch ref, squashing all history into one commit
- **Remove the squash call** in `commitAndPushLocked` — commits now retain their natural parent chain
- **Remove `Force: true`** from `Push` — force push was only needed because orphan commits diverge from remote history; normal fast-forward pushes work correctly now
- add support for specifying git branch (via GITSTORE_GIT_BRANCH) when using gitstore

## Impact

- Commits accumulate naturally with full history preserved
- No more force pushes overwriting the remote
- GC (`maybeRunGC`) is unchanged — it continues to prune loose objects on the regular interval

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -137,6 +137,7 @@ func main() {
 		gitStoreRemoteURL    string
 		gitStoreUser         string
 		gitStorePassword     string
+		gitStoreBranch       string = "main"
 		gitStoreLocalPath    string
 		gitStoreInst         *store.GitTokenStore
 		gitStoreRoot         string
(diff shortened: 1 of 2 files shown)


hkfires:
Not preserving the Git history was intentional.

PR #2013, opened 2026-03-10
Fetch model catalog from network

hkfires:


sdk/cliproxy/service.go
```diff
@@ -829,7 +829,22 @@ func (s *Service) registerModelsForAuth(a *coreauth.Auth) {
...
 	case "codex":
-		models = registry.GetOpenAIModels()
+		codexPlanType := ""
+		if a.Attributes != nil {
+			codexPlanType = strings.TrimSpace(a.Attributes["plan_type"])
+		}
+		switch strings.ToLower(codexPlanType) {
+		case "pro":
+			models = registry.GetCodexProModels()
+		case "plus":
```
excelwang, line 839: Defaulting missing or unknown `plan_type` to `pro` looks dangerous here. This makes any Codex auth with missing / unparsable tier metadata advertise `pro`-only models such as `gpt-5.4` even though the account may actually be `team` / `free`. Because the tier is reconstructed from auth metadata on reload, this is not just a theoretical case for older or hand-authored files. I would strongly prefer a safe fallback here (lowest/common tier, or no tier-only expansion until the tier is known) plus a regression test for the missing-tier path.

internal/registry/model_updater.go
```diff
@@ -0,0 +1,198 @@
...
+	if err := loadModelsFromBytes(embeddedModelsJSON, "embed"); err != nil {
+		panic(fmt.Sprintf("registry: failed to parse embedded models.json: %v", err))
+	}
+}
+
+// StartModelsUpdater runs a one-time models refresh on startup.
+// It blocks until the startup fetch attempt finishes so service initialization
+// can wait for the refreshed catalog before registering auth-backed models.
+// Safe to call multiple times; only one refresh will run.
+func StartModelsUpdater(ctx context.Context) {
```
excelwang, line 49: Blocking startup on a network fetch here is a real operational regression. In the worst case this now waits up to roughly 60s across the two URLs before the server is fully up, and this client does not use the proxy settings that the rest of CLIProxyAPI honors. In environments that require an outbound proxy, startup will always pay that penalty and still fall back to the embedded catalog. I think this should either run asynchronously after startup, or use the existing config-aware HTTP transport so it behaves like the rest of the system.

.github/workflows/pr-test-build.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Refresh models catalog
```
excelwang, line 15: Refreshing `models.json` from the network during CI means the PR is no longer tested against the catalog content actually committed in this branch. The same commit can produce different results on different days depending on the external file state. That makes CI and releases non-reproducible. I would rather pin the catalog to a reviewed file in the repo (or at least a specific content SHA) and update it through a separate refresh workflow / PR.

internal/registry/model_definitions.go
```diff
@@ -35,7 +160,7 @@ func GetStaticModelDefinitionsByChannel(channel string) []*ModelInfo {
 	case "aistudio":
 		return GetAIStudioModels()
 	case "codex":
```
excelwang, line 162: The catalog is now split into `codex-free/team/plus/pro`, but the static lookup path still treats `codex` as `pro` only. That means the management static definitions endpoint and `LookupStaticModelInfo()` will drift from the new tiered catalog design as soon as there is a model that exists only in a non-pro tier or differs across tiers. I think the static lookup path should either merge all Codex tiers for lookup purposes or be made explicitly tier-aware too.

PR #2024, opened 2026-03-10
fix(codex): harden tier-aware model catalog refresh

excelwang:
## Summary
This is a follow-up hardening PR for #2013.

It keeps the network-backed Codex model catalog direction, but fixes the parts that make model visibility and routing unsafe or inconsistent.

## What this changes
- adds a dedicated Codex tier catalog with embedded fallback data
- refreshes the Codex catalog asynchronously instead of blocking service startup
- honors configured outbound proxy settings during catalog refresh
- restores `plan_type` from Codex OAuth metadata / `id_token` and persists it back to auth files
- applies tier-aware model registration only when `plan_type` is actually known
- falls back to the full Codex union model set when `plan_type` is unknown, preserving manual model routing compatibility
- makes static `codex` model lookup search the full Codex union instead of a single implicit tier

## Why this is needed
After #2013, the model catalog direction is correct, but there are still a few sharp edges:

1. Codex OAuth auths need tier-aware registration, otherwise low-tier accounts can expose higher-tier models.
2. Codex API keys do not have `chatgpt_plan_type`, so applying OAuth tier logic to them breaks existing manual model selection in config.
3. Startup should not block on external catalog fetches.
4. Static Codex model lookup should be closed over all Codex tiers, not just one bucket.

This PR fixes those issues without changing the overall network-catalog design.

## Behavior after this PR
- explicit `codex.models[]` config still has highest priority
- known Codex OAuth `plan_type` => tier-aware registration
- unknown `plan_type` => full Codex union fallback
- Codex API keys without explicit `models[]` keep manual model-name compatibility

## Tests
Added coverage for:
- recovering `plan_type` from Codex `id_token`
- tier-aware Codex model registration for free/team/plus
- unknown-plan fallback to union
- API-key fallback to union when no explicit `models[]` are configured
- explicit `models[]` overriding union fallback


luispater (changes requested):
Summary

Thanks for the hardening follow-up. The tier-aware registration and fallback direction looks right, but I found one blocking issue and one compatibility issue before merge.

Key findings
- Blocking: startup is still synchronously blocked by `registry.StartModelsUpdater(...)` in `cmd/server/main.go`, so the new async Codex refresh does not actually remove the existing network wait on process startup.
- Non-blocking: plan aliases like `business`, `enterprise`, and `edu` are normalized to `team` for routing, but device login still persists the raw JWT plan into the credential filename and metadata, which can create inconsistent filenames and skip the team-account hash naming path.

Test plan
- go test ./...

excelwang:
Addressed the two blocking points from the latest review.

What changed
- Rebased the branch onto current `dev` so the follow-up now applies to the code path introduced by #2013.
- Removed the separate codex-only catalog store/updater and now reuse the shared `models.json` catalog already introduced on `dev`.
- Moved startup model refresh off the synchronous `cmd/server/main.go` path. `registry.StartModelsUpdater(...)` now runs once in the background and the service triggers a model rebind callback after the startup attempt finishes, so process startup is no longer blocked on the network fetch.
- Normalized Codex plan aliases like `business`, `enterprise`, `edu`, and `education` to `team` before persistence. This now applies to filename generation and to persisted `plan_type` metadata / attributes for device login and the management login flow.

Tests
- `go build -o /tmp/cliproxyapi-pr2024-test ./cmd/server`
- `go test ./... -count=1`

Additional coverage added
- startup updater returns immediately and still invokes the callback after the startup attempt
- alias plan types normalize to `team` in metadata persistence
- device login record generation persists normalized `team` and uses the team hashed filename path

Please take another look when convenient.

excelwang:
Follow-up on xkonjin's review:

I picked up the two actionable points from that comment set.

Updated
- `sdk/auth/filestore.go`: the codex plan normalization write-back on read no longer silently ignores persistence failures. It now logs explicit warnings when the normalized metadata cannot be marshaled or written back.
- `internal/registry/model_updater.go`: `StartModelsUpdater(...)` no longer has an implicit first-callback-only behavior. Multiple callers can now register callbacks for the same one-time startup refresh attempt, and each callback fires after that attempt completes. I also added coverage for callbacks registered after the refresh has already started.

On the nil-check point: `EnsureCodexPlanTypeMetadata(...)` already had a nil guard before this follow-up, so I did not add a redundant change there.

Validation
- `go test ./internal/registry ./sdk/auth -count=1`
- `go test ./... -count=1`

luispater (approved):
Summary: The previous blockers are resolved. The startup refresh is now kicked off asynchronously from `sdk/cliproxy/service.go`, and Codex plan aliases are normalized before filename generation and auth persistence in `internal/auth/codex/filename.go`, `sdk/auth/codex_device.go`, `sdk/auth/filestore.go`, and `internal/watcher/synthesizer/file.go`. The service also falls back to the Codex union when no plan type is known, which preserves API-key compatibility.

Verification: `go test ./internal/registry ./sdk/auth ./sdk/cliproxy/...` and `go build -o test-output ./cmd/server && rm -f test-output`

luispater:
This PR is approved and already targets `dev`, but it currently has merge conflicts. Please resolve the conflicts, keep the branch ready, and wait for merge.

PR #2052, opened 2026-03-11
fix(auth): persist refresh status/message and reduce redundant refresh attempts

xx299x:
- Persist a standardized `refresh_message` (and `refresh_status`) when Codex token refresh fails, so `main` exposes consistent, actionable failure context via the management API/UI.
- Avoid repeated refresh work after a permanent failure by short-circuiting subsequent refresh attempts, improving stability and reducing wasted concurrent refresh traffic on `main`.

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -332,6 +332,21 @@ func (h *Handler) listAuthFilesFromDisk(c *gin.Context) {
 				emailValue := gjson.GetBytes(data, "email").String()
 				fileData["type"] = typeValue
 				fileData["email"] = emailValue
+				if pv := gjson.GetBytes(data, "priority"); pv.Exists() {
+					switch pv.Type {
+					case gjson.Number:
+						fileData["priority"] = int(pv.Int())
+					case gjson.String:
+						if parsed, errAtoi := strconv.Atoi(strings.TrimSpace(pv.String())); errAtoi == nil {
+							fileData["priority"] = parsed
+						}
+					}
+				}
+				if nv := gjson.GetBytes(data, "note"); nv.Exists() && nv.Type == gjson.String {
+					if trimmed := strings.TrimSpace(nv.String()); trimmed != "" {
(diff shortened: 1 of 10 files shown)


xkonjin:
## Code Review

Solid improvement to Codex refresh error classification and persistence. Found two blocking issues.

### Blocking Issues

**1. Inconsistent metadata persistence across backends**

`refresh_status` / `refresh_message` are injected into `auth.Storage` in `FileTokenStore`:

```go
auth.Metadata["refresh_status"] = status
auth.Metadata["refresh_message"] = msg
saveTokenToFile(auth.Storage)
```

But other backends write `auth.Storage` directly:
- `GitTokenStore.SaveToken(auth.Storage)`
- `ObjectTokenStore.Put(auth.Storage)`
- `PostgresStore.SaveToken(auth.Storage)`

These backends will DROP the new metadata instead of persisting it.

Fix: Ensure all backends include `auth.Metadata` when persisting, or unify persistence through a common layer.

**2. Permanent failures treated as successful refreshes**

`CodexExecutor.Refresh`:

```go
if isNonRetryableRefreshError(err) || auth.RefreshStatus == RefreshStatusPermanentFailure {
    return auth, nil  // Returns nil error!
}
```

`Manager.refreshAuth`:

```go
auth, err := executor.Refresh(ctx, cred.Auth)
if err == nil {  // nil error = success
    cred.Auth.LastRefreshedAt = now
    cred.Auth.LastError = ""
    saveCredential(...)
}
```

Problems:
1. Permanent failures update `LastRefreshedAt` as if successful, breaking its documented meaning
2. Failed auths will keep re-entering the refresh loop every `refresh-interval` since `LastError` is cleared
3. The persisted "permanent failure" state gets overwritten with "success"

Fix: Return an error from `Refresh` for non-retryable/permanent failures, so `Manager.refreshAuth` correctly handles them.

### Summary

The direction is right, but these two issues break the durability and semantics of refresh failure tracking. Fix persistence consistency and error handling before merge.

PR #2063, opened 2026-03-11
feat: add configurable messages-path for Claude upstream (fixes #2055)

admccc:
## Summary

Add a configurable `messages-path` field to `ClaudeKey` so that relay services exposing Claude-compatible APIs at non-standard paths (e.g., new-api, one-api) can be configured without hardcoding `/v1/messages`.

### Root Cause

The executor hardcoded `/v1/messages` appended to `base-url`, causing `…/v1/v1/messages` when relay services already include `/v1` in their base URL (e.g., `https://relay.example.com/v1`).

### Solution

1. **`messages-path` config field** — users can set a custom path per Claude key; defaults to `/v1/messages`.
2. **`normalizeBaseURL()`** — strips trailing `/v1` from `base-url` before combining with `messages-path`, preventing the double-path issue automatically.
3. **Path normalization** — `messages-path` values are normalized at every entry point (synthesizer, management API): leading `/` is enforced and trailing `/` is stripped, so values like `messages`, `v1/messages`, or `/messages/` all produce correct URLs.

### Changes

| File | Change |
|---|---|
| `internal/config/config.go` | Add `MessagesPath` field to `ClaudeKey` |
| `internal/watcher/synthesizer/config.go` | Propagate normalized `messages-path` into auth attributes |
| `internal/runtime/executor/claude_executor.go` | Add `normalizeBaseURL()` + normalize `claudeMessagesPath()` output; apply to all 3 URL construction sites |
| `internal/runtime/executor/claude_executor_test.go` | Add `TestNormalizeBaseURL`; expand `TestClaudeMessagesPath` with normalization cases |
| `internal/api/handlers/management/config_lists.go` | Normalize `messages-path` in `normalizeClaudeKey()` and `PatchClaudeKey` |
| `internal/tui/keys_tab.go` | Display `[path: …]` in keys tab when `messages-path` is set |
| `config.example.yaml` | Document `messages-path` field with usage examples |

### Example configuration

```yaml
claude:
  - api-key: sk-ant-xxxxx
    base-url: https://relay.example.com/v1   # trailing /v1 is stripped automatically
    # messages-path defaults to /v1/messages → final URL: https://relay.example.com/v1/messages

  - api-key: sk-ant-yyyyy
    base-url: https://new-api.example.com
    messages-path: /messages                  # custom path for non-standard relay
```

## Test plan

- [x] `TestNormalizeBaseURL` — covers `/v1` stripping, nested paths, non-matching suffixes
- [x] `TestClaudeMessagesPath` — covers nil auth, missing leading slash, trailing slash, whitespace
- [x] Existing tests unaffected
- [x] Manual verification: relay with `base-url` ending in `/v1` produces correct URL

Fixes: https://github.com/router-for-me/CLIProxyAPI/issues/2055

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -143,6 +143,11 @@ nonstream-keepalive-interval: 0
 #   - api-key: "sk-atSM..."
 #     prefix: "test" # optional: require calls like "test/claude-sonnet-latest" to target this credential
 #     base-url: "https://www.example.com" # use the custom claude API endpoint
+#     messages-path: "/messages"           # optional: custom messages API path (default: "/v1/messages")
+#                                          # use this when your relay service (e.g., new-api) does not
+#                                          # use the /v1 prefix. Example: set to "/messages" if the relay
+#                                          # exposes the endpoint at {base-url}/messages instead of
+#                                          # the standard {base-url}/v1/messages
 #     headers:
 #       X-Custom-Header: "custom-value"
 #     proxy-url: "socks5://proxy.example.com:1080" # optional: per-key proxy override
--- a/internal/api/handlers/management/config_lists.go
+++ b/internal/api/handlers/management/config_lists.go
@@ -979,7 +979,11 @@ func normalizeClaudeKey(entry *config.ClaudeKey) {
 	}
 	entry.APIKey = strings.TrimSpace(entry.APIKey)
 	entry.BaseURL = strings.TrimSpace(entry.BaseURL)
-	entry.MessagesPath = strings.TrimSpace(entry.MessagesPath)
+	mp := strings.TrimSpace(entry.MessagesPath)
+	if mp != "" && !strings.HasPrefix(mp, "/") {
+		mp = "/" + mp
+	}
+	entry.MessagesPath = strings.TrimRight(mp, "/")
 	entry.ProxyURL = strings.TrimSpace(entry.ProxyURL)
 	entry.Headers = config.NormalizeHeaders(entry.Headers)
 	entry.ExcludedModels = config.NormalizeExcludedModels(entry.ExcludedModels)
--- a/internal/api/handlers/management/config_lists.go
+++ b/internal/api/handlers/management/config_lists.go
@@ -274,6 +274,7 @@ func (h *Handler) PatchClaudeKey(c *gin.Context) {
 		APIKey         *string               `json:"api-key"`
 		Prefix         *string               `json:"prefix"`
 		BaseURL        *string               `json:"base-url"`
+		MessagesPath   *string               `json:"messages-path"`
 		ProxyURL       *string               `json:"proxy-url"`
 		Models         *[]config.ClaudeModel `json:"models"`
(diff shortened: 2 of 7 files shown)


luispater (changes requested):
Summary:
This change moves Claude upstream routing in the right direction, but there are still compatibility gaps in the final URL construction. I found one blocking issue that means the original `/v1/v1/messages` failure mode is still possible, plus one normalization gap on the new `messages-path` field.

Key findings:
- Blocking: `internal/runtime/executor/claude_executor.go` now concatenates `baseURL` and `messages-path` directly. If a user keeps the common relay configuration from issue #2055 (`base-url: https://relay.example.com/v1`) and relies on the default path, the final URL is still `.../v1/v1/messages`. That means the original bug is not actually fixed for existing configs. The new test only checks path selection, not the final URL assembly, so this regression is currently untested.
- Non-blocking but important: `messages-path` is only `TrimSpace`'d. Values like `messages`, `v1/messages`, or `/messages/` will produce malformed URLs such as `https://hostmessages`, `https://hostv1/messages`, or `.../messages//count_tokens`. Since this is a new user-facing config field, it should be normalized (at least enforce a leading `/` and trim a trailing `/`) and covered by tests.

Test plan:
- Reviewed the final diff in a clean PR worktree.
- Ran `go test ./internal/runtime/executor`.
- Ran `go test ./internal/watcher/synthesizer ./internal/api/handlers/management ./internal/tui`.
- Ran `go build -o test-output ./cmd/server && rm test-output`.

xkonjin:
## Code Review

Good direction on supporting non-standard Claude relay paths. The messages-path config field and claudeMessagesPath() helper are clean and well-tested.

### Issues

**Blocking: URL construction does not fix the original bug**

The concatenation in claude_executor.go still produces double /v1 for existing configs:

```go
url := fmt.Sprintf("%s%s?beta=true", baseURL, claudeMessagesPath(auth))
```

If user has base-url: https://relay.example.com/v1 (from #2055) and uses default messages-path, final URL becomes:
https://relay.example.com/v1/v1/messages?beta=true

The original issue persists. The new test only checks claudeMessagesPath() behavior, not final URL assembly.

**Non-blocking: Normalize messages-path field**

messages-path is only TrimSpace-ed, so these malformed inputs will break:
- "messages" → https://hostmessages (missing /)
- "v1/messages" → https://hostv1/messages (missing /)
- "/messages/" → .../messages//count_tokens (trailing /)

Enforce leading / and trim trailing / during normalization, and add test coverage for URL construction with various base-url + messages-path combos.

### Summary

Approve with fixes to (1) prevent double /v1 for existing configs and (2) normalize path format.

luispater (approved):
Summary: The previous blockers are resolved. The executor now combines `normalizeBaseURL()` with the configurable `messages-path`, and the management/synthesizer paths normalize the field consistently before it reaches the runtime.

Test plan:
- `go test ./internal/runtime/executor ./internal/api/handlers/management ./internal/watcher/synthesizer ./internal/tui`

luispater:

PR #2066, opened 2026-03-11
fix(management): revive codex auth after quota refresh

### LINKED ISSUE #2065: bug(management): quota refresh does not revive codex auth after official usage recovers
## Summary

Refreshing Codex quota from the management quota page fetches the latest official usage successfully, but it does not revive auth entries that are currently blocked by `usage_limit_reached` cooldown.

## Expected behavior

When the management quota refresh confirms that a Codex auth is no longer exhausted, the runtime auth state should be updated so the auth can be scheduled again immediately.

## Actual behavior

The management quota page can fetch the latest official usage via the management `api-call` path, but the auth remains:

- `status=error`
- `unavailable=true`
- `quota.exceeded=true`
- `next_retry_after` unchanged

As a result, the auth still cannot be selected until the old cooldown expires, even though the official usage response already shows the credential is available again.

## Reproduction

1. Let a Codex auth hit `usage_limit_reached` so it enters cooldown.
2. Open `management.html#/quota`.
3. Click the refresh button for that auth/quota card.
4. Observe that the page can fetch the latest official usage successfully.
5. Check `GET /v0/management/auth-files` for that auth.

## Current result

The auth stays blocked in runtime state.

## Expected result

If the refreshed official usage clearly shows the auth is no longer rate-limited, the runtime auth state should be revived immediately instead of waiting for the previous cooldown deadline.

excelwang:
Fixes #2065

## Summary

When the management quota page refreshes Codex usage successfully and the official `wham/usage` response shows the credential is available again, the runtime auth state should be cleared immediately instead of waiting for the old cooldown to expire.

## What this changes

- detect successful Codex `wham/usage` refreshes in management `api-call`
- conservatively revive the auth only when the official response clearly indicates the credential is schedulable again
- clear runtime quota/cooldown state and refresh the scheduler snapshot
- add regression tests for both revive and still-blocked cases

## Notes

This does not change normal request-time quota handling. It only lets the management quota refresh path bring runtime state back in sync after official usage has already recovered.

### DIFF
--- a/internal/api/handlers/management/api_tools.go
+++ b/internal/api/handlers/management/api_tools.go
@@ -209,6 +209,8 @@ func (h *Handler) APICall(c *gin.Context) {
 		return
 	}
 
+	h.reviveAuthFromOfficialUsage(c.Request.Context(), auth, parsedURL, resp.StatusCode, respBody)
+
 	c.JSON(http.StatusOK, apiCallResponse{
 		StatusCode: resp.StatusCode,
 		Header:     resp.Header,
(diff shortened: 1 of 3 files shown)


luispater:
No plans to support this.

Any quota-related features won't be merged.

PR #2084, opened 2026-03-12
perf(auth): reduce P2 watcher, scheduler, and refresh overhead

shenshuoyaoyouguang:
## Summary
- replace watcher DeepEqual auth comparisons with cached semantic fingerprints so non-semantic field churn no longer emits modify updates
- add a due-at refresh index so auto refresh scans only auths that are ready to refresh instead of snapshotting the full pool each tick
- add a model-scoped scheduler update path for MarkResult so model-specific failures only rebuild the current shard and existing fallback indexes
- add regression and benchmark coverage for watcher fingerprinting, refresh scheduling, and model-scoped cooldown updates

## Testing
- go test ./...
- go vet ./...


shenshuoyaoyouguang:
Closing because PR quality is poor.

PR #2087, opened 2026-03-12
fix(perf): reduce p1 hot-path blocking

shenshuoyaoyouguang:
## Summary
- move watcher auth-file heavy I/O out of the main fsnotify loop into per-path workers while keeping dispatch serialization intact
- propagate stream cancellation through shared sendChunk helpers for OpenAI-compatible, Kimi, and Qwen executors
- shrink Manager.MarkResult hot lock scope by persisting auth snapshots after in-memory/scheduler updates
- add regression coverage for watcher burst ordering, stream cancellation timing, and concurrent MarkResult + pickNext/pickNextMixed behavior

## Impacted Areas
- internal/watcher
- internal/runtime/executor
- sdk/cliproxy/auth

## Verification
- go test ./internal/watcher ./internal/runtime/executor ./sdk/cliproxy/auth
- go test ./...
- go vet ./...

## Notes
- keeps cooldown, fair rotation, Gemini virtual parent handling, and Codex websocket preference semantics unchanged
- packaged as the P1 hot-path mitigation slice for low-risk rollback

### DIFF
--- a/internal/runtime/executor/kimi_executor.go
+++ b/internal/runtime/executor/kimi_executor.go
@@ -271,17 +271,21 @@ func (e *KimiExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.Aut
 			}
 			chunks := sdktranslator.TranslateStream(ctx, to, from, req.Model, opts.OriginalRequest, body, bytes.Clone(line), &param)
 			for i := range chunks {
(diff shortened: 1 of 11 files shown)


shenshuoyaoyouguang:
Closing because PR quality is poor.

PR #2121, opened 2026-03-13
Reconcile registry model states on auth changes

destinoantagonista-wq:
Add Manager.ReconcileRegistryModelStates to clear stale per-model runtime failures for models currently registered in the global model registry. The method finds models supported for an auth, resets non-clean ModelState entries, updates aggregated availability, persists changes, and pushes a snapshot to the scheduler. Introduce modelStateIsClean helper to determine when a model state needs resetting. Call ReconcileRegistryModelStates from Service paths that register/refresh models (applyCoreAuthAddOrUpdate and refreshModelRegistrationForAuth) to keep the scheduler and global registry aligned after model re-registration.


luispater (changes requested):
Summary
This moves model-state cleanup into the auth re-registration paths, which is the right direction. However, there is still a stale-state hole when the failed model disappears or is renamed during registry refresh.

Key findings
- Blocking: `ReconcileRegistryModelStates()` only resets entries for models still returned by `GetModelsForClient()`. If the stale failure belongs to a model that was removed or renamed during refresh, the loop skips it, `changed` stays false, and auth-level availability/status is never recomputed. That leaves stale `Unavailable` / `StatusMessage` / `LastError` on the auth even though the refreshed registry snapshot is healthy. This still affects management output and websocket availability fallback paths.

Test plan
- Reviewed the changed paths in `sdk/cliproxy/auth/conductor.go` and `sdk/cliproxy/service.go`
- Ran `go test ./sdk/cliproxy/auth ./sdk/cliproxy/...`

Follow-ups
- I think this should either prune unsupported `ModelStates` during reconciliation or recompute aggregated auth state using only the currently supported registry models before persisting.

destinoantagonista-wq:
Updated this PR with the follow-up fixes from review.

What changed:
- prune unsupported ModelStates during reconciliation so renamed/removed models cannot keep stale auth-level status behind
- recompute and clear aggregated auth availability when reconciliation leaves no active model states
- stop swallowing persist errors in the reconcile path; persist failures now log a warning with request_id/auth_id context

Validation:
- go test ./sdk/cliproxy/auth ./sdk/cliproxy/...

I did not apply the canonicalModelKey simplification suggestion because canonicalModelKey already returns the trimmed raw model when suffix parsing does not change it, and the current branch intentionally prunes unsupported entries instead of skipping them.

luispater (commented):
Summary: The follow-up patch fixes the original stale-state bug. `ReconcileRegistryModelStates` now prunes unsupported model states, resets supported ones, and clears aggregate auth availability when nothing remains.

Blocking findings:
- None.

Non-blocking findings:
- The PR no longer includes the dedicated reconcile regression tests that were briefly added in the follow-up commit, so the renamed/removed-model edge cases are not backed by automated coverage in the final diff.
- The refresh path uses `context.Background()` for the new reconcile calls, so persistence there will not respect caller cancellation.

Test plan:
- `go test ./sdk/cliproxy/auth ./sdk/cliproxy/...`

PR #2124, opened 2026-03-14
feat(api): expose priority and note in GET /auth-files response

RGBadmin:
## 概述 / Summary

`GET /auth-files` 列表接口此前不返回凭证文件中的 `priority` 和 `note` 字段，导致前端无法在卡片上展示优先级和备注，也无法按优先级排序。本 PR 将这两个字段暴露到列表响应中，并扩展 `PATCH /auth-files/fields` 以支持写入 `note`。

The `GET /auth-files` list endpoint previously omitted the `priority` and `note` fields stored inside each auth file's JSON content. This meant the frontend could not display priority badges or notes on credential cards, nor sort by priority. This PR exposes both fields in the list response and extends `PATCH /auth-files/fields` to support writing the `note` field.

### 1. `buildAuthFileEntry()` — 正常路径 / Normal path
- 从 `auth.Attributes["priority"]`（合成阶段已写入）读取 priority，转为 int 返回
- 从 `auth.Metadata["note"]` 读取 note（string），非空时返回
- Reads `priority` from `auth.Attributes["priority"]` (already set by synthesizer), converts to int
- Reads `note` from `auth.Metadata["note"]` (string), includes when non-empty

### 2. `listAuthFilesFromDisk()` — 回退路径 / Fallback path
- 通过 gjson 从磁盘文件中额外读取 `priority` 和 `note` 字段
- Uses gjson to additionally read `priority` and `note` from on-disk JSON files

### 3. `synthesizeFileAuths()` — 合成阶段 / Synthesis
- 在 priority 读取之后，补充从 `metadata["note"]` 读取 note 并存入 `auth.Attributes["note"]`
- After the existing priority extraction, reads `note` from `metadata["note"]` into `auth.Attributes["note"]`

### 4. `PatchAuthFileFields()` — 字段写入 / Field update
- 请求结构体新增 `Note *string` 字段
- 空字符串时从 Metadata 中删除 note，非空时写入
- Adds `Note *string` to the request struct
- Empty string removes note from Metadata; non-empty string writes it

## 改动文件 / Changed files

| 文件 / File | 改动 / Change |
|---|---|
| `internal/api/handlers/management/auth_files.go` | 列表响应增加 priority/note；PATCH 支持 note 写入 |
| `internal/watcher/synthesizer/file.go` | 合成阶段读取 note 到 Attributes |

## 测试计划 / Test plan

- [ ] 含 `priority` 的凭证文件，列表接口返回 `priority` 整数值 / Auth files with `priority` return it as integer in list response
- [ ] 含 `note` 的凭证文件，列表接口返回 `note` 字符串 / Auth files with `note` return it as string in list response
- [ ] 不含 priority/note 的凭证文件，响应中不包含这两个字段 / Auth files without priority/note omit these fields
- [ ] `PATCH /auth-files/fields` 传入 `note` 可正常写入和清除 / `PATCH /auth-files/fields` with `note` writes and clears correctly
- [ ] auth manager 不可用时，回退路径同样返回 priority/note / Fallback disk-read path also returns priority/note when auth manager is unavailable

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -342,7 +342,7 @@ func (h *Handler) listAuthFilesFromDisk(c *gin.Context) {
 						}
 					}
 				}
-				if nv := gjson.GetBytes(data, "note"); nv.Exists() {
+				if nv := gjson.GetBytes(data, "note"); nv.Exists() && nv.Type == gjson.String {
 					if trimmed := strings.TrimSpace(nv.String()); trimmed != "" {
 						fileData["note"] = trimmed
 					}


luispater (changes requested):
Thanks for the PR. I'm requesting changes mainly around one correctness gap.

- Blocking: `GET /auth-files` still drops `note` for multi-project Gemini credentials. `buildAuthFileEntry()` only reads `note` from `auth.Metadata`, while the new synthesizer code stores it in `Attributes`, and `SynthesizeGeminiVirtualAuths()` does not copy `note` into the virtual auth metadata. Since the disabled primary auth is hidden and the virtual auths are what the UI actually lists, those cards will still not receive `note`.
- Non-blocking: the fallback disk-read path now uses `gjson.Result.Int()` for any existing `priority`, which means malformed string values become `0` instead of being omitted. It would be better to keep this behavior aligned with the normal auth-manager path.
- Tests: I ran `go test ./internal/watcher/synthesizer ./internal/api/handlers/management -count=1` in a detached worktree for PR `2124`, and both packages passed. I didn't find any new regression coverage for the added `note` behavior.

Once the Gemini virtual-auth path is handled and the parsing behavior/tests are tightened up, I'm happy to take another look.

PR #2128, opened 2026-03-14
fix: apiCallTransport fallback 应保留环境变量代理

zhoubeiqing:
## 问题

`apiCallTransport`（`api_tools.go`）在凭证和全局配置均未设置代理时，fallback 分支显式将 `Proxy` 设为 `nil`，禁用了环境变量代理（`HTTP_PROXY` / `HTTPS_PROXY`）。

而 `newProxyAwareWebsocketDialer`（`codex_websockets_executor.go`）在相同场景下保留了 `http.ProxyFromEnvironment` 作为默认行为。

两者行为不一致，导致管理面板通过 `/v0/management/api-call` 代理的请求（如获取 Codex 额度）在仅配置了环境变量代理的部署环境中无法到达上游服务，触发超时。

## 修复

当没有显式代理配置时，直接返回 `http.DefaultTransport`（默认包含 `ProxyFromEnvironment`），与 WebSocket 执行器保持一致。

```diff
- transport, ok := http.DefaultTransport.(*http.Transport)
- if !ok || transport == nil {
-     return &http.Transport{Proxy: nil}
- }
- clone := transport.Clone()
- clone.Proxy = nil
- return clone
+ return http.DefaultTransport
```

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -28,7 +28,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Generate Build Metadata
         run: |
-          echo VERSION=`git describe --tags --always --dirty` >> $GITHUB_ENV
+          echo "VERSION=${GITHUB_REF_NAME}" >> $GITHUB_ENV
           echo COMMIT=`git rev-parse --short HEAD` >> $GITHUB_ENV
           echo BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` >> $GITHUB_ENV
       - name: Build and push (amd64)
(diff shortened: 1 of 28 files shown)


zhoubeiqing (commented):
It looks good

zhoubeiqing:
> Thanks for the fix. The behavior change in `apiCallTransport` looks directionally correct and matches the websocket dialer by preserving `ProxyFromEnvironment` when no explicit proxy is configured.
> 
> I have one blocking concern before this should merge:
> 
> * Please add a regression test for the exact scenario this PR is fixing: when both `auth.ProxyURL` and `cfg.ProxyURL` are empty, `apiCallTransport` should inherit the environment proxy settings instead of disabling them. Right now `internal/api/handlers/management/api_tools_test.go` only covers `direct` and the invalid-auth-falls-back-to-global-proxy path, so the reported bug can regress again without any test failing.
> 
> Test plan I ran:
> 
> * `go test ./internal/api/handlers/management`
> * `go test ./internal/runtime/executor`
> * `go test ./sdk/proxyutil`
> 
> Once that coverage is in place, this looks good to me.

@luispater  Added the regression test for the no-explicit-proxy case in api_tools_test.go.
When both auth.ProxyURL and cfg.ProxyURL are empty, apiCallTransport now inherits environment proxy settings.
Could you take another look?

PR #2133, opened 2026-03-14
fix(auth): prevent stale runtime state inheritance from disabled auth entries

### LINKED ISSUE #2061: Re-created auth file inherits stale runtime state when re-added with the same path

DragonFSKY:
## Summary

- Gate runtime state inheritance (`ModelStates`, `LastRefreshedAt`, `NextRefreshAfter`) in both `Manager.Update` and `Service.applyCoreAuthAddOrUpdate` so that state is only carried over when **both** the existing and incoming auth are non-disabled
- Prevents re-created auth files from inheriting stale cooldown/backoff state from previously disabled entries

## Root Cause

When an auth file is deleted, `applyCoreAuthRemoval` marks the runtime entry as `Disabled=true` / `Status=StatusDisabled` but does not remove it from `coreManager`. When the same file path is re-added, the update branch unconditionally inherited `ModelStates` from the old disabled entry, causing the new auth to carry over stale cooldown/backoff state and become unroutable.

## Changes

- `sdk/cliproxy/auth/conductor.go` — Add disabled guard to `ModelStates` inheritance in `Manager.Update`
- `sdk/cliproxy/service.go` — Add disabled guard to `LastRefreshedAt`, `NextRefreshAfter`, and `ModelStates` inheritance in `Service.applyCoreAuthAddOrUpdate`
- `sdk/cliproxy/auth/conductor_update_test.go` — Add 4 test cases covering all disabled/active transition combinations

## Test plan

- [x] Disabled existing + Disabled incoming → no state inheritance
- [x] Active existing + Disabled incoming → no state inheritance
- [x] Disabled existing + Active incoming → no state inheritance
- [x] Active existing + Active incoming → normal inheritance preserved
- [x] Full `go test ./sdk/cliproxy/auth/` and `go test ./sdk/cliproxy/` pass with no regressions

Closes #2061

### DIFF
--- a/sdk/cliproxy/service_stale_state_test.go
+++ b/sdk/cliproxy/service_stale_state_test.go
@@ -1,0 +1,85 @@
+package cliproxy
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
+	coreauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
+	"github.com/router-for-me/CLIProxyAPI/v6/sdk/config"
+)
+
+func TestServiceApplyCoreAuthAddOrUpdate_DeleteReAddDoesNotInheritStaleRuntimeState(t *testing.T) {
+	service := &Service{
+		cfg:         &config.Config{},
+		coreManager: coreauth.NewManager(nil, nil, nil),
+	}
+
+	authID := "service-stale-state-auth"
+	modelID := "stale-model"
+	lastRefreshedAt := time.Date(2026, time.March, 1, 8, 0, 0, 0, time.UTC)
+	nextRefreshAfter := lastRefreshedAt.Add(30 * time.Minute)
+
+	t.Cleanup(func() {
+		GlobalModelRegistry().UnregisterClient(authID)
+	})
+
+	service.applyCoreAuthAddOrUpdate(context.Background(), &coreauth.Auth{
+		ID:               authID,
+		Provider:         "claude",
+		Status:           coreauth.StatusActive,
+		LastRefreshedAt:  lastRefreshedAt,
+		NextRefreshAfter: nextRefreshAfter,
+		ModelStates: map[string]*coreauth.ModelState{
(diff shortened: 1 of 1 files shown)


DragonFSKY:
Thanks for the review.

I added the suggested service-level regression test for the watcher delete -> re-add path, covering `LastRefreshedAt`, `NextRefreshAfter`, and stale `ModelStates` inheritance on re-add.

PR #2141, opened 2026-03-14
fix: skip empty assistant message in tool call translation (#2132)

### LINKED ISSUE #2132: OpenAI `chat.completions` 工具调用在路由到 Codex/Responses 时失败，报错 `No tool output found for   function call`
  [ ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received
  an unexpected error.
  [x] No, it's another issue.

  Not a manually crafted payload issue. This happens when using an OpenAI-compatible client through CLIProxyAPI.

  **Describe the bug**
  When using an OpenAI-compatible client with standard `chat.completions` tool calling, normal chat works, but
  tool calling fails after the model emits a function call.

  The client sends a normal OpenAI tool-calling conversation:
  1. `/v1/chat/completions` with `tools`
  2. assistant returns `tool_calls`
  3. client executes the tool
  4. client sends the next `/v1/chat/completions` request including:
     - the previous assistant message with `tool_calls`
     - a `role: "tool"` message with the matching `tool_call_id`

  After CLIProxyAPI routes this to Codex/Responses, upstream returns:

  ```json
  {
    "error": {
      "message": "No tool output found for function call call_StZoTqlGCVxf9egdCqNuAwU3.",
      "param": "input",
      "code": null
    }
  }

  The key point is that the error points to param: "input", which strongly suggests the issue happens during
  Chat Completions -> Responses/Codex translation, not at the original client payload level.

  From code reading, the suspected area is the OpenAI Chat Completions to Codex/Responses conversion path,
  especially:

  - sdk/api/handlers/openai/openai_handlers.go
  - internal/translator/codex/openai/chat-completions/codex_openai_request.go
  - internal/translator/codex/openai/chat-completions/codex_openai_response.go

  My suspicion is:

  - assistant tool_calls are translated into function_call
  - but the matching role:"tool" message is not always preserved/mapped correctly into function_call_output
  - or the translated call_id does not exactly match the one expected by upstream

  CLI Type
  codex

  Model Name
  gpt-5.4

  ```md
  **LLM Client**
  Obsidian YOLO plugin

  **Request Information**
  I do not currently have a sanitized curl payload ready.

  However, this is reproducible through Obsidian YOLO using OpenAI-compatible `/v1/chat/completions` tool
  calling.

  Normal chat works.
  Once tool calling is triggered, the follow-up request fails with:

  ```json
  {
    "error": {
      "message": "No tool output found for function call call_StZoTqlGCVxf9egdCqNuAwU3.",
      "param": "input",
      "code": null
    }
  }

  If needed, I can try to provide request logs later, but based on current behavior and code reading this looks
  like a translator issue rather than a malformed manually-written payload.

  Expected behavior
  Standard OpenAI chat.completions tool calling should continue to work when routed through CLIProxyAPI to
  Codex/Responses.

  If the client sends:

  - assistant tool_calls
  - followed by role:"tool" with matching tool_call_id

  then CLIProxyAPI should translate that correctly into Responses/Codex format so upstream accepts it.

  Screenshots
  N/A

  OS Type

  - OS: Windows
  - Version: Win11 

  Additional context
  I also checked the client-side plugin code (obsidian-yolo), and it appears to follow the normal OpenAI tool-
  calling flow:

  - sends tools
  - receives assistant.tool_calls
  - sends back role:"tool" with tool_call_id

  So this does not look like the plugin simply forgot to send tool output.

  The main reason I suspect CLIProxyAPI is:

  - the upstream error explicitly points to param: "input"
  - which suggests the problem appears after CLIProxyAPI converts chat-completions history into Responses-style
    input

  Possibly related files:

  - sdk/api/handlers/openai/openai_handlers.go
  - internal/translator/codex/openai/chat-completions/codex_openai_request.go
  - internal/translator/codex/openai/chat-completions/codex_openai_response.go

Muran-prog:
Fixes #2132

When assistant has tool_calls but null/empty `content`, the translator emitted an empty {type: "message", role: "assistant", content: []} into the Responses API input array before `function_call` items. The API couldn't match function_call_output to its function_call by `call_id`.

## Fix
Only emit assistant messages that have content parts - 4-line change:

if role != "assistant" || len(gjson.Get(msg, "content").Array()) > 0 {
    out, _ = sjson.SetRaw(out, "input.-1", msg)
}

Tests
9 new tests in codex_openai_request_test.go.

### DIFF
--- a/internal/translator/codex/openai/chat-completions/codex_openai_request_test.go
+++ b/internal/translator/codex/openai/chat-completions/codex_openai_request_test.go
@@ -1,7 +1,6 @@
 package chat_completions
 
 import (
-	"strings"
 	"testing"
 
 	"github.com/tidwall/gjson"
@@ -623,14 +622,9 @@ func TestToolsDefinitionTranslated(t *testing.T) {
 		t.Fatal("no tools found in output")
 	}
 
-	// look for "search" tool
 	found := false
 	for _, tool := range tools {
-		name := tool.Get("name").String()
-		if name == "" {
-			name = tool.Get("function.name").String()
-		}
-		if strings.Contains(name, "search") {
+		if tool.Get("name").String() == "search" {
 			found = true
 			break
 		}


Muran-prog:
I traced this through ConvertOpenAIRequestToCodex() - when the assistant message has tool_calls but null/empty content, an empty {type: "message", role: "assistant", content: []} gets emitted before the function_call items. The Responses API then can't match function_call_output to its function_call by call_id.

Muran-prog:
Fix with tests: #2141

Muran-prog:
Applied the suggestion - switched to exact match and dropped the function.name fallback.

Muran-prog:
Ready to merge - the translator-path-guard failure is expected since this touches translator code. Build passes.

Issue #1531, opened 2026-02-10
Invalid JSON payload received: Unknown name \"deprecated\"


iRbouh:
**Is it a request payload issue?**
[  ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.

**If it's a request payload issue, you MUST know**
```
API Error: 400 {"error":{"code":400,"message":"Invalid JSON payload received. Unknown name \"deprecated\" at
     'request.tools[0].function_declarations[45].parameters.properties[1].value': Cannot find field.\nInvalid JSON payload received. Unknown name \"deprecated\" at
     'request.tools[0].function_declarations[45].parameters.properties[2].value': Cannot find field.","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.BadRequ
     est","fieldViolations":[{"field":"request.tools[0].function_declarations[45].parameters.properties[1].value","description":"Invalid JSON payload received. Unknown name \"deprecated\"
     at 'request.tools[0].function_declarations[45].parameters.properties[1].value': Cannot find
     field."},{"field":"request.tools[0].function_declarations[45].parameters.properties[2].value","description":"Invalid JSON payload received. Unknown name \"deprecated\" at
     'request.tools[0].function_declarations[45].parameters.properties[2].value': Cannot find field."}]}]}}
```

**CLI Type**
Claude Code

**Model Name**
Antigravity claude-opus-4-6-thinking

**LLM Client**
Claude Code

**OS Type**
 - OS: Ubuntu
 - Version 24

ed222Light:
yes,I have the sam one.

0x0a0d:
try to disable mcp stitch

FadhliRajwaa:
test

Issue #1424, opened 2026-02-04
bug: Claude → Gemini translation fails due to unsupported JSON Schema fields ($id, patternProperties)


lonr-6:
## Bug Description

When using Claude API format with Gemini backend (e.g., `gemini-claude-opus-4-5-thinking`), the request fails with JSON Schema validation errors.

## Error Message

```
Invalid JSON payload received. Unknown name "$id" at 'request.tools[0].function_declarations[135].parameters.properties[0].value': Cannot find field.
Invalid JSON payload received. Unknown name "patternProperties" at 'request.tools[0].function_declarations[135].parameters.properties[0].value.properties[5].value': Cannot find field.
```

## Root Cause

The `ConvertClaudeRequestToGemini` function in `internal/translator/gemini/claude/gemini_claude_request.go` does not sanitize the tool's `input_schema` before converting it to `parametersJsonSchema`.

Gemini API's function calling feature does not support certain JSON Schema keywords:
- `$id`
- `$schema`
- `$ref`
- `patternProperties`
- `propertyNames`
- `additionalProperties` (in some contexts)

## Affected Code

```go
// internal/translator/gemini/claude/gemini_claude_request.go
if inputSchemaResult := toolResult.Get("input_schema"); inputSchemaResult.Exists() {
    inputSchema := inputSchemaResult.Raw  // ← Raw schema passed without cleaning
    tool, _ := sjson.SetRaw(tool, "parametersJsonSchema", inputSchema)
    // ...
}
```

## Suggested Fix

Add a recursive schema cleaning function similar to what was proposed in #385:

```go
func cleanSchemaForGemini(schemaJSON string) string {
    fieldsToRemove := []string{
        "$id", "$schema", "$ref", "$defs",
        "patternProperties", "propertyNames",
        "additionalProperties",
    }
    
    result := schemaJSON
    for _, field := range fieldsToRemove {
        result, _ = sjson.Delete(result, field)
        // Also need recursive deletion for nested objects
    }
    return result
}
```

Then apply it before setting `parametersJsonSchema`:

```go
inputSchema := cleanSchemaForGemini(inputSchemaResult.Raw)
tool, _ = sjson.SetRaw(tool, "parametersJsonSchema", inputSchema)
```

## Environment

- Client: OpenCode with multiple MCP servers (cloudbase, filesystem, github, etc.)
- Provider config: `@ai-sdk/anthropic` with custom baseURL pointing to CLIProxyAPI
[...]

PR #2142, opened 2026-03-14
fix: strip uniqueItems from Gemini function_declarations (#2123)

### LINKED ISSUE #2123: Bug: Claude→Gemini translation fails: uniqueItems not stripped from function_declarations
## Bug Description

When using OpenAI-compatible API (`/v1/chat/completions`) with Gemini backend models (e.g., `gemini-3-flash`), requests fail with 400 errors because `uniqueItems` is not stripped from the tool schema before converting to Gemini `function_declarations`.

This is the same class of issue as #1424 (`$id`, `patternProperties`) and #1531 (`deprecated`), which were fixed previously. However, `uniqueItems` was not included in the sanitization list.

## Error Message

```
400 Invalid JSON payload received. Unknown name "uniqueItems" at 'request.tools[0].function_declarations[43].parameters.properties[2].value': Cannot find field.
Invalid JSON payload received. Unknown name "uniqueItems" at 'request.tools[0].function_declarations[53].parameters.properties[2].value': Cannot find field.
Invalid JSON payload received. Unknown name "uniqueItems" at 'request.tools[0].function_declarations[57].parameters.properties[2].value': Cannot find field.
```

## Reproduction

- **CLI Type**: OpenClaw (OpenAI-compatible format)
- **Model**: `gemini-3-flash` (via antigravity OAuth)
- **CLIProxyAPI Version**: v6.8.52 (Docker `eceasy/cli-proxy-api:latest`, built 2026-03-13)
- **Trigger**: OpenClaw v2026.3.11 added `uniqueItems` to some tool parameter schemas. Previous OpenClaw version (v2026.2.6) did not include this field, so the issue only appeared after upgrading.

## Expected Behavior

CLIProxyAPI should strip `uniqueItems` (and ideally all unsupported JSON Schema keywords) from tool schemas before forwarding to Gemini API, similar to how `$id`, `patternProperties`, and `deprecated` are already handled.

## Suggested Fix

In the schema sanitization logic (e.g., `ConvertClaudeRequestToGemini` or equivalent OpenAI→Gemini translator), add `uniqueItems` to the list of fields to strip. 

Consider using an allowlist approach instead of a blocklist — only keep Gemini-supported schema fields (`type`, `description`, `properties`, `required`, `items`, `enum`, `format`, `nullable`, `minimum`, `maximum`, `minItems`, `maxItems`) and strip everything else.

Muran-prog:
Fixes #2123

Gemini API rejects `uniqueItems` in tool schemas:
400 Unknown name "uniqueItems" at 'request.tools0.function_declarations43.parameters.properties2.value'

Added `uniqueItems` to `unsupportedConstraints` in `gemini_schema.go` alongside `minItems`/`maxItems` - same approach as #1424 and #1531.

All existing schema sanitization tests pass.

### DIFF
--- a/internal/util/gemini_schema_test.go
+++ b/internal/util/gemini_schema_test.go
@@ -1046,3 +1046,27 @@ func TestRemoveExtensionFields(t *testing.T) {
 		})
 	}
 }
+
+// uniqueItems should be stripped and moved to description hint (#2123).
+func TestCleanJSONSchemaForAntigravity_UniqueItemsStripped(t *testing.T) {
+	input := `{
+		"type": "object",
+		"properties": {
+			"ids": {
+				"type": "array",
+				"description": "Unique identifiers",
+				"items": {"type": "string"},
+				"uniqueItems": true
+			}
+		}
+	}`
+
+	result := CleanJSONSchemaForAntigravity(input)
+
+	if strings.Contains(result, `"uniqueItems"`) {
+		t.Errorf("uniqueItems should be removed from schema")
+	}
+	if !strings.Contains(result, "uniqueItems: true") {
+		t.Errorf("uniqueItems hint missing in description")
+	}
+}


Muran-prog:
Added a test in the follow-up commit - verifies uniqueItems is stripped and the description hint is added.

PR #2143, opened 2026-03-14
Fix/max tokens to max completion tokens 2101

### LINKED ISSUE #2101: 400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
Is it a request payload issue?

[ ] No, it's another issue.

If it's a request payload issue, you MUST know

Describe the bug

When using GitHub Models+ as an OpenAI-compatible provider (Base URL: https://models.inference.ai.azure.com), any newer model triggers the following 400 error:

400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.

The frontend/client is still sending the legacy max_tokens parameter in the request payload, but GitHub Models (Azure backend) has deprecated it for o1 series, GPT-5 and all newer models. It now strictly requires max_completion_tokens.

CLI Type

openai-compatibility

Model Name

o4-mini、o3、gpt-5、gpt-5-mini、gpt-5-nano

LLM Client

Management WebUI connect test.

Request Information

.

Expected behavior

The OpenAI-compatible layer should automatically convert max_tokens → max_completion_tokens for GitHub Models / Azure endpoints (just like LiteLLM does), or allow per-model parameter override in the provider settings.

OS Type

OS: render
Version: v6.8.51
Additional context

This issue started appearing after GitHub Models updated their newer model deployments. Many OpenAI-compatible frontends are affected. Temporary workaround is to use LiteLLM as a proxy layer (which auto-maps the parameter).

Muran-prog:
Fixes #2101

Newer OpenAI models (o3, o4-mini, gpt-5, etc.) reject the legacy max_tokens parameter:
400 Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.

Added promoteMaxTokens() in the openai-compat executor - renames max_tokens to max_completion_tokens before forwarding to upstream. If max_completion_tokens is already set, just removes the old field.

Applied in both Execute and ExecuteStream paths.

### DIFF
--- a/internal/translator/openai/openai/responses/openai_openai-responses_response_test.go
+++ b/internal/translator/openai/openai/responses/openai_openai-responses_response_test.go
@@ -1,0 +1,81 @@
+package responses
+
+import (
+	"context"
+	"testing"
+
+	"github.com/tidwall/gjson"
+)
+
+// minimal chat completion response for non-stream tests
+const minimalChatCompletion = `{
+	"id":"chatcmpl-test",
+	"object":"chat.completion",
+	"created":1700000000,
+	"model":"gpt-5",
+	"choices":[{"index":0,"message":{"role":"assistant","content":"hi"},"finish_reason":"stop"}],
+	"usage":{"prompt_tokens":5,"completion_tokens":1,"total_tokens":6}
+}`
+
+// When the translated request carries max_output_tokens (Responses-native field),
+// the response must echo it back.
+func TestNonStream_MaxOutputTokens_Direct(t *testing.T) {
+	req := []byte(`{"model":"gpt-5","max_output_tokens":4096}`)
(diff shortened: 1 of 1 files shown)


Muran-prog:
Updated based on feedback:

- Responses converter (both stream and non-stream paths in openai_openai-responses_response.go) now uses max_completion_tokens as a fallback for max_output_tokens.
- Fallback priority: max_output_tokens > max_completion_tokens > max_tokens.
- promoteMaxTokens() now returns the original payload on sjson.SetBytes errors to prevent silent data loss.
- Added 5 tests for the non-stream Responses converter covering all token limit field combinations (direct, promoted, legacy, etc.).
- Kept provider scoping global for now, as max_completion_tokens is the standard for current OpenAI-compatible models.

PR #2148, opened 2026-03-15
fix: add default transport timeouts to prevent indefinite hangs (#2113)

### LINKED ISSUE #2113: Application Freezes in Offline Environment Due to Missing HTTP Client Timeou
## Description

CLIProxyAPI freezes/hangs indefinitely when running in an offline or network-restricted environment. The application becomes unresponsive because HTTP requests to unreachable endpoints never time out.

## Root Cause

After analyzing the source code, I found that the HTTP client in CLIProxyAPI does not have proper timeout configurations:

**File: `internal/runtime/executor/proxy_helpers.go`**

```go
func newProxyAwareHTTPClient(ctx context.Context, cfg *config.Config, auth *cliproxyauth.Auth, timeout time.Duration) *http.Client {
    httpClient := &http.Client{}
    if timeout > 0 {
        httpClient.Timeout = timeout
    }
    // ...
}
```

When called with `timeout = 0` (which is the default in most executors):

```go
// openai_compat_executor.go:115, 175
httpClient := newProxyAwareHTTPClient(ctx, e.cfg, auth, 0)  // timeout = 0 means NO timeout!
```

The Go `http.Client` has **no default timeout**, which causes:

1. TCP connection attempts hang indefinitely
2. TLS handshake never times out
3. Waiting for response headers never times out

Additionally, `buildProxyTransport()` creates `http.Transport` without any timeout settings:

```go
transport = &http.Transport{
    DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
        return dialer.Dial(network, addr)  // No Dialer.Timeout!
    },
    // Missing: TLSHandshakeTimeout, ResponseHeaderTimeout, etc.
}
```

## Reproduction Steps

1. Deploy CLIProxyAPI in an environment without internet access (e.g., behind a firewall that blocks GitHub)
2. Configure `remote-management.disable-control-panel: false` (default)
3. Start the application
4. The application hangs when trying to download the management panel from GitHub

## Affected Scenarios

| Scenario | Trigger | Result |
|----------|---------|--------|
| Management panel download | Startup or first access to management UI | Hangs on GitHub request |
| OAuth token refresh | Using OAuth providers | Hangs on auth server request |
| Upstream API call | Request to unreachable endpoint | Hangs indefinitely |
| Version update check | Startup | Hangs on GitHub API request |

## Suggested Fix

Add default timeout values to `newProxyAwareHTTPClient()` and `buildProxyTransport()`:

```go
func newProxyAwareHTTPClient(ctx context.Context, cfg *config.Config, auth *cliproxyauth.Auth, timeout time.Duration) *http.Client {
    // Set default timeout if not specified
    if timeout <= 0 {
        timeout = 60 * time.Second
    }
    
    httpClient := &http.Client{Timeout: timeout}
    
    // ... existing proxy logic ...
    
    if transport != nil {
        transport.DialContext = (&net.Dialer{
            Timeout:   10 * time.Second,
            KeepAlive: 30 * time.Second,
        }).DialContext
        transport.TLSHandshakeTimeout = 10 * time.Second
        transport.ResponseHeaderTimeout = 30 * time.Second
        transport.IdleConnTimeout = 90 * time.Second
    }
    
    return httpClient
}
```

## Workaround for Users

For offline environments, set in `config.yaml`:

```yaml
remote-management:
  disable-control-panel: true  # Prevent GitHub download attempts
```

## Environment

- CLIProxyAPI Version: v6.8.51
- OS: Windows/Linux
- Network: Offline/restricted environment

## Related Files

- `internal/runtime/executor/proxy_helpers.go`
- `internal/runtime/executor/openai_compat_executor.go`
- `internal/runtime/executor/claude_executor.go`
- `internal/runtime/executor/antigravity_executor.go`
- All other executors that call `newProxyAwareHTTPClient()`

Muran-prog:
Fixes #2113 - App freezes in offline/restricted environments due to missing timeouts in `newProxyAwareHTTPClient()`.

Changes:
- Added transport-level timeouts (10s dial/TLS, 30s header, 90s idle) to `buildProxyTransport()` and `newDefaultTransport()`.
- `newProxyAwareHTTPClient` now returns a transport with timeouts by default instead of a bare `http.Client{}`.
- `http.Client.Timeout` is kept at 0 to avoid breaking SSE/streaming.
- `applyTransportTimeouts` preserves existing fields to avoid overwriting proxy settings.

Files:
- `internal/runtime/executor/proxy_helpers.go` (logic)
- `internal/runtime/executor/proxy_helpers_test.go` (4 new tests)

All executor tests pass.

### DIFF
--- a/internal/runtime/executor/proxy_helpers.go
+++ b/internal/runtime/executor/proxy_helpers.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"context"
+	"net"
 	"net/http"
 	"strings"
 	"time"
@@ -10,6 +11,15 @@ import (
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
 	"github.com/router-for-me/CLIProxyAPI/v6/sdk/proxyutil"
 	log "github.com/sirupsen/logrus"
+)
+
+// Connection-setup timeouts. Safe for streaming — only affect dial/TLS/headers, not body reads.
+const (
+	defaultDialTimeout           = 10 * time.Second
+	defaultDialKeepAlive         = 30 * time.Second
+	defaultTLSHandshakeTimeout   = 10 * time.Second
+	defaultResponseHeaderTimeout = 30 * time.Second
+	defaultIdleConnTimeout       = 90 * time.Second
 )
 
 // newProxyAwareHTTPClient creates an HTTP client with proper proxy configuration priority:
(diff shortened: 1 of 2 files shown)


Muran-prog:
Good catch - fixed in 0e823ec. newDefaultTransport now sets Proxy: http.ProxyFromEnvironment and ForceAttemptHTTP2: true to match http.DefaultTransport behavior

PR #2158, opened 2026-03-16
fix(antigravity): resolve empty functionResponse.name for toolu_* tool_use_id format

sususu98:
## 问题描述

Antigravity 的 Claude→Gemini 翻译器在处理 `tool_result` 时，通过字符串分割 `tool_use_id` 来推导 `functionResponse.name`。当 `tool_use_id` 按 `-` 分割后恰好只有 2 段时（如 Claude SDK 标准格式 `toolu_tool-<uuid>`），`[0:len-2]` 产生空切片，导致 `name` 为空字符串。Gemini API 要求 `functionResponse.name` 非空，返回 400 `REQUIRED_FIELD_MISSING`。

## 触发场景

Claude Code 客户端发送多轮对话请求，对话历史中包含两种格式的 `tool_use_id`：

| 格式 | 示例 | split(`-`) 段数 | 旧逻辑结果 |
|------|------|:-:|------|
| Claude SDK 标准格式 | `toolu_tool-48fca351f12844eabf49dad8b63886d2` | **2** | `""` → **400 错误** |
| Claude Code 自定义格式 | `Read-1773420180464065165-1327` | 3 | `"Read"` ✅ |

只有第一种格式触发 bug。第二种恰好能工作是因为 3 段去掉 2 段还剩 1 段。

## 根因分析

```go
// 旧代码 — 对 2 段 ID 失败
funcName := toolCallID
toolCallIDs := strings.Split(toolCallID, "-")
if len(toolCallIDs) > 1 {
    funcName = strings.Join(toolCallIDs[0:len(toolCallIDs)-2], "-")
    // "toolu_tool-<uuid>" → 2 段 → [0:0] → 空切片 → ""
}
```

从 ID 字符串猜测工具名本身就是错误的设计。Claude API 中 `tool_result` 通过 `tool_use_id` 引用前面 `tool_use` 块的 `id`，工具名应从对应的 `tool_use.name` 获取。

## 修复方案

用 `tool_use.id → tool_use.name` 的 lookup map 替代字符串分割启发式：

1. **在主循环中增量构建映射**：处理 `tool_use` 块时，记录 `id → name` 到 `toolNameByID` map
2. **处理 `tool_result` 时查表**：从 map 中查找对应的工具名
3. **Fallback**：查不到时退回使用完整 `tool_use_id` 作为 name（保证非空），并输出 warn 日志

```go
// 新代码
toolNameByID := make(map[string]string)

// 处理 tool_use 时记录映射
if functionID != "" && functionName != "" {
    toolNameByID[functionID] = functionName
}

// 处理 tool_result 时查找
funcName := toolNameByID[toolCallID]
if funcName == "" {
    funcName = toolCallID
    log.Warnf("antigravity claude request: tool_result references unknown tool_use_id=%s, using ID as function name", toolCallID)
}
```

这个方案是语义正确的：`tool_result` 的名字就应该来自对应 `tool_use.name`，而不是从 ID 字符串猜。

## 测试覆盖

- `TestToolResult` — 更新为包含前置 `tool_use` 块，验证 name 正确解析
- `TestToolResultName_TouluFormat` — 新增，覆盖原始 bug 场景（`toolu_tool-*` 格式，多工具并行）
- `TestToolResultName_CustomFormat` — 新增，覆盖 `Read-timestamp-seq` 格式
- 所有既有测试通过，无回归

## 影响范围

- 仅修改 `internal/translator/antigravity/claude/antigravity_claude_request.go`
- 所有走 antigravity + Claude API 格式的模型受益（claude-sonnet-4-5, claude-opus-4-6, claude-haiku-4-5 等）

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -12,6 +12,7 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/thinking"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/translator/gemini/common"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
+	log "github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 )
@@ -67,6 +68,10 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 	// contents
 	contentsJSON := "[]"
 	hasContents := false
+
+	// tool_use_id → tool_name lookup, populated incrementally during the main loop.
+	// Claude's tool_result references tool_use by ID; Gemini requires functionResponse.name.
+	toolNameByID := make(map[string]string)
 
 	messagesResult := gjson.GetBytes(rawJSON, "messages")
 	if messagesResult.IsArray() {
@@ -170,6 +175,10 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 						argsResult := contentResult.Get("input")
 						functionID := contentResult.Get("id").String()
 
+						if functionID != "" && functionName != "" {
+							toolNameByID[functionID] = functionName
+						}
+
 						// Handle both object and string input formats
 						var argsRaw string
 						if argsResult.IsObject() {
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Thanks for the PR. I’m requesting changes mainly around the new fallback behavior for `tool_result`.

The `toolNameByID` lookup is the right fix for the `toolu_*` case, but the new miss path now sets `functionResponse.name` to the raw `tool_use_id`. That is a behavior change from the current translator, which still handles standalone `tool_result` turns by deriving a semantic tool name from IDs like `get_weather-call-123`. With this patch, the same input would now emit `functionResponse.name = "get_weather-call-123"`, which is not the actual tool name and can still break downstream matching/validation for partial or truncated histories.

There is also a test gap here: the previous `TestConvertClaudeRequestToAntigravity_ToolResult` covered a `tool_result` without a preceding `tool_use`, but this PR rewrites that test to include the matching `tool_use` and therefore no longer exercises the fallback path.

I’d keep the new ID-to-name map, but preserve a semantic fallback when the matching `tool_use` is missing (for example: old heuristic first, raw ID only as a last resort), and add an explicit regression test for the no-matching-`tool_use` case.

Test plan:
- Not run locally in this shared workspace.
- Reviewed the PR diff and the passing `pr-test-build / build` check only.

sususu98:
Thanks for the review! All three points have been addressed in the latest force-push:

**1. Preserved semantic fallback (heuristic → raw ID as last resort)**

The miss path now uses a 3-tier fallback chain:

```go
funcName, ok := toolNameByID[toolCallID]
if !ok {
    // Tier 2: old heuristic — strip last 2 dash-segments
    parts := strings.Split(toolCallID, "-")
    if len(parts) > 2 {
        funcName = strings.Join(parts[:len(parts)-2], "-")
    }
    // Tier 3: raw ID only when heuristic produces empty string
    if funcName == "" {
        funcName = toolCallID
    }
}
```

- `get_weather-call-123` (3 segments) → `"get_weather"` ✅ (same as before)
- `toolu_tool-<uuid>` (2 segments) → `"toolu_tool-<uuid>"` (raw ID, non-empty — was `""` before this PR)

**2. Added regression tests for no-matching-`tool_use` fallback**

- `TestToolResultName_NoMatchingToolUse_Heuristic` — standalone `tool_result` with 3-segment ID, asserts heuristic derives `"get_weather"`
- `TestToolResultName_NoMatchingToolUse_RawID` — standalone `tool_result` with 2-segment `toolu_*` ID, asserts raw ID is used and name is non-empty

**3. Also adopted Gemini Code Assist's suggestion**

Changed map lookup from zero-value check to comma-ok idiom (`funcName, ok := toolNameByID[...]`) for more idiomatic Go.

PR #2173, opened 2026-03-16
fix: strip markdown code fences from non-streaming Claude responses

willie:
## Summary
- Claude models sometimes wrap JSON responses in markdown code fences (` ```json ... ``` `) when accessed through the OpenAI-compatible translation layer
- Downstream apps that expect raw JSON (e.g. [Hindsight](https://github.com/vectorize-io/hindsight)) fail to parse the response and must retry, wasting tokens and adding latency
- Adds `StripMarkdownCodeFences()` to `internal/util` that detects when the entire response is a single fenced block and strips the fence lines
- Applied to the non-streaming paths in both chat completions (`/v1/chat/completions`) and OpenResponses (`/v1/responses`) translators

## Test plan
- [x] Unit tests for `StripMarkdownCodeFences` covering: json/xml fences, no language tag, plain text passthrough, empty string, trailing/leading whitespace, multiline content, inline backticks, missing closing fence, single-line backticks
- [x] Full project builds cleanly
- [x] Existing translator tests pass
- [x] End-to-end tested with Hindsight via Docker — retain calls succeed without JSON parse errors

### DIFF
--- a/internal/translator/claude/openai/chat-completions/claude_openai_response.go
+++ b/internal/translator/claude/openai/chat-completions/claude_openai_response.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 )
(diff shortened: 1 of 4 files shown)


willie:
Closing — translator changes require maintainer action per CI policy. Filed as an issue instead.

PR #2197, opened 2026-03-18
fix: unify Anthropic TLS fingerprint to Bun BoringSSL

lichengzhe:
## Summary

- Unify all Anthropic connections (API requests + OAuth token refresh) to use **Bun BoringSSL TLS fingerprint** (JA3: `50027c67d7d68e24c00d233bca146d88`), matching real Claude Code CLI
- Replace Chrome_Auto fingerprint with custom `BunBoringSSLSpec()` captured from Bun v1.3.8
- Switch from HTTP/2 (manual `h2.ClientConn` pool) to HTTP/1.1 (`http.Transport` + `DialTLSContext`) — Bun's ALPN only offers `http/1.1`
- Route all Claude executor API calls through `newClaudeHTTPClient()` using the utls transport, eliminating the independent bare Go TLS path
- Extract `ResolveProxyURL()` as single source of truth for proxy priority (`auth.ProxyURL > cfg.ProxyURL > env`), used by both request and refresh paths

## Context

CLIProxyAPI currently has two independent HTTP client paths for Anthropic connections: API requests use standard Go `http.Transport` (native Go TLS fingerprint), while token refresh uses utls with Chrome fingerprint. This creates a TLS fingerprint inconsistency — the `User-Agent` header claims Claude Code CLI (Bun/Node.js runtime), but the TLS ClientHello doesn't match what Bun actually produces.

This PR unifies both paths to use a single Bun BoringSSL fingerprint, ensuring full consistency between the declared client identity and the observed TLS behavior. This improves transparency and reduces the risk of being flagged by upstream bot detection systems that cross-reference User-Agent with TLS fingerprints.

## Files Changed

| File | Change |
|------|--------|
| `internal/auth/claude/bun_tls_spec.go` | **New** — Bun BoringSSL ClientHello spec |
| `internal/auth/claude/utls_transport.go` | Chrome→Bun, h2→h1.1, simplified architecture |
| `internal/auth/claude/anthropic_auth.go` | Optional `proxyURL` param on `NewClaudeAuth` |
| `internal/runtime/executor/proxy_helpers.go` | `ResolveProxyURL()` + `newClaudeHTTPClient()` |
| `internal/runtime/executor/claude_executor.go` | Use `newClaudeHTTPClient` for all API calls |
| `sdk/cliproxy/auth/conductor.go` | Elevate refresh failure logs DEBUG→WARN |

## Test plan

- [ ] CI build passes (Go compilation + existing tests)
- [ ] Deploy to test instance, verify TLS fingerprint with `tls.peet.ws` matches JA3 `50027c67d7d68e24c00d233bca146d88`
- [ ] Verify API requests succeed through proxy
- [ ] Verify OAuth token refresh works through per-account proxy

Supersedes #1947.

### DIFF
--- a/internal/auth/claude/anthropic_auth.go
+++ b/internal/auth/claude/anthropic_auth.go
@@ -50,19 +50,26 @@ type ClaudeAuth struct {
 }
 
 // NewClaudeAuth creates a new Anthropic authentication service.
-// It initializes the HTTP client with a custom TLS transport that uses Firefox
-// fingerprint to bypass Cloudflare's TLS fingerprinting on Anthropic domains.
+// It initializes the HTTP client with a custom TLS transport that uses Bun BoringSSL
+// fingerprint to match real Claude Code CLI behavior.
+//
+// An optional proxyURL may be provided to override the proxy in cfg.SDKConfig.
+// This allows callers to pass a pre-resolved proxy URL (e.g. from ResolveProxyURL)
+// so that both request and refresh paths share the same proxy priority logic.
 //
 // Parameters:
 //   - cfg: The application configuration containing proxy settings
+//   - proxyURL: Optional pre-resolved proxy URL that overrides cfg.SDKConfig.ProxyURL
 //
 // Returns:
 //   - *ClaudeAuth: A new Claude authentication service instance
-func NewClaudeAuth(cfg *config.Config) *ClaudeAuth {
-	// Use custom HTTP client with Firefox TLS fingerprint to bypass
-	// Cloudflare's bot detection on Anthropic domains
+func NewClaudeAuth(cfg *config.Config, proxyURL ...string) *ClaudeAuth {
+	pURL := cfg.SDKConfig.ProxyURL
+	if len(proxyURL) > 0 && proxyURL[0] != "" {
+		pURL = proxyURL[0]
+	}
 	return &ClaudeAuth{
-		httpClient: NewAnthropicHttpClient(&cfg.SDKConfig),
+		httpClient: NewAnthropicHttpClient(pURL),
 	}
 }
 
--- a/internal/auth/claude/bun_tls_spec.go
+++ b/internal/auth/claude/bun_tls_spec.go
@@ -1,0 +1,131 @@
+// Bun v1.3.x (BoringSSL) TLS ClientHello Spec for utls
+// Captured from real Claude Code 2.1.71 / Bun v1.3.8 via tls.peet.ws
+//
+// JA3 Hash: 50027c67d7d68e24c00d233bca146d88
+// JA3: 771,4865-4866-4867-49195-49199-49196-49200-52393-52392-49161-49171-49162-49172-156-157-47-53,0-65037-23-65281-10-11-35-16-5-13-18-51-45-43-21,29-23-24,0
+// JA4: t13d1715h1_5b57614c22b0_7baf387fc6ff
+//
+// Key differences from Chrome:
+//   - No GREASE (Chrome injects random GREASE values)
+//   - ALPN: http/1.1 only (Chrome: h2 + http/1.1)
+//   - Fewer extensions (no compress_certificate, no delegated_credentials)
+//   - ECH extension 65037 present (BoringSSL-specific)
+
+package claude
+
+import (
+	tls "github.com/refraction-networking/utls"
+)
+
+// BunBoringSSLSpec returns a utls ClientHelloSpec that exactly matches
+// Bun v1.3.x's BoringSSL TLS fingerprint, as used by Claude Code CLI.
+//
+// This ensures TLS fingerprint consistency between API requests and
+// token refresh, matching what Anthropic sees from real Claude Code users.
+func BunBoringSSLSpec() *tls.ClientHelloSpec {
+	return &tls.ClientHelloSpec{
+		TLSVersMin: tls.VersionTLS12,
+		TLSVersMax: tls.VersionTLS13,
+		CipherSuites: []uint16{
+			// TLS 1.3 suites
+			tls.TLS_AES_128_GCM_SHA256,       // 0x1301
+			tls.TLS_AES_256_GCM_SHA384,       // 0x1302
+			tls.TLS_CHACHA20_POLY1305_SHA256,  // 0x1303
+			// TLS 1.2 ECDHE suites
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,       // 0xC02B
+			tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,         // 0xC02F
+			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,       // 0xC02C
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,         // 0xC030
+			tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, // 0xCCA9
+			tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,   // 0xCCA8
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,          // 0xC009
+			tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,            // 0xC013
+			tls.TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,          // 0xC00A
+			tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,            // 0xC014
+			// TLS 1.2 RSA suites
+			tls.TLS_RSA_WITH_AES_128_GCM_SHA256, // 0x009C
+			tls.TLS_RSA_WITH_AES_256_GCM_SHA384, // 0x009D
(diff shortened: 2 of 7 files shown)


lichengzhe:
Addressed all review feedback in 10c1ef5a:

1. **Env proxy fallback** (Gemini + Codex P1): Changed default dialer from `proxy.Direct` to `proxy.FromEnvironment()`, so deployments relying on `HTTPS_PROXY` without explicit `proxy-url` config continue to work.

2. **Context support in dialTLS** (Codex P2): Now uses `proxy.ContextDialer` when available for context-aware TCP dial, and propagates context deadline to TLS handshake via `conn.SetDeadline()` to prevent indefinite hangs on unreachable endpoints.

luispater (changes requested):
Summary:
Thanks for unifying the Claude TLS fingerprint path. I found two blocking proxy regressions in the new uTLS transport, so I don't think this is safe to merge yet.

Key findings:
- Blocking: explicit `proxy-url` values now go through `proxyutil.BuildDialer()`, which delegates to `golang.org/x/net/proxy.FromURL()`. That dialer only supports `socks5/socks5h`, so `http://...` and `https://...` proxies now fail and get ignored for Claude API / refresh traffic.
- Blocking: the environment fallback is no longer equivalent to the previous `net/http` behavior. `proxy.FromEnvironment()` only checks `ALL_PROXY` / `NO_PROXY`, so standard `HTTP_PROXY` / `HTTPS_PROXY` deployments stop working. `ALL_PROXY=http://...` also fails for the same reason above, so only SOCKS via `ALL_PROXY` still works here.

Test plan:
- Add regression coverage for `proxy-url=http(s)://...`
- Add regression coverage for `proxy-url=socks5://...`
- Add regression coverage for `proxy-url=direct`
- Add regression coverage for `HTTPS_PROXY`
- Add regression coverage for `ALL_PROXY=socks5://...`

lichengzhe:
@luispater Thanks for catching the proxy regressions — both were real issues. Fixed in 8461a8af:

**1. HTTP/HTTPS proxy support:**
Replaced `proxyutil.BuildDialer` (SOCKS5-only) with explicit proxy handling in `dialTLS`:
- SOCKS5: `proxy.SOCKS5` dialer with `ContextDialer` support
- HTTP/HTTPS: manual CONNECT tunneling with `Proxy-Authorization` header
- Direct: `net.Dialer.DialContext`

**2. Environment variable compatibility:**
Replaced `proxy.FromEnvironment()` (only reads `ALL_PROXY`) with `proxyFromEnv()` that checks `HTTPS_PROXY`, `HTTP_PROXY`, `ALL_PROXY` (case-insensitive) and respects `NO_PROXY` — matching standard `http.Transport` behavior.

**Architecture:** `resolveProxy(targetAddr)` determines the effective proxy per-request (explicit config > env vars > direct), then `dialTLS` handles the tunnel/dial + utls handshake in a single path. This keeps the fingerprint consistent regardless of proxy type.

I don't have Go test infrastructure for proxy integration tests (would need a test SOCKS5/HTTP proxy), but verified locally:
- `go build ./...` passes
- `go test ./internal/runtime/executor/... ./sdk/proxyutil/...` passes
- Deployed and running in production with SOCKS5 proxy

lichengzhe:
@luispater Pushed two more commits addressing both blocking issues:

**8461a8af** — HTTP/HTTPS proxy + env var fix:
- Replaced `proxyutil.BuildDialer` (SOCKS5-only) with explicit handling per proxy type in `dialTLS`:
  - SOCKS5: `proxy.SOCKS5` dialer with `ContextDialer` support
  - HTTP/HTTPS: manual CONNECT tunneling with `Proxy-Authorization`
  - Direct: `net.Dialer.DialContext`
- Replaced `proxy.FromEnvironment()` (`ALL_PROXY` only) with custom `proxyFromEnv()` checking `HTTPS_PROXY`, `HTTP_PROXY`, `ALL_PROXY` and respecting `NO_PROXY`

**c4bc2e9e** — Self-review polish:
- Replaced the hand-rolled `proxyFromEnv()` with `http.ProxyFromEnvironment` for correct `NO_PROXY` matching (CIDR, port, wildcard) — cleaner and more robust
- Added default port for HTTP/HTTPS proxies (`:80`/`:443`) when the proxy URL has no explicit port
- Properly close CONNECT `resp.Body`
- Handle `bufio.Reader` buffered bytes after CONNECT handshake with `bufferedConn` wrapper

All tests pass locally (`go build ./...` + `go test ./internal/runtime/executor/... ./sdk/proxyutil/...`).

lichengzhe:
Pushed commit 5 with two bug fixes found during code review:

**Fix 1 (P1): HTTPS proxy TLS wrapping**
`dialViaHTTPConnect()` was sending raw CONNECT requests to `https://` proxies over plain TCP. HTTPS proxies expect TLS first, then CONNECT inside the encrypted tunnel. Fixed by wrapping the TCP connection in standard `crypto/tls` before sending CONNECT. (utls fingerprinting is only for the final `api.anthropic.com` connection.)

**Fix 2 (P2): CONNECT deadline**
The CONNECT request/response had no deadline — if a proxy accepted TCP but never replied, the goroutine would hang indefinitely. Fixed by propagating the context deadline to the connection before CONNECT.

@luispater ready for re-review — these are the only remaining issues from the code review.

lichengzhe:
@luispater Thanks for the thorough review — both blocking issues have been addressed:

**1. HTTP/HTTPS proxy support** (commit 8461a8a)
- Implemented manual CONNECT tunneling in `dialViaHTTPConnect()` for `http://` and `https://` proxies
- HTTPS proxies are properly wrapped in standard `crypto/tls` before the CONNECT handshake (commit 222a684)
- SOCKS5 continues to work via `proxy.SOCKS5` dialer
- Default port inference (80 for http, 443 for https) when proxy URL has no explicit port

**2. Environment variable compatibility** (commit c4bc2e9)
- Replaced `proxy.FromEnvironment()` (which only reads `ALL_PROXY`) with `http.ProxyFromEnvironment` in `resolveProxy()`
- This correctly handles `HTTPS_PROXY`, `HTTP_PROXY`, `ALL_PROXY`, and respects `NO_PROXY` (CIDR, port, wildcard) — matching standard `http.Transport` behavior

**Additional improvements in latest commit** (46ae697):
- **Connection reuse**: `NewAnthropicHttpClient` now caches `*http.Client` per proxyURL via `sync.Map`, so the `http.Transport` connection pool is reused across requests. Previously each API call created a fresh transport, meaning a full TLS handshake every time.
- **Transport pool config**: `MaxIdleConns=100`, `MaxIdleConnsPerHost=10`, `IdleConnTimeout=90s`, `ExpectContinueTimeout=1s` — matching `http.DefaultTransport` defaults.
- **Context RoundTripper**: Documented as an intentional design decision. `newClaudeHTTPClient` does not honor `cliproxy.roundtripper` because an injected RoundTripper would replace the TLS layer and break the Bun fingerprint. SDK embedders should use `proxy-url` instead. `newProxyAwareHTTPClient` continues to honor it for non-Claude providers.

Ready for re-review when you get a chance.

lichengzhe:
## Refactor: Extract proxy dialing into `proxy_dial.go`

Pushed `aaed93ba` — separates proxy tunneling from TLS fingerprinting for cleaner review.

### What moved

| → `proxy_dial.go` (new, 202 lines) | Stays in `utls_transport.go` (113 lines) |
|---|---|
| `ProxyDialer` struct + `NewProxyDialer()` | `utlsRoundTripper` + `newUtlsRoundTripper()` |
| `resolveProxy()` (env/direct/explicit dispatch) | `dialTLS()` (utls handshake only) |
| `DialContext()` (SOCKS5/CONNECT/direct switch) | `RoundTrip()` |
| `dialViaSocks5()`, `dialViaHTTPConnect()` | `NewAnthropicHttpClient()` + `sync.Map` cache |
| `bufferedConn` | |

### What didn't change
- `bun_tls_spec.go` — unchanged
- Public API — `NewAnthropicHttpClient(proxyURL string)` signature preserved
- All other files — unchanged

Build ✅ Tests ✅

lichengzhe:
@luispater Thanks for the thorough review! Both blocking issues have been addressed in subsequent commits:

1. **HTTP/HTTPS proxy regression** — Fixed in c4bc2e9e + 222a684d. No longer uses `proxyutil.BuildDialer()`. `proxy_dial.go` handles all proxy types manually: SOCKS5 via `proxy.SOCKS5`, HTTP via TCP+CONNECT, HTTPS via TCP+TLS+CONNECT.

2. **Environment proxy fallback** — Fixed in c4bc2e9e. `resolveProxy()` delegates to `http.ProxyFromEnvironment()` (reads HTTPS_PROXY/HTTP_PROXY/ALL_PROXY, respects NO_PROXY) instead of `proxy.FromEnvironment()`.

Latest commit aaed93ba is a pure refactor (extract proxy dialing into own file) with no behavioral changes. Build + tests pass.

Could you re-review when you get a chance?

luispater (changes requested):
Summary: The refactor now correctly handles Bun TLS fingerprinting, HTTP/HTTPS proxy tunneling, and deadline propagation in the proxy handshake path.

Blocking: `resolveProxy` in `proxy_dial.go` strips the target port before delegating to `http.ProxyFromEnvironment`, so port-qualified `NO_PROXY` entries cannot work as intended. That leaves a correctness gap in the environment proxy fallback path.

Verification: `go test ./internal/auth/claude ./internal/runtime/executor ./sdk/cliproxy/auth`

PR #2203, opened 2026-03-18
feat: add gpt-5.4-mini codex model definitions

WormW:
Summary
Add gpt-5.4-mini to the Codex model catalogs so the registry exposes it consistently.

Changes
• Added gpt-5.4-mini entries to internal/registry/models/models.json
• Included the model in codex-team, codex-plus, and codex-pro
• Reused the existing gpt-5.4 metadata shape for display name, context length, completion tokens, and thinking support

Notes
• This is a metadata-only change for registry/model listing support
• No runtime behavior changes beyond model discovery and model info lookup

### DIFF
--- a/internal/registry/models/models.json
+++ b/internal/registry/models/models.json
@@ -1602,6 +1602,29 @@
       }
     },
     {
+      "id": "gpt-5.4-mini",
+      "object": "model",
+      "created": 1772668800,
+      "owned_by": "openai",
+      "type": "openai",
+      "display_name": "GPT 5.4 Mini",
+      "version": "gpt-5.4-mini",
+      "description": "Stable version of GPT 5.4 Mini",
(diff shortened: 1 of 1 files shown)


Sixzero:
+1

hkfires:
The model list is mainly updated in another repository, while the one embedded in the program is only used as a fallback and will no longer be updated frequently.

richard-mayers:
@hkfires how do we provide the model list to the built? 

Is there a way?

hkfires:
> @hkfires how do we provide the model list to the built?
> 
> Is there a way?

No need to provide it separately; after the program starts, it will retrieve it from https://github.com/router-for-me/models

richard-mayers:
@hkfires I see, my problem was that I needed to reload the program. Thanks a lot!!!

PR #2210, opened 2026-03-18
fix(auth): stream bootstrap errors now trigger auth rotation

### LINKED ISSUE #2189: [Bug] Failover/retry not triggered when upstream Claude API relay returns empty stream or undefined error (context canceled)
[ ] Yes, this is a request payload issue.
[x] No, it's another issue.

**Describe the bug**

When using multiple `claude-api-key` entries pointing to different third-party Claude API relays (configured for redundancy), the failover/retry mechanism fails to activate when the first upstream relay returns certain non-standard errors. Only the first relay is attempted; the remaining backup relays are never tried.

**Setup:** 3 Claude API relay keys configured via `claude-api-key` (different providers at different price tiers) to achieve high availability at lower cost than official Claude API.

**What happened:**

1. One of the three upstream relays experienced a temporary outage.
2. CLIProxyAPI's error log showed:
```json
{"error":{"message":"context canceled","type":"server_error","code":"internal_server_error"}}
```
3. The downstream client (Open WebUI) received: `empty_stream: upstream stream closed before first payload`
4. **The management panel showed only the first relay was activated and errored — the other two backup relays were never attempted.**
5. Switching `routing.strategy` from `fill-first` to `round-robin` did not fix the issue — still only one relay was tried.
6. Manually disabling the broken relay restored service immediately (the remaining two relays worked fine).

**Root cause investigation:**

I bypassed CLIProxyAPI and sent requests directly to the broken relay using Cherry Studio. The relay returned an error with type `Undefined` (no standard HTTP error code or recognizable error body). This appears to be an empty or malformed upstream response.

**Key finding — the retry logic is error-type dependent:**

To verify, I deliberately set an invalid API key for the broken relay in CLIProxyAPI config. This caused that relay to return a clear, well-formed error response ("invalid key"). **With this explicit error, the failover worked correctly** — CLIProxyAPI properly rotated to the next relay.

This confirms that the failover mechanism works fine when the upstream returns a recognizable error, but fails to trigger when the upstream returns an empty stream, undefined error, or closes the connection before sending any payload.

In summary, CLIProxyAPI's retry/failover logic does not recognize certain upstream failure modes:
- Empty response body
- `Undefined` error type
- `context canceled` before any payload is sent
- Possibly: upstream closes connection before sending any SSE data

In these cases, the error is not matched against the retriable status codes (403, 408, 500, 502, 503, 504 per `config.example.yaml`), so no retry is attempted even though the request clearly failed.

**CLI Type**
claude-api-key (third-party Claude API relays via `claude-api-key` with custom `base-url`)

**Model Name**
Claude Opus 4.6 | Anthropic

**LLM Client**
Open WebUI (connecting to CLIProxyAPI as OpenAI-compatible endpoint)

**Request Information**

CLIProxyAPI error log at the time of failure:
```
=== API RESPONSE ===
Timestamp: 2026-03-17T18:06:52.850475624+08:00
{"error":{"message":"context canceled","type":"server_error","code":"internal_server_error"}}

=== RESPONSE ===
Status: 500
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
Access-Control-Allow-Headers: *

{"error":{"message":"context canceled","type":"server_error","code":"internal_server_error"}}
```

Relevant config:
```yaml
routing:
  strategy: "round-robin"  # also tested with fill-first, same result

request-retry: 3  # default

claude-api-key:
  - api-key: "sk-relay-A..."
    base-url: "https://relay-a.example.com"
  - api-key: "sk-relay-B..."
    base-url: "https://relay-b.example.com"
  - api-key: "sk-relay-C..."
    base-url: "https://relay-c.example.com"
```

**Expected behavior**

When the first upstream relay fails (regardless of error type — including empty streams, context cancellation, or undefined errors), CLIProxyAPI should retry with the next available relay credential, up to `request-retry` times. The current behavior only retries on recognized HTTP error codes, which means a relay that fails in a "silent" way (empty body, connection reset before payload, undefined error) will block all requests without triggering failover — defeating the purpose of multi-relay redundancy.

Specifically:
1. Any upstream response that results in `empty_stream` / `context canceled` should be treated as a retriable failure.
2. If the upstream closes the connection before sending any SSE data, it should trigger credential rotation.
3. The failover mechanism should not depend on receiving a well-formed error response from the upstream.

**Version**
CLIProxyAPI v6.8.55 (Commit: db63f9b5, Built: 2026-03-16T14:33:45Z)

DragonFSKY:
## Summary

- When `executeStreamWithModelPool` hit a bootstrap error or empty stream on the last model in the pool, it wrapped the error into a `StreamChunk` channel and returned `(result, nil)`. The `nil` error prevented `executeStreamMixedOnce` from trying the next auth credential.
- This caused streaming requests (including WebSocket/Codex) to only try one auth per `ExecuteStream` call. Handler-level `bootstrap-retries` provided additional attempts, but each still picked only one auth — wasting retries on already-failed credentials.
- Remove the last-model special case so all models follow the same error path: call `MarkResult`, discard remaining stream chunks, and propagate the error. The inner auth rotation loop then picks the next available credential, matching the existing HTTP behavior.

Fixes #2189

## Changes

`sdk/cliproxy/auth/conductor.go` — Remove last-model special-case branches in `executeStreamWithModelPool`:
  - Bootstrap error path (was line 580-595): removed `if idx < len(execModels)-1` guard and the `wrapStreamResult` fallback. All models now call `MarkResult` + `discardStreamChunks` + `continue`.
  - Empty stream path (was line 598-610): same treatment.
  - `idx` variable no longer needed, changed to `_`.

`sdk/cliproxy/auth/conductor_stream_retry_test.go` — New tests for cross-auth rotation:
  - `TestExecuteStream_RotatesAuthOnBootstrapError`: auth1 returns 429, verifies auth2 is tried and succeeds.
  - `TestExecuteStream_RotatesAuthOnEmptyStream`: auth1 returns empty stream, verifies auth2 is tried and succeeds.
  - `TestExecuteStream_AllAuthsFailReturnsError`: both auths fail, verifies `(nil, error)` is returned (not wrapped stream).

## Test plan

- [x] Existing `TestManagerExecuteStream_*` tests pass (model-pool rotation within single auth)
- [x] New auth rotation tests pass
- [x] Full `sdk/cliproxy/auth` package tests pass with no regression
- [ ] Manual verification with multiple Codex auth entries where some are rate-limited

### DIFF
--- a/sdk/api/handlers/handlers.go
+++ b/sdk/api/handlers/handlers.go
@@ -589,25 +589,30 @@ func (h *BaseAPIHandler) ExecuteStreamWithAuthManager(ctx context.Context, handl
 		SourceFormat:    sdktranslator.FromString(handlerType),
 	}
 	opts.Metadata = reqMeta
-	streamResult, err := h.AuthManager.ExecuteStream(ctx, providers, req, opts)
+	maxBootstrapRetries := StreamingBootstrapRetries(h.Cfg)
+	streamResult, initialBootstrapRetries, err := h.executeStreamWithBootstrapRetry(ctx, providers, req, opts, maxBootstrapRetries)
 	if err != nil {
 		err = enrichAuthSelectionError(err, providers, normalizedModel)
-		errChan := make(chan *interfaces.ErrorMessage, 1)
-		status := http.StatusInternalServerError
-		if se, ok := err.(interface{ StatusCode() int }); ok && se != nil {
-			if code := se.StatusCode(); code > 0 {
-				status = code
+		if shouldWrapImmediateStreamError(err) {
+			streamResult = streamResultFromError(err)
+		} else {
+			errChan := make(chan *interfaces.ErrorMessage, 1)
+			status := http.StatusInternalServerError
+			if se, ok := err.(interface{ StatusCode() int }); ok && se != nil {
+				if code := se.StatusCode(); code > 0 {
+					status = code
+				}
 			}
-		}
-		var addon http.Header
-		if he, ok := err.(interface{ Headers() http.Header }); ok && he != nil {
-			if hdr := he.Headers(); hdr != nil {
-				addon = hdr.Clone()
+			var addon http.Header
+			if he, ok := err.(interface{ Headers() http.Header }); ok && he != nil {
+				if hdr := he.Headers(); hdr != nil {
+					addon = hdr.Clone()
+				}
 			}
+			errChan <- &interfaces.ErrorMessage{StatusCode: status, Error: err, Addon: addon}
+			close(errChan)
+			return nil, nil, errChan
 		}
-		errChan <- &interfaces.ErrorMessage{StatusCode: status, Error: err, Addon: addon}
-		close(errChan)
-		return nil, nil, errChan
 	}
 	passthroughHeadersEnabled := PassthroughHeadersEnabled(h.Cfg)
 	// Capture upstream headers from the initial connection synchronously before the goroutine starts.
@@ -626,8 +631,7 @@ func (h *BaseAPIHandler) ExecuteStreamWithAuthManager(ctx context.Context, handl
 		defer close(dataChan)
 		defer close(errChan)
 		sentPayload := false
-		bootstrapRetries := 0
-		maxBootstrapRetries := StreamingBootstrapRetries(h.Cfg)
+		bootstrapRetries := initialBootstrapRetries
 
 		sendErr := func(msg *interfaces.ErrorMessage) bool {
 			if ctx == nil {
@@ -652,20 +656,6 @@ func (h *BaseAPIHandler) ExecuteStreamWithAuthManager(ctx context.Context, handl
 				return false
 			case dataChan <- chunk:
 				return true
-			}
-		}
-
-		bootstrapEligible := func(err error) bool {
-			status := statusFromError(err)
-			if status == 0 {
-				return true
-			}
-			switch status {
-			case http.StatusUnauthorized, http.StatusForbidden, http.StatusPaymentRequired,
-				http.StatusRequestTimeout, http.StatusTooManyRequests:
-				return true
-			default:
-				return status >= http.StatusInternalServerError
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


luispater (changes requested):
Summary:
The production fix in `executeStreamWithModelPool` looks directionally correct: removing the last-model special-case allows bootstrap/empty-stream failures to propagate as errors so auth rotation can continue in `executeStreamMixedOnce`.

Blocking issue:
The newly added tests are flaky due to shared global registry state under `t.Parallel()` with reused auth IDs. This can make CI nondeterministic and masks real regressions. Please isolate test identities per test (unique auth IDs/provider/model), or avoid parallelization for this file, and keep cleanup strictly scoped.

Non-blocking improvement:
`TestExecuteStream_AllAuthsFailReturnsError` should validate which auth IDs were attempted (set equality / order expectation where appropriate), not only the number of attempts.

Test evidence:
- Reviewed PR diff and checks (ignoring `translator-path-guard` per request).
- Reproduced flakiness with repeated runs:
  - `go test ./sdk/cliproxy/auth -run 'TestExecuteStream_(RotatesAuthOnBootstrapError|RotatesAuthOnEmptyStream|AllAuthsFailReturnsError)$' -count=200`
  - Observed intermittent failures including `auth_not_found`, `empty_stream`, and single-attempt auth traces.

DragonFSKY:
Thanks, this was an oversight on my side.

I isolated the new retry tests with unique provider/model/auth IDs to avoid conflicts under `t.Parallel()`, and I tightened the assertions to verify the exact auth IDs attempted.

I also fixed the related stream bootstrap error handling while revalidating this path.

luispater (approved):
Summary: The previous blocking issue is resolved. The new tests isolate provider/model/auth IDs per test and assert the exact auth attempts, so parallel runs no longer collide on shared registry state. The updated handler bootstrap retry path and auth rotation regressions pass the targeted package tests.

Test plan:
- `go test ./sdk/cliproxy/auth ./sdk/api/handlers`

luispater:

luispater:
Thanks for the fix and tests here.

I rechecked the original issue (#2189) against the current `dev`/`main` branch. The same failure mode is already covered in the current code: stream bootstrap errors and empty streams now return a non-nil bootstrap error from `executeStreamWithModelPool`, allowing `executeStreamMixedOnce` to rotate to the next auth credential instead of treating the stream as successful.

That fix reached the current release line through #2222, and #2189 has been closed as completed. Since this PR is now superseded by the already-merged implementation, I am closing it to keep the PR queue clean.

PR #2224, opened 2026-03-19
Add OpenAI-compatible audio transcription endpoint

madwiki:
## Summary

This PR adds support for OpenAI-compatible audio transcription requests via `POST /v1/audio/transcriptions`.

The new endpoint is implemented as a provider-native multipart proxy path, reusing the existing auth selection flow instead of introducing any client-specific behavior or additional credential requirements.

## What Changed

- Added `POST /v1/audio/transcriptions`
- Accepted multipart form fields:
  - `model`
  - `file`
  - `prompt` (optional)
  - `language` (optional)
- Normalized responses to JSON with at least a `text` field
- Added a raw HTTP execution path in the auth manager for provider-native endpoints that do not fit the existing JSON translation pipeline
- Preserved existing Codex/OpenAI auth selection behavior for audio requests
- Updated Codex HTTP request preparation so multipart requests keep their original content type while still applying provider-specific auth headers
- Exposed `POST /v1/audio/transcriptions` in the root endpoint list
- Registered `gpt-4o-mini-transcribe` in Codex model sets
- Added tests covering route exposure, multipart validation, response normalization, and Codex request preparation

## Why

CLIProxyAPI already supports OpenAI-compatible chat and responses endpoints, but it did not expose the speech-to-text transcription endpoint.

Because audio transcription uses `multipart/form-data`, it does not fit cleanly into the existing JSON translator flow. This change adds the missing capability with minimal impact on the current architecture by routing it through a dedicated raw HTTP execution path.

## Testing

- `go test ./...`

### DIFF
--- a/internal/registry/model_definitions.go
+++ b/internal/registry/model_definitions.go
@@ -13,6 +13,7 @@ type staticModelsJSON struct {
 	Vertex      []*ModelInfo `json:"vertex"`
 	GeminiCLI   []*ModelInfo `json:"gemini-cli"`
 	AIStudio    []*ModelInfo `json:"aistudio"`
+	CodexShared []*ModelInfo `json:"codex-shared,omitempty"`
 	CodexFree   []*ModelInfo `json:"codex-free"`
 	CodexTeam   []*ModelInfo `json:"codex-team"`
 	CodexPlus   []*ModelInfo `json:"codex-plus"`
@@ -50,22 +51,22 @@ func GetAIStudioModels() []*ModelInfo {
 
 // GetCodexFreeModels returns model definitions for the Codex free plan tier.
 func GetCodexFreeModels() []*ModelInfo {
-	return cloneModelInfos(getModels().CodexFree)
+	return cloneModelInfos(mergeStaticModelSections(getModels().CodexShared, getModels().CodexFree))
 }
 
 // GetCodexTeamModels returns model definitions for the Codex team plan tier.
 func GetCodexTeamModels() []*ModelInfo {
-	return cloneModelInfos(getModels().CodexTeam)
+	return cloneModelInfos(mergeStaticModelSections(getModels().CodexShared, getModels().CodexTeam))
 }
 
 // GetCodexPlusModels returns model definitions for the Codex plus plan tier.
 func GetCodexPlusModels() []*ModelInfo {
-	return cloneModelInfos(getModels().CodexPlus)
+	return cloneModelInfos(mergeStaticModelSections(getModels().CodexShared, getModels().CodexPlus))
 }
 
 // GetCodexProModels returns model definitions for the Codex pro plan tier.
 func GetCodexProModels() []*ModelInfo {
-	return cloneModelInfos(getModels().CodexPro)
+	return cloneModelInfos(mergeStaticModelSections(getModels().CodexShared, getModels().CodexPro))
 }
 
 // GetQwenModels returns the standard Qwen model definitions.
@@ -96,6 +97,31 @@ func cloneModelInfos(models []*ModelInfo) []*ModelInfo {
 	out := make([]*ModelInfo, len(models))
 	for i, m := range models {
 		out[i] = cloneModelInfo(m)
+	}
+	return out
+}
+
+func mergeStaticModelSections(sections ...[]*ModelInfo) []*ModelInfo {
+	var out []*ModelInfo
+	seen := make(map[string]struct{})
+	for _, section := range sections {
+		for _, model := range section {
+			if model == nil {
+				continue
+			}
+			modelID := strings.TrimSpace(model.ID)
+			if modelID == "" {
+				continue
+			}
+			if _, exists := seen[modelID]; exists {
+				continue
+			}
+			seen[modelID] = struct{}{}
+			out = append(out, model)
+		}
+	}
+	if len(out) == 0 {
+		return nil
 	}
 	return out
 }
@@ -156,7 +182,7 @@ func LookupStaticModelInfo(modelID string) *ModelInfo {
 		data.Vertex,
 		data.GeminiCLI,
 		data.AIStudio,
-		data.CodexPro,
+		mergeStaticModelSections(data.CodexShared, data.CodexPro),
 		data.Qwen,
 		data.IFlow,
 		data.Kimi,
--- a/internal/registry/model_definitions_test.go
+++ b/internal/registry/model_definitions_test.go
@@ -1,0 +1,131 @@
+package registry
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestCodexTierModelsIncludeSharedTranscriptionModels(t *testing.T) {
+	testCases := []struct {
+		name   string
+		models []*ModelInfo
+	}{
+		{name: "free", models: GetCodexFreeModels()},
+		{name: "team", models: GetCodexTeamModels()},
+		{name: "plus", models: GetCodexPlusModels()},
+		{name: "pro", models: GetCodexProModels()},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if !modelListContainsID(tc.models, "gpt-4o-mini-transcribe") {
+				t.Fatalf("%s tier missing gpt-4o-mini-transcribe", tc.name)
+			}
+			if !modelListContainsID(tc.models, "whisper-1") {
+				t.Fatalf("%s tier missing whisper-1", tc.name)
+			}
+		})
+	}
+}
+
+func TestLookupStaticModelInfoFindsSharedCodexModels(t *testing.T) {
+	for _, modelID := range []string{"gpt-4o-mini-transcribe", "whisper-1"} {
+		info := LookupStaticModelInfo(modelID)
+		if info == nil {
+			t.Fatalf("LookupStaticModelInfo(%q) = nil", modelID)
+		}
+		if info.ID != modelID {
+			t.Fatalf("LookupStaticModelInfo(%q).ID = %q, want %q", modelID, info.ID, modelID)
+		}
+	}
+}
+
+func TestValidateModelsCatalogRejectsCodexSharedOverlap(t *testing.T) {
+	data := newStaticModelsCatalogFixture()
+	data.CodexShared = []*ModelInfo{testStaticModel("shared-transcribe")}
+	data.CodexFree = []*ModelInfo{testStaticModel("shared-transcribe")}
+
+	err := validateModelsCatalog(data)
+	if err == nil || !strings.Contains(err.Error(), `codex-shared overlaps codex-free on model id "shared-transcribe"`) {
+		t.Fatalf("validateModelsCatalog() error = %v, want overlap error", err)
+	}
+}
+
+func TestDetectChangedProvidersTreatsCodexSharedAndDuplicatedSchemasAsEquivalent(t *testing.T) {
+	oldData := newStaticModelsCatalogFixture()
+	oldData.CodexFree = []*ModelInfo{
+		testStaticModel("free-only"),
+		testStaticModel("gpt-4o-mini-transcribe"),
+		testStaticModel("whisper-1"),
+	}
+	oldData.CodexTeam = []*ModelInfo{
+		testStaticModel("team-only"),
+		testStaticModel("gpt-4o-mini-transcribe"),
+		testStaticModel("whisper-1"),
+	}
+	oldData.CodexPlus = []*ModelInfo{
+		testStaticModel("plus-only"),
+		testStaticModel("gpt-4o-mini-transcribe"),
+		testStaticModel("whisper-1"),
+	}
+	oldData.CodexPro = []*ModelInfo{
+		testStaticModel("pro-only"),
+		testStaticModel("gpt-4o-mini-transcribe"),
+		testStaticModel("whisper-1"),
+	}
+
+	newData := newStaticModelsCatalogFixture()
+	newData.CodexShared = []*ModelInfo{
+		testStaticModel("gpt-4o-mini-transcribe"),
+		testStaticModel("whisper-1"),
+	}
+	newData.CodexFree = []*ModelInfo{testStaticModel("free-only")}
+	newData.CodexTeam = []*ModelInfo{testStaticModel("team-only")}
+	newData.CodexPlus = []*ModelInfo{testStaticModel("plus-only")}
+	newData.CodexPro = []*ModelInfo{testStaticModel("pro-only")}
+
+	changed := detectChangedProviders(oldData, newData)
+	if stringSliceContains(changed, "codex") {
+		t.Fatalf("detectChangedProviders() = %v, want codex unchanged", changed)
+	}
+}
+
+func newStaticModelsCatalogFixture() *staticModelsJSON {
+	return &staticModelsJSON{
+		Claude:      []*ModelInfo{testStaticModel("claude-base")},
+		Gemini:      []*ModelInfo{testStaticModel("gemini-base")},
+		Vertex:      []*ModelInfo{testStaticModel("vertex-base")},
+		GeminiCLI:   []*ModelInfo{testStaticModel("gemini-cli-base")},
+		AIStudio:    []*ModelInfo{testStaticModel("aistudio-base")},
+		CodexFree:   []*ModelInfo{testStaticModel("free-base")},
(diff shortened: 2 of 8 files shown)


madwiki:
> ## Code Review
> Good work adding the audio transcription endpoint. The overall structure is solid — proper multipart parsing, file validation, auth routing, and response normalization. A few observations:
> 
> ### Potential Issues
> 1. **Unbounded file read into memory** — `io.ReadAll(file)` in `parseAudioTranscriptionRequest` and `io.ReadAll(upstreamResp.Body)` in `AudioTranscriptions` have no size cap beyond the 32MB multipart form limit. If an upstream response is unexpectedly large (e.g. verbose JSON from a misconfigured provider), this could spike memory. Consider adding a `LimitedReader` on the upstream response body.
> 2. **`cliCancel` called with body bytes** — In the success path, `cliCancel(normalizedBody)` passes the entire response body as the cancel reason. If this is for logging/tracing, fine, but make sure the downstream context handler can handle `[]byte` (not just `error` or `string`). If it expects a specific type, this could silently fail or panic depending on the assertion.
> 3. **Model registration duplication** — `gpt-4o-mini-transcribe` is copy-pasted into 4 tier blocks in `models.json` (codex-team, codex-plus, codex-pro, and the unnamed block). If tiers diverge or a new transcription model is added, this is easy to get out of sync. Not blocking, but worth noting for maintainability.
> 
> ### Nits
> * `applyCodexHeaders` now calls `applyCodexPreparedHeaders` and then unconditionally sets `Content-Type` to `application/json`. The `PrepareRequest` path intentionally skips setting Content-Type (so multipart is preserved). This split is correct but fragile — a comment on `applyCodexHeaders` noting it is only for JSON request paths would help future readers.
> * `resolveAudioTranscriptionURL` falls back to `https://chatgpt.com/backend-api/transcribe` for codex auth without a base_url. If that endpoint ever changes or requires different auth headers, this hardcoded URL could silently break. Consider making it configurable or at least adding a constant with a comment.
> 
> ### Good
> * File validation checks both extension and MIME type with a fallback to `http.DetectContentType` — solid defensive approach.
> * Tests cover both the happy path (plain-text wrapping) and rejection of unsupported formats.
> * The `HTTPResponseError` type with `StatusCode()` and `Headers()` is clean and will compose well with the existing retry machinery.
> 
> Overall: looks ready with the memory/size concern as the main thing to address before merge.

Thanks for the review. Addressed in 78b3a104.

This follow-up keeps the external API unchanged and tightens the implementation around memory usage, bounds, and maintainability.

Changes made:
- Replaced the previous ParseMultipartForm + full in-memory buffering path with a streaming MultipartReader flow.
- The uploaded audio is now staged into a temp file, and the outbound multipart request is rebuilt by reopening that staged file instead of buffering the whole file in memory.
- Bounded upstream response reads and now return a clear 502-style error if the upstream transcription payload exceeds the configured limit.
- Replaced the relevant numeric and URL literals with named constants/helpers, including the default Codex transcription endpoint.
- Added focused tests for response normalization edge cases, default Codex URL resolution, oversized upstream responses, auth failover on retriable upstream HTTP failures, and staged temp-file cleanup.
- Added a short comment clarifying that applyCodexHeaders is the JSON-request helper, while the prepared-request path intentionally preserves caller-provided content types such as multipart/form-data.

On the cliCancel point, I did not change that path because the existing cancel helper already handles []byte payloads explicitly, so this is not a new issue introduced by this endpoint.

madwiki:
Addressed in ea79e77 in response to the Codex review feedback.

This follow-up fixes the compatibility and scheduler issues called out there:
- explicit non-JSON transcription formats are now preserved instead of always being wrapped as JSON
- non-file multipart fields are now bounded
- the raw HTTP path now applies resolved upstream models before request construction
- auth-specific request-build failures now fall through to the next eligible model/auth when appropriate
- Codex transcription URL derivation now follows the existing Codex base_url semantics

I intentionally kept this scoped to the issues that affected correctness, compatibility, or robustness.

madwiki:
Addressed in 8cdeddc in response to the latest Codex review feedback.

This follow-up fixes the two error-path issues called out there:
- raw HTTP upstream error bodies are now bounded before constructing `HTTPResponseError`
- upstream error headers are now filtered through the same header filter before being surfaced to clients when passthrough headers are enabled

I did not add transcription streaming support in this pass, since that would be a broader compatibility expansion beyond the current non-streaming transcription flow.

madwiki:
Addressed in 35a54e4 in response to the latest Codex review feedback.

This follow-up fixes the remaining `model=auto` routing gap for audio transcription. Instead of resolving `auto` from a global transcription-looking model list, the audio path now resolves it against auths that can actually serve `/v1/audio/transcriptions` and reuses the auth-manager's existing execution-model candidate logic.

That means:
- `auto` no longer picks models from unsupported providers just because they look transcription-capable in the registry
- OpenAI-compatible alias / model-pool setups are now respected when choosing the audio route model
- the selected route model still expands to the correct upstream transcription model during raw HTTP execution

I also added regression coverage for both cases: avoiding a newer non-audio chat model and resolving through an OpenAI-compatible alias setup.

madwiki:
> ## Code Review
> **Summary:** Large, well-structured PR adding an OpenAI-compatible `/v1/audio/transcriptions` endpoint. Handles multipart file upload, temp file staging, content-type preservation, auto-model resolution, and response normalization. Includes comprehensive tests.
> 
> ### What looks good
> * Solid input validation: file size limits, content-type sniffing, supported format allowlist
> * Temp file cleanup with `defer` and the `cleanupOnError` pattern is correct
> * `filteredErrorHeaders` refactoring removes duplicated header-extraction code across 3 call sites
> * Header preservation in `applyCodexPreparedHeaders` correctly avoids overwriting multipart content types
> * Test coverage is thorough (874 lines of tests)
> 
> ### Issues to consider
> 1. **Temp file race window on crash.** If the process crashes between creating the temp file and `Cleanup()`, orphaned files with prefix `cliproxy-audio-transcription-*` accumulate in the OS temp dir. Consider using a dedicated subdirectory that can be swept on startup, or setting a TTL-based cleanup in a background goroutine.
> 2. **`resolveAutoAudioModelBase` iterates all auths and all models.** For deployments with many auth entries and large model registries, this could be slow on every `auto` request. Consider caching the resolved model with a short TTL (e.g., 30s) since the model registry does not change frequently.
> 3. **`audioTranscriptionUploadLimitBytes` is 32MB but no `Content-Length` pre-check.** The limit is enforced after reading, so a malicious client can still force the server to read up to 32MB before rejection. Adding a quick `Content-Length` header check before streaming would short-circuit obviously oversized uploads.
> 4. **Models registry duplication.** The `gpt-4o-mini-transcribe` entry is copy-pasted identically across 4 tier blocks in `models.json`. If this model needs updating, all 4 must be touched. Consider a shared definition or post-processing step.
> 5. **Missing `gpt-4o-transcribe` model.** Only `gpt-4o-mini-transcribe` is added. OpenAI also offers the full `gpt-4o-transcribe` model. Intentional omission, or should it be included?
> 
> ### Nitpicks
> * `audioTranscriptionContentSniffBytes = 512` is fine but `http.DetectContentType` only reads the first 512 bytes by design, so this is a perfect match. Worth a comment noting this coupling.
> * The `countingWriter` is clean but could be a shared utility since similar patterns exist elsewhere in the codebase.
> 
> Solid work overall. The core flow is secure and well-tested. The main concern is temp file lifecycle management.

Addressed the remaining audio-transcription follow-ups in:
- c8a585b (`Realign audio transcription with auth manager semantics`)
- fc3a342 (`Document audio passthrough and validate response formats`)

This update tightens the audio transcription path so it follows the existing auth-manager / selector semantics more closely and closes the remaining compatibility gaps from review.

Changes included:
- added a read-only execution-model preview path so audio `model=auto` no longer mutates alias/model-pool rotation during inspection
- pinned the auth selected by audio `auto` into the actual raw HTTP execution path, so preview and execution no longer diverge across OpenAI-compatible providers
- made provider-specific modality metadata authoritative for transcription-capability detection, with name-based fallback only when metadata is absent
- added explicit `whisper-1` registry support so direct `model=whisper-1` requests resolve through the normal provider/model path
- changed successful JSON normalization to stage and stream from temp files instead of reading the full response back into memory
- documented in route setup that audio transcription uses a direct multipart passthrough path rather than the JSON translator stack
- added early `response_format` validation so unsupported formats now fail fast with an OpenAI-style invalid request error instead of being passed through silently

Validation:
- `go test ./sdk/api/handlers/openai ./sdk/cliproxy/auth`
- `go test ./...`

xkonjin (commented):
## Code Review — Audio Transcription Endpoint

Nice work on this feature. The multipart passthrough approach is the right call for binary payloads, and the test coverage is solid. A few observations:

### Potential Issues

1. **Model registry duplication** — `gpt-4o-mini-transcribe` and `whisper-1` entries are copy-pasted across 4 tier blocks (openai, codex-team, codex-plus, codex-pro). If a model gets added or updated, you need to touch 4 places. Consider extracting shared model definitions or at minimum leaving a comment pointing at the canonical block.

2. **Temp file cleanup on panic paths** — `stageFilePart` and `normalizeAudioTranscriptionResponseFromReader` both use a `keepTempFile` bool pattern with deferred cleanup. If any goroutine panics between setting `keepTempFile = true` and the caller registering cleanup, the temp file leaks. The `writeMultipartBody` goroutine is fire-and-forget with no recovery, so a panic there could leave the pipe hanging. Consider adding a `recover()` in the goroutine.

3. **`normalizeAudioTranscriptionResponse` (in-memory version)** is defined but only used in tests. The production path uses `normalizeAudioTranscriptionResponseFromReader` which stages to disk. If the in-memory version drifts out of sync with the file-based one, test coverage becomes misleading. Worth either removing the in-memory version or adding a property test asserting equivalence.

4. **`audioTranscriptionUploadLimitBytes` is 32MB** — the newer `gpt-4o-mini-transcribe` model may support larger files. Worth making this configurable or documenting the assumption.

### Minor

- `filteredErrorHeaders` helper is a good DRY refactor across three call sites.
- `applyCodexPreparedHeaders` nil-guard on `r.Header` is defensive in a good way.
- Content-length test (`lastLength > 0`) catches the regression where multipart bodies sent with unknown length.

### Security

- File type validation via both extension and MIME sniffing is solid.
- `audioTranscriptionNonFileFieldsLimitBytes` cap at 1MB prevents abuse via giant prompt fields.
- No path traversal risk since filenames are only used in multipart Content-Disposition headers upstream.

Main thing to address before merge: goroutine panic safety in `writeMultipartBody`.

madwiki:
> ## Code Review — Audio Transcription Endpoint
> Nice work on this feature. The multipart passthrough approach is the right call for binary payloads, and the test coverage is solid. A few observations:
> 
> ### Potential Issues
> 1. **Model registry duplication** — `gpt-4o-mini-transcribe` and `whisper-1` entries are copy-pasted across 4 tier blocks (openai, codex-team, codex-plus, codex-pro). If a model gets added or updated, you need to touch 4 places. Consider extracting shared model definitions or at minimum leaving a comment pointing at the canonical block.
> 2. **Temp file cleanup on panic paths** — `stageFilePart` and `normalizeAudioTranscriptionResponseFromReader` both use a `keepTempFile` bool pattern with deferred cleanup. If any goroutine panics between setting `keepTempFile = true` and the caller registering cleanup, the temp file leaks. The `writeMultipartBody` goroutine is fire-and-forget with no recovery, so a panic there could leave the pipe hanging. Consider adding a `recover()` in the goroutine.
> 3. **`normalizeAudioTranscriptionResponse` (in-memory version)** is defined but only used in tests. The production path uses `normalizeAudioTranscriptionResponseFromReader` which stages to disk. If the in-memory version drifts out of sync with the file-based one, test coverage becomes misleading. Worth either removing the in-memory version or adding a property test asserting equivalence.
> 4. **`audioTranscriptionUploadLimitBytes` is 32MB** — the newer `gpt-4o-mini-transcribe` model may support larger files. Worth making this configurable or documenting the assumption.
> 
> ### Minor
> * `filteredErrorHeaders` helper is a good DRY refactor across three call sites.
> * `applyCodexPreparedHeaders` nil-guard on `r.Header` is defensive in a good way.
> * Content-length test (`lastLength > 0`) catches the regression where multipart bodies sent with unknown length.
> 
> ### Security
> * File type validation via both extension and MIME sniffing is solid.
> * `audioTranscriptionNonFileFieldsLimitBytes` cap at 1MB prevents abuse via giant prompt fields.
> * No path traversal risk since filenames are only used in multipart Content-Disposition headers upstream.
> 
> Main thing to address before merge: goroutine panic safety in `writeMultipartBody`.

Thanks a lot for the careful review, and sorry for the churn on this PR. You were right to call out the remaining rough edges.

I addressed the follow-ups in `fc3a342d` and `07f44a8c`:

- documented the audio transcription route as a direct multipart passthrough path and tightened early `response_format` validation
- unified audio `response_format` handling and added `diarized_json`, so validation, `Accept` selection, and response handling now stay consistent
- fixed audio `model=auto` selection so globally cooling-down auths are no longer treated as selectable when model-specific state is missing
- hardened the multipart writer goroutine by recovering panics and surfacing them as request errors
- removed the duplicate test-only normalization path and switched those tests to exercise the real file-backed normalization flow
- cleaned up the transcription-model duplication introduced in this PR by adding an optional shared Codex model layer while keeping compatibility with the existing duplicated remote catalog shape

I also added regression coverage around these paths and re-ran `go test ./...`.

On the broader temp-file orphan point: I agree the whole-process crash window is real. I kept this pass focused on request-path cleanup and panic hardening, since fully solving crash-time orphan cleanup would need a separate process-lifecycle policy rather than more endpoint-local logic.

Thanks again for the thoughtful review.

nlgtuankiet:
Good work @madwiki 👍 , really look forward to this PR getting merged

nlgtuankiet:
Look like all good, can you please take a look @luispater?

luispater:
No plans to support this.

nlgtuankiet:
I’m building an AI assistant, and Codex / GPT-5.x models don’t support audio input yet.
Right now I fall back to Gemini for that, but it’s super slow. 

Any concern by adding support for `gpt-4o-mini-transcribe` @luispater ?

smwbev:
I saw the [earlier decision not to support this](https://github.com/router-for-me/CLIProxyAPI/pull/2224#issuecomment-4111960396), so I am following up here rather than opening a duplicate. With the related Grok STT work in #5386, would a narrowly scoped OpenAI-compatible file-transcription route now be worth reconsidering?

My use case is a self-hosted Hermes agent: I would like to send recorded audio through the proxy instead of running local faster-whisper.

New evidence from a relay deployment (2026-09-22):
- Authenticated `GET /v1/models` returned HTTP 200 and listed `meta/muse-voice-transcribe-1.0`.
- The multipart request below, with a generated one-second silent WAV, returned **HTTP 404, empty body, no Content-Type header**. The deployment SHA is unknown, so this observation is separate from the source inspection.

```sh
curl -i -X POST 'https://RELAY.example/v1/audio/transcriptions' \
  -H 'Authorization: Bearer ***' \
  -F 'model=meta/muse-voice-transcribe-1.0' \
  -F 'response_format=json' \
  -F 'file=@probe.wav;type=audio/wav'
```

At inspected commit `e01806f971b1758b23bb067d93f7d2acd73d2c70`, the [OpenAI route registrations](https://github.com/router-for-me/CLIProxyAPI/blob/e01806f971b1758b23bb067d93f7d2acd73d2c70/internal/api/server_routes.go#L61-L100) lack `/v1/audio/transcriptions`; realtime transcription sessions do exist. Also, the merged Meta/Muse Code provider in #5502 and the relay catalog entry do **not** establish that native Meta credentials or the current executor support voice transcription; that endpoint and entitlement still need verification.

The requested scope would be `POST /v1/audio/transcriptions` using the OpenAI multipart contract: `file`, `model`, optional `language`, `prompt`, `response_format` (`json`, `text`, `srt`, `verbose_json`, `vtt`), `temperature`, and `timestamp_granularities[]`. Reuse existing auth/model mapping and route only to verified transcription-capable upstreams (initially Whisper-class/OpenAI-compatible providers; Muse Voice only if its contract and credentials are confirmed). Formats/options should respect upstream capabilities rather than imply universal support. Eligible models could optionally appear in `/v1/models`; `/v1/audio/translations` can remain a separate follow-up.

If the earlier decision still stands, understood. Otherwise, I would be happy to help test multipart uploads and Hermes integration. Thank you!

PR #2252, opened 2026-03-20
fix(antigravity): always include text field in thought parts to prevent Google 500

sususu98:
## Summary

When Claude sends redacted thinking with empty text (`"thinking": ""`), the translator omits the `"text"` field from thought parts. Google Antigravity API requires this field, causing 500 "Unknown Error" responses.

## Root Cause

In `antigravity_claude_request.go`, the thought block construction had a conditional:

```go
if thinkingText != "" {
    partJSON, _ = sjson.SetBytes(partJSON, "text", thinkingText)
}
```

This produces `{"thought": true, "thoughtSignature": "..."}` without a `"text"` field when thinking text is empty (redacted thinking). Google's API fails to parse this and returns a generic 500.

## Fix

Always include the `"text"` field, even when empty:

```go
partJSON, _ = sjson.SetBytes(partJSON, "text", thinkingText)
```

## Verification

Analyzed 270 log files (173 error + 97 success):

| | Empty thought (no text) | Thought with text | No thought |
|---|---|---|---|
| **500 Unknown Error** | **129** | 0 | **0** |
| **200 Success** | **0** | 61 | 35 |

After deploying the fix: **0 new "Unknown Error" 500s** observed, confirming the root cause.

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -104,59 +104,59 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 
 						// Always try cached signature first (more reliable than client-provided)
 						// Client may send stale or invalid signatures from different sessions
-						signature := ""
-						if thinkingText != "" {
-							if cachedSig := cache.GetCachedSignature(modelName, thinkingText); cachedSig != "" {
-								signature = cachedSig
-								// log.Debugf("Using cached signature for thinking block")
-							}
+			signature := ""
+					if thinkingText != "" {
+						if cachedSig := cache.GetCachedSignature(modelName, thinkingText); cachedSig != "" {
+							signature = cachedSig
+							// log.Debugf("Using cached signature for thinking block")
 						}
+					}
 
-						// Fallback to client signature only if cache miss and client signature is valid
-						if signature == "" {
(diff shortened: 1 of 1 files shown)


xkonjin (commented):
## Code Review

**Summary:** Always includes the `text` field in thought parts sent to Google Antigravity API, even when empty. Prevents Google 500 errors from missing required fields.

### Issues Found

**🐛 Indentation regression (blocking)**
The entire thinking-block handling section (lines ~107-160) has been re-indented from tabs to mixed spaces, breaking the Go formatting convention used throughout this file. The functional change is a single-line removal of the conditional:
```go
// Before: text was conditional
if thinkingText != "" {
    partJSON, _ = sjson.SetBytes(partJSON, "text", thinkingText)
}
// After: text is always set (the actual fix)
partJSON, _ = sjson.SetBytes(partJSON, "text", thinkingText)
```

The fix itself is correct and minimal, but the surrounding whitespace changes make the diff 5x larger than needed and introduce inconsistent indentation (spaces where the rest of the file uses tabs). Please run `gofmt` or `goimports` on this file and split the formatting into a separate commit, or revert the whitespace-only changes.

**✅ Logic is sound**
- Unconditionally setting `text` (even to `""`) is the right fix for the Google API requirement.
- No behavioral change to signature validation or unsigned-block dropping.

### Verdict
The one-line fix is correct and addresses a real production issue. The whitespace noise should be cleaned up before merge to keep blame/history useful.

PR #2266, opened 2026-03-22
fix(websocket): skip stale state merge after client-side compact

### LINKED ISSUE #2207: bug(codex): `invalid_request_error` after auto compact
[x] No, it's another issue.

**Describe the bug**
Use codex  with websocket

```
supports_websockets = true
```
`invalid_request_error` always occurs after codex auto compact.

```
Context compacted

⚠ Heads up: Long threads and multiple compactions can cause the model to be less accurate. Start a new thread when possible to keep threads small and targeted.

■ {"type":"error","status":400,"error":{
    "message": "No tool output found for function call call_rPQoj8lo6TtbE0VmFZMzOdMT.",
    "param": "input",
    "code": null
  }}
```

DragonFSKY:
## Summary
- After Codex CLI compact, the client sends a full conversation transcript as input (with `compaction`/`compaction_summary` items). `normalizeResponseSubsequentRequest()` was unconditionally merging this with stale `lastRequest`/`lastResponseOutput`, breaking `function_call`/`function_call_output` pairings and causing 400 errors ("No tool output found for function call").
- Added `inputContainsFullTranscript()` heuristic to detect full transcripts (compaction items or assistant messages), bypassing the stale merge when present.
- Normal incremental flow (user messages / function_call_output only) is unaffected.

## Test plan
- [x] Unit tests: 5 new tests covering detection heuristic + merge bypass + incremental merge preservation
- [x] Production verification: deployed to staging, triggered compact via Codex CLI, confirmed `full transcript detected, skipping stale merge (input items=127)` in logs with no subsequent 400 errors

Fixes #2207

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket.go
@@ -315,20 +315,32 @@ func normalizeResponseSubsequentRequest(rawJSON []byte, lastRequest []byte, last
 		}
 	}
 
-	existingInput := gjson.GetBytes(lastRequest, "input")
-	mergedInput, errMerge := mergeJSONArrayRaw(existingInput.Raw, normalizeJSONArrayRaw(lastResponseOutput))
-	if errMerge != nil {
-		return nil, lastRequest, &interfaces.ErrorMessage{
-			StatusCode: http.StatusBadRequest,
-			Error:      fmt.Errorf("invalid previous response output: %w", errMerge),
+	// When the client sends a full conversation transcript (e.g. after compact),
+	// the input already contains the complete history including assistant messages.
+	// In that case, skip merging with stale lastRequest/lastResponseOutput to avoid
+	// breaking function_call / function_call_output pairings.
+	// See: https://github.com/router-for-me/CLIProxyAPI/issues/2207
+	var mergedInput string
+	if inputContainsFullTranscript(nextInput) {
+		log.Infof("responses websocket: full transcript detected, skipping stale merge (input items=%d)", len(nextInput.Array()))
+		mergedInput = nextInput.Raw
+	} else {
+		existingInput := gjson.GetBytes(lastRequest, "input")
+		var errMerge error
+		mergedInput, errMerge = mergeJSONArrayRaw(existingInput.Raw, normalizeJSONArrayRaw(lastResponseOutput))
+		if errMerge != nil {
+			return nil, lastRequest, &interfaces.ErrorMessage{
+				StatusCode: http.StatusBadRequest,
+				Error:      fmt.Errorf("invalid previous response output: %w", errMerge),
+			}
 		}
-	}
 
-	mergedInput, errMerge = mergeJSONArrayRaw(mergedInput, nextInput.Raw)
-	if errMerge != nil {
-		return nil, lastRequest, &interfaces.ErrorMessage{
-			StatusCode: http.StatusBadRequest,
-			Error:      fmt.Errorf("invalid request input: %w", errMerge),
+		mergedInput, errMerge = mergeJSONArrayRaw(mergedInput, nextInput.Raw)
+		if errMerge != nil {
+			return nil, lastRequest, &interfaces.ErrorMessage{
+				StatusCode: http.StatusBadRequest,
+				Error:      fmt.Errorf("invalid request input: %w", errMerge),
+			}
 		}
 	}
 	dedupedInput, errDedupeFunctionCalls := dedupeFunctionCallsByCallID(mergedInput)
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Thanks for the fix and the focused tests around compact replay behavior.

I’m requesting changes for one correctness risk in the new detection heuristic:

- **Blocking:** The new `inputContainsFullTranscript()` logic marks any assistant-role message as “full transcript,” which causes `normalizeResponseSubsequentRequest()` to skip merging prior state. That can drop required historical context in incremental (non-compact) requests that legitimately contain assistant items, resulting in behavior changes outside the compact scenario this PR targets.

**Suggested adjustment**
- Narrow full-transcript detection to stronger replay signals (for example, `compaction` / `compaction_summary`), or gate assistant-message detection with additional conditions that distinguish replay from incremental append.

**Tests**
- Please add one regression test where incremental input includes an assistant message but should still preserve prior merged history. Current tests cover compact replay and user/function_call_output incremental flows, but not this ambiguity.

Once this heuristic is tightened, I’m happy to re-review.

DragonFSKY:
Thanks, that was a valid regression risk.

I tightened the full-transcript detection so we now only treat compact replay markers (`compaction` / `compaction_summary`) as a signal to skip the stale merge. Assistant messages alone no longer bypass the merge path.

I also added a regression test covering incremental assistant input to make sure prior history is still merged in that case, while keeping the compact replay test in place.

Targeted verification:
`go test ./sdk/api/handlers/openai -run 'Test(InputContainsFullTranscript|NormalizeSubsequentRequest)'`

luispater (approved):
The follow-up correctly narrows replay detection to compaction markers, keeps incremental assistant-message flows on the merge path, and adds a Codex-only bypass for compact replays. The new tests cover both the bypass and the unsupported-downstream fallback, so I do not see a remaining blocking correctness issue in `sdk/api/handlers/openai/openai_responses_websocket.go` or `sdk/api/handlers/openai/openai_responses_websocket_test.go`.

GitHub checks are green, so this looks ready.

luispater:

KajiMaCN:
> This PR is approved but currently has merge conflicts. Please change the base branch to `dev`, resolve the conflicts, and then wait for merge.

@DragonFSKY @luispater 

I pulled this branch over a week ago, and it had been working fine up until today. But now the 400 error has suddenly reappeared:

```json
{
  "type": "error",
  "status": 400,
  "error": {
    "message": "No tool call found for custom tool call output with call_id call_xxx.",
    "param": "input",
    "code": null
  }
}
```
What seems a bit different from before is that if I ask Codex to modify the text, the content in the document does get changed, but a 400 error appears in the Codex chat window, and GPT’s response message does not show up.

Also, this now happens in every conversation. Only during the first 2–3 seconds do normal messages appear. After that, it always turns into a 400 error. Codex’s reply also disappears, leaving only the error code shown above.

After I switched to the main branch and rebuilt the container locally, the 400 error disappeared.

Could you help check and confirm what might be causing this?

DragonFSKY:
@KajiMaCN Hi, thanks for the detailed report.

The 400 error you're seeing (`No tool call found for custom tool call output with call_id`) is caused by tool-call / tool-call-output pairing breakage in the websocket transcript — exactly the class of bug this PR targets.

The reason it re-appeared on this branch is that the branch was based on the **old main** and was missing several complementary fixes that have since landed on `dev`:

- `c1d75998` — handle transcript replacement after websocket compaction
- `d1c07a09` — add websocket **tool call repair** with caching (auto-patches orphaned tool-call outputs)
- `b6c6379b` — fix repair logic for `custom_tool_call` specifically

These fixes work together with this PR's compaction-bypass logic to cover the full range of edge cases.

I've just **rebased the branch onto the latest `dev`**, so all of the above fixes are now included. Please re-pull and rebuild — the issue should be resolved.

If it still reproduces after the rebase, please share the request log (with `LOG_LEVEL=debug`) so I can trace the exact merge path.

KajiMaCN:
> @KajiMaCN Hi, thanks for the detailed report.
> 
> The 400 error you're seeing (`No tool call found for custom tool call output with call_id`) is caused by tool-call / tool-call-output pairing breakage in the websocket transcript — exactly the class of bug this PR targets.
> 
> The reason it re-appeared on this branch is that the branch was based on the **old main** and was missing several complementary fixes that have since landed on `dev`:
> 
> * `c1d75998` — handle transcript replacement after websocket compaction
> * `d1c07a09` — add websocket **tool call repair** with caching (auto-patches orphaned tool-call outputs)
> * `b6c6379b` — fix repair logic for `custom_tool_call` specifically
> 
> These fixes work together with this PR's compaction-bypass logic to cover the full range of edge cases.
> 
> I've just **rebased the branch onto the latest `dev`**, so all of the above fixes are now included. Please re-pull and rebuild — the issue should be resolved.
> 
> If it still reproduces after the rebase, please share the request log (with `LOG_LEVEL=debug`) so I can trace the exact merge path.

Understood. Thank you very much for your patient explanation and for fixing the bug.

luispater (approved):

PR #2272, opened 2026-03-22
fix: ensure usage records are published for 0-token responses

### LINKED ISSUE #2273: Usage statistics not recorded when upstream returns 0 tokens
### Describe the bug

When an upstream model API returns a successful response but explicitly sets all token usage metrics to `0` (e.g., `"total_tokens": 0`), the request is entirely excluded from CLIProxyAPI's usage statistics aggregation.

While it is correct to filter out empty chunks from streams to prevent spam, a completed request that happens to report `0` tokens should still be recorded as a valid request/transaction in the usage statistics (so the request count increments), rather than being silently dropped.

### Steps to reproduce

1. Configure a provider that returns `"usage": {"prompt_tokens": 0, "completion_tokens": 0, "total_tokens": 0}` in its successful JSON response.
2. Enable `usage-statistics-enabled: true` in the config.
3. Make a chat completion request to this model.
4. Observe that the token/request usage in the management dashboard remains unchanged for this key/model.

### Expected behavior

The request should be recorded in the in-memory usage statistics (incrementing the request count) even if the token delta is `0`.

### Related PR

A fix for this has been submitted in PR #2272, which adds the missing `ensurePublished` safety calls to the affected executors.

LucasInsight:
When an upstream model returns `0` tokens, CLIProxyAPI's `publishWithOutcome` correctly drops the chunk to prevent empty streaming spam. However, several executors were missing the `defer reporter.ensurePublished(ctx)` safety call at the end of their execution paths to guarantee that the request itself is logged exactly once before closing the connection. This PR patches all missing executors to ensure 0-usage responses still record the transaction cleanly.

Resolves #2273

### DIFF
--- a/internal/runtime/executor/aistudio_executor.go
+++ b/internal/runtime/executor/aistudio_executor.go
@@ -117,6 +117,7 @@ func (e *AIStudioExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth,
 	baseModel := thinking.ParseSuffix(req.Model).ModelName
 	reporter := newUsageReporter(ctx, e.Identifier(), baseModel, auth)
 	defer reporter.trackFailure(ctx, &err)
+	defer reporter.ensurePublished(ctx)
 
 	translatedReq, body, err := e.translateRequest(req, opts, false)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


luispater (changes requested):
Summary:
This PR adds a safety `ensurePublished(ctx)` fallback across executors so 0-token / missing-usage upstream responses still emit a single usage record. Good direction, and CI looks green.

Blocking:
- The defer order is currently:
  - `defer reporter.trackFailure(ctx, &err)`
  - `defer reporter.ensurePublished(ctx)`
  In Go, defers run LIFO. Since `usageReporter` uses `sync.Once`, `ensurePublished` will run first and publish a success record (`Failed=false`), which can prevent `trackFailure` from ever publishing a failure record (`Failed=true`) when `err != nil`.
  Please swap the order everywhere to:
  - `defer reporter.ensurePublished(ctx)`
  - `defer reporter.trackFailure(ctx, &err)`

Non-blocking:
- Consider a small helper to register these defers consistently to avoid reintroducing the ordering bug.

Test plan:
- CI: pr-test-build (passed)
- Recommended locally: `go test ./...`

PR #2280, opened 2026-03-23
Fix token under-reporting when translating from Anthropic to OpenAI

Traviis:
### Summary
When translating Anthropic API responses to OpenAI chat completions format, prompt_tokens excludes cache_read_input_tokens, causing it to report only the non-cached portion of the prompt on cache hits. total_tokens is similarly undercounted.

### The Bug
In internal/translator/claude/openai/chat-completions/claude_openai_response.go, both the streaming and non-streaming paths compute:
prompt_tokens = inputTokens + cacheCreationInputTokens  // missing cacheReadInputTokens
total_tokens  = inputTokens + outputTokens              // should be promptTokens + outputTokens

On a cache hit, Anthropic returns something like:
- input_tokens: 13 (non-cached portion)
- cache_read_input_tokens: 22000 (cached portion)
- cache_creation_input_tokens: 0

The current code produces prompt_tokens: 13 instead of the correct 22013.

### Impact
Per the OpenAI spec (https://platform.openai.com/docs/api-reference/chat/object), prompt_tokens should reflect the total number of prompt tokens regardless of caching, with the cached breakdown reported separately in prompt_tokens_details.cached_tokens.

This breaks any downstream client that relies on prompt_tokens for context window management. Specifically, I discovered this while using CLIProxyAPI behind LiteLLM (https://github.com/BerriAI/litellm) (as an openai provider) with OpenCode (https://github.com/anomalyco/opencode) as the client. OpenCode uses usage.prompt_tokens from the API response to determine when to trigger auto-compaction of conversation history. With the underreported token counts, compaction never triggers and conversations eventually exceed the model's context window.

LiteLLM also reports the under-counted token usage, so it's not limited to Opencode.

### Reproduction
Send the same large prompt twice through the /v1/chat/completions endpoint to trigger prompt caching, then compare the usage in the response:
First request (cold cache): prompt_tokens: 22044 — correct  
Second request (cache hit): prompt_tokens: 11 — wrong, should be ~22044

The Codex→OpenAI translator (internal/translator/codex/openai/) does not have this bug because OpenAI's native input_tokens already includes cached tokens.

In my case, I just switched to the Anthropic API for liteLLM, but figured this would help someone else.

### DIFF
--- a/internal/translator/claude/openai/chat-completions/claude_openai_response_test.go
+++ b/internal/translator/claude/openai/chat-completions/claude_openai_response_test.go
@@ -1,0 +1,136 @@
+package chat_completions
+
+import (
+	"context"
+	"testing"
+
+	"github.com/tidwall/gjson"
+)
+
+// TestStreamingUsage_CacheHit verifies that prompt_tokens includes
+// cache_read_input_tokens when translating a Claude message_delta event
+// with a cache hit to OpenAI streaming format.
+func TestStreamingUsage_CacheHit(t *testing.T) {
+	// Simulate a message_delta event from Anthropic with a cache hit:
+	// input_tokens=13 (non-cached), cache_read_input_tokens=22000 (cached), output_tokens=5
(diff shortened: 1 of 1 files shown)


Traviis:
Didn't notice the translator change block, i'll open an issue for this.

PR #2288, opened 2026-03-23
Align Gemini CLI requests with official Code Assist

ddmww:


luispater (changes requested):
Summary:
This PR moves the Gemini CLI executor closer to the official Code Assist request shape, but there are still a few blocking regressions that make it unsafe to merge as-is.

Key findings:
- Blocking: `.github/workflows/docker-image.yml` now publishes to `ghcr.io/ddmww/cliproxyapi`. This repository belongs to `router-for-me`, so merging this would either push release images into a contributor-owned namespace or fail when the upstream repo tries to publish.
- Blocking: `buildGeminiCLICountTokensPayload` only keeps `request.contents` and `request.model`, but OpenAI/Gemini system or developer messages are translated into `request.systemInstruction`. That means `countTokens` will undercount whenever a system prompt is present, so it no longer matches the actual generate request shape.
- Major: the new stable `session_id` behavior is internally inconsistent. The implementation/test expect the same prompt to reuse the same `session_id`, but `scripts/test_gemini_cli_official.py` still generates a fresh UUID every time, which matches the current local behavior instead. The new Go test also fails locally for this reason.

Test plan:
- Ran `go test ./internal/runtime/executor -run 'TestGeminiCLI'` and it failed at `TestGeminiCLIExecute_UsesStableSessionIDForSamePrompt`.
- Ran `go test ./internal/runtime/executor -run 'TestGeminiCLI(Execute_SendsOfficialHeadersAndBody|Execute_RetriesUsingConfiguredRequestRetry|ExecuteStream_UsesSSEPathWithoutLocalRetry|CountTokens_UsesOfficialBodyAndRetryConfig)$'` and those passed.

PR #2293, opened 2026-03-23
fix(security): harden management panel asset updater

Xvvln:
## Summary

The management panel auto-updater (`internal/managementasset/updater.go`) has several security weaknesses that could allow a man-in-the-middle attacker to replace the management panel HTML with malicious content. This PR hardens the updater with minimal, focused changes:

- **Abort on digest mismatch**: Previously, when the SHA256 digest of the downloaded `management.html` did not match the digest from the GitHub release API, the code only logged a warning and **proceeded to write the file anyway**. Now it logs an error and aborts the update.
- **Download size limit**: `downloadAsset()` now uses `io.LimitReader` with a 10 MB cap to prevent OOM from oversized responses.
- **Opt-in auto-update**: Added `auto-update-panel` config option under `remote-management` (default: `false`). The periodic background updater (every 3 hours) now only runs when explicitly enabled. The panel is still downloaded on first access if the local file is missing — this change only affects the silent background update loop.

## Motivation

The `management.html` file contains JavaScript that runs in the user's browser with access to the management API (API keys, credentials, config). An attacker who can intercept the download could inject malicious JS. The existing code made this easier than necessary by:
1. Ignoring hash verification failures
2. Silently auto-updating every 3 hours by default (larger attack window)
3. Having no download size limit

No existing issues or PRs have raised these concerns.

## Changes

| File | Change |
|------|--------|
| `internal/managementasset/updater.go` | Abort on digest mismatch; add `maxAssetDownloadSize` (10 MB) limit; check `AutoUpdatePanel` flag |
| `internal/config/config.go` | Add `AutoUpdatePanel bool` field to `RemoteManagement` struct |
| `config.example.yaml` | Document `auto-update-panel` option |

## Test plan

- [x] `go build ./cmd/server/` — compiles successfully
- [x] `go test ./internal/config/...` — all tests pass
- [x] `go test ./internal/managementasset/...` — no test files (existing behavior)
- [ ] CI `pr-test-build` check passes
- [ ] Manual: set `auto-update-panel: false` (default), verify no background updates occur
- [ ] Manual: set `auto-update-panel: true`, verify background updater runs normally

### DIFF
--- a/internal/managementasset/updater.go
+++ b/internal/managementasset/updater.go
@@ -288,6 +288,9 @@ func ensureFallbackManagementHTML(ctx context.Context, client *http.Client, loca
 		return false
 	}
 
+	log.Warnf("management asset downloaded from fallback URL without digest verification (hash=%s) — "+
+		"consider setting auto-update-panel: true to receive verified updates from GitHub", downloadedHash)
+
 	if err = atomicWriteFile(localPath, data); err != nil {
 		log.WithError(err).Warn("failed to persist fallback management control panel page")
 		return false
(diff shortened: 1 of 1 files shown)


Xvvln:
**Note on the fallback mechanism**

The existing `ensureFallbackManagementHTML` path (downloads from `https://cpamc.router-for.me/` when GitHub API is unreachable and no local file exists) does not perform digest verification. This PR intentionally leaves that behavior unchanged to avoid breaking first-run experience, but adds a `Warn`-level log when the fallback is used so operators are aware of the reduced security guarantee.

If the maintainers prefer, this could be addressed in a follow-up by either:
1. Removing the fallback entirely (safest, but first-run without GitHub access would show 404)
2. Hardcoding a known-good hash for the fallback asset (requires updating it on each release)

luispater:
This PR has been merged, but it violates **Rule No.2**.

You added a new config without maintaining the original logic.

So, I've renamed `auto-update-panel` to `disable-auto-update-panel`.

PR #2307, opened 2026-03-24
fix: preserve responses continuation state for openai-compatible upstreams

xpcrack:
## Summary
- preserve HTTP `/v1/responses` continuation state for openai-compatible upstreams that do not support `previous_response_id`
- cache completed response turns for both non-streaming and streaming responses
- normalize string shorthand `input` into a user message before merging continuation input
- add regression coverage for HTTP continuation and streaming completion caching

## Testing
- verified `/v1/responses` continuation against a live `gpt-5.4` openai-compatible upstream for both non-streaming and streaming flows

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -25,6 +25,7 @@ import (
 // It holds a pool of clients to interact with the backend service.
 type OpenAIResponsesAPIHandler struct {
 	*handlers.BaseAPIHandler
+	turnState responsesTurnStateCache
 }
 
 // NewOpenAIResponsesAPIHandler creates a new OpenAIResponses API handlers instance.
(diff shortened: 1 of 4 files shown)


luispater (changes requested):
Summary:
This PR improves `/v1/responses` continuation for OpenAI-compatible upstreams by caching completed turns and reconstructing continuation input when `previous_response_id` is present. The core behavior and targeted regression tests are solid.

Key findings:
- Blocking: The new turn-state cache is unbounded in practice. Entries expire logically, but cleanup only happens on exact-key reads. Since response IDs are generally not read again after expiration, memory can grow without bound over time.
- Non-blocking: Continuation normalization currently requires `input` to be an array and may reject valid string shorthand input in cache-hit flows.
- Non-blocking: Missing cache-eviction/boundedness tests.

Test plan:
- Reviewed PR metadata, file-level diff, and full patch via `gh`/script context collection.
- Reviewed CI checks (intentionally ignored `translator-path-guard` per instruction).
- Did not run local tests or checkout the PR branch in this review pass.

Requested changes:
Please add a deterministic cache cleanup/bounding strategy (for example periodic sweep, size cap + eviction policy, or both) and corresponding tests before approval.

PR #2316, opened 2026-03-24
Add per-model thinking support for OpenAI compatibility

hkfires:
Introduce a new configuration option for models to specify their thinking capabilities, allowing for more granular control over reasoning levels. Default levels are set to ["low", "medium", "high"] if not specified.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -206,7 +206,9 @@ nonstream-keepalive-interval: 0
 #       - api-key: "sk-or-v1-...b781" # without proxy-url
 #     models: # The models supported by the provider.
 #       - name: "moonshotai/kimi-k2:free" # The actual model name.
-#         alias: "kimi-k2" # The alias used in the API.
+#         alias: "kimi-k2"               # The alias used in the API.
+#         thinking:                      # optional: omit to default to levels ["low","medium","high"]
+#           levels: ["low", "medium", "high"]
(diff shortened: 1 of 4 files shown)


jroth1111:
Tested further, and I want to narrow my earlier comment.

I no longer think the Z.A.I case is strong evidence of a general `openai-compatibility` runtime bug.

What I found after more testing is that Z.A.I exposes two different integration paths, and for the GLM models I was testing (`glm-4.7`, `glm-5`, `glm-5-turbo`), the better path is the Anthropic-compatible endpoint:

- `base-url: https://api.z.ai/api/anthropic`
- configured under `claude-api-key`

That path worked cleanly for me both direct and proxied, including with `max_tokens: 32`.

So the original Z.A.I report I linked is better understood as an endpoint/recommendation issue than as strong evidence that `openai-compatibility` itself is generally wrong.

I still think this PR is useful, because provider/model-specific thinking metadata is still a real need. But I would no longer use the Z.A.I case as the main argument for a broader `openai-compatibility` request-shaping bug.

The more accurate takeaway from my testing is:

- NVIDIA-style thinking / effort mismatches still look like a real class of issue
- Z.A.I GLM models seem better served by the Anthropic-compatible endpoint than the OpenAI-compatible coding endpoint
- documenting the recommended Z.A.I integration path would probably help users more than treating that specific case as a core transport/runtime bug

PR #2346, opened 2026-03-25
fix(codex): Treat Codex capacity errors as retryable

pjpjq:
## Summary
- map Codex "Selected model is at capacity" failures into retryable 429 status errors
- keep existing reset-based retry-after handling for real usage limit responses
- add a regression test covering capacity errors returned as 400s

## Testing
- go test ./internal/runtime/executor -run 'TestParseCodexRetryAfter|TestNewCodexStatusErrTreatsCapacityAsRetryableRateLimit' -count=1\n- go test ./internal/runtime/executor -count=1

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -685,11 +685,37 @@ func applyCodexHeaders(r *http.Request, auth *cliproxyauth.Auth, token string, s
 }
 
 func newCodexStatusErr(statusCode int, body []byte) statusErr {
-	err := statusErr{code: statusCode, msg: string(body)}
-	if retryAfter := parseCodexRetryAfter(statusCode, body, time.Now()); retryAfter != nil {
+	errCode := statusCode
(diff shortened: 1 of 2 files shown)


luispater (commented):
Summary:
This PR is focused and aligned with its stated intent: treat Codex “model at capacity” failures as retryable by mapping them to 429, while preserving existing retry-after parsing for true usage-limit responses.

Key findings:
- Blocking: none.
- Non-blocking: capacity detection is text-based and status-agnostic; adding a status guard (e.g., only remap on 400/429) would reduce false-positive risk.

Test plan:
- Verified the PR adds a regression test:
  - `TestNewCodexStatusErrTreatsCapacityAsRetryableRateLimit`
- Existing retry-after behavior remains covered by `TestParseCodexRetryAfter`.

Follow-ups:
- Optional hardening: add a negative test ensuring non-capacity errors containing similar text are not remapped unintentionally.

luispater (approved):

PR #2349, opened 2026-03-25
fix(codex): preserve image content in tool outputs

DarkNoah:
Ensure tool-role message content can be translated from structured arrays so image_url items are emitted as input_image parts instead of being dropped from function_call_output payloads.


xkonjin (commented):
Nice fix overall, especially the new coverage for mixed text + image tool outputs. One correctness edge case still worries me: when a tool message comes through as an array with a content item type other than `text` or `image_url`, this translation now silently drops that part and can emit an empty/partial `output` array. That is a lossy behavior change compared to the old string passthrough path, and it could break providers/tools that return richer structured parts in `content` arrays.

I would either preserve unknown parts in a fallback form, or add a regression test proving we intentionally reject them. Right now the failure mode is quiet data loss, which is the risky version.

luispater (changes requested):
Summary:
This PR significantly improves multimodal translation for Codex tool outputs (image/file content in structured arrays) and adds solid coverage for the new paths. The main risk is a compatibility regression in `function_call_output` construction for non-string/non-array tool content.

Blocking issue:
- In `ConvertOpenAIRequestToCodex`, tool-role handling now sets `output` only for string or array content. If `content` is `null` (or other non-array/non-string), `output` is omitted entirely. Previously this path always emitted `output` via string coercion.
- Please add a fallback branch that always sets `output` (for example empty string or serialized raw content), then add regression tests.

Test plan:
- Ran: `go test ./internal/translator/codex/openai/chat-completions` (pass)
- Requested extra test coverage:
  - tool message with `content: null` should still produce a valid `function_call_output` with `output` present.

luispater (changes requested):
Summary:
The latest commits fix the original null-output issue and add broader multimodal coverage for tool outputs, but one tool-file path is still unsafe.

Blocking findings:
- `internal/translator/codex/openai/chat-completions/codex_openai_request.go:177` still emits `input_file` for tool content when only `filename` is present. That can produce an `input_file` without any real payload (`file_id`, `file_data`, or `file_url`), which is the same invalid shape the user-content path now avoids. Please require a real source field there too, or fall back to raw text/JSON, and add a regression test for filename-only tool files.

Test plan:
- `go test ./internal/translator/codex/openai/chat-completions -count=1`

PR #2366, opened 2026-03-27
Add Qoder provider support and docs

PKFireBarry:
- Add Qoder provider support across auth, registry, watcher, and executor wiring
- Expose Qoder OAuth device-flow login via CLI flag and management endpoint
- Add Qoder model catalog entries and config examples
- Update docs (EN/CN/JA) and note management UI asset refresh

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,18 +39,20 @@ Get 10% OFF GLM CODING PLAN：https://z.ai/subscribe?ic=8JVLJQFSKB
 - OpenAI Codex support (GPT models) via OAuth login
 - Claude Code support via OAuth login
 - Qwen Code support via OAuth login
+- Qoder support via OAuth login
 - iFlow support via OAuth login
 - Amp CLI and IDE extensions support with provider routing
 - Streaming and non-streaming responses
 - Function calling/tools support
 - Multimodal input support (text and images)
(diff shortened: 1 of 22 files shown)


luispater (changes requested):
Summary:
This PR adds broad Qoder support across auth, management routes, model registry, and executor wiring. The integration shape is good, but there are two correctness/reliability issues that should be fixed before merge.

Blocking issues:
1. `internal/runtime/executor/qoder_executor.go`: `http.Client{Timeout: 180 * time.Second}` in the Qoder executor introduces hard cutoff risk for long streaming sessions. This can break valid responses and conflicts with existing timeout policy for established upstream streams.
2. `internal/auth/qoder/api.go`: refresh persistence uses `storage.SaveTokenToFile("")`, which does not provide a writable auth file path. This will fail file creation and causes token refresh persistence to fail.

Suggested tests:
- Auth refresh regression test: expired/nearly-expired Qoder token should refresh and persist successfully.
- Streaming executor test: long-running SSE stream should not be aborted by client timeout.
- Tool-calls path test: verify tool call chunks are accumulated and returned in non-stream response mode.

Decision:
Request changes due to the two blocking reliability bugs above.

FojleRabbiRabib:
Any updates on this?

njuptlzf:
I am willing to continue this work based on this foundation. #3098

Issue #1962, opened 2026-03-08
[FEATURE] The fill-first routing policy adds the option to lock a unique credential for each user.


CyberRookie-X:
The fill-first routing policy adds the option to lock a unique credential for each user.

- When multiple users use the fill-first strategy, rate limiting for a single credential is easily triggered. Switching to a different credential triggers the rate limit again, and soon all credentials will also be subject to rate limiting. Credentials that have been rate-limited will have to wait several hours before they can be used again.
- The round-robin routing strategy will result in multiple times more cache writes.

PR #2374, opened 2026-03-27
fix(codex): restore prompt cache continuity for Codex requests

### LINKED ISSUE #2373: Codex prompt caching misses on repeated chat-completions requests
## Problem

Repeated Codex requests sent through the proxy do not reuse upstream prompt cache state, even when the caller repeats the same prompt sequence against the same model and provider.

In practice this shows up as:

- repeated requests behaving like fresh turns
- no observable cache read / cached-token reuse
- higher latency and token cost than expected
- cache continuity working in a reference client but not through this proxy

## Expected behavior

When repeated requests target the same Codex auth/account context and the request shape is otherwise stable, the proxy should preserve the continuity signals needed for upstream prompt caching so cache reads can occur on later requests.

## Actual behavior

The proxy was breaking Codex continuity in multiple ways:

- executor-layer request preparation stripped `prompt_cache_retention`
- Codex request preparation used non-matching continuity headers instead of the `session_id` shape used by the reference client
- OpenAI Chat Completions -> Codex translation did not preserve `prompt_cache_retention`
- repeated chat-completions requests often reached Codex without a stable `prompt_cache_key`
- when no explicit continuity key was present, `session_id` fell back to a fresh random UUID per request
- auth selection could rotate across equivalent Codex auths, which changes upstream identity and can break cache reuse even when prompts are otherwise identical

## Root cause

The proxy did not preserve the same continuity envelope expected by Codex for cache reuse.

The key issue was not just `prompt_cache_retention`. The larger problem was that repeated requests did not consistently share a stable continuity identity. In affected flows:

- `prompt_cache_key` could be absent
- `session_id` changed on every request
- auth/account selection could drift across repeated requests

This combination caused upstream Codex to treat repeated requests as unrelated instead of as part of the same reusable cache context.

## Why this matters

Without stable continuity:

- prompt caching becomes ineffective
- repeated requests pay repeated prompt-processing cost
- latency increases
- proxy behavior diverges from reference Codex/OpenAI-compatible clients
- users cannot rely on cache-friendly workflows through the proxy

## Proposed fix

Implement Codex continuity handling explicitly and consistently:

- preserve caller-provided `prompt_cache_retention` on Codex request paths
- preserve `prompt_cache_retention` for OpenAI Chat Completions -> Codex translation
- use exact `session_id` continuity header for Codex requests
- derive a stable continuity key for Codex requests using the best available source, with precedence such as:
  - explicit `prompt_cache_key`
  - execution session identity
  - explicit idempotency key
  - stable request/auth affinity
  - stable client principal fallback
  - stable auth-ID fallback
- apply that continuity key to both:
  - request header `session_id`
- keep auth selection sticky for repeated request sequences so cache continuity is not broken by auth rotation

## Scope

This issue is intentionally focused on Codex routes and continuity behavior only.

Out of scope:

- auth routing redesign
- non-Codex providers
- broader request translator redesign outside the required Codex pass-through
- unrelated cache-control semantics for other providers

## Acceptance criteria

- repeated Codex requests can reuse upstream prompt cache context when request/auth continuity is stable
- `prompt_cache_retention` is no longer stripped on Codex execution paths
- OpenAI Chat Completions -> Codex preserves `prompt_cache_retention`
- Codex request preparation uses stable `session_id`
- Codex request preparation injects stable `prompt_cache_key` when available or derivable
- auth selection remains stable enough across repeated request sequences to avoid breaking cache continuity
- targeted regression tests cover continuity precedence and retention preservation

## Implementation notes

<details>
<summary>Continuity requirements observed during investigation</summary>

The reference behavior that matters for Codex cache reuse is not just retention flags. Cache reuse depends on a continuity envelope that includes:

- stable account/auth context
- stable `session_id`
- stable `prompt_cache_key`
- stable request body defaults for Codex-compatible requests

The proxy must preserve these signals across repeated requests instead of allowing them to vary arbitrarily per request.

</details>

<details>
<summary>Key failure modes found during debugging</summary>

The most important failure modes were:

1. `prompt_cache_retention` was stripped at executor level for Codex request paths.
2. Chat Completions -> Codex translation dropped `prompt_cache_retention`.
3. Continuity headers used a shape that did not match the working client behavior.
4. Repeated requests often had no stable `prompt_cache_key`.
5. `session_id` fell back to a fresh random UUID on each request.
6. Codex auth selection could rotate between requests, changing upstream identity and defeating cache reuse.

Any one of these can reduce cache effectiveness; together they made cache reuse unreliable or absent.

</details>

<details>
<summary>Why stable auth selection is part of the fix</summary>

Even if the prompt body is identical, upstream cache reuse can still fail when requests switch between different Codex auth/account contexts.

That means continuity must be preserved not only in the request payload/header shape, but also in the selected upstream auth identity for a repeated request sequence.

This is why sticky auth selection is justified here as part of cache continuity, not as a generic routing feature expansion.

</details>

## Risks

- Codex upstream cache behavior is not fully documented, so some continuity heuristics are based on validated runtime behavior rather than an official contract
- if upstream changes its cache bucketing rules, continuity precedence may need adjustment
- richer diagnostics may be desirable temporarily until the fix has proven stable in wider use

## Validation plan

- run targeted executor/translator/auth tests
- manually verify repeated Codex requests against the same model/account context
- confirm that repeated requests show stable continuity signals
- confirm cache reads or cached-token reuse on subsequent requests

VooDisss:
## Summary

This PR fixes Codex prompt cache continuity through CLIProxyAPI.

While this PR fixes #2373 directly, it also addresses the same underlying class of continuity problems discussed in #1962 and #1998, and is adjacent to the Codex session/account pinning issues in #2230 and #2256.

In the broken state, repeated Codex requests could reach the upstream without the same continuity envelope, so upstream prompt caching did not reliably reuse prior prompt state. The most visible repro was with OpenCode using Codex through CLIProxyAPI, where cache reads worked in the reference client but not through the proxy. The fix is not specific to OpenCode, but that integration made the failure mode easy to validate.

At a high level, this PR:

- preserves `prompt_cache_retention` on Codex request paths instead of stripping it
- preserves `prompt_cache_retention` when translating OpenAI Chat Completions requests to Codex
- aligns Codex continuity headers to `session_id`
- introduces explicit Codex continuity resolution so repeated requests reuse a stable upstream cache/session identity
- keeps auth selection sticky for repeated request sequences so continuity is not accidentally broken by auth rotation
- adds targeted regression tests covering the continuity policy and retention preservation

## Why this change is needed

Repeated Codex requests were missing the stable continuity signals needed for upstream cache reuse.

Observed failure modes included:

- executor-layer Codex request preparation removed `prompt_cache_retention`
- Chat Completions -> Codex translation dropped `prompt_cache_retention`
- Codex request preparation used header shapes that did not match the working client behavior
- OpenAI-style Codex requests could be sent without a stable `prompt_cache_key`
- when no stable continuity key existed, `session_id` fell back to a fresh random UUID per request
- auth selection could drift across repeated requests, which changes upstream identity and can also break cache reuse

The combined effect was that repeated requests looked like unrelated turns to the upstream.

## What this PR changes

- preserves caller-provided `prompt_cache_retention` on Codex HTTP, stream, and websocket request preparation paths
- preserves `prompt_cache_retention` in the OpenAI Chat Completions -> Codex translator
- applies Codex continuity using exact `session_id`
- centralizes continuity resolution in a dedicated helper instead of scattering the policy across executor flow code
- derives a stable continuity key with explicit precedence:
  - request `prompt_cache_key`
  - execution session metadata
  - explicit `Idempotency-Key`
  - request/auth affinity key
  - stable client-principal hash
  - stable auth-ID hash
- applies the chosen continuity key to both:
  - request header `session_id`
- adds sticky auth affinity so repeated request sequences reuse the same Codex auth when possible
- keeps websocket behavior aligned with the same continuity model

## Validation result

This change was validated in both tests and live runtime behavior.

Expected outcome after the fix:

- repeated Codex requests sent through CLIProxyAPI can reuse upstream cache continuity instead of behaving like fresh turns
- `session_id` is stable across repeated requests in the same continuity context
- prompt cache reads now occur through CLIProxyAPI in the previously failing OpenCode + Codex scenario

## Tests run

- `go test ./sdk/cliproxy/auth ./internal/runtime/executor ./internal/translator/codex/openai/responses ./internal/translator/codex/openai/chat-completions`

Result:
- all targeted packages passed

## Built artifact

- Windows executable was built successfully for manual verification during development

<details>
<summary>Detailed root cause analysis</summary>

The investigation showed that the problem was not just `prompt_cache_retention`.

The more important issue was that repeated Codex requests did not share a stable continuity identity when forwarded by the proxy. In the failing path:

- `prompt_cache_key` could be absent
- `session_id` could change on every request
- auth/account selection could rotate across repeated requests

This meant that even identical prompts could be treated by the upstream as unrelated requests.

The final working behavior required preserving a continuity envelope that includes:

- stable auth/account context
- stable `session_id`
- stable `prompt_cache_key`
- stable Codex-compatible request defaults

The fix encoded that continuity policy explicitly instead of relying on incidental request context.

</details>

<details>
<summary>Implementation notes for gatekeeper review</summary>

This PR intentionally keeps the scope focused on Codex continuity and avoids broader redesign.

Architectural choices:

- continuity policy was extracted into a dedicated helper so executor flow remains orchestration-focused
- auth affinity was added only because auth rotation was a real, observed cache-breaking failure mode
- translator changes are intentionally minimal and limited to preserving `prompt_cache_retention`
- websocket behavior was kept aligned with the same continuity semantics

The goal was to fix the real bug with the smallest behaviorally complete change set, not to redesign request routing or provider architecture.

</details>

<details>
<summary>Behavioral details of continuity resolution</summary>

For OpenAI-style requests forwarded to Codex, continuity resolution now prefers the strongest available stable signal in this order:

1. explicit `prompt_cache_key`
2. execution session metadata
3. explicit `Idempotency-Key`
4. auth/request affinity metadata
5. stable client-principal-derived hash
6. stable auth-ID-derived hash

The chosen continuity key is then applied to:

- `prompt_cache_key` in the request body
- `session_id` in the request headers

This ensures repeated requests reuse the same upstream continuity identity even when the original caller does not provide all continuity fields directly.

</details>

<details>
<summary>Runtime validation outcome</summary>

Manual runtime validation during debugging confirmed the pre-fix and post-fix behavior difference.

Before the fix, repeated Codex requests showed continuity drift such as:

- empty `prompt_cache_key`
- changing `session_id`
- occasional auth rotation across repeated requests

After the fix, repeated requests in the validated scenario reused stable continuity values and upstream cache reads started working through CLIProxyAPI.

This was specifically confirmed in the OpenCode -> CLIProxyAPI -> Codex flow, which previously missed prompt cache reuse.

</details>

## Risk assessment

This PR does change request continuity behavior for Codex, so it is not zero-risk, but the risk is bounded and justified.

Main considerations:

- continuity precedence is partly based on validated runtime behavior rather than a formal upstream contract
- sticky auth affinity changes repeated-request behavior intentionally, but only to preserve continuity
- diagnostics were retained because they were useful during validation and make future regressions easier to inspect

Despite that, the changes are targeted, test-covered, and constrained to the exact area that was failing.

## Out of scope

This PR does not attempt to:

- redesign provider routing generally
- change non-Codex providers
- redesign translator behavior beyond the necessary Codex retention pass-through
- change unrelated cache-control semantics

### DIFF
--- a/internal/runtime/executor/codex_continuity.go
+++ b/internal/runtime/executor/codex_continuity.go
@@ -14,8 +14,6 @@ import (
 	"github.com/tidwall/sjson"
 )
 
-const codexAuthAffinityMetadataKey = "auth_affinity_key"
-
 type codexContinuity struct {
 	Key    string
 	Source string
--- a/internal/runtime/executor/codex_executor_cache_test.go
+++ b/internal/runtime/executor/codex_executor_cache_test.go
@@ -178,18 +178,3 @@ func TestCodexExecutorCacheHelper_ClaudePreservesCacheContinuity(t *testing.T) {
 		t.Fatalf("session_id = %q, want %q", got, continuity.Key)
 	}
 }
-
-func TestResolveCodexContinuity_DoesNotForwardAuthAffinityKey(t *testing.T) {
-	req := cliproxyexecutor.Request{Payload: []byte(`{"model":"gpt-5.4"}`)}
-	opts := cliproxyexecutor.Options{Metadata: map[string]any{"auth_affinity_key": "principal:raw-client-secret"}}
-	auth := &cliproxyauth.Auth{ID: "codex-auth-1", Provider: "codex"}
-
-	continuity := resolveCodexContinuity(context.Background(), auth, req, opts)
-
-	if continuity.Source != "auth_id" {
-		t.Fatalf("continuity.Source = %q, want %q", continuity.Source, "auth_id")
-	}
-	if continuity.Key == "principal:raw-client-secret" {
-		t.Fatal("continuity.Key leaked raw auth affinity key")
-	}
-}


VooDisss:
@codex review

VooDisss:
@codex review

VooDisss:
@codex review

VooDisss:
@codex review

VooDisss:
@codex please re-review the current PR head only.

This PR has been narrowed substantially since the earlier review comments. The auth-affinity routing changes that triggered the later scheduler/pinning concerns have been removed from the branch.

The remaining diff is intentionally limited to the Codex continuity/cache fix:
- preserve `prompt_cache_retention` on Codex paths
- preserve `prompt_cache_retention` in OpenAI Chat Completions -> Codex translation
- stabilize `prompt_cache_key`
- stabilize `session_id`
- align HTTP and websocket continuity behavior
- keep focused regression coverage around continuity and retention behavior

Please evaluate the current head commit as a scoped Codex prompt-cache continuity fix, not the earlier affinity-routing iterations.

VooDisss:
@codex please re-review the current head commit only.

The PR has been narrowed further to comply with repository policy: all internal/translator changes were removed. The remaining diff is only the executor-level Codex continuity/cache fix.

VooDisss:
@luispater this makes OpenCode work with OpenAI and cache to hit. It works for me, when compiling the .exe and the originally released .exe does not...

VooDisss:
@luispater I made it work for myself, but I wanted it to work for all the other people. I can't keep battling these agents forever, cause they will ALWAYS find something to latch onto, as they are inherently non-deterministic

PR #2398, opened 2026-03-29
fix: preserve SSE event boundaries for Responses streams

7RPH:
## Summary
- preserve SSE event boundaries for `/v1/responses` streaming when chunks contain only `data:` lines
- prevent adjacent `response.output_item.done` and `response.completed` chunks from being concatenated into one client parse unit
- add a regression test covering data-only SSE chunk separation
## Test Plan
- [x] `go test ./sdk/api/handlers/openai -run TestForwardResponsesStreamSeparatesDataOnlySSEChunks -count=1`
- [x] `go test ./sdk/api/handlers/openai -run "TestForwardResponsesStream" -count=1`
- [x] `go test ./internal/translator/openai/openai/responses -count=1`

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -32,14 +32,12 @@ func writeResponsesSSEChunk(w io.Writer, chunk []byte) {
 	if bytes.HasSuffix(chunk, []byte("\n\n")) {
 		return
 	}
+	suffix := []byte("\n\n")
 	if bytes.HasSuffix(chunk, []byte("\n")) {
-		if _, err := w.Write([]byte("\n")); err != nil {
-			return
-		}
-	} else {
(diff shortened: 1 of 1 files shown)


7RPH:
Updated per review.
- simplified `writeResponsesSSEChunk` to the minimal error-aware form
- moved the non-error SSE separation test into `openai_responses_handlers_stream_test.go`
Verified with:
- `go test ./sdk/api/handlers/openai -run "TestForwardResponsesStream" -count=1`
- `go test ./internal/translator/openai/openai/responses -count=1`

7RPH:
Updated per follow-up review.
- added error checks for the trailing newline writes in `writeResponsesSSEChunk`
- made the SSE test assert exact event boundaries and payloads instead of loose substring matches
Verified with:
- `go test ./sdk/api/handlers/openai -run "TestForwardResponsesStream" -count=1`
- `go test ./internal/translator/openai/openai/responses -count=1`

luispater (approved):

PR #2400, opened 2026-03-29
Revert "fix(codex): restore prompt cache continuity for Codex requests"

luispater:
Reverts router-for-me/CLIProxyAPI#2374


VooDisss:
Great, now it does not cache requests again... Basically reintroduced the problem, and I wonder why only my patched version caches tokens... Yesterday 6.30 version cached tokens but it was a fluke and the next day it does not cache them anymore even of subsequential requests...

@luispater I understand its complex to keep track of many changes and perhaps I should just create my own minimal proxy that does the job of caching tokens in between requests and such. Not asking for much, I think.

PR #2412, opened 2026-03-30
feat: configurable signature cache toggle for Antigravity/Claude thinking blocks

sususu98:
## Summary

Adds a configurable toggle (`antigravity-signature-cache-enabled`) to control signature cache behavior for Claude thinking blocks in the Antigravity translator. Supports two modes:

- **Cache mode** (default, `true`): Cached signatures are preferred and validated — existing behavior, no change.
- **Bypass mode** (`false`): Client-provided signatures are used directly after format normalization. Useful for testing and deployments where signature caching is not desired.

### Config & Runtime
- New config field `antigravity-signature-cache-enabled` (default: `true`, hot-reloadable)
- `internal/cache/signature_cache.go`: Thread-safe `atomic.Bool` toggle with `SetSignatureCacheEnabled()` / `SignatureCacheEnabled()`
- `internal/api/server.go`: `applySignatureCacheConfig()` for initial load and hot-reload via `UpdateClients()`

### Request Translator (bypass mode)
- When cache is disabled and client sends Anthropic-native signature (`E...` prefix, 1-layer Base64), re-encodes to 2-layer Base64 (`R...`) for the Antigravity backend
- Cache mode format (`claude#R...`) continues to work as before

### Response Translator (bypass mode)
- `decodeSignature()`: Decodes `R...` (2-layer Base64) back to `E...` (Anthropic native) for client consumption
- Bypass mode outputs raw signature without cache group prefix
- **Bug fix**: Flushes co-located thinking text before emitting signature delta — previously, text arriving in the same chunk as a signature was silently dropped

### Tests
- `TestTextAndSignatureInSameChunk`: Verifies co-located text+signature chunks emit both correctly
- `TestSignatureOnlyChunk`: Verifies signature-only chunks (normal case) still cache correctly

## Signature Format Reference

| Format | Encoding | Context |
|--------|----------|---------|
| `E...` | 1-layer Base64 (Anthropic native) | Client ↔ Anthropic API |
| `R...` | 2-layer Base64 (Antigravity) | CPA ↔ Antigravity backend |
| `claude#R...` | Cache mode wrapper | CPA cache mode client output |

## Testing

- All existing tests pass
- Cross-provider signature compatibility verified (Vertex/Anthropic/Bedrock signatures from haiku accepted by opus)
- Hot-reload toggle verified at runtime

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -115,6 +115,16 @@ nonstream-keepalive-interval: 0
 #   keepalive-seconds: 15   # Default: 0 (disabled). <= 0 disables keep-alives.
 #   bootstrap-retries: 1    # Default: 0 (disabled). Retries before first byte is sent.
 
+# Signature cache validation for thinking blocks (Antigravity/Claude).
+# When true (default), cached signatures are preferred and validated.
+# When false, client signatures are used directly after normalization (bypass mode for testing).
+# antigravity-signature-cache-enabled: true
+
+# Bypass mode signature validation strictness (only applies when signature cache is disabled).
+# When true, validates full Claude protobuf tree (Field 2 -> Field 1 structure).
+# When false (default), only checks R/E prefix + base64 + first byte 0x12.
+# antigravity-signature-bypass-strict: false
+
 # Gemini API keys
 # gemini-api-key:
 #   - api-key: "AIzaSy...01"
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -24,6 +24,7 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/api/middleware"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/api/modules"
 	ampmodule "github.com/router-for-me/CLIProxyAPI/v6/internal/api/modules/amp"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/cache"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/config"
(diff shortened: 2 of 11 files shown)


sususu98:
Please don't merge this yet — I'm still running additional tests. I'll update the PR once testing is complete.

luispater (commented):
## Summary
This PR adds a hot-reloadable toggle for Antigravity/Claude thinking signature handling:
- `antigravity-signature-cache-enabled` (default `true`): current cache/validation behavior.
- `antigravity-signature-bypass-strict` (default `false`): optional strict protobuf-tree validation when cache is disabled.

It also fixes a streaming edge case where thinking text co-located with a signature in the same chunk could be dropped.

## Key findings
- Cache mode appears to preserve existing semantics: prefer cached signatures; fall back to client-provided only if valid.
- Bypass mode normalizes client signatures to the backend-expected 2-layer base64 (`R...`) and (for Claude clients) decodes back to Anthropic-native `E...` on responses when cache is disabled.
- Strict bypass precheck (`400 Bad Request`) happens before any upstream call (good), and is gated behind both cache disabled + strict enabled.

## Tests
Added coverage for:
- Streaming: text + signature in the same chunk emits both and caches signature for full thinking text.
- Executor: strict bypass rejects invalid signatures before any upstream request; non-strict/cache mode skip precheck.
- Request/response translators: bypass normalization and signature validation helpers.

## Suggestions (non-blocking)
- Consider reducing Info-level logging for bypass strict mode when cache mode is enabled (or make it conditional on bypass being active) to avoid noisy logs.

(Translator-path-guard is failing due to expected translator changes; not considered for this recommendation.)

PR #2422, opened 2026-03-30
fix(codex): add default instructions for /responses/compact

MonsterQiu:
Summary:
- add default empty instructions for Codex /responses/compact requests
- add a regression test to verify compact requests include instructions

Why:
- /responses already injects default instructions when missing
- /responses/compact did not, which could trigger upstream 400 "Instructions are required"

Verification:
- go test ./sdk/api/handlers/openai ./internal/runtime/executor -run 'Test(OpenAIResponsesCompact|CodexExecutorCompactAddsDefaultInstructions|OpenAICompatExecutorCompactPassthrough)' -count=1\n- verified locally that /v1/responses/compact no longer fails with "Instructions are required"\n\nRelated to #2408


MonsterQiu:
Updated the compact request normalization to coerce both missing and explicit `null` `instructions` values to an empty string.

Also expanded the regression test to cover both cases and re-ran:
- `go test ./internal/runtime/executor -run 'TestCodexExecutorCompactAddsDefaultInstructions' -count=1`\n- `go test ./sdk/api/handlers/openai ./internal/runtime/executor -run 'Test(OpenAIResponsesCompact|CodexExecutorCompactAddsDefaultInstructions|OpenAICompatExecutorCompactPassthrough)' -count=1`

PR #2441, opened 2026-03-31
fix(auth): resolve oauth aliases before suspension checks

### LINKED ISSUE #2421: Bug: Model alias not resolved before suspension check — causes false no_auth for aliased models
## Bug Report

**Summary:** When a model is suspended (e.g., due to upstream 404), the proxy checks suspension state **before** resolving `oauth-model-alias` mappings. This causes aliased models to return `no_auth` even though the alias target model is healthy and available.

## Reproduction

1. Configure an alias in `oauth-model-alias`:
   ```yaml
   oauth-model-alias:
     antigravity:
       - name: claude-opus-4-6-thinking
         alias: claude-opus-4-6
         fork: true
   ```

2. Have an upstream provider return 404 for the raw model name (`claude-opus-4-6`)

3. Proxy suspends the model:
   ```
   [model_registry.go:696] Suspended client antigravity-*.json for model claude-opus-4-6: not_found
   ```

4. All subsequent requests for `claude-opus-4-6` return `no_auth` immediately — the alias to `claude-opus-4-6-thinking` is never evaluated

## Impact

- **Claude Code** sends `opus[1m]` → resolves internally to `claude-opus-4-6` → hits proxy → suspended → `no_auth`
- **Direct curl** to `opus[1m]` → proxy resolves its own alias to `gemini-3.1-pro-high` → works fine
- Other models on the same provider continue to work (suspension is model-specific, not client-wide) — only the aliased path is broken

## Expected Behavior

Alias resolution should happen **before** suspension checks. If `claude-opus-4-6` is aliased to `claude-opus-4-6-thinking`, and `claude-opus-4-6-thinking` is available and not suspended, the request should succeed.

## Actual Behavior

```
{"error":{"message":"auth_unavailable: no auth available","type":"server_error","code":"internal_server_error"}}
```

## Timeline (real incident)

- Antigravity upstream returned 404 for `claude-opus-4-6` around 03:56 UTC
- Proxy suspended the model
- `claude-opus-4-6-thinking`, `gemini-3.1-pro-high`, `sonnet[1m]` — all continued to work
- Only `opus[1m]` (which Claude Code resolves to `claude-opus-4-6`) failed with `no_auth`
- Restarting the proxy temporarily fixes it (clears in-memory suspension), but the issue recurs if the upstream 404 persists

## Suggested Fix

In the conductor/request routing path, resolve `oauth-model-alias` mappings before querying the model registry for suspension state. This ensures aliased models route to their healthy targets even if the original model name is suspended.

## Environment

- CLIProxyAPI version: dev (latest main)
- Config: `disable-cooling: true` on auth file
- Providers: antigravity (OAuth), claude (API key), openai-compatibility (GLM, OpenRouter)

MonsterQiu:
Summary:
- resolve oauth model aliases before auth availability checks when selecting credentials
- use the resolved upstream model for execution-state checks when oauth alias routing changes the model
- add a regression test covering a blocked route model with a healthy alias target

Why:
- requests for aliased oauth models could return `auth_unavailable: no auth available` when the client-visible route model was suspended, even though the alias target model was healthy
- the routing code selected auths and checked model suspension against the unresolved route model instead of the alias-resolved upstream model

Verification:
- go test ./sdk/cliproxy/auth -run 'TestManagerExecute_OAuthAliasBypassesBlockedRouteModel' -count=1
- go test ./sdk/cliproxy/auth -count=1
- go test ./...

Closes #2421

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -450,6 +450,10 @@ func (m *Manager) selectionModelForAuth(auth *Auth, routeModel string) string {
 	return resolvedModel
 }
 
+func (m *Manager) selectionModelKeyForAuth(auth *Auth, routeModel string) string {
+	return canonicalModelKey(m.selectionModelForAuth(auth, routeModel))
+}
+
 func (m *Manager) stateModelForExecution(auth *Auth, routeModel, upstreamModel string, pooled bool) string {
 	stateModel := executionResultModel(routeModel, upstreamModel, pooled)
 	selectionModel := m.selectionModelForAuth(auth, routeModel)
@@ -507,8 +511,7 @@ func (m *Manager) availableAuthsForRouteModel(auths []*Auth, provider, routeMode
 	availableByPriority := make(map[int][]*Auth)
 	cooldownCount := 0
 	var earliest time.Time
-	for i := 0; i < len(auths); i++ {
-		candidate := auths[i]
+	for _, candidate := range auths {
 		checkModel := m.selectionModelForAuth(candidate, routeModel)
 		blocked, reason, next := isAuthBlockedForModel(candidate, checkModel, now)
 		if !blocked {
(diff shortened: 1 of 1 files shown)


MonsterQiu:
Follow-up pushed in `f611dd6e`.

This keeps the same fix but tightens the implementation in two places:
- deduplicates route-aware model support checks behind a helper
- preserves the original `model` argument for non-built-in selectors instead of always blanking it in the legacy path

Verification rerun after the refactor:
- go test ./sdk/cliproxy/auth -run 'TestManagerExecute_OAuthAliasBypassesBlockedRouteModel' -count=1
- go test ./sdk/cliproxy/auth -count=1
- go test ./...

PR #2472, opened 2026-04-01
fix(amp): stop suppressing thinking blocks in streaming mode

### LINKED ISSUE #2471: Amp TUI crashes with 'undefined is not an object' when using local provider for Claude models with thinking
## Description

When using a local provider (e.g., antigravity) to serve `claude-opus-4-6` with extended thinking enabled, the Amp TUI crashes with:

```
Error: undefined is not an object (evaluating 'r.type')
```

or

```
Error: undefined is not an object (evaluating 'e.type')
```

## Root Cause

`ResponseRewriter.rewriteStreamEvent()` unconditionally calls `suppressAmpThinking()` which strips all thinking-related SSE events (`content_block_start` for thinking, `thinking_delta`) from the stream. However, it does **not** suppress `signature_delta` events, resulting in:

1. **Orphaned `signature_delta` events** — the client receives a signature for a content block that was never started
2. **Index gap** — subsequent content blocks (e.g., text at index 1) are sent, but index 0 never existed from the client's perspective

The ampcode.com proxy path does not use the `ResponseRewriter`, so thinking blocks pass through intact and the client works correctly.

## Expected Behavior

Streaming thinking blocks should pass through the rewriter with signature injection (via `ensureAmpSignature`), matching the behavior of ampcode.com responses. Only non-streaming responses should suppress thinking blocks (when `tool_use` is present).

## Steps to Reproduce

1. Configure a local provider (antigravity) for `claude-opus-4-6`
2. Send a request from the Amp TUI that triggers extended thinking
3. Observe the crash in the client

## Related

- Commit `279cbbbb` on `origin/dev` already contains this fix (by CharTyr) but was not merged to `main`

jveko:
## Summary

- Remove unconditional `suppressAmpThinking` call from `rewriteStreamEvent` in streaming mode
- Thinking blocks now pass through with signature injection via `ensureAmpSignature`, matching ampcode.com proxy behavior
- Non-streaming responses still suppress thinking when `tool_use` is present via `rewriteModelInResponse`

## Problem

When using a local provider (e.g., antigravity) for `claude-opus-4-6` with extended thinking, the Amp TUI crashes with `undefined is not an object (evaluating 'r.type')`. The streaming thinking suppression strips `content_block_start` and `thinking_delta` events but leaves `signature_delta` events orphaned, breaking SSE index alignment.

## Test plan

- [ ] Verify local provider path with thinking-enabled models no longer crashes Amp TUI
- [ ] Verify ampcode.com proxy path still works unchanged
- [ ] Verify non-streaming responses still suppress thinking blocks when tool_use is present

Closes #2471

### DIFF
--- a/internal/api/modules/amp/response_rewriter_test.go
+++ b/internal/api/modules/amp/response_rewriter_test.go
@@ -100,23 +100,89 @@ func TestRewriteStreamChunk_MessageModel(t *testing.T) {
 	}
 }
 
-func TestRewriteStreamChunk_SuppressesThinkingContentBlockFrames(t *testing.T) {
+func TestRewriteStreamChunk_PassesThinkingBlocksWithSignatureInjection(t *testing.T) {
 	rw := &ResponseRewriter{suppressedContentBlock: make(map[int]struct{})}
 
 	chunk := []byte("event: content_block_start\ndata: {\"type\":\"content_block_start\",\"index\":0,\"content_block\":{\"type\":\"thinking\",\"thinking\":\"\"}}\n\nevent: content_block_delta\ndata: {\"type\":\"content_block_delta\",\"index\":0,\"delta\":{\"type\":\"thinking_delta\",\"thinking\":\"abc\"}}\n\nevent: content_block_stop\ndata: {\"type\":\"content_block_stop\",\"index\":0}\n\nevent: content_block_start\ndata: {\"type\":\"content_block_start\",\"index\":1,\"content_block\":{\"type\":\"tool_use\",\"name\":\"bash\",\"input\":{}}}\n\n")
 	result := rw.rewriteStreamChunk(chunk)
 
-	if contains(result, []byte("\"thinking\"")) || contains(result, []byte("\"thinking_delta\"")) {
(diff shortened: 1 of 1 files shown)


jveko:
Thanks for the review @xkonjin! Addressed both points:

1. **Signature injection**: Added `TestRewriteStreamChunk_PassesThinkingBlocksWithSignatureInjection` — confirms `ensureAmpSignature` injects `"signature":""` on thinking blocks after suppression removal.

2. **Interleaved thinking + content**: Added `TestRewriteStreamChunk_ThinkingInterleavedWithContent_PreservesIndices` — verifies index 0 (thinking) and index 1 (text) both pass through with all event types (`thinking_delta`, `signature_delta`, `text_delta`).

3. **Thinking as last chunk (edge case)**: Added `TestRewriteStreamChunk_ThinkingBlockAsLastChunk` — verifies stream terminates correctly with `message_delta` and `message_stop` when thinking is the final content block.

All tests passing: dd9bc3e0

jveko:
Closing in favor of #2474 which includes this same streaming thinking passthrough fix plus additional Gemini-related fixes (tool schema sanitization, trailing unanswered function call stripping).

PR #2474, opened 2026-04-01
fix(gemini): clean tool schemas, stream thinking passthrough, and strip trailing unanswered function calls

aikins01:
## Summary

- Delegate tool schema sanitization to `util.CleanJSONSchemaForGemini` and remove brittle `bytes.Replace` hack
- Strip `eager_input_streaming` from Claude tool declarations before sending to Gemini
- Let thinking blocks pass through in SSE streaming (aligned with #2472)
- Strip trailing model turns with unanswered function calls to prevent Gemini empty responses
- Update tests to expect thinking blocks to pass through during streaming

## Problem

1. **Tool schema validation errors**: The `gemini` translator sent Claude tool schemas directly to Gemini without cleaning unsupported keys (`$schema`, `format`, `patternProperties`, `eager_input_streaming`), causing 400 Bad Request errors. Mirrors fix from PR #2108 applied to `gemini-cli` translator.

2. **Broken SSE stream**: The `ResponseRewriter` was suppressing thinking blocks during streaming, which corrupted the SSE event sequence (orphaned `event:` lines, missing `content_block_start` for indices Amp expected). Thinking blocks now pass through intact in streaming, matching ampcode.com behavior (ref #2471).

3. **Empty Gemini responses**: When conversation history ended with a model turn containing an unanswered `functionCall`, Gemini returned empty responses with `finishReason: STOP`. The translator now strips these trailing turns.

### DIFF
--- a/internal/translator/gemini/claude/gemini_claude_request.go
+++ b/internal/translator/gemini/claude/gemini_claude_request.go
@@ -6,6 +6,7 @@
 package claude
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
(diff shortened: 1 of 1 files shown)


aikins01:
i was actually just working on addressing the blockers and fixing a regression we caught in the streaming logic but i see this got merged

PR #2491, opened 2026-04-02
test(claude): add compatibility sentinels and centralize builtin fallback handling

mpfo0106:
## Summary
- add Claude Code-informed compatibility sentinel tests for representative SDK/control shapes
- extract Claude builtin tool fallback names into a small internal helper used by `applyClaudeToolPrefix`
- preserve the current fallback behavior while making builtin-prefix handling easier to extend and review
- add focused executor regression tests for fallback-seeded builtins across `tool_choice`, `tool_use`, `tool_reference`, and nested `tool_result` paths

## Motivation
This is a safe first step toward keeping CLIProxyAPI grounded against the now-public Claude Code source without introducing broader runtime risk.

The scope is intentionally narrow:
- keep current default behavior intact for existing users
- improve coverage around builtin tool prefixing edge cases, especially history-only builtin references
- add lightweight source-informed sentinel coverage for a few Claude Code message/control shapes so future drift is easier to notice

This is compatibility scaffolding, not a broader Claude Code control-plane/session alignment change.

## Why this is low-risk
- the runtime change is intentionally small and confined to Claude builtin prefix handling
- the previous hardcoded builtin fallback is still preserved; it is now centralized rather than replaced by a broader new runtime contract
- the new helper is augmentative, not replacement-only: typed builtins still extend the set found in the current request body
- no stable public route behavior is intentionally changed beyond the narrow builtin-prefix safety fix
- the patch is covered by focused regression tests and a full `go test ./...` pass

## Why this is useful now that Claude Code source is public
- the public source gives a concrete reference for a few SDK/control concepts that were previously inferred indirectly
- this patch captures a small subset of those concepts as compatibility sentinels, so future drift is easier to notice during maintenance
- it also makes builtin handling easier to reason about and extend without broadening the runtime surface area in the same PR
- in other words, this change improves maintainability and observability first, while deliberately avoiding a larger control-plane/session redesign

## Notes
- the builtin fallback remains small and authoritative in this slice: `web_search`, `code_execution`, `text_editor`, `computer`
- this PR should be read as a safe compatibility hardening patch, not as full Claude Code alignment
- larger follow-up work (if needed) should happen separately for runtime/session/control-plane behavior

## Tests
- `go test ./test/...`
- `go test ./sdk/translator/...`
- `go test ./internal/runtime/executor -run 'Claude|Builtin|Tool'`
- `go test ./...`


luispater:
Hi @mpfo0106 — could you please move the following files into `internal/runtime/executor/helps/` (the `helps` package) to keep helper utilities centralized?

- `internal/runtime/executor/claude_builtin_tools.go` → `internal/runtime/executor/helps/claude_builtin_tools.go`
- `internal/runtime/executor/claude_builtin_tools_test.go` → `internal/runtime/executor/helps/claude_builtin_tools_test.go`

`git mv` is preferred to preserve history.

Thanks!

PR #2495, opened 2026-04-02
feat: PG store config override + cross-replica auth sync

xandr0s:
## Summary

Two features for Kubernetes multi-replica deployments with PostgreSQL store:

### 1. Config override via PGSTORE_CONFIG_OVERRIDE_PATH

When this env var is set, the file at that path is read on bootstrap and written into PG store, overriding any previously stored config. This lets k8s ConfigMaps take precedence over PG-stored config on every pod restart.

**Problem:** PG store seeds config from file on first run, then always reads from DB — ignoring ConfigMap updates on subsequent deploys.

### 2. Auth sync polling via StartAuthSync()

Periodically polls auth_store table and incrementally updates local auth files. When one replica adds/refreshes OAuth credentials, other replicas pick them up within 30s.

**Implementation details:**
- Incremental sync: compares content before writing, only writes changed files
- Mutex-protected: prevents races with concurrent file access
- Cleans up local files removed from PG
- Works with existing fsnotify file watcher for seamless client reload

### Changes

- internal/store/postgresstore.go: Add StartAuthSync() + incrementalAuthSync(), config override in syncConfigFromDatabase()
- cmd/server/main.go: Call StartAuthSync() after PG store bootstrap

Both features are opt-in and fully backward compatible.

## Test plan

- [ ] Deploy with PGSTORE_CONFIG_OVERRIDE_PATH — verify ConfigMap overrides PG config on restart
- [ ] Deploy 2 replicas — add OAuth account on replica 1, verify replica 2 picks it up within 30s
- [ ] Deploy without env var — verify existing behavior unchanged
- [ ] Verify no excessive fsnotify events when auth files unchanged

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/joho/godotenv"
 	configaccess "github.com/router-for-me/CLIProxyAPI/v6/internal/access/config_access"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/auth/kiro"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/buildinfo"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/cmd"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/config"
@@ -50,6 +51,19 @@ func init() {
 	buildinfo.BuildDate = BuildDate
 }
 
+// setKiroIncognitoMode sets the incognito browser mode for Kiro authentication.
+// Kiro defaults to incognito mode for multi-account support.
+// Users can explicitly override with --incognito or --no-incognito flags.
+func setKiroIncognitoMode(cfg *config.Config, useIncognito, noIncognito bool) {
+	if useIncognito {
+		cfg.IncognitoBrowser = true
+	} else if noIncognito {
+		cfg.IncognitoBrowser = false
+	} else {
+		cfg.IncognitoBrowser = true // Kiro default
+	}
+}
+
 // main is the entry point of the application.
 // It parses command-line flags, loads configuration, and starts the appropriate
 // service based on the provided flags (login, codex-login, or server mode).
@@ -62,18 +76,35 @@ func main() {
 	var codexDeviceLogin bool
 	var claudeLogin bool
 	var qwenLogin bool
+	var kiloLogin bool
 	var iflowLogin bool
 	var iflowCookie bool
+	var gitlabLogin bool
+	var gitlabTokenLogin bool
 	var noBrowser bool
 	var oauthCallbackPort int
 	var antigravityLogin bool
 	var kimiLogin bool
+	var cursorLogin bool
+	var kiroLogin bool
+	var kiroGoogleLogin bool
+	var kiroAWSLogin bool
+	var kiroAWSAuthCode bool
+	var kiroImport bool
+	var kiroIDCLogin bool
+	var kiroIDCStartURL string
+	var kiroIDCRegion string
+	var kiroIDCFlow string
+	var githubCopilotLogin bool
+	var codeBuddyLogin bool
 	var projectID string
 	var vertexImport string
 	var configPath string
 	var password string
 	var tuiMode bool
 	var standalone bool
+	var noIncognito bool
+	var useIncognito bool
 	var localModel bool
 
 	// Define command-line flags for different operation modes.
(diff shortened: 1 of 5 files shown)


xandr0s:
Thanks for the thorough review! Regarding the P2 concern about poll clobbering local edits:

In our deployment, auth credentials are always added via management UI or CLI login — both write to PG store first, then mirror to disk. The poll only syncs PG → local, never the reverse direction. So the race between poll and local file edits doesn't apply to our use case.

That said, adding an `updated_at` comparison (skip overwrite if DB record is older than local file mtime) would make this more robust for mixed workflows. Happy to add that if you think it's worth it for the general case.

xkonjin (commented):
## Code Review: PG Store Config Override + Cross-Replica Auth Sync

### Summary
Well-architected feature for multi-replica deployments. The auth sync mechanism enables horizontal scaling without sticky sessions, and the config override provides necessary flexibility for k8s environments.

### Strengths
- **Smart incremental sync** — only writes changed files, avoiding fsnotify storms
- **Atomic file operations** — proper temp file + rename pattern in incrementalAuthSync
- **Config override is well-designed** — PGSTORE_CONFIG_OVERRIDE_PATH provides clean k8s integration
- **Orphan cleanup** — removes local auth files not present in DB, preventing stale credential accumulation

### Issues

**1. Potential race condition in auth sync startup**
In main.go, StartAuthSync is called immediately after pgStoreInst creation, but there's a small window where the bootstrap may not have completed. Consider starting sync after Bootstrap() returns.

**2. Error handling in orphan cleanup is silent**
If removal fails (permissions, filesystem issues), it's silently ignored. Consider logging at WARN level.

**3. No backoff for sync failures**
The sync loop logs errors but retries immediately on the next interval. If PG is temporarily unavailable, this could spam logs. Consider adding a simple backoff or at least a counter.

**4. Missing test coverage**
- No tests for incrementalAuthSync logic
- No tests for config override fallback path (when override file read fails)
- No tests for orphan cleanup behavior

### Security
- Proper file permissions (0o600 for files, 0o700 for dirs)
- SQL queries use parameterized queries implicitly via QueryContext
- Consider adding authentication/authorization on the auth_store table access if multiple untrusted pods share the PG instance

### Performance
The 30-second poll interval is reasonable for auth propagation, but consider making it configurable for latency-sensitive environments

### Verdict
Good foundational work for multi-replica support. The suggestions are improvements, not blockers. Recommend adding tests before merge if time permits.

luispater (approved):
Summary:
Adds two Postgres-store features aimed at multi-replica/Kubernetes deployments: (1) `PGSTORE_CONFIG_OVERRIDE_PATH` to force an on-disk config to win on bootstrap, and (2) auth sync polling to mirror `auth_store` changes into local auth files.

Key findings:
- Config override is applied early during bootstrap and keeps the DB + local spool consistent.
- Auth sync is incremental (content-compare before write) and serialized to reduce race risk and unnecessary fsnotify churn.
- The poll loop uses `time.After(interval)` to avoid ticker pileups if a sync iteration runs long.

Non-blocking suggestions:
- In the cleanup pass, consider normalizing `rel` with `filepath.ToSlash(rel)` before checking against `seen` to avoid OS-separator mismatches.
- Consider starting `StartAuthSync()` with a cancellable lifecycle context instead of `context.Background()` for graceful shutdown (optional).
- Consider guarding/documenting `interval > 0` to avoid a tight loop if misconfigured (optional).

Test plan:
- go test ./...

xandr0s:
Pushed follow-up commit addressing the remaining review feedback:

**Applied (this push)**
- Interval guard in `StartAuthSync` (non-positive values log a WARN and return) — @luispater nit #3
- `filepath.ToSlash(rel)` in orphan cleanup key (Windows path separator consistency) — @luispater nit #1
- WARN log on orphan `os.Remove` failures — @xkonjin review #2
- Unit tests via `DATA-DOG/go-sqlmock`: `incrementalAuthSync` (write new / skip unchanged / orphan removal / query error), `StartAuthSync` (interval=0 disabled, ctx cancel stops goroutine), `absoluteAuthPath` traversal rejection — @xkonjin review #4. 8/8 green locally.

**Already landed in dc88a48**
- Duplicate `StartAuthSync` call removed (gemini-code-assist / codex P2)
- `ApplyCustomHeadersFromMetadata` restored in `List` (codex P1)
- Orphan cleanup now recursive (`filepath.WalkDir`) with correct full-path key (gemini P2)
- `time.NewTicker` → `time.After` in select, no pileup (gemini P2)

**Intentionally deferred (with rationale)**
- **Bounded/cancellable context in `main.go`** (@luispater nit #2, codex P1): every other background goroutine in the project uses `context.Background()` (`StartAutoUpdater`, `StartModelsUpdater`, `PersistConfig`, `sdkAuth.*`). Introducing lifecycle-ctx only in `StartAuthSync` creates inconsistency; a cross-cutting refactor to threaded shutdown ctx is a better fit for a standalone PR.
- **Exponential backoff on sync failures** (@xkonjin review #3): at the 30s poll interval, a transient PG outage produces one log line every 30s, which is not a log-spam concern. Backoff state would complicate the loop with no clear payoff.
- **`updated_at` vs local mtime reconciliation** (codex P2 clobbering): in our deployment and the intended use case (management UI / CLI login), auth writes always go PG-first, disk-mirror-second, so the clobbering window does not exist. Happy to revisit as a follow-up if mixed-workflow users need it — the schema addition + compare logic is straightforward.

@luispater @xkonjin — could you take another look when you get a chance?

PR #2510, opened 2026-04-03
feat(usage): persist statistics across restarts

Janenesome:
## Summary

Automatically restore and save usage statistics snapshots so history survives service restarts.

## Changes

- Add `PersistenceManager` in `internal/usage/persistence.go` to handle snapshot save/restore/autosave lifecycle
- Add two new config fields: `usage-statistics-persistence-enabled` (default: `true`) and `usage-statistics-persistence-file` (optional path override)
- Integrate persistence into `sdk/cliproxy/service.go`: restore on startup, autosave every 5 minutes, save on shutdown and config reload
- Use atomic write with `.bak` fallback to prevent data corruption
- Dedup-safe merge via existing `MergeSnapshot` logic

## Testing

- `internal/usage/persistence_test.go`: 562 lines covering save/restore/autosave/backup/atomic write
- `sdk/cliproxy/service_usage_persistence_test.go`: 417 lines covering full service lifecycle (startup restore, shutdown persist, config reload)
- `internal/usage/logger_plugin_test.go`: added provider fallback identifier test

## Notes

Persistence is only active when both `usage-statistics-enabled: true` and `usage-statistics-persistence-enabled: true` are set. No behavior change for users who don't opt in.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -66,6 +66,13 @@ error-logs-max-files: 10
 # When false, disable in-memory usage statistics aggregation
 usage-statistics-enabled: false
 
+# When true, automatically restore and periodically persist usage statistics snapshots.
+usage-statistics-persistence-enabled: true
+
+# Optional path override for the usage statistics snapshot file.
+# Empty uses WRITABLE_PATH/data/usage-statistics.json or <config dir>/data/usage-statistics.json.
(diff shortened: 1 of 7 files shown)


luispater:
No plans to support this.

PR #2522, opened 2026-04-03
fix(amp): strip signature from tool_use blocks before forwarding to Claude

aikins01:
ensureAmpSignature injects `signature:""` into tool_use blocks so the Amp TUI does not crash on `P.signature.length`. when Amp sends the conversation back in subsequent turns, Claude rejects the extra field with 400:

```
tool_use.signature: Extra inputs are not permitted
```

this strips the proxy-injected `signature` from tool_use blocks in `SanitizeAmpRequestBody` before forwarding to the upstream API.

also fixes a latent bug where mixing `sjson.DeleteBytes` on the raw JSON with array reconstruction via `block.Value()` would undo the deletion when both an invalid thinking block and a tool_use block existed in the same message.

### DIFF
--- a/internal/api/modules/amp/response_rewriter.go
+++ b/internal/api/modules/amp/response_rewriter.go
@@ -2,6 +2,7 @@ package amp
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


aikins01:
good catch on the caller stripping - it was initially added defensively since caller is also not part of claude api schema, but you are right that it is client-provided metadata from amp and is not currently causing any errors. removed it to keep the change scoped to the actual bug (signature). also fixed gofmt.

luispater (approved):
Summary: The previous blocker is resolved. `SanitizeAmpRequestBody` now strips the proxy-injected `signature` field only from `tool_use` blocks, and it preserves the existing thinking-block cleanup path without mutating unrelated fields.

Test plan:
- `go test ./internal/api/modules/amp`

PR #2533, opened 2026-04-04
feat(claude): configurable prompt cache TTL and automatic image compression

gitrvc-hub:
## What this does

Two improvements to the Claude executor that address real pain points when running the proxy in front of clients that don't handle caching or image limits themselves.

### Configurable prompt cache TTL

The proxy already auto-injects `cache_control` breakpoints when the client sends none — which is great. The TTL has always been hardcoded to the default 5-minute lifetime though.

This adds a `prompt-cache-ttl` option that accepts `""` (keep the current behaviour, 5 min) or `"1h"` for the extended cache lifetime. It can be set at two levels:

```yaml
# Global default for all Claude requests (OAuth + api-key)
prompt-cache-ttl: "1h"

# Per-credential override under claude-api-key
claude-api-key:
  - api-key: "sk-..."
    prompt-cache-ttl: "1h"
```

The `prompt-caching-scope-2026-01-05` beta that gates the 1h TTL is already in the default `Anthropic-Beta` header, so no header changes needed.

### Automatic image compression

Claude enforces a hard 5 MB per-image cap and rejects anything wider or taller than 8000px. Clients that forward screenshots or high-res images hit opaque 400 errors with no recovery path.

The executor now checks every base64 image block in the messages payload before sending upstream. Images within limits pass through unchanged. Oversized ones are re-encoded as JPEG at decreasing quality levels (85 → 70 → 50 → 30) until they fit. No new dependencies — uses only Go stdlib image packages (`image/jpeg`, `image/png`, `image/gif`).

## Changes

- `internal/config/config.go` — `PromptCacheTTL` field on both root `Config` and `ClaudeKey`
- `internal/watcher/synthesizer/config.go` — propagate per-key TTL to auth attributes
- `internal/runtime/executor/claude_executor.go` — `resolvePromptCacheTTL`, `buildCacheControl`, updated inject functions, `compressImagesInPayload`
- `config.example.yaml` — documented both options
- Tests for TTL resolution, `buildCacheControl`, inject behaviour, image compression helpers

## Testing

```
go test ./internal/runtime/executor/... ./internal/watcher/synthesizer/... ./internal/config/...
```

All existing tests pass. New tests cover:
- `buildCacheControl` with and without `ttl`
- `resolvePromptCacheTTL` precedence (per-key over global, nil fallback)
- `ensureCacheControl` injects correct TTL
- `scaleToFit` edge cases
- `maybeCompressImage` (small image unchanged, oversized compressed, invalid base64 safe)
- `compressImagesInPayload` end-to-end


luispater (changes requested):
Thanks for the PR — the cache TTL option and image compression look useful, and the targeted tests/build pass locally.

I found one blocking correctness issue in the per-credential TTL override path:

- `resolvePromptCacheTTL` now treats `auth.Attributes["prompt_cache_ttl"]` key presence as an explicit override, including the empty string case.
- But `synthesizeClaudeKeys` only materializes `prompt_cache_ttl` when `ck.PromptCacheTTL` is non-empty.
- Because `ClaudeKey.PromptCacheTTL` is a plain `string` populated via `yaml.Unmarshal`, `prompt-cache-ttl: ""` and an omitted field collapse to the same value before synthesis.

In practice, a `claude-api-key` entry cannot reset a global `prompt-cache-ttl: "1h"` back to the default 5-minute TTL, even though the new comments/tests claim that this should work.

Test plan:
- `go test ./internal/runtime/executor/... ./internal/watcher/synthesizer/... ./internal/config/...`

gitrvc-hub:
Fixed.

You were right that the original config path could not distinguish between:
- omitted `prompt-cache-ttl`
- explicit `prompt-cache-ttl: ""`

So a per-key entry could not actually reset a global `prompt-cache-ttl: "1h"` back to the default 5-minute TTL.

What changed:
- `ClaudeKey.PromptCacheTTL` is now a `*string` so the config keeps tri-state semantics
  - `nil` = inherit global
  - `""` = explicit reset to default 5m
  - `"1h"` = explicit override to 1h
- `synthesizeClaudeKeys()` now materializes `prompt_cache_ttl` whenever the field is explicitly set, including the empty-string case
- added regression coverage for:
  - explicit empty per-key override
  - omitted per-key value falling back to global

Pushed in: `2c09f7c3`

PR #2542, opened 2026-04-04
fix(management): support explicit proxy overrides in api-call

AstroQore:
## Summary
- accept `proxy_url` / `proxyUrl` / `ProxyURL` on management `/api-call` requests
- prioritize an explicit request proxy before auth-level and global proxy settings while preserving `direct` / `none` semantics
- add transport tests that cover explicit direct and explicit proxy override precedence

## Tests
- `go test ./internal/api/handlers/management -run 'TestAPICallTransport|TestAuthByIndexDistinguishesSharedAPIKeysAcrossProviders'`\n\n## Notes\n- this is the backend half of the provider proxy fix so the management UI can forward per-request provider proxies without needing an auth index\n- no existing request shape changes are required for callers that do not send a proxy override\n- pairs with router-for-me/Cli-Proxy-API-Management-Center#190 and the frontend fix in router-for-me/Cli-Proxy-API-Management-Center#194

### DIFF
--- a/internal/api/handlers/management/api_tools.go
+++ b/internal/api/handlers/management/api_tools.go
@@ -43,6 +43,9 @@ type apiCallRequest struct {
 	AuthIndexSnake  *string           `json:"auth_index"`
 	AuthIndexCamel  *string           `json:"authIndex"`
 	AuthIndexPascal *string           `json:"AuthIndex"`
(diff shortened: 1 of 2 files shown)


LTbinglingfeng:
https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/194#issuecomment-4187897375

PR #2574, opened 2026-04-06
feat(temporal): anti-drift injection for all provider requests

dwillitzer:
## Summary
- Adds temporal anti-drift system that injects `<temporal>` tags into every outbound request to all providers
- Format-aware injection: Claude format (top-level `system` array) vs OpenAI format (`messages` array) — no 400 errors
- Single injection point in conductor covers all providers (Antigravity, Claude, GLM, Qwen, OpenRouter) without per-executor changes

## Why This Matters
In long agent sessions, AI models hallucinate dates — wrong days, wrong weeks, wrong months. This cascades into incorrect tool calls, wrong file paths, and stale context decisions. By injecting fresh temporal context on every request, models stay grounded in the current date/time.

Pattern validated in `pi-rs` where `TemporalDriftDetector` found real drift in sessions >20 turns.

## Changes
- **New:** `internal/temporal/temporal.go` — temporal tag builder + format-aware payload injection (~120 lines)
- **Modified:** `sdk/cliproxy/auth/conductor.go` — injects before `Execute()` and `ExecuteStream()`
- Zero PII — only UTC timestamps, no user data

## Test plan
- [x] `go build -o cliproxyapi ./cmd/server/` — compilation succeeds
- [x] Claude API requests no longer return 400 (format-aware system injection)
- [x] OpenAI-format requests receive temporal system message in messages array
- [x] All providers route correctly with injection enabled
- [ ] Smoke test: `curl -s -X POST localhost:8317/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{"model":"claude-sonnet-4-6","messages":[{"role":"user","content":"what day is it"}],"max_tokens":100}'`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -14,9 +14,10 @@ import (
 	"syscall"
 
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
-	log "github.com/sirupsen/logrus"
-	"golang.org/x/crypto/bcrypt"
-	"gopkg.in/yaml.v3"
+	temporal "github.com/router-for-me/CLIProxyAPI/v6/internal/temporal"
+ 	log "github.com/sirupsen/logrus"
+ 	"golang.org/x/crypto/bcrypt"
+ 	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -141,6 +142,10 @@ type Config struct {
 
 	// Payload defines default and override rules for provider payload parameters.
 	Payload PayloadConfig `yaml:"payload" json:"payload"`
+
+	// Temporal controls temporal anti-drift injection into LLM request payloads.
+	// nil means use defaults (enabled, every request).
+	Temporal *temporal.Config `yaml:"temporal" json:"temporal"`
 
 	legacyMigrationPending bool `yaml:"-" json:"-"`
 }
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -28,6 +28,7 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/misc"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/runtime/executor/helps"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/temporal"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/thinking"
 	antigravityclaude "github.com/router-for-me/CLIProxyAPI/v6/internal/translator/antigravity/claude"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
@@ -505,6 +506,12 @@ func (e *AntigravityExecutor) Execute(ctx context.Context, auth *cliproxyauth.Au
 		return resp, errValidate
 	}
 	req.Payload = originalPayload
+	// Re-apply temporal injection after signature validation (which forces
+	// req.Payload back to the raw original). Without this, the conductor's
+	// anti-drift hook is silently stripped for every Antigravity request.
+	if temporal.ShouldInject(temporal.ConfigOrDefault(e.cfg.Temporal)) {
+		req.Payload = temporal.InjectIntoPayload(req.Payload, baseModel)
+	}
 	token, updatedAuth, errToken := e.ensureAccessToken(ctx, auth)
 	if errToken != nil {
 		return resp, errToken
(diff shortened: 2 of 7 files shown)


luispater (changes requested):
Summary:
I found two blocking issues in this change set.

Key findings:
- Blocking: the temporal injector is enabled unconditionally via `temporal.DefaultConfig()` in the execution path, so every proxied request now gets a hidden system prompt with no user opt-in and no configuration escape hatch. That changes request semantics for all providers and breaks API compatibility rather than preserving it.
- Blocking: `internal/registry/openrouter_enrichment.go` adds a new `http.Client{Timeout: 30 * time.Second}` / `context.WithTimeout` fetch path. This repo explicitly restricts new timeout-driven network behavior to a small set of documented exceptions, and this enrichment path is not one of them.

Test plan:
- Reviewed the temporal injection flow and the new registry enrichment loop plus management endpoints.

luispater (changes requested):
Summary:
I like the goal, but this PR currently mixes the temporal injection feature with unrelated registry/management work and introduces a couple of blockers.

Blocking findings:
- `sdk/cliproxy/auth/conductor.go` hard-codes `temporal.DefaultConfig()` at the injection call sites instead of wiring the feature to runtime config, which makes this always-on request mutation impossible to disable or roll back if it causes provider regressions.
- `internal/registry/openrouter_enrichment.go` adds a new long-lived background HTTP poller with its own timeout/refresh policy. That is a separate runtime/network behavior change from the temporal feature and should be reviewed independently instead of landing as part of this PR.

Test plan:
- Reviewed the temporal injection path, the new management endpoints, and the OpenRouter enrichment worker.

dwillitzer:
@luispater — force-pushed a full rewrite addressing both blocking findings. This PR is now temporal-only; the OpenRouter enrichment you flagged separately is split out to #2880.

## Blocker #1 — unconditional injection / no opt-in
- `DefaultConfig()` now returns `{Enabled: false}`. With the `temporal:` YAML block absent, behavior is identical to a non-temporal build. No surprise request mutation.
- Opt-in is explicit:
  ```yaml
  temporal:
    enabled: true
    inject_interval: 0   # 0 = every request; N = every Nth request
  ```
- `Config.Temporal` is `*temporal.Config` (pointer) so YAML-omitted and explicit-zero are distinguishable at runtime.

## Blocker #2 — OpenRouter enrichment
- Removed from this PR entirely. See #2880 for the standalone enrichment PR with a note about the `http.Client{Timeout: …}` pattern vs. `internal/misc/antigravity_version.go` precedent.

## Scope
- Single commit, +389 / -5 (was 9 commits, +667 / -29).
- Only touches: `internal/temporal/`, `internal/config/config.go` (adds `Temporal *temporal.Config` field), `sdk/cliproxy/auth/conductor.go` (hook + live config resolution + reload log), `internal/runtime/executor/claude_executor.go` + test (preserve temporal across Claude Code-shaped system rewrite), `internal/runtime/executor/antigravity_executor.go` (re-inject after signature validation).

## Also fixed under this scope
Two injection-stripping paths that were silent in the prior revision:
1. **Claude OAuth cloaking** — `checkSystemInstructionsWithSigningMode` rebuilds the system array and the OAuth sanitizer drops forwarded user-system content. The rebuilder now preserves any injected `<temporal_context>` block and excludes it from the sanitizer.
2. **Antigravity signature validation** — `req.Payload = originalPayload` after validation reverted the conductor's mutation. The executor now re-applies injection at all three generation sites (`Execute`, streaming variant, retry stream). `CountTokens` intentionally left alone.

## Verified
End-to-end against a running instance:
- GLM (openai-compat): `prompt_tokens` 33 → 120; `<temporal_context>` visible in request-log.
- Claude via Antigravity: `input_tokens` 34 → 145; model reproduced the element verbatim.
- Unit tests cover: injection, dedup guard, image-model skip, opt-in default, explicit enable, disabled, Claude OAuth preservation regression.

Ready for re-review when you have time.

dwillitzer:
Resolved merge conflict with upstream `main` in `internal/runtime/executor/claude_executor_test.go`.

**Conflict:** upstream renamed a test comment ("survive conversion" → "survive forwarding"). Our temporal preservation regression test was adjacent.

**Resolution:** Kept the `TestCheckSystemInstructionsWithSigningMode_PreservesTemporalContext` test, adopted upstream's comment wording. Build passes, all executor + temporal tests pass.

Still awaiting @luispater re-review of the clean rewrite (both original blockers addressed in `375e840`).

dwillitzer:
Addressed both findings from the Codex review on `827545ff`:

**P1 — Validate temporal marker before bypassing OAuth sanitization** (`389f2e30`)
- Added `temporal.IsValidTemporalBlock()` which regex-validates that preserved content matches the `BuildTemporalTag` output shape: `<temporal_context><temporal day="..." date="YYYY-MM-DD" .../></temporal_context>`.
- The Claude executor now calls this before preserving a temporal block in the rebuilt system array. Arbitrary content wrapped in `<temporal_context>` tags is no longer forwarded unsanitized — it falls through to the normal sanitization path.

**P2 — Match escaped temporal tags in dedup guard** (`389f2e30`)
- The `InjectIntoPayload` dedup check now also matches `<` (`<`-escaped `<`) in addition to raw `<temporal `. Prevents double injection when an upstream serializer encodes angle brackets.

**Test fix**: The `PreservesTemporalContext` test was constructing malformed JSON (unescaped inner quotes in the temporal tag text). gjson silently truncated it. Fixed to properly escape inner quotes via `strings.ReplaceAll`.

Build passes, all temporal and executor tests green.

PR #2575, opened 2026-04-06
fix(codex): improve OAuth websocket parity for fast/priority responses

Willhong:
## Summary

Improve Codex OAuth websocket parity so `service_tier=priority` behaves closer to official Codex `/fast` behavior.

This patch focuses on Codex OAuth websocket routing and request-shape parity, and is related to the same Codex websocket transport area touched by #2230 / #2256, but addresses a different problem:

- #2230 / #2256 focused on websocket session pinning and auth switching after quota failures.
- This patch focuses on **Codex OAuth websocket eligibility + request/header/session parity for fast/priority behavior**.

### 1) Default file-backed Codex OAuth auths to websocket-capable

Files:
- `internal/watcher/synthesizer/file.go`
- `sdk/auth/filestore.go`

Behavior:
- Preserve explicit `websockets` metadata when present.
- Otherwise synthesize `Attributes["websockets"] = "true"` for provider `codex`.

Why:
- File-backed Codex OAuth auths were otherwise routed through HTTP fallback instead of the Codex websocket path.

### 2) Normalize Codex websocket request bodies in both execute paths

File:

Behavior:
- For both `Execute` and `ExecuteStream`:
  - ensure `model`
  - ensure `stream=true` in streaming path
  - drop unsupported carryover fields
  - ensure non-empty `instructions`

Why:
- Codex websocket upstream is stricter than generic OpenAI Responses handling and rejects some under-normalized payloads.

### 3) Add session-aligned websocket metadata/header parity

Files:
- `internal/runtime/executor/codex_executor.go`

Behavior:
- Populate `prompt_cache_key` from the execution/session id when absent.
- Populate `client_metadata.x-codex-turn-metadata`.
- Mirror the same execution/session identity into websocket headers:
  - `Session_id`
  - `x-client-request-id`
  - `x-codex-turn-metadata`
- Stop forcing `Conversation_id` on Codex websocket requests.
- Use an originator closer to the official app-server/debug path.

Why:
- Official Codex websocket requests carry request/session metadata that was missing or mismatched in the proxy path.
- This metadata appears to matter for Codex OAuth websocket behavior, especially when using `service_tier=priority`.

## Investigation summary

During investigation, the following concrete issues were confirmed:

1. Official Codex `/fast` sends `service_tier: "priority"` over websocket.
2. Official default behavior omits the tier field rather than sending `"default"`.
3. File-backed Codex OAuth auths in CLIProxyAPI were not websocket-enabled by default.
4. Proxy websocket requests were missing Codex-like session metadata and header parity.
5. Sending literal `service_tier: "default"` can be rejected upstream on the Codex websocket path.

### Official Codex baseline
Using traced official Codex websocket requests:
- default/no tier -> request omits `service_tier`
- fast -> request sends `service_tier: "priority"`

### Proxy behavior after this patch
The proxy websocket path:
- routes Codex OAuth auths over websocket instead of HTTP fallback
- sends `service_tier: "priority"` upstream
- includes prompt cache + turn metadata + request/session headers
- avoids invalid default-tier semantics by using omitted tier as baseline

### Latency benchmark
Short websocket prompt:
`Reply with exactly: OK and nothing else.`

Patched proxy, comparing **omitted tier** vs **priority**:

**omit(null)**
- total median: **1.94s**
- first-token median: **1.28s**

**priority**
- total median: **1.81s**
- first-token median: **1.12s**

This is not a dramatic improvement, but it is measurable and consistently in the expected direction after the websocket parity fixes.

## Notes / scope

- Upstream still returns `response.completed.service_tier = default`, so that field alone does not appear reliable for verifying Codex fast behavior on this route.
- This PR does **not** claim perfect 1:1 parity with official Codex internals.
- It does provide validated transport/session/request-shape fixes that move Codex OAuth websocket behavior materially closer to official `/fast` behavior.

## Related

- Related to #2230
- Related transport area: #2256

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	codexUserAgent  = "codex-tui/0.118.0 (Mac OS 26.3.1; arm64) iTerm.app/3.6.9 (codex-tui; 0.118.0)"
-	codexOriginator = "codex-tui"
+	codexOriginator = "codex-toy-app-server"
 )
 
 var dataTag = []byte("data:")
--- a/internal/runtime/executor/codex_websockets_executor.go
+++ b/internal/runtime/executor/codex_websockets_executor.go
@@ -190,9 +190,10 @@ func (e *CodexWebsocketsExecutor) Execute(ctx context.Context, auth *cliproxyaut
 	body, _ = sjson.DeleteBytes(body, "previous_response_id")
 	body, _ = sjson.DeleteBytes(body, "prompt_cache_retention")
 	body, _ = sjson.DeleteBytes(body, "safety_identifier")
-	if !gjson.GetBytes(body, "instructions").Exists() {
-		body, _ = sjson.SetBytes(body, "instructions", "")
-	}
+	body = normalizeCodexWebsocketInstructions(body)
+	executionSessionID := executionSessionIDFromOptions(opts)
+	var turnMeta string
+	body, turnMeta = enrichCodexWebsocketSessionMetadata(body, executionSessionID)
 
 	httpURL := strings.TrimSuffix(baseURL, "/") + "/responses"
 	wsURL, err := buildCodexResponsesWebsocketURL(httpURL)
@@ -201,6 +202,13 @@ func (e *CodexWebsocketsExecutor) Execute(ctx context.Context, auth *cliproxyaut
 	}
 
 	body, wsHeaders := applyCodexPromptCacheHeaders(from, req, body)
+	if strings.TrimSpace(executionSessionID) != "" {
+		wsHeaders.Set("Session_id", executionSessionID)
+		wsHeaders.Set("x-client-request-id", executionSessionID)
+	}
+	if strings.TrimSpace(turnMeta) != "" {
+		wsHeaders.Set("x-codex-turn-metadata", turnMeta)
+	}
 	wsHeaders = applyCodexWebsocketHeaders(ctx, wsHeaders, auth, apiKey, e.cfg)
 
 	var authID, authLabel, authType, authValue string
@@ -210,7 +218,6 @@ func (e *CodexWebsocketsExecutor) Execute(ctx context.Context, auth *cliproxyaut
 		authType, authValue = auth.AccountInfo()
 	}
 
-	executionSessionID := executionSessionIDFromOptions(opts)
 	var sess *codexWebsocketSession
 	if executionSessionID != "" {
 		sess = e.getOrCreateSession(executionSessionID)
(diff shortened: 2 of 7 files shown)


Willhong:
Addressed the review feedback.

This follow-up update includes:
- fixing the `executionSessionID` vs body/header metadata mismatch
  - the effective session id is now used consistently for body/header metadata
  - the original execution session id is still used only for session-store lookups, so we do not pollute the store with one-off generated sessions
- refactoring duplicated websocket body preparation in `Execute` / `ExecuteStream` into `prepareCodexWebsocketRequestBody(...)`
- refactoring the default Codex OAuth websocket attribute logic into the shared helper `sdk/auth.CodexWebsocketsAttributeValue(...)`
- preserving `previous_response_id` so incremental websocket turns do not regress

I also pushed the updated branch, and the live instance I am testing against is already running this patched build.

luispater (changes requested):
Summary:
I found one blocking issue in the websocket request normalization path.

Key findings:
- Blocking: `normalizeCodexWebsocketInstructions` now injects `You are a helpful assistant.` whenever `instructions` is missing or blank. That silently changes the effective prompt for requests that previously had no system instructions, so the websocket path stops being behaviorally neutral relative to the existing empty-string handling in the Codex executors. For a compatibility-layer fix, that prompt mutation is too invasive to ship.

Test plan:
- Reviewed the websocket request preparation changes and the Codex auth routing updates.

luispater (commented):
Summary:
The websocket parity direction makes sense, but this is a protocol-sensitive path and I do not have enough confidence to approve it without regression coverage.

Key findings:
- Non-blocking: please add targeted tests around websocket request-body normalization and header/session shaping so future parity work does not silently drift again.
- Non-blocking: defaulting file-backed Codex auths to websocket-capable changes routing behavior; a focused regression test around that fallback would make this much safer.

Test plan:
- Reviewed the Codex websocket/body/header changes and the auth synthesis updates.

Willhong:
Addressed the requested change and pushed an update.

What changed:
- restored behaviorally neutral websocket instruction normalization by delegating to the existing Codex normalization logic (`missing/null -> ""`)
- removed the default assistant prompt injection from the websocket path
- added regression tests for websocket request-body normalization and session metadata shaping
- added focused regression coverage for the file-backed Codex auth websocket fallback / explicit override

Validation:
- `/usr/local/go/bin/go test ./internal/runtime/executor -run 'Test(BuildCodexWebsocketRequestBodyPreservesPreviousResponseID|PrepareCodexWebsocketRequestBody.*|ApplyCodexWebsocketHeaders.*|ApplyCodexHeaders.*)'`\n- `/usr/local/go/bin/go test ./internal/watcher/synthesizer -run 'Test(FileSynthesizer_Synthesize_CodexAuth.*)'`\n\nCould you take another look when you have a moment?

Willhong:
@luispater Addressed the requested change and pushed an update.

What changed:
- restored behaviorally neutral websocket instruction normalization by delegating to the existing Codex normalization logic (`missing/null -> ""`)
- removed the default assistant prompt injection from the websocket path
- added regression tests for websocket request-body normalization and session metadata shaping
- added focused regression coverage for the file-backed Codex auth websocket fallback / explicit override

Validation:
- `/usr/local/go/bin/go test ./internal/runtime/executor -run 'Test(BuildCodexWebsocketRequestBodyPreservesPreviousResponseID|PrepareCodexWebsocketRequestBody.*|ApplyCodexWebsocketHeaders.*|ApplyCodexHeaders.*)'`
- `/usr/local/go/bin/go test ./internal/watcher/synthesizer -run 'Test(FileSynthesizer_Synthesize_CodexAuth.*)'`

Could you take another look when you have a moment?

Willhong:
Closing this as effectively superseded/stale.

Most of the websocket protocol/request/header/session parity work has since landed via newer changes on `dev` (notably #3208), and this branch now conflicts heavily with that area. The remaining unique file-backed Codex auth websocket defaulting behavior is not worth carrying as part of this stale PR for my current use case.

Thanks for the review and feedback.

PR #2601, opened 2026-04-08
fix(thinking): validate suffix before treating as thinking config

l1i1:
## Summary

Fixes model ID being incorrectly stripped when model names contain parentheses that are not valid thinking suffixes (e.g., `stepfun-ai/step-3.5-flash(free)` → `stepfun-ai/step-3.5-flash`).

## Problem

The `ParseSuffix` function in `internal/thinking/suffix.go` was treating ANY text in parentheses as a thinking suffix. This caused models like `stepfun-ai/step-3.5-flash(free)` to have their `(free)` part stripped, resulting in the wrong model being sent to the upstream API.

## Solution

Modified `ParseSuffix` to only recognize valid thinking suffixes:
- Numeric budgets: `16384`, `0`, `08192`
- Special values: `none`, `auto`, `-1`  
- Level names: `minimal`, `low`, `medium`, `high`, `xhigh`, `max`

Model names with other parentheses (e.g., `(free)`, `(pro)`) are now preserved as-is.

## Test Cases

| Input | Before | After |
|-------|--------|-------|
| `stepfun-ai/step-3.5-flash(free)` | `stepfun-ai/step-3.5-flash` | `stepfun-ai/step-3.5-flash(free)` |
| `claude-sonnet-4-5(16384)` | `claude-sonnet-4-5` | `claude-sonnet-4-5` |
| `gpt-5.2(high)` | `gpt-5.2` | `gpt-5.2` |
| `gemini-2.5-pro` | `gemini-2.5-pro` | `gemini-2.5-pro` |

### DIFF
--- a/internal/thinking/suffix.go
+++ b/internal/thinking/suffix.go
@@ -40,29 +40,12 @@ func ParseSuffix(model string) SuffixResult {
 	modelName := model[:lastOpen]
 	rawSuffix := model[lastOpen+1 : len(model)-1]
 
-	// Validate that the suffix is a valid thinking suffix
-	// Numeric suffix
-	if _, err := strconv.Atoi(rawSuffix); err == nil {
-		return SuffixResult{
-			ModelName: modelName,
-			HasSuffix: true,
-			RawSuffix: rawSuffix,
-		}
-	}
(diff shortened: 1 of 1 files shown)


luispater:
After discussing it internally, we've determined this PR is not needed.

Closed.

PR #2605, opened 2026-04-08
fix(management): support external base url for webui oauth callbacks

cikichen:
## Summary
- add `remote-management.external-base-url` so webui oauth callbacks can use a configured public base URL
- route Codex, Claude, Gemini CLI, and Antigravity webui oauth flows through provider-specific external callback paths when configured
- add config and management handler tests to cover external callback resolution and preserve localhost fallback behavior

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -22,6 +22,9 @@ remote-management:
   # Leave empty to disable the Management API entirely (404 for all /v0/management routes).
   secret-key: ""
 
+  # External base URL for remote management clients when the server is accessed behind a proxy or public domain.
+  external-base-url: ""
+
   # Disable the bundled management control panel asset download and HTTP route when true.
   disable-control-panel: false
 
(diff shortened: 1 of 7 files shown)


luispater (approved):
### Summary
- Adds `remote-management.external-base-url` to support externally reachable OAuth callback URLs for WebUI flows.
- Routes Codex/Claude/Gemini CLI/Antigravity WebUI OAuth through provider-specific external callback paths when configured, and skips localhost callback forwarders in that case.
- Adds tests to cover external callback URL resolution and preserve localhost fallback behavior.

### Key findings
- Blocking: None.
- Non-blocking:
  - Consider validating `external-base-url` scheme to `http`/`https` only to prevent accidental misconfiguration.
  - Tests currently rely on `time.Sleep(600ms)`; polling with a timeout could reduce flakiness on slow runners.

### Test plan
- `go test ./internal/api/handlers/management ./internal/config ./internal/auth/claude`

PR #2620, opened 2026-04-08
fix(responses): keep tool-call/result chains contiguous when converting to chat/completions

davidwushi1145:
## Summary

Buffer interleaved non-tool messages while Responses-originated tool calls are still pending, then flush them only after the matching tool results are emitted downstream to `/v1/chat/completions`.

## Problem

The current `/v1/responses` compatibility path can generate an invalid upstream chat-completions transcript when a non-tool message is inserted between:

- an `assistant` message with `tool_calls`
- the matching `tool` message(s)

This was reproduced from the captured request artifacts:

Example of the malformed sequence:

```text
assistant(tool_calls=[call_GgABCmWvPYuP5ds9a4Fp0XST])
user("Bash reported `command not found`...")
tool(tool_call_id=call_GgABCmWvPYuP5ds9a4Fp0XST)
```

Strict OpenAI-compatible chat-completions executors can reject that transcript and return `500 Internal Server Error`.

## Root Cause

The Responses compatibility flow preserved item order too literally when converting Responses `input` items into upstream chat-completions `messages`.

That allowed injected diagnostic messages to land between pending tool calls and their matching tool outputs, violating the chat-completions tool-call contract.

## Fix

Added a small sanitizer in the OpenAI handler layer for chat-completions payloads produced from Responses-style input.

It now:

- buffers interleaved non-tool messages while tool results are pending
- flushes buffered messages only after all pending tool call IDs are satisfied
- merges consecutive assistant tool-call messages into a single assistant turn
- leaves already-valid transcripts unchanged

## Testing

Verified locally with:

```bash
gofmt -w sdk/api/handlers/openai/openai_handlers.go sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer.go sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer_test.go
go test ./sdk/api/handlers/openai -count=1
go build -o test-output ./cmd/server && rm test-output
```

Regression coverage added for:

- interleaved developer/user messages between `assistant.tool_calls` and matching `tool`
- consecutive assistant tool-call messages that should become one assistant turn
- already-valid transcripts remaining unchanged

---

# Incident Report: Responses-to-Chat-Completions Tool Call Ordering Bug

## Summary

A `/v1/responses` conversation could be converted into an invalid upstream `/v1/chat/completions` transcript when a non-tool message appeared between an assistant tool call and the matching tool result.

This was reproduced with the following captured artifacts:

- `request-742c1251.log`
- `request-f79a035c.log`
- `request-body-gid___axonhub_Request_104.json`
- `request-headers-gid___axonhub_Request_104.json`

The failing upstream request used the correct model (`skd/gpt5.4` on the local side, rewritten to `GPT-5.4` upstream), so this incident is **not** the earlier model-alias mismatch issue (`skd/gpt-5.4`).

## Impact

OpenAI-compatible upstream executors that strictly validate tool-call sequencing can reject the malformed transcript and return `500 Internal Server Error` instead of generating a model response.

## Symptoms

The replayed request in `request-f79a035c.log` failed with:

- local endpoint: `/v1/responses`
- upstream endpoint: `/v1/chat/completions`
- status: `500`
- body: `{"error":{"message":"Internal Server Error","type":"api_error"}}`

## Concrete Reproduction Evidence

The exported upstream payload in `request-body-gid___axonhub_Request_104.json` contains this invalid order near the end of the `messages` array:

1. `messages[19]`: `assistant` with `tool_calls = ["call_GgABCmWvPYuP5ds9a4Fp0XST"]`
2. `messages[20]`: `user` message beginning with `Bash reported \`command not found\``
3. `messages[21]`: `tool` message with `tool_call_id = "call_GgABCmWvPYuP5ds9a4Fp0XST"`

That sequence violates the chat-completions tool-call contract: once an assistant emits `tool_calls`, the transcript must provide the matching `tool` messages before any `user`, `system`, or additional `assistant` turns.

## Root Cause

The `/v1/responses` compatibility flow converted Responses input items to chat-completions messages in strict linear order.

That behavior was insufficient for tool-call chains because Responses payloads can contain intermediate non-tool messages (for example injected diagnostics such as:

- `Bash reported \`command not found\`, \`permission denied\`, or a missing file/path...`

) between a `function_call` item and the eventual `function_call_output` item.

The converter forwarded that intermediate message directly into the upstream `messages` array, which broke the required adjacency between:

- `assistant.tool_calls`
- matching `tool` messages

## Reproduction Procedure

1. Capture or replay a Responses-originated request that is routed into the OpenAI-compatibility `/v1/chat/completions` executor path.
2. Confirm the replay fails, as shown in `request-f79a035c.log`.
3. Inspect the exported upstream request body (`request-body-gid___axonhub_Request_104.json`).
4. Validate the message ordering with a simple pending-tool-call checker:

```python
import json
from pathlib import Path

body = json.loads(Path("request-body-gid___axonhub_Request_104.json").read_text())
pending = set()
for i, message in enumerate(body["messages"]):
    role = message.get("role")
    if pending and role != "tool":
        print("invalid order at", i, "role=", role, "pending=", sorted(pending))
        break
    if role == "assistant":
        for tool_call in message.get("tool_calls", []):
            if tool_call.get("id"):
                pending.add(tool_call["id"])
    if role == "tool":
        pending.discard(message.get("tool_call_id"))
```

Expected result before the fix: the checker reports the interleaved `user` message that appears before the pending tool result is supplied.

## Why This Was Hard to Spot

Large payload size, long AGENTS/system prompts, and many tool definitions were initially suspicious. Direct replay experiments showed those factors were not sufficient on their own to cause the failure. The decisive issue was the malformed message ordering inside the converted upstream chat-completions transcript.

## Fix

The handler now sanitizes chat-completions payloads produced from Responses-style input before forwarding them upstream.

The sanitizer does two things:

1. **Buffers interleaved non-tool messages** while tool calls are still pending.
2. **Flushes buffered messages only after all pending tool results arrive**, restoring a valid tool-call chain.
3. **Merges consecutive assistant tool-call messages** into a single assistant turn so parallel tool calls remain valid for strict chat-completions backends.

## Files Changed

- `sdk/api/handlers/openai/openai_handlers.go`
- `sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer.go`
- `sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer_test.go`

## Regression Coverage

Added tests for:

- interleaved developer/user messages between tool call and tool result
- consecutive assistant tool-call messages that should become a single assistant turn
- no-op behavior for already-valid transcripts

## Verification

Recommended verification commands:

```bash
gofmt -w sdk/api/handlers/openai/openai_handlers.go sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer.go sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer_test.go
go test ./sdk/api/handlers/openai -count=1
go build -o test-output ./cmd/server && rm test-output
```

## Remaining Risk

This fix preserves interleaved non-tool messages by reordering them after the pending tool-result chain. That is the safest chat-completions-compatible behavior, but it still changes transcript order relative to the original Responses payload. If a future backend supports richer mixed tool/message sequencing natively, the sanitizer may become unnecessary for that backend.

### DIFF
--- a/sdk/api/handlers/openai/openai_handlers.go
+++ b/sdk/api/handlers/openai/openai_handlers.go
@@ -117,6 +117,7 @@ func (h *OpenAIAPIHandler) ChatCompletions(c *gin.Context) {
 	if shouldTreatAsResponsesFormat(rawJSON) {
 		modelName := gjson.GetBytes(rawJSON, "model").String()
 		rawJSON = responsesconverter.ConvertOpenAIResponsesRequestToOpenAIChatCompletions(modelName, rawJSON, stream)
+		rawJSON = sanitizeConvertedResponsesChatCompletions(rawJSON)
 		stream = gjson.GetBytes(rawJSON, "stream").Bool()
 	}
 
--- a/sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer.go
+++ b/sdk/api/handlers/openai/openai_responses_chatcompletions_sanitizer.go
@@ -1,0 +1,258 @@
+package openai
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+// sanitizeConvertedResponsesChatCompletions normalizes chat-completions payloads
+// produced from /v1/responses-style inputs before they are forwarded upstream.
+//
+// Some Responses conversations can contain non-tool messages between an
+// assistant tool call and the matching tool result. OpenAI-compatible
+// chat-completions backends expect every pending tool call to be satisfied by
+// tool messages before any other role appears. This sanitizer keeps tool-call
+// chains contiguous by buffering interleaved non-tool messages until all
+// pending tool results arrive. It also merges consecutive assistant tool-call
+// messages into a single assistant turn so parallel calls stay valid for strict
+// chat-completions executors.
+func sanitizeConvertedResponsesChatCompletions(rawJSON []byte) []byte {
+	decoder := json.NewDecoder(bytes.NewReader(rawJSON))
+	decoder.UseNumber()
+
+	var payload map[string]any
+	if err := decoder.Decode(&payload); err != nil {
+		return rawJSON
+	}
+
+	rawMessages, ok := payload["messages"].([]any)
+	if !ok || len(rawMessages) == 0 {
+		return rawJSON
+	}
+
+	messages, changed := normalizeChatCompletionsToolMessageOrder(rawMessages)
+	if !changed {
+		return rawJSON
+	}
+
+	payload["messages"] = messages
+	out, err := json.Marshal(payload)
+	if err != nil {
+		return rawJSON
+	}
+	return out
+}
+
+func normalizeChatCompletionsToolMessageOrder(messages []any) ([]any, bool) {
+	remaining := append([]any(nil), messages...)
+	normalized := make([]any, 0, len(messages))
+	buffered := make([]any, 0)
+	pendingToolCalls := make(map[string]struct{})
+	lastAssistantToolIndex := -1
+	changed := false
+
+	prependBuffered := func() {
+		if len(buffered) == 0 {
+			return
+		}
+		remaining = append(append(make([]any, 0, len(buffered)+len(remaining)), buffered...), remaining...)
+		buffered = buffered[:0]
+		lastAssistantToolIndex = -1
+	}
+
+	for len(remaining) > 0 {
+		if len(pendingToolCalls) == 0 && len(buffered) > 0 {
+			prependBuffered()
+		}
+
+		rawMessage := remaining[0]
+		remaining = remaining[1:]
+
+		message, ok := rawMessage.(map[string]any)
+		if !ok {
+			if len(pendingToolCalls) > 0 {
+				buffered = append(buffered, rawMessage)
+				changed = true
+			} else {
+				normalized = append(normalized, rawMessage)
+				lastAssistantToolIndex = -1
+			}
+			continue
+		}
+
+		role, _ := message["role"].(string)
+		toolCalls := messageToolCalls(message)
+
+		if role == "assistant" && len(toolCalls) > 0 {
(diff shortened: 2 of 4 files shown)


luispater (approved):
Summary:
- Adds a small sanitizer on the Responses→ChatCompletions compatibility path to keep `assistant.tool_calls` ↔ `tool` result chains contiguous for strict upstream executors.
- Buffers interleaved non-tool messages until all pending tool results arrive, and merges consecutive assistant tool-call turns into a single assistant message.

Key findings:
- Scope is appropriately narrow: only applies when `/v1/responses`-style input is converted inside `ChatCompletions`.
- `json.Decoder.UseNumber()` avoids large-integer precision loss during sanitize re-marshal (covered by a regression test).
- Tests cover the core interleaving case, assistant tool-call merging, a captured fixture shape, and no-op behavior for already-valid transcripts.

Non-blocking:
- Behavior note: this intentionally reorders buffered non-tool messages relative to the original Responses item order (necessary for chat-completions tool-call contract).
- Performance: sanitize does a full JSON decode and (when changed) re-marshal; if this path becomes hot, a cheap pre-scan for violations could skip work on already-valid payloads.
- If tool-call IDs can ever be missing/empty in converted payloads, consider documenting that assumption and/or adding a focused test.

Test plan:
- Not run locally.
- CI: `build` ✅ (per PR checks)
- Suggested locally: `go test ./sdk/api/handlers/openai -count=1` and `go build -o test-output ./cmd/server && rm test-output`

xi10-a11y:
pls merge

davidwushi1145:
https://github.com/router-for-me/CLIProxyAPI/issues/2798

xisungod:
When to merge

luispater:
You added the code in the wrong place.

You should fix the issue in the translator.

Closed.

PR #2623, opened 2026-04-08
fix(codex): rebuild response.output in ExecuteStream

### LINKED ISSUE #2585: OpenAI Responses function_call stream emits inconsistent ids/names, causing AMP gpt-5.4 deep mode to end with empty output
## Summary
When AMP uses CLIProxyAPI for `openai/gpt-5.4` in deep mode through `/api/provider/openai/v1/responses`, the response stream contains inconsistent function-call events.

`response.output_item.added` contains a valid tool/function call with a non-empty `name`, but a later `response.function_call_arguments.delta` event appears with a different id and missing `name`. AMP then logs `Skipping tool_use normalization due to missing name` and the final result becomes empty (`outputCount=0`).

This looks very similar to the recent empty-response regressions around GPT-5.4 / codex / responses, but this report includes event-level evidence showing that the streamed function-call metadata becomes inconsistent mid-stream.

## Environment
- CLIProxyAPI: latest as of 2026-04-07
- Client: AMP CLI `0.0.1775538866-g2b4395`
- Model: `openai/gpt-5.4`
- Mode: `deep`
- Endpoint path seen by bridge: `/api/provider/openai/v1/responses`
- Route: `cliproxyapi`

## Repro
Run AMP in deep mode against a bridge that routes `gpt-5.4` to CLIProxyAPI.

Example AMP command:
```bash
~/.amp/bin/amp --log-level debug --no-ide --no-jetbrains -m deep --stream-json -x '请只回复 AFTER_RESTART_TOOLITEMS_OK，不要输出任何别的内容。'
```

## Expected
The function/tool call events should remain internally consistent across the stream:
- same logical call id
- name preserved on later function_call_arguments events
- AMP should be able to normalize the tool_use and produce a non-empty final output

## Actual
AMP finishes successfully but returns an empty final result.

AMP logs show:
```text
function_call item added: hasName=true, name="skill", callId="call_Xtz2ezUjYYFfAcXH6KGIolza", outputIndex=1
function_call arguments done: existingName="skill", outputIndex=1
Skipping tool_use normalization due to missing name
response.completed outputCount=0
```

Bridge logs for the same request show:
```text
[bridge-openai-sse] ...
toolItems=response.output_item.added@1:function_call:call_Xtz2ezUjYYFfAcXH6KGIolza:skill:args0|response.function_call_arguments.delta@1:function_call:fc_014e7bc7bae84d300169d4a0b6543c8191ac1ac9226e14312d:-:args...
missingToolName=true
```

## Important observation
For the same logical tool/function call in one streamed response:
- `response.output_item.added` uses `call_Xtz2ezUjYYFfAcXH6KGIolza` and `name="skill"`
- later `response.function_call_arguments.delta` appears as `fc_014e7bc7bae84d300169d4a0b6543c8191ac1ac9226e14312d` with missing name

That mismatch strongly suggests the streamed Responses translation is not preserving function-call identity and/or name consistently across events.

## Why this likely matters
Even if some clients are tolerant, AMP's Responses normalization treats the later missing-name event as malformed and drops the tool_use, which leads to an empty completed output.

## Related issues
This appears related to recent GPT-5.4 / empty-response reports such as:
- #2582
- #2583

But this report adds event-level evidence that the function-call metadata itself becomes inconsistent within the same stream.

## Request
Could you check the OpenAI Responses streaming translation for function-call events, specifically:
- mapping consistency between `response.output_item.added` and later `response.function_call_arguments.delta/done`
- preservation of `name`
- whether `call_*` and `fc_*` ids are being mixed for the same logical call

If helpful, I can provide the full AMP debug log and bridge SSE summary log for this request.

lenstr:
This fixes Codex streaming responses when the upstream backend sends an empty `response.output` in the final `response.completed` event.

After the upstream protocol change, CLIProxyAPI was updated to handle this in the non-streaming path in #2583. The streaming path was left unchanged, so `ExecuteStream` still forwarded a final `response.completed` event with empty `response.output`.

Clients that rely on `response.completed.response.output` then see no final output. In practice this broke Amp deep with `gpt-5.4`, where the stream completed but the final result was empty.

This change collects earlier `response.output_item.done` items during SSE processing and uses them to rebuild `response.completed.response.output` when needed. Usage reporting stays unchanged.

Related:
- related to #2583
- fixes #2585

Tested with a streamed `/v1/responses` request against the current backend behavior. Before the change, the final `response.completed` event had `output: []`. After the change, it contains the reconstructed assistant message and downstream clients return a non-empty result.

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -168,8 +168,7 @@ func (e *CodexExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, re
 	helps.AppendAPIResponseChunk(ctx, e.cfg, data)
 
 	lines := bytes.Split(data, []byte("\n"))
-	outputItemsByIndex := make(map[int64][]byte)
-	var outputItemsFallback [][]byte
+	outputCollector := newCodexResponseOutputCollector()
 	for _, line := range lines {
 		if !bytes.HasPrefix(line, dataTag) {
 			continue
@@ -179,16 +178,7 @@ func (e *CodexExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, re
 		eventType := gjson.GetBytes(eventData, "type").String()
 
 		if eventType == "response.output_item.done" {
-			itemResult := gjson.GetBytes(eventData, "item")
-			if !itemResult.Exists() || itemResult.Type != gjson.JSON {
-				continue
-			}
-			outputIndexResult := gjson.GetBytes(eventData, "output_index")
-			if outputIndexResult.Exists() {
-				outputItemsByIndex[outputIndexResult.Int()] = []byte(itemResult.Raw)
-			} else {
-				outputItemsFallback = append(outputItemsFallback, []byte(itemResult.Raw))
-			}
+			outputCollector.AddOutputItemDone(eventData)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


luispater (approved):
Summary
- Fixes Codex streaming cases where the final `response.completed` frame can contain an empty `response.output` by reconstructing it from prior `response.output_item.done` events (streaming + non-streaming paths).

Key findings
- Blocking: none.
- Non-blocking: minor hardening suggestions noted above.

Implementation notes
- Collects output items by `output_index` (sorted) plus fallback items without an index.
- Only patches when `response.output` is missing/empty; otherwise leaves the completed frame untouched.
- On the streaming path, rebuilds the SSE `data:` line when a patch was applied so downstream clients see the reconstructed `response.output`.

Test plan
- Ran: `go test ./internal/runtime/executor -run TestCodex -v`
- Not run: `go test ./...`

Amoodaa:
@luispater is it ready?

Amoodaa:
bump

lenstr:
Looks like this has effectively landed upstream already.

dev now contains bb8408ce and b6781d69, which seem to cover the same fix as this PR.

I also re-tested the current release locally on v6.9.35. The original failure mode is gone: streamed /v1/responses now returns populated response.completed.response.output, and amp -m deep --stream-json produces a non-empty final result again.

So from my side this looks superseded now. If there is no remaining gap here, I think it is reasonable to close this PR.

PR #2624, opened 2026-04-09
fix(executor): recover empty codex stream completions

ckckck:
## Summary
- cache Codex stream output items and function call state while consuming SSE events
- recover `response.output` when `response.completed` arrives with an empty output array
- add regression coverage for message output recovery and function call recovery in stream mode

## Background
Some Codex/OpenAI Responses streams emit `response.output_item.done` or function call argument events, but end with `response.completed.response.output=[]`.

The non-stream path already had recovery logic for empty completed output, while `ExecuteStream` did not. That mismatch can leave downstream clients with an empty final response even though the stream already carried recoverable output state.

## Testing
- `go test ./internal/runtime/executor -run 'TestCodexExecutorExecute(Stream)?_Empty' -count=1`

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -36,6 +36,220 @@ const (
 
 var dataTag = []byte("data:")
 
+type codexStreamFunctionCallState struct {
+	ItemID      string
+	CallID      string
+	Name        string
+	Arguments   string
+	OutputIndex int64
+}
+
+type codexStreamCompletionState struct {
+	outputItemsByIndex  map[int64][]byte
+	outputItemsFallback [][]byte
+	functionCallsByItem map[string]*codexStreamFunctionCallState
+}
+
+func newCodexStreamCompletionState() *codexStreamCompletionState {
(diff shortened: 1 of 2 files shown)


luispater (approved):
Summary:
This improves Codex streaming robustness by caching `response.output_item.done` items and function-call argument state, then reconstructing `response.output` when the terminal `response.completed` event arrives with `response.output: []`.

Key findings:
- Recovery is gated to the `response.completed` event and only triggers when `response.output` is empty, minimizing impact on normal streams.
- Adds regression tests for both message output recovery and function-call recovery in stream mode.

Non-blocking suggestions:
- Consider assigning fallback indexes monotonically after the max observed `output_index` (instead of `len(indexes)+len(recovered)`) to avoid edge-case reordering if upstream emits sparse/non-zero-based indexes.
- The `Warnf` logs might be noisy if this happens frequently; consider downgrading/rate-limiting or emitting a metric.

Test plan:
- `go test ./internal/runtime/executor -run 'TestCodexExecutorExecute(Stream)?_Empty' -count=1`

PR #2637, opened 2026-04-09
feat(registry): add gemini-3-flash-agent

arrrrny:
gemini-3-flash-agent is a great model and it has its own quota. 
 If there is no other reason not to list, better add it along with 3.1 flash-lite. 
 
 This is different quoata from the 3.1 quota

### DIFF
--- a/internal/registry/models/models.json
+++ b/internal/registry/models/models.json
@@ -2761,6 +2761,72 @@
       }
     },
     {
+      "id": "gemini-3-flash-agent",
+      "object": "model",
+      "owned_by": "antigravity",
+      "type": "antigravity",
+      "display_name": "Gemini 3 Flash Agent",
+      "name": "gemini-3-flash-agent",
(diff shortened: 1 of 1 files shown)


luispater (approved):
### Summary
Adds three Antigravity model definitions to the static registry (`internal/registry/models/models.json`):
- `gemini-3-flash-agent`
- `gemini-3-flash-lite`
- `gemini-3.1-flash-lite`

### Key findings
- Data-only change (static model registry); no runtime code paths modified.
- Non-blocking: the PR title/body emphasize `gemini-3-flash-agent`, but the diff adds two Lite variants as well—consider updating the PR metadata for clarity.
- Non-blocking: please confirm the Lite variants’ `thinking.levels` and token limits match what Antigravity actually supports to avoid advertising unsupported reasoning levels.

### Test plan
- Not run locally (CI is green).

PR #2657, opened 2026-04-10
fix(usage): skip null usage node in streaming response parsing

Davy-Jessi:
## Summary
- Fix `ParseOpenAIStreamUsage` and `ParseClaudeStreamUsage` to skip chunks where `usage` is `null`
- Previously, chunks with `"usage": null` would pass the `Exists()` check but return 0 for all token fields
- This caused streaming requests to record zero tokens when upstream returns null usage in intermediate chunks

## Test plan
- [x] Verified fix with MiniMax-M2.7 streaming requests - tokens now recorded correctly
- [x] All existing tests pass

## Fix
```go
// Before
if !usageNode.Exists() { ... }

// After  
if !usageNode.Exists() || usageNode.Type != gjson.JSON { ... }
```

### DIFF
--- a/internal/runtime/executor/helps/usage_helpers.go
+++ b/internal/runtime/executor/helps/usage_helpers.go
@@ -241,7 +241,7 @@ func ParseOpenAIStreamUsage(line []byte) (usage.Detail, bool) {
 		return usage.Detail{}, false
 	}
 	usageNode := gjson.GetBytes(payload, "usage")
-	if !usageNode.Exists() || usageNode.Type != gjson.JSON {
+	if !usageNode.Exists() || !usageNode.IsObject() {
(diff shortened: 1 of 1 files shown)


luispater (commented):
## Summary
The null/non-object guard looks directionally correct, and I did not find a blocking logic issue in the parser changes themselves. I am not comfortable approving it yet because the regression that motivated the fix is still untested.

## Key findings
- Non-blocking: `usage_helpers_test.go` still does not cover any of the streaming parsers or the `usage: null` / non-object cases handled here. Please add focused unit tests for the affected streaming helpers, especially OpenAI and Claude, and ideally the Gemini-family helpers touched in the second commit as well.

## Test plan

PR #2669, opened 2026-04-10
fix(auth): load and persist prefix for OAuth providers

thebtf:
## Summary

`FileTokenStore.readAuthFile` never populated `Auth.Prefix` from metadata, so a provider-prefix value written via `PatchAuthFileFields` was lost on every restart. This PR closes the save/load asymmetry.

## Symptom

With multiple Qwen OAuth auths configured (each with its own `prefix`), after a restart:

```
request: qwenai/qwen-3.5-plus -> unknown provider for model
```

Until the user nudges the prefix via the management UI PATCH again. Same pattern affects any OAuth provider that uses `Auth.Prefix` for routing.

## Root cause

The write path persists `prefix`:

```go
// PatchAuthFileFields (already in main)
targetAuth.Prefix = prefix
targetAuth.Metadata["prefix"] = prefix
```

The read path only pulls `email`, `project_id`, `access_token`, and custom headers from metadata — `prefix` is never applied to the typed field when the file is re-read on startup.

## Fix

Three symmetrical changes:

1. **`sdk/auth/filestore.go`** — when `readAuthFile` unmarshals metadata, pull `metadata["prefix"]` into `auth.Prefix` (trim-checked so empty strings do not clobber a default).

2. **`internal/api/handlers/management/auth_files.go` (`RequestQwenToken`)** — when the OAuth handshake creates the initial Qwen record, set both the typed field (`Prefix: "qwenai"`) and the metadata mirror (`Metadata["prefix"] = "qwenai"`) so the first `readAuthFile` after save finds it.

3. **`internal/api/handlers/management/auth_files.go` (`buildAuthFileEntry`)** — expose `"prefix"` in the API response so the management UI can display / edit the current value. The existing PATCH handler already accepts this field; it was write-only in the response.

## Scope

Not Qwen-specific — every OAuth provider that uses `Auth.Prefix` for routing (or wants to start using it) now survives a restart correctly. Qwen is just the one where the bug surfaced because the default prefix `qwenai` is expected by the legacy routing map.

## Test plan

- [x] `go build ./...`
- [x] `go test ./sdk/auth/... ./internal/api/handlers/management/...`
- [ ] Manual: create two Qwen OAuth auths with the `qwenai` prefix, restart, verify `qwenai/qwen-3.5-plus` still routes (**this is the reproducer**).

No new unit tests: the fix is a missing field assignment and a missing response key. The PATCH round-trip path is already covered by existing tests; this PR restores the file-load side of the same round-trip.

## Related

The same change is carried as a fork-local patch on `thebtf/CLIProxyAPIPlus` and has been deployed + smoke-tested.

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -475,6 +475,7 @@ func (h *Handler) buildAuthFileEntry(auth *coreauth.Auth) gin.H {
 		"name":           name,
 		"type":           strings.TrimSpace(auth.Provider),
 		"provider":       strings.TrimSpace(auth.Provider),
+		"prefix":         strings.TrimSpace(auth.Prefix),
 		"label":          auth.Label,
 		"status":         auth.Status,
 		"status_message": auth.StatusMessage,
--- a/internal/store/gitstore.go
+++ b/internal/store/gitstore.go
@@ -496,11 +496,7 @@ func (s *GitTokenStore) readAuthFile(path, baseDir string) (*cliproxyauth.Auth, 
 	if email, ok := metadata["email"].(string); ok && email != "" {
 		auth.Attributes["email"] = email
 	}
-	cliproxyauth.ApplyCustomHeadersFromMetadata(auth)
-	if disabled, ok := metadata["disabled"].(bool); ok && disabled {
-		auth.Disabled = true
-		auth.Status = cliproxyauth.StatusDisabled
-	}
+	cliproxyauth.HydrateAuthFromMetadata(auth)
 	return auth, nil
 }
 
--- a/internal/store/objectstore.go
+++ b/internal/store/objectstore.go
@@ -603,11 +603,7 @@ func (s *ObjectTokenStore) readAuthFile(path, baseDir string) (*cliproxyauth.Aut
 		LastRefreshedAt:  time.Time{},
 		NextRefreshAfter: time.Time{},
 	}
-	cliproxyauth.ApplyCustomHeadersFromMetadata(auth)
-	if disabled, ok := metadata["disabled"].(bool); ok && disabled {
-		auth.Disabled = true
-		auth.Status = cliproxyauth.StatusDisabled
-	}
+	cliproxyauth.HydrateAuthFromMetadata(auth)
 	return auth, nil
 }
 
--- a/internal/store/postgresstore.go
+++ b/internal/store/postgresstore.go
@@ -318,11 +318,7 @@ func (s *PostgresStore) List(ctx context.Context) ([]*cliproxyauth.Auth, error) 
 			LastRefreshedAt:  time.Time{},
 			NextRefreshAfter: time.Time{},
 		}
-		cliproxyauth.ApplyCustomHeadersFromMetadata(auth)
-		if disabled, ok := metadata["disabled"].(bool); ok && disabled {
-			auth.Disabled = true
-			auth.Status = cliproxyauth.StatusDisabled
-		}
+		cliproxyauth.HydrateAuthFromMetadata(auth)
 		auths = append(auths, auth)
 	}
 	if err = rows.Err(); err != nil {
--- a/sdk/auth/filestore.go
+++ b/sdk/auth/filestore.go
@@ -319,15 +319,10 @@ func (s *FileTokenStore) readAuthFiles(path, baseDir string) ([]*cliproxyauth.Au
 		LastRefreshedAt:  time.Time{},
 		NextRefreshAfter: time.Time{},
 	}
-	if prefix, ok := metadata["prefix"].(string); ok {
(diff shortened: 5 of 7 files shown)


luispater (changes requested):
Thanks for chasing this; the file-backed path looks better, but I don't think the fix is complete yet.

Blocking:
- `AuthManager.Load()` restores credentials through the configured store backend, not just `sdk/auth/filestore.go`.
- This PR teaches the filesystem store to hydrate `Auth.Prefix`, but `internal/store/postgresstore.go`, `internal/store/gitstore.go`, and `internal/store/objectstore.go` still rebuild `Auth` without restoring `metadata["prefix"]`.
- As a result, the same restart regression still exists whenever the repo is using one of the supported non-filesystem auth stores.

Please extend the same prefix rehydration to the other store implementations (or move it into shared deserialization logic), then this looks much closer.

Non-blocking:
- A small regression test around store reload preserving `prefix` would make this safer to maintain.

thebtf:
@luispater thanks — you were right about the store gap. Pushed a second commit that addresses both points.

### Blocking fix

Introduced `HydrateAuthFromMetadata()` in `sdk/cliproxy/auth/metadata_hydrate.go` as the shared deserialization entry point. It currently hydrates:

1. **Prefix** — with the stricter validation from `watcher/synthesizer/file.go`: strips whitespace + leading/trailing slashes, rejects values containing slashes (a prefix is a single namespace segment like `"qwenai"`, not a path).
2. **Custom headers** — delegates to the existing `ApplyCustomHeadersFromMetadata` internally.

All four store backends now call `HydrateAuthFromMetadata(auth)` instead of `ApplyCustomHeadersFromMetadata(auth)`:

| Store | File | Change |
|-------|------|--------|
| Filesystem | `sdk/auth/filestore.go` | Replaced inline prefix extraction + `ApplyCustomHeaders` with single `HydrateAuthFromMetadata` |
| Postgres | `internal/store/postgresstore.go` | `ApplyCustomHeaders` → `HydrateAuthFromMetadata` |
| Git | `internal/store/gitstore.go` | same |
| Object (S3/MinIO) | `internal/store/objectstore.go` | same |

The watcher/synthesizer is left unchanged — it already handles prefix correctly inline and has additional provider-specific logic (`excluded_models`, `disable_cooling`, codex JWT parsing, etc.) that doesn't belong in the generic hydration function.

`ApplyCustomHeadersFromMetadata` remains exported for backward compatibility.

### Regression tests

`sdk/cliproxy/auth/metadata_hydrate_test.go`:
- 12 prefix cases: nil/empty metadata, simple prefix, whitespace trimming, leading/trailing/embedded slashes, non-string values
- nil-Auth safety (must not panic)
- Custom-headers integration (verifies the full `HydrateAuthFromMetadata` path applies both prefix and headers)

### Why a shared function instead of inline patches

`ApplyCustomHeadersFromMetadata` is already a shared deserialization helper called from all 5 construction sites — it just covers one metadata field. This extends the pattern so the next metadata-to-typed-field mapping (e.g., `proxy_url` is also missing from 3 stores) only needs one change in `HydrateAuthFromMetadata` instead of patching 5 files independently.

thebtf:
Rebased on latest `main` after the Qwen removal (`8fac2963 chore: remove Qwen support from SDK and internal components`). The PR now contains **only the generic shared-deserialization** improvement — the Qwen-specific `Prefix: "qwenai"` default in `RequestQwenToken` is dropped because the function no longer exists in `main`.

### What remains (still relevant for every OAuth provider with a Prefix)

| Change | Reason |
|--------|--------|
| `sdk/cliproxy/auth/metadata_hydrate.go` — `HydrateAuthFromMetadata()` | Single source of truth for metadata-to-typed-field mapping |
| `sdk/auth/filestore.go` — call `HydrateAuthFromMetadata` | Hydrate `Auth.Prefix` on filesystem store load |
| `internal/store/postgresstore.go` — same | Hydrate on Postgres store load |
| `internal/store/gitstore.go` — same | Hydrate on Git store load |
| `internal/store/objectstore.go` — same | Hydrate on S3/MinIO store load |
| `internal/api/handlers/management/auth_files.go` — expose `"prefix"` in `buildAuthFileEntry` | So the management UI can read/display the current prefix for any provider |

### Impact with Qwen gone from main

Qwen was the specific provider where the `Prefix` save/load asymmetry surfaced most visibly. The asymmetry itself is provider-agnostic: **any OAuth provider in `main` that relies on `Auth.Prefix` for routing loses its prefix on restart** because `readAuthFile` populates `Metadata` but never pulls `metadata["prefix"]` into `auth.Prefix`. Today this primarily matters when `Prefix` is set via the management `PATCH` endpoint (already persists `"prefix"` to metadata) and must survive a process restart.

Diff is now 7 files / +102 / -4. Build + tests green on current `main` tip.

PR #2699, opened 2026-04-11
feat(retry): detect and retry transient HTTP 400 errors from upstream

MoonTiwzy:
## Problem

Some upstream providers (notably Chinese API endpoints) return HTTP 400 with temporary error messages like:
```json
{"error":{"message":"当前模型暂时无法使用,请稍后再试","type":"<nil>"},"requestId":"..."}
```
("The current model is temporarily unavailable, please try again later")

These are **not** true client validation errors but transient upstream failures. Currently, `request-retry` only retries 403, 408, 500, 502, 503, 504 — so these 400s are immediately returned as errors without retry.

## Solution

- Added `isTransientBadRequest()` function that detects known transient error patterns in both Chinese and English
- Updated `isRequestInvalidError()` to exclude transient 400s, allowing retry logic to handle them
- Added `case 400` in both conductor state switches to apply 1-minute cooldown for transient 400s (same as 5xx)
- Non-transient 400s (actual `invalid_request_error`) remain unaffected

### Detected patterns:
- `当前模型暂时无法使用` (model temporarily unavailable)
- `请稍后再试` (try again later) 
- `temporarily unavailable`
- `model is currently unavailable`
- `service is temporarily`
- `try again later`

## Tests

5 new tests covering transient detection, real invalid requests, and nil safety — all passing.

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"path/filepath"
@@ -2051,6 +2052,17 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 								shouldSuspendModel = true
 								setModelQuota = true
 							}
+						case 400:
+							if result.Error != nil && isTransientBadRequest(fmt.Errorf("%s", result.Error.Message)) {
+								if disableCooling {
+									state.NextRetryAfter = time.Time{}
+								} else {
+									next := now.Add(1 * time.Minute)
+									state.NextRetryAfter = next
+								}
+							} else {
+								state.NextRetryAfter = time.Time{}
+							}
 						case 408, 500, 502, 503, 504:
 							if disableCooling {
 								state.NextRetryAfter = time.Time{}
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Thanks for tackling this — the retry classification change is useful, but I found one blocking correctness gap.

Blocking issue:
- `isTransientBadRequestMessage()` now only matches decoded Chinese text. The executor layer passes upstream 400 bodies through as raw JSON strings (`string(b)` in the OpenAI-compatible/Qwen executors), so a provider response that contains escaped Unicode like `\u5f53\u524d\u6a21\u578b...` will not match anymore. In that case the request is still treated as a permanent 400 and never retried, which means the PR misses the transient failure mode it is trying to fix.

Please either keep supporting the raw escaped pattern as well, or normalize/unescape the body before matching, and add a regression test that exercises the raw `\uXXXX` payload path.

Test plan:

MoonTiwzy:
Fixed. I added support for raw `\uXXXX`-escaped transient 400 payloads, which is the path the executor can pass through via `string(body)`. The matcher now checks both decoded text and raw escaped Unicode, and I added regression tests covering the raw escaped JSON/body case.

Tested with:

This should preserve the retry/cooldown path for transient upstream 400s even when the provider body is still unicode-escaped.

MoonTiwzy:
@luispater I pushed a follow-up fix for the unicode-escaped `\uXXXX` path and added regression coverage for the raw escaped payload case. The auth package tests are passing now. Could you take another look?

PR #2706, opened 2026-04-11
Recover responses stream disconnects

KuekHaoYang:
## Summary
- recover upstream Responses transport errors by rewriting recoverable terminal `event: error` frames into synthetic `response.completed`
- recover late transport breaks even when the stream is inside an in-progress `custom_tool_call` edit, by rebuilding `custom_tool_call.input` from incremental deltas
- recover terminal stream failures that surface through the proxy error channel instead of the data channel, so long streams no longer leak `event:error` just because the upstream parser promoted the failure into `errChan`
- stop downstream streaming immediately when writing to the client fails instead of silently continuing to drain upstream output
- make streaming request logs record only bytes that were actually written to the client so debug logs stop overstating delivery success
- add regression coverage for message streams, tool-call streams, terminal-error recovery, and downstream write-failure cancellation across SSE and websocket outputs

## Root Cause
There turned out to be four separate failure modes behind `stream closed before response.completed`:
- some upstream Responses streams ended with transport-level `event: error` frames after useful content had already been emitted
- downstream write failures were still swallowed locally, so the proxy could keep draining upstream output after the desktop client had already lost the connection
- the stream recovery state tracked assistant text and `function_call.arguments`, but not `custom_tool_call.input`, so long Codex edit streams could still fail recovery if the transport broke during an in-progress tool edit
- even after that state fix, not every late transport failure arrived as a stream chunk; some were promoted into the proxy's terminal `errChan` path, which previously bypassed recovery entirely and wrote `event:error` straight to the client

That last gap exactly matches the newest reproduced trace from April 12, 2026 around 16:09 local time: the stream had already emitted `response.output_item.added` for a `custom_tool_call`, then thousands of `response.custom_tool_call_input.delta` events, then a late recoverable upstream transport failure. The proxy now attempts completion recovery in both the chunk path and the terminal error path.

## Impact
Responses streaming is now resilient across the full long-edit path:
- recoverable upstream transport breaks are normalized into a final completion event when enough state exists to rebuild it
- in-progress `custom_tool_call` edits now survive late transport errors because their partial input is preserved and emitted in the synthetic completion
- terminal upstream failures routed through `errChan` are now treated the same way as recoverable `event:error` chunks instead of leaking straight through to the desktop client
- downstream disconnects now terminate the stream promptly instead of ghost-writing more SSE frames into logs after the client is gone
- request logs are now trustworthy for partial-write cases because they only capture the successfully written prefix

## Validation
- `go test ./sdk/api/handlers`
- `go test ./internal/api/middleware`
- `go test ./...` still has pre-existing failures in `internal/runtime/executor` unrelated to this change

### DIFF
--- a/internal/api/middleware/response_writer.go
+++ b/internal/api/middleware/response_writer.go
@@ -77,6 +77,10 @@ func (w *ResponseWriterWrapper) Write(data []byte) (int, error) {
 
 	// CRITICAL: Write to client first (zero latency)
 	n, err := w.ResponseWriter.Write(data)
+	if n <= 0 {
+		return n, err
+	}
+	written := data[:n]
 
 	// THEN: Handle logging based on response type
 	if w.isStreaming && w.chunkChannel != nil {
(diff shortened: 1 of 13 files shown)


KuekHaoYang:
Closing this experimental recovery PR after restoring my local deployment back to official v6.9.22.

PR #2709, opened 2026-04-11
feat(auth): support cloak config in OAuth credential JSON files

nyakojiru:
## Summary

- Adds support for reading `cloak` configuration from OAuth credential JSON files (created via `--claude-login`)
- Previously, `cloak.mode` was only configurable for `claude-api-key` entries in `config.yaml` — OAuth credentials always used the default `"auto"` mode, injecting the Claude Code system prompt into every request
- This change extracts the `cloak` object from auth file metadata and populates auth attributes (`cloak_mode`, `cloak_strict_mode`, `cloak_sensitive_words`, `cloak_cache_user_id`) so that `getCloakConfigFromAuth()` can read them at runtime

### Usage

Add a `"cloak"` field to the OAuth credential JSON file in the auth directory:

```json
{
  "type": "claude",
  "email": "user@example.com",
  "cloak": {
    "mode": "never"
  }
}
```

Supported fields mirror the `claude-api-key` cloak config:
- `mode`: `"auto"` (default), `"always"`, or `"never"`
- `strict-mode`: boolean
- `sensitive-words`: array of strings
- `cache-user-id`: boolean

### Problem

When using OAuth credentials (`--claude-login`), downstream applications sending their own system prompts had the Claude Code system prompt (~1,378 tokens) prepended to every request. This caused:
- Token waste on every request
- Conflicting identity framing (model identified as "Claude Code" instead of following the app's system prompt)
- No way to disable this behavior for OAuth credentials

### Changes

- `internal/watcher/synthesizer/file.go`: Extract cloak config from metadata during auth synthesis (runtime code path)
- `sdk/auth/filestore.go`: Extract cloak config from metadata during file reading (SDK code path)

## Test plan

- [x] Verified `cloak_mode` attribute is populated from auth JSON file on startup
- [x] Tested with `cloak.mode: "never"` — prompt tokens dropped from ~1,477 to 23 (clean passthrough)
- [x] Confirmed all cloak field types parse correctly (string mode, bool strict-mode, array sensitive-words, bool cache-user-id)
- [x] Build passes with no errors

### DIFF
--- a/internal/watcher/synthesizer/file.go
+++ b/internal/watcher/synthesizer/file.go
@@ -157,6 +157,30 @@ func synthesizeFileAuths(ctx *SynthesisContext, fullPath string, data []byte) []
 			}
 		}
 	}
+	// Extract cloak configuration from auth file metadata into attributes
+	// so that getCloakConfigFromAuth can read them for OAuth credentials.
+	if cloakObj, ok := metadata["cloak"].(map[string]any); ok {
+		if mode, ok := cloakObj["mode"].(string); ok && mode != "" {
+			a.Attributes["cloak_mode"] = mode
+		}
+		if strict, ok := cloakObj["strict-mode"].(bool); ok && strict {
+			a.Attributes["cloak_strict_mode"] = "true"
+		}
+		if words, ok := cloakObj["sensitive-words"].([]any); ok && len(words) > 0 {
+			var parts []string
+			for _, w := range words {
+				if s, ok := w.(string); ok {
+					parts = append(parts, s)
+				}
+			}
+			if len(parts) > 0 {
+				a.Attributes["cloak_sensitive_words"] = strings.Join(parts, ",")
+			}
+		}
+		if cache, ok := cloakObj["cache-user-id"].(bool); ok && cache {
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Summary:
This change moves OAuth `cloak` metadata into auth attributes so `getCloakConfigFromAuth()` can consume it, which is the right integration point.

Blocking findings:
- The same auth-file parsing logic still exists in the Git/Object/Postgres stores, and those paths do not populate `cloak_mode`, `cloak_strict_mode`, `cloak_sensitive_words`, or `cloak_cache_user_id`. As written, the feature only works for the file watcher / SDK filestore paths and stays broken for the supported alternative auth backends.
- `buildAuthFromFileData()` in the management auth-file handler still drops the new `cloak` metadata before upserting the auth into the manager, so files uploaded or updated through the management API will not honor the new settings immediately.

Non-blocking:
- Please add focused regression tests for the watcher and SDK auth-file loaders so this mapping is covered explicitly.

Recommendation:
Please move the cloak extraction into a shared helper and reuse it across every auth JSON ingestion path before merging.

PR #2720, opened 2026-04-12
feat(auth): warm allowlisted providers during fill-first routing

cookie223:
## Summary
- trigger a tiny background warmup request for allowlisted secondary providers after fill-first mixed routing selects the primary auth
- start with an `openai` allowlist entry and deduplicate warmups per provider/auth with a short TTL so concurrent requests do not fan out
- add conductor tests covering async warmup, round-robin skip, failure isolation, and TTL deduplication

## Test plan
- [x] `PATH=\"/opt/homebrew/opt/go/bin:$PATH\" && go test ./sdk/cliproxy/auth/...`
- [x] `PATH=\"/opt/homebrew/opt/go/bin:$PATH\" && go build -o /tmp/cliproxyapi-test-build ./cmd/server && rm /tmp/cliproxyapi-test-build`
- [ ] `PATH=\"/opt/homebrew/opt/go/bin:$PATH\" && go test ./...` *(currently fails in `internal/runtime/executor/claude_executor_test.go` on pre-existing system prompt block assertions unrelated to this change)*

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -2955,16 +2955,23 @@ func (m *Manager) maybeWarmOtherMixedProvidersAsync(ctx context.Context, provide
 		}
 		targets = append(targets, providerKey)
 	}
-	now := time.Now()
 	for _, providerKey := range targets {
-		if !m.tryStartMixedProviderWarmup(providerKey, now) {
+		auth, _, err := m.pickNext(ctx, providerKey, routeModel, cliproxyexecutor.Options{}, nil)
+		if err != nil {
+			logEntryWithRequestID(ctx).Debugf("mixed provider warmup skipped provider=%s model=%s error=%v", providerKey, routeModel, err)
 			continue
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
The mixed-provider warmup flow is headed in the right direction, but I found one blocking correctness issue in the new background request path.

Blocking:
- `warmMixedProvider()` currently treats any HTTP response as a successful warmup. The executor contract returns a raw `*http.Response`, so 401/429/5xx responses do not come back as errors here. Because the TTL reservation is recorded before the request and is not cleared on a bad status, a failed warmup suppresses follow-up warmups for that auth for the next minute. In practice, that means the feature can silently stop warming a secondary provider exactly when the provider rejects the request.

Please make non-2xx responses fail the warmup path and avoid keeping the TTL reservation for unsuccessful attempts.

Test plan:
- `PATH="/opt/homebrew/opt/go/bin:$PATH" go test ./sdk/cliproxy/auth/...`

PR #2726, opened 2026-04-12
fix: ensure Qwen refreshed tokens are persisted to disk

aeddk-bnh:
## Summary
Updated `QwenExecutor.Refresh` to synchronize the new `access_token` and rotated `refresh_token` into the underlying storage object. This ensures that `FileTokenStore` correctly persists the latest credentials to disk, preventing "invalid token" errors after server restarts or cache misses.

## Key Changes
- Synchronize refreshed token data (Access/Refresh/Expiry) to `auth.Storage`.
- Injected `refreshTokens` function to `QwenExecutor` for easier unit testing.
- Added comprehensive unit tests in `qwen_executor_test.go` to verify storage updates and graceful fallback when storage is nil.

## Test plan
- [x] Ran `go test -v ./internal/runtime/executor/qwen_executor_test.go` - PASS.
- [x] Verified that Metadata is updated correctly.
- [x] Verified that `QwenTokenStorage` struct is updated when present.
- [x] Verified that the executor doesn't panic when `auth.Storage` is nil.

### DIFF
--- a/internal/auth/qwen/qwen_auth.go
+++ b/internal/auth/qwen/qwen_auth.go
@@ -349,11 +349,17 @@ func (o *QwenAuth) CreateTokenStorage(tokenData *QwenTokenData) *QwenTokenStorag
 	return storage
 }
 
-// UpdateTokenStorage updates an existing token storage with new token data
+// UpdateTokenStorage updates an existing token storage with new token data.
+// It only overwrites RefreshToken and ResourceURL if the new values are non-empty,
+// preserving existing credentials when the server omits them during rotation.
 func (o *QwenAuth) UpdateTokenStorage(storage *QwenTokenStorage, tokenData *QwenTokenData) {
 	storage.AccessToken = tokenData.AccessToken
-	storage.RefreshToken = tokenData.RefreshToken
+	if tokenData.RefreshToken != "" {
+		storage.RefreshToken = tokenData.RefreshToken
+	}
 	storage.LastRefresh = time.Now().Format(time.RFC3339)
-	storage.ResourceURL = tokenData.ResourceURL
+	if tokenData.ResourceURL != "" {
+		storage.ResourceURL = tokenData.ResourceURL
+	}
 	storage.Expire = tokenData.Expire
 }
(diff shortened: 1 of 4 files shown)


aeddk-bnh:
Thank you for the detailed feedback! I've updated the implementation to address the concerns:

- **Preserving Data**: Updated `UpdateTokenStorage` to only overwrite `RefreshToken` and `ResourceURL` if the new values are non-empty. This prevents existing credentials from being wiped when the server omits optional fields during refresh token rotation.
- **Synchronized Auth Timestamps**: Added updates for `auth.LastRefreshedAt` and `auth.UpdatedAt` using `time.Now().UTC()` to ensure the system's internal state is consistent.
- **New Unit Test**: Added `TestQwenExecutorRefresh_PreservesExistingStorageWhenResponseHasEmptyFields` to verify that old storage values are correctly preserved when the refresh response contains empty optional fields.

Force-pushed the updated commit.

luispater (changes requested):
Summary:
This PR improves the local `QwenExecutor.Refresh()` state update logic, but it still does not fix the persistence path described in the PR.

Blocking findings:
- The 429 immediate-retry flow refreshes only the cloned `Auth` passed into `executor.Execute()`. That refreshed copy is never written back through `Manager.Update()`, so the latest tokens are still not persisted for the long-lived manager state or for restart recovery.
- After restart, `FileTokenStore.readAuthFile()` reconstructs Qwen auths from `Metadata` only and leaves `Storage` unset. Updating/cloning `auth.Storage` inside `QwenExecutor.Refresh()` therefore does not address the restart/cache-miss scenario called out in the PR description.

Non-blocking findings:
- `FileTokenStore.Save()` already calls `SetMetadata(auth.Metadata)`, and `MergeMetadata()` applies metadata after serializing the storage struct. That means metadata already wins during file writes, so the new storage mutation is not the critical missing piece.
- The new tests validate executor-local storage mutation, but they do not cover the manager/store integration path where persistence actually happens.

Test plan:
- Not run; review based on code inspection of the refresh, manager, and filestore paths.

PR #2746, opened 2026-04-12
fix(executor): use per-request collision-aware rename maps for OAuth tool name remapping

aikins01:
Replace the global `oauthToolRenameReverseMap` and heuristic-based `clientTitleCaseTools` with explicit `forwardMap`/`reverseMap` in `oauthRenameContext`. This ensures:

- Collision handling is consistent across `tools[]`, `tool_choice`, and message references (`tool_use`, `tool_reference`, nested `tool_reference`)
- Reverse mapping only reverses names that were actually renamed, preventing incorrect lowercasing of client-native TitleCase tools
- Dead code removed: `oauthToolRenameReverseMap` global, redundant `renamed` bool return value

### Problem

The proxy remaps tool names like `bash` -> `Bash` upstream for OAuth traffic to avoid third-party fingerprinting. However:

1. The reverse mapping unconditionally lowercased TitleCase names on the response path, even when the client already sent TitleCase -- breaking smart-mode permissions in downstream clients
2. Collision handling (when both `Glob` and `glob` exist) was only applied to `tools[]` but not to `tool_choice` or message references, causing internal payload inconsistencies

### Fix

- Build a collision-aware effective rename map once per request by scanning `tools[]`
- Use `forwardMap` consistently for all forward rewrites (tools, tool_choice, messages)
- Use `reverseMap` (inverse of actual renames) for all reverse rewrites
- Early return in reverse functions when `reverseMap` is empty (no renames occurred)

### Tests added

- `Collision_ToolChoiceNotRenamed` -- tool_choice respects collisions
- `Collision_MessageRefsNotRenamed` -- message refs respect collisions
- `Mixed_LowercaseAndTitleCase` -- only lowercase tools renamed/reversed
- `Stream_MixedCase` -- SSE stream reverse for mixed-case payloads

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -66,15 +66,6 @@ var oauthToolRenameMap = map[string]string{
 	"notebookedit": "NotebookEdit",
 }
 
-// oauthToolRenameReverseMap is the inverse of oauthToolRenameMap for response decoding.
-var oauthToolRenameReverseMap = func() map[string]string {
-	m := make(map[string]string, len(oauthToolRenameMap))
-	for k, v := range oauthToolRenameMap {
-		m[v] = k
-	}
-	return m
-}()
-
 // oauthToolsToRemove lists tool names that must be stripped from OAuth requests
 // even after remapping. Currently empty — all tools are mapped instead of removed.
 var oauthToolsToRemove = map[string]bool{}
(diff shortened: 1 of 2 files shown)


luispater (approved):
### Summary
This refactor makes OAuth tool-name remapping request-scoped and collision-aware by building an effective `forwardMap`/`reverseMap` from `tools[]`. Reverse mapping now only applies to names that were actually renamed, preventing accidental lowercasing of client-native TitleCase tools.

### Key findings
- ✅ Collision-safe renaming is applied consistently across `tools[]`, `tool_choice`, and message tool references.
- ✅ Reverse remap is now precise (uses per-request `reverseMap`) and safely no-ops when no renames occurred.
- ✅ Test coverage was expanded for collisions, mixed-case requests, and SSE stream reverse mapping.

### Non-blocking notes
- `forwardMap` is derived only from `tools[]`. If a request contains historical tool references in `messages` for tools not present in `tools[]`, those references will no longer be renamed. This seems aligned with the PR intent, but is worth confirming as an intended behavior change.

### Test plan
- CI is green.
- Recommend: `go test ./...` (or at least `go test ./internal/runtime/executor -run TestRemapOAuthToolNames -v`).

luispater (changes requested):
## Blocking

## Recommendation

PR #2748, opened 2026-04-13
feat: add anthropic-compatibility provider type

Yeachan-Heo:
## Problem

Providers that implement the Anthropic messages API format (`x-api-key` auth, `/v1/messages` endpoint) at a custom base URL could not be registered correctly:

- Adding them as a `claude-api-key` entry fails because the executor only sends `x-api-key` for `api.anthropic.com` and falls back to `Authorization: Bearer` for any other host.
- Adding them as `openai-compatibility` won't work if the upstream does not support OpenAI format.

There was no first-class config section for "Anthropic-format API at a custom URL."

## Solution

Add an `anthropic-compatibility` config section — parallel to `openai-compatibility` and `vertex-compat-api-key` — for providers that speak the Anthropic messages protocol at a non-Anthropic base URL.

```yaml
anthropic-compatibility:
  - name: MyProvider
    base-url: https://my-provider.example.com
    api-key-entries:
      - api-key: sk-...
    models:
      - name: upstream-model-id
        alias: ant-my-model
```

### Config (`internal/config/config.go`, `sdk/config/config.go`)
- New `AnthropicCompatibility`, `AnthropicCompatibilityAPIKey`, `AnthropicCompatibilityModel` structs with YAML tags
- `AnthropicCompatibilityModel` implements `GetName()`/`GetAlias()` for the `buildConfigModels[T]` generic
- `Config.AnthropicCompatibility []AnthropicCompatibility` field
- SDK re-exports the three new types

### Synthesizer (`internal/watcher/synthesizer/config.go`)
- `synthesizeAnthropicCompat()` creates `Auth` entries with `Provider: "claude"` and `anthropic_compat: "true"` attribute, matching the same structure as other synthesizers (priority, prefix, headers, excluded-models, models hash)

### Model hash (`internal/watcher/diff/model_hash.go`)
- `ComputeAnthropicCompatModelsHash()` for hot-reload diff detection

### Executor (`internal/runtime/executor/claude_executor.go`)
- `applyClaudeHeaders`: extend `isAnthropicBase` to include `anthropic_compat == "true"` so `x-api-key` is sent to custom-URL providers instead of `Authorization: Bearer`

### Service (`sdk/cliproxy/service.go`)
- `openAICompatInfoFromAuth`: early-return for `anthropic_compat` entries so they reach `case "claude":` instead of the default OpenAI-compat handler
- `registerModelsForAuth` `case "claude":` branches on `anthropic_compat` to call `resolveConfigAnthropicCompatEntry` + `buildAnthropicCompatConfigModels`

### Conductor (`sdk/cliproxy/auth/conductor.go`)
- `rebuildAPIKeyModelAliasLocked` and the slow-path upstream model resolver both had `case "claude":` blocks that only consulted `cfg.ClaudeKey`; aliases from Anthropic-compat entries were silently dropped and forwarded verbatim to upstream
- Added `resolveAnthropicCompatAPIKeyConfig` + `resolveUpstreamModelForAnthropicCompatAPIKey` and branch on `anthropic_compat` in both switch cases

### Startup log (`internal/api/server.go`)
- Anthropic-compat key count added to the startup summary line

## Backward compatibility

No existing config fields or behavior are changed. `anthropic-compatibility` is a new optional section; omitting it is a no-op.

## Testing

Validated locally with a provider (Apitopia) that implements `/v1/messages` at a custom base URL:
- Model alias resolves correctly through the alias table
- `x-api-key` header is sent instead of `Authorization: Bearer`
- Existing `openai-compatibility` entry for the same provider continues to work in parallel

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,6 +20,10 @@ auths/*
 logs/*
 conv/*
 config.yaml
+.omc/*
+.omx/*
+.env
+.env.local
 
 # Development/editor
 bin/*
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ cli-proxy-api
 # Configuration
 config.yaml
 .env
+.env.local
 
 # Generated content
 bin/*
@@ -12,6 +13,7 @@ logs/*
 conv/*
 temp/*
 refs/*
+.omc/*
 
 # Storage backends
 pgstore/*
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/store"
 	_ "github.com/router-for-me/CLIProxyAPI/v7/internal/translator"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/tui"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/usage"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/util"
 	sdkAuth "github.com/router-for-me/CLIProxyAPI/v7/sdk/auth"
 	coreauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
@@ -485,6 +486,7 @@ func main() {
 		}
 	}
 	redisqueue.SetUsageStatisticsEnabled(cfg.UsageStatisticsEnabled)
+	usage.SetStatisticsEnabled(cfg.UsageStatisticsEnabled)
 	redisqueue.SetRetentionSeconds(cfg.RedisUsageQueueRetentionSeconds)
 	coreauth.SetQuotaCooldownDisabled(cfg.DisableCooling)
 
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -68,6 +68,7 @@ func main() {
 	var antigravityLogin bool
 	var kimiLogin bool
 	var xaiLogin bool
+	var grokLogin bool
 	var projectID string
 	var vertexImport string
 	var vertexImportPrefix string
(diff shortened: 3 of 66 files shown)


luispater (changes requested):
## Summary

This PR adds a first-class `anthropic-compatibility` config section for providers that implement the Anthropic Messages API (`/v1/messages`, `x-api-key`) at a non-`api.anthropic.com` base URL, wiring it through config → synthesizer → Claude executor header behavior → SDK model registration and model alias resolution.

## Blocking

1) `auth_kind` is set to a non-standard value (`"anthropic-compat"`) for these API-key entries.
- In `internal/watcher/synthesizer/config.go`, `synthesizeAnthropicCompat()` calls `ApplyAuthExcludedModelsMeta(..., "anthropic-compat")`.
- Several code paths treat only `auth_kind == "apikey"` as API-key and everything else as OAuth-ish (e.g. `oauthExcludedModels`, `OAuthModelAliasChannel` for `claude`).
- This can cause Anthropic-compat API-key auths to incorrectly receive OAuth excluded-models and OAuth model-alias behavior.
Suggested fix: keep `auth_kind: apikey` for Anthropic-compat entries and rely on `anthropic_compat=true` to distinguish the subtype (or add a separate attribute like `auth_subtype=anthropic_compat`).

2) Reusing `compat_name` / `provider_key` triggers OpenAI-compat detection in the auth manager.
- `sdk/cliproxy/auth/conductor.go:isOpenAICompatAPIKeyAuth()` returns true whenever `compat_name` is present.
- When `openai-compatibility` and `anthropic-compatibility` share the same `name`, OpenAI-compat model pooling/alias resolution can be accidentally applied to Anthropic-compat requests.
Suggested fix: explicitly exclude `anthropic_compat=true` in OpenAI-compat detection (at least `isOpenAICompatAPIKeyAuth`, and any other compat-name-based switches), or use distinct attribute keys for Anthropic-compat to avoid colliding with OpenAI-compat semantics.

## Non-blocking

- Please consider documenting whether `models` is required for `anthropic-compatibility`. Today an empty list means no models are registered for that auth (effectively disabling it).
- Tests would help: a small unit test ensuring Anthropic-compat auths are not treated as OAuth/OpenAI-compat, and that alias resolution behaves correctly when both compat sections coexist.

## Test plan

- Not run locally (per review constraints: no PR checkout). CI looks green for the relevant checks.

PR #2750, opened 2026-04-13
feat(openai-compat): add configurable endpoint routing for providers

Dear-Tao:
## Summary
Adds support for configuring the upstream endpoint mode for OpenAI-compatible providers. This allows dynamic routing between `/chat/completions` and `/responses` based on configuration.

## Motivation
Some upstream providers use different endpoint families, such as OpenAI Responses API (`/v1/responses`) versus the standard Chat Completions API (`/v1/chat/completions`). Previously, the endpoint was hardcoded or determined solely by the incoming request format. This change gives users explicit control over which endpoint family to use.

## Changes
- **New `endpoint-mode` config field**: Added to the `openai-compatibility` section to specify the upstream endpoint type.
- **Supported modes**:
  - `auto` (default): Selects the endpoint based on the incoming request format.
  - `chat-completions`: Forces routing to `/chat/completions`.
  - `responses`: Forces routing to `/responses`.
  - Empty: Falls back to legacy behavior.
- **Watcher support**: Added `endpoint-mode` to the config signature calculation so changes trigger a hot-reload.
- **Tests**: Added unit tests for configuration parsing and endpoint resolution logic.

## Configuration Example
```yaml
openai-compatibility:
  - name: my-provider
    base-url: https://api.example.com
    # Supported values: auto, chat-completions, responses
    endpoint-mode: auto
    api-key-entries:
      - key: sk-xxx
```

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -224,6 +224,7 @@ nonstream-keepalive-interval: 0
 #   - name: "openrouter" # The name of the provider; it will be used in the user agent and other places.
 #     prefix: "test" # optional: require calls like "test/kimi-k2" to target this provider's credentials
 #     base-url: "https://openrouter.ai/api/v1" # The base URL of the provider.
+#     endpoint-mode: "auto" # optional override: leave empty to keep provider default/legacy behavior; use auto | chat-completions | responses to force routing
 #     headers:
 #       X-Custom-Header: "custom-value"
 #     api-key-entries:
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -508,6 +508,11 @@ type OpenAICompatibility struct {
 	// BaseURL is the base URL for the external OpenAI-compatible API endpoint.
 	BaseURL string `yaml:"base-url" json:"base-url"`
 
+	// EndpointMode controls which upstream endpoint family is used.
+	// Supported values: "auto", "chat-completions", "responses".
+	// When empty, it defaults to "auto".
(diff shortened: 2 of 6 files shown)


luispater (changes requested):
### Summary
Adds `openai-compatibility[].endpoint-mode` to control upstream routing between `/chat/completions` and `/responses`, updates watcher signatures, and introduces unit tests + a responses-stream passthrough path.

### Blocking
- The stated defaults are inconsistent: PR description and `internal/config/config.go` imply empty => `auto`, but for configured compat entries the executor treats empty as legacy chat-completions routing. Please align docs/comments with the actual behavior (or adjust behavior to match the docs).
- When no compat config entry is resolved (`resolveCompatConfig` returns nil), the executor currently defaults to `auto`, which can change existing behavior for auths not backed by `openai-compatibility` config (e.g., file-based/custom providers): `/v1/responses` requests may start hitting upstream `/responses` instead of being translated to chat completions. Please confirm intent and add coverage.
- `/responses` streaming usage accounting looks incomplete: `ParseOpenAIStreamUsage` only reads `usage.prompt_tokens`/`completion_tokens` at root, so Responses usage fields (`input_tokens`/`output_tokens`, commonly under `response.usage`) won’t be captured reliably. Suggest mirroring `ParseOpenAIUsage`’s fallback logic for streaming (and adding a unit test).

### Non-blocking
- Normalization logic for endpoint mode is duplicated in config + executor; consider reusing one to avoid drift.

### Test plan
- CI checks are green. Recommended locally: `go test ./...` (especially executor + handler streaming paths).

PR #2762, opened 2026-04-13
fix(openai-compat): ignore usage:null in stream token accounting

LuZhuanz:
## Summary
- Fix OpenAI-compatible stream usage parsing to ignore empty usage chunks.
- Specifically ignore:
  - `usage: null`
  - `usage: {}`
  - usage objects where all token counts are zero.
- Add regression tests for the above scenarios.

## Root Cause
Some OpenAI-compatible providers emit intermediate stream chunks with empty usage payloads (`usage:null`, `usage:{}`, or all-zero usage fields) before the final chunk with real usage.

Previously, these empty chunks could trigger early usage publish with zero tokens.
Because usage publishing is once-only, later real usage chunks were ignored, causing management usage totals to remain `0`.

## Fix
In `ParseOpenAIStreamUsage`:
- Parse usage only when `usage` exists and is a JSON object.
- After extracting token fields, return `ok=false` when all token counters are zero:
  - `input_tokens`
  - `output_tokens`
  - `reasoning_tokens`
  - `cached_tokens`
  - `total_tokens`

This prevents empty intermediate chunks from consuming the once-only publish slot.
`EnsurePublished` at stream end still guarantees request counting if no valid usage appears.

## Tests
Added/updated tests in `internal/runtime/executor/helps/usage_helpers_test.go`:
- `TestParseOpenAIStreamUsageNullUsageIgnored`
- `TestParseOpenAIStreamUsageEmptyUsageObjectIgnored`
- `TestParseOpenAIStreamUsageZeroUsageObjectIgnored`
- Existing non-empty usage tests remain passing.

## Validation
- `go test ./internal/runtime/executor/helps -run 'ParseOpenAI(Stream)?Usage|NullUsage|EmptyUsageObject|ZeroUsageObject' -v`
- Production validation confirmed token accounting is no longer stuck at `0`.


kaisa911:
希望可以解决这个问题

luispater (approved):
Summary
- Fixes OpenAI-compatible streaming usage parsing to ignore `usage: null`, `{}`, and all-zero usage chunks so the final real usage isn't dropped by the once-only publisher.
- Aligns streaming parsing with non-streaming `ParseOpenAIUsage` by supporting both `prompt_tokens`/`completion_tokens` (Chat Completions) and `input_tokens`/`output_tokens` (Responses), plus cached/reasoning token aliases.
- Adds regression tests for the above edge cases.

Key findings
- Non-blocking: The all-zero guard could be simplified to `if detail == (usage.Detail{}) { ... }` for readability (since `usage.Detail` is comparable).
- Non-blocking: If any provider emits multiple non-empty usage chunks with increasing counts, the current once-only publish will still capture the first non-empty chunk; worth confirming provider behavior (follow-up only).

Test plan
- CI: green.
- Suggested: `go test ./internal/runtime/executor/helps -run TestParseOpenAIStreamUsage -v`

luispater (changes requested):
## Blocking

## Recommendation

PR #2763, opened 2026-04-13
fix: derive Antigravity User-Agent OS/arch from runtime

RaviTharuma:
## Summary

`AntigravityUserAgent()` was hardcoded to `darwin/arm64`:

```go
return fmt.Sprintf("antigravity/%s darwin/arm64", AntigravityLatestVersion())
```

On a Linux x64 VPS, the User-Agent still claimed `darwin/arm64` — a fingerprint mismatch since Google can correlate the request IP's datacenter location (Linux server) with the claimed OS (macOS).

## Fix

Use `runtime.GOOS` and `runtime.GOARCH` to derive the platform dynamically:

```go
return fmt.Sprintf("antigravity/%s %s/%s", AntigravityLatestVersion(), runtime.GOOS, runtime.GOARCH)
```

Examples:
- macOS M1: `antigravity/1.21.9 darwin/arm64` (unchanged)
- Linux VPS: `antigravity/1.21.9 linux/amd64` (was incorrectly `darwin/arm64`)
- Windows: `antigravity/1.21.9 windows/amd64`

## Risk

Minimal. The version number (auto-fetched from releases API) is unchanged. Only the OS/arch suffix is now accurate. Existing macOS users see identical output.

## Files
- `internal/misc/antigravity_version.go` (+12/-1)
- `go vet` clean

### DIFF
--- a/internal/misc/antigravity_version.go
+++ b/internal/misc/antigravity_version.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


RaviTharuma:
Closing this PR. The hardcoded darwin/arm64 is intentional and correct.

CLIProxyAPI currently works without bans on Google/Gemini/Antigravity. Real Antigravity is a desktop coding tool — most users are on macOS. Claiming darwin/arm64 from a datacenter IP is actually less suspicious than linux/amd64, because Google can correlate the request origin (datacenter) with the claimed OS. Reporting the true server OS would be a more detectable fingerprint, not less.

The version auto-refresh from the releases API already keeps the version number current. The OS/arch suffix should stay darwin/arm64.

PR #2766, opened 2026-04-13
fix(amp): preserve tool calls in proxied streams

edlsh:
## Summary
- normalize known Amp tool names in rewritten responses
- preserve the lowercase `glob` name that Amp expects instead of rewriting it to `Glob`
- repair OpenAI Responses SSE completions whose final `response.output` is empty even though `response.output_item.done` already streamed tool-call items
- preserve streamed `function_call.name` and arguments in repaired `response.completed.response.output` so Amp can normalize and execute the tool call
- add regression coverage for both tool-name normalization and empty completed output repair

## Root Cause
Amp deep GPT/Codex mode could receive a completed Responses stream where the tool call appeared in streamed `response.output_item.done` events, but the final `response.completed.response.output` was `[]`. Amp then dropped the tool use during normalization because the completed output no longer carried the named call, making the CLI look stuck after the model responded.

## Testing
- `go build -o /tmp/cliproxyapi-amp-glob-fixed ./cmd/server`
- live localhost `/v1/responses` forced-tool smoke test returned a completed output containing `function_call` name `shell` and args `{"cmd":"pwd"}`

## Notes
- Local `go test ./...` still has unrelated existing failures in `internal/api/modules/amp` and `internal/runtime/executor` due local/global config state and Claude executor expectations.

### DIFF
--- a/internal/api/modules/amp/response_rewriter.go
+++ b/internal/api/modules/amp/response_rewriter.go
@@ -123,6 +123,52 @@ func (rw *ResponseRewriter) Flush() {
 
 var modelFieldPaths = []string{"message.model", "model", "modelVersion", "response.model", "response.modelVersion"}
 
+// ampCanonicalToolNames maps tool names to the exact casing expected by the
+// Amp mode tool whitelist (case-sensitive match).
+var ampCanonicalToolNames = map[string]string{
+	"bash":  "Bash",
+	"read":  "Read",
+	"grep":  "Grep",
+	"glob":  "glob",
+	"task":  "Task",
+	"check": "Check",
+}
+
+// normalizeAmpToolNames fixes tool_use block names to match Amp's canonical casing.
+// Some upstream models return lowercase tool names (e.g. "bash" instead of "Bash")
+// which causes Amp's case-sensitive mode whitelist to reject them.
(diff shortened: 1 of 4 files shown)


luispater (approved):
## Summary
This PR normalizes known Amp tool names in rewritten Anthropic-style responses, while explicitly preserving the lowercase `glob` tool name that Amp expects. It adds focused unit coverage for normalized and untouched tool names.

## Key findings
- Correctness: Normalization is applied to both non-streaming `content[].name` tool_use blocks and streaming `content_block_start` (`content_block.name`).
- Compatibility: `glob` remains lowercase (and mixed/upper-case variants will normalize to `glob`).
- Safety: Unknown tool names are left untouched.

## Test plan
- Not run locally (per review constraints; no PR checkout).
- CI is green; PR description reports running:
  - `go test ./internal/api/modules/amp`
  - `go test ./...`

## Non-blocking suggestion
- Add a unit test asserting `Glob`/`GLOB` becomes `glob` to lock in the key casing nuance.

edlsh:
Consolidated into #2896 (which now contains both the glob lowercase tool casing fix and the openai responses deep-mode repair).

PR #2773, opened 2026-04-14
fix(payload-filter): resolve parameter filtering bugs for OpenAI-comp…

zhchang:
…atible providers

Fix three critical bugs preventing parameter filtering from working correctly with OpenAI-compatible providers like Grok and DeepSeek:

## Bug #1: Executor ordering - ApplyThinking re-adds filtered parameters
- Move ApplyPayloadConfigWithRoot() to run AFTER ApplyThinking() in openai_compat_executor
- Ensures filter has final authority over parameters sent to upstream provider
- Fixes: reasoning_effort re-injected after filter removes it for suffix models (e.g., grok-3(medium))

## Bug #2: Protocol mismatch - Responses API not recognized by filter rules
- Normalize "openai-response" protocol to "openai" in payloadModelRulesMatch()
- Filter rules with protocol: "openai" now match both Chat Completions and Responses API
- Fixes: Silent filter failures for Responses API requests

## Bug #3: Case-sensitive model matching
- Make matchModelPattern() case-insensitive by lowercasing both pattern and model
- Pattern "grok-*" now correctly matches "Grok-3", "GROK-3", etc.
- Improves robustness for real-world model name variations

## Testing
- Added 19 comprehensive unit tests in payload_helpers_test.go
- Added 9 QA reproduction tests in test/reasoning_effort_filter_test.go
- All tests passing (28/28), zero regressions
- Verified end-to-end with real provider scenarios

## Documentation
- Added FILTER_ARCHITECTURE.md documenting filter system design
- Provides reference for future maintenance and extensibility

### DIFF
--- a/FILTER_ARCHITECTURE.md
+++ b/FILTER_ARCHITECTURE.md
@@ -1,0 +1,578 @@
+# CLIProxyAPI Parameter Filtering System - Architecture Documentation
+
+## Executive Summary
+
+CLIProxyAPI implements a configuration-driven **payload filtering system** that allows administrators to:
+- Set default parameters (only if missing)
+- Override existing parameters
+- **Filter/remove parameters** that incompatible providers don't support
+- Apply these rules selectively based on model name patterns and target protocol
+
+The filtering happens **after request translation** and **before sending to upstream providers**, ensuring that provider-specific constraints are respected.
+
+---
+
+## 1. Architecture Overview
+
+### High-Level Request Flow
+
+```
+┌─────────────────────────────────────────────────────────────────────────────┐
+│ Client Request (OpenAI/Gemini/Claude Format)                                 │
+└──────────────────────────────────────────────────────────────────────────────┘
+                                    ↓
+                    ┌───────────────────────────────┐
+                    │ API Handler (route-specific)   │
+                    │ - Authenticate                 │
+                    │ - Extract model name           │
+                    │ - Determine source protocol    │
+                    └───────────────────────────────┘
+                                    ↓
+                    ┌───────────────────────────────┐
+                    │ Request Translation             │
+                    │ source_format → target_format  │
+                    └───────────────────────────────┘
+                                    ↓
+    ╔═══════════════════════════════════════════════════════════════╗
+    ║ *** PAYLOAD FILTERING (ApplyPayloadConfigWithRoot) ***       ║
+    ║                                                               ║
+    ║ 1. Apply Default Rules (first-write-wins)                   ║
+    ║ 2. Apply DefaultRaw Rules (JSON values)                     ║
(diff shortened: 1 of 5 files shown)


luispater (commented):
Summary
- Fixes payload filter correctness for OpenAI-compatible providers by running `ApplyPayloadConfigWithRoot()` after `thinking.ApplyThinking()`, normalizing protocol matching so `protocol: openai` also applies to Responses API (`openai-response`), and making model glob matching case-insensitive.

Non-blocking
- `payloadModelRulesMatch()`: only normalizes the incoming protocol. Any existing config using `protocol: openai-response` will no longer match; consider treating it as an alias on the rule side too, or explicitly document it as unsupported.
- `OpenAICompatExecutor.Execute` (`alt=responses/compact`): `stream` is deleted before `ApplyPayloadConfigWithRoot()`, but payload default/override rules now run later and could re-introduce `stream`. Consider deleting `stream` again after applying payload rules (or add a regression test).

Test plan
- Not run locally (review-only as requested); CI `build` is green.

luispater (changes requested):
## Summary

This PR fixes three bugs in the payload filtering system for OpenAI-compatible providers. However, it has significant merge conflicts with the current `dev` branch and one of its three fixes is already applied upstream.

## Merge Conflicts (Blocking)

The PR targets `dev` but was created on April 14 and the branch has diverged substantially:

1. `ApplyPayloadConfigWithRoot` has been renamed/refactored to `ApplyPayloadConfigWithRequest` with additional parameters (`fromProtocol`, `requestPath`, `headers`).
2. `payloadModelRulesMatch` signature has changed to include `fromProtocol`, `headers`, `payload`, and `root` parameters.
3. The executor ordering fix (Bug #1) is already present on `dev` -- `ApplyThinking` already runs before `ApplyPayloadConfigWithRequest`.

The PR cannot be merged without a full rebase.

## What is still valuable

- **Bug #2** (protocol normalization): The "openai-response" to "openai" normalization is not yet on `dev` and addresses a real gap.
- **Bug #3** (case-insensitive matching): `matchModelPattern` on `dev` is still case-sensitive. The `strings.ToLower` fix is valid and useful.
- **Tests**: The unit tests for `matchModelPattern` and the filter behavior are valuable, though they need updating to match the current API.

## Recommendations

1. Rebase onto current `dev` and resolve conflicts.
2. Drop the executor ordering change (Bug #1) since it is already applied.
3. Resubmit Bug #2 and Bug #3 fixes against the current function signatures.
4. Consider splitting `FILTER_ARCHITECTURE.md` into a separate PR or removing it -- 578 lines of architecture docs in a bug-fix PR adds maintenance burden.
5. Update tests to use `ApplyPayloadConfigWithRequest` and ensure they compile against current `dev`.

PR #2842, opened 2026-04-16
feat(antigravity): deferred credits activation with pool-based gating

sususu98:
## Problem

When an antigravity auth hits `429 QUOTA_EXHAUSTED`, the old implementation
retries with credits immediately inside the executor (`enabledCreditTypes=["GOOGLE_ONE_AI"]`).
In large auth pools this wastes paid credits while other free-quota auths are still available.

**Old behavior:**

```
Auth A: 429 QUOTA_EXHAUSTED → executor injects credits immediately → success (paid)
        But auths B-Z still have free quota!
```

**New behavior:**

```
Auth A: 429 → return to conductor → try auth B (free) → success
Auth B: 429 → return to conductor → try auth C (free) → success
...
All free auths exhausted → conductor finds auth with credits → last-resort retry with credits
```

## Design

Credits fallback is moved from executor to conductor level. After all normal auths
are exhausted with qualifying errors (429/503/auth unavailable/cooldown), the conductor
searches for antigravity auths with available Google One AI Credits, sets the credits
context flag, and retries as a last resort.

### Credits fallback trigger conditions

| Condition | Description |
|-----------|-------------|
| `cfg.QuotaExceeded.AntigravityCredits == true` | Config toggle must be enabled |
| `lastErr` is 429/503/auth_not_found/auth_unavailable/model_cooldown | Only specific error types |
| providers is empty or contains `antigravity` | No explicit provider or antigravity included |
| providers explicitly set to non-antigravity (e.g. `anthropic`) | **Does not trigger** |

### Attempt limits

- `maxRetryCredentials` configured → bounds the number of credits auths attempted
- Not configured (default 0) → tries all eligible credits auths

## Changes

| File | Change |
|------|--------|
| `sdk/cliproxy/auth/antigravity_credits.go` | New: `WithAntigravityCredits` context helper, `AntigravityCreditsHint` storage, `antigravityCreditsAvailableForModel` |
| `sdk/cliproxy/auth/antigravity_credits_test.go` | New: tests verifying credits auth remains blocked during cooldown |
| `sdk/cliproxy/auth/conductor.go` | Inject credits fallback into `Execute`/`ExecuteCount`/`ExecuteStream`; add `findAllAntigravityCreditsCandidateAuths`, `shouldAttemptAntigravityCreditsFallback`, `tryAntigravityCreditsExecute/Count/Stream` |

## Test plan

- [ ] Credits fallback triggers after all normal auths return 429/503
- [ ] Credits fallback does NOT trigger when providers explicitly set to `["anthropic"]`
- [ ] Credits fallback works when providers is empty or contains `"antigravity"`
- [ ] `maxRetryCredentials` config limits credits auth attempts when set
- [ ] Normal auth selection flow is unaffected

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -99,6 +99,7 @@ quota-exceeded:
   switch-project: true # Whether to automatically switch to another project when a quota is exceeded
   switch-preview-model: true # Whether to automatically switch to a preview model when a quota is exceeded
   antigravity-credits: true # Whether to retry Antigravity quota_exhausted 429s once with enabledCreditTypes=["GOOGLE_ONE_AI"]
+  antigravity-credits-threshold: 0 # Pool exhaustion ratio before activating credits (0=immediate, 0.8=80% exhausted, 1.0=all exhausted)
 
 # Routing strategy for selecting credentials when multiple match.
 routing:
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -80,6 +80,15 @@ type Config struct {
 	// MaxRetryInterval defines the maximum wait time in seconds before retrying a cooled-down credential.
 	MaxRetryInterval int `yaml:"max-retry-interval" json:"max-retry-interval"`
 
+	// TTFTTimeout defines the global Time-To-First-Token timeout for streaming requests, in seconds.
+	// If no first token (including thinking tokens) is received within this duration, the request
+	// is aborted and retried with a different credential/model. 0 = disabled (default).
+	TTFTTimeout int `yaml:"ttft-timeout" json:"ttft-timeout"`
+	// TTFTTimeoutModels provides per-model TTFT timeout overrides, in seconds.
+	// Keys are client-facing model names (the model name in the request, before alias resolution).
+	// Values override the global ttft-timeout for that specific model.
+	TTFTTimeoutModels map[string]int `yaml:"ttft-timeout-models,omitempty" json:"ttft-timeout-models,omitempty"`
+
 	// QuotaExceeded defines the behavior when a quota is exceeded.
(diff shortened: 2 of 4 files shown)


luispater (changes requested):
## Summary
This PR adds two-layer gating for Antigravity credits usage:
1) per-auth credits balance gate from `loadCodeAssist` metadata
2) per-model pool exhaustion gate via `quota-exceeded.antigravity-credits-threshold`

It also improves quota reset handling by using `quotaResetTimeStamp` when present, and adds substantial test coverage around pool gating and balance parsing.

## Blocking
- `clearAntigravityCreditsPermanentlyDisabled()` currently deletes the whole `antigravityCreditsFailureByAuth` entry. Since `updateAntigravityCreditsBalance()` runs on every token refresh (~10 min), this also clears non-permanent cooldown (`DisabledUntil`), effectively defeating the existing 5h auto-disable protection after transient credits failures.
  - Suggestion: only clear when the stored state is actually `PermanentlyDisabled` (and/or `ExplicitBalanceExhausted`), rather than unconditional delete.

## Non-blocking suggestions
- Consider guarding `updateAntigravityCreditsBalance()` behind `quota-exceeded.antigravity-credits` to avoid extra external calls when credits retry is disabled.
- Consider explicitly clearing `credit_amount` / `min_credit_amount` when `paidTier.availableCredits` is missing or doesn’t include `GOOGLE_ONE_AI`, to avoid stale balance decisions.
- Note: `antigravityModelPools` never evicts pools; likely fine if model names are bounded, but worth keeping in mind.

## Tests
Good additions in `internal/runtime/executor/antigravity_executor_credits_test.go` covering threshold gating, sample size behavior, and balance parsing.

Issue #2853, opened 2026-04-17
Claude Code 2.1.112 compatibility regression on /v1/chat/completions


fisher1006:
## Summary
Using CLIProxyAPI as the local adapter for Claude Code, Claude Code 2.1.112 regressed compared with 2.1.110/2.1.111 on the same machine, same provider path, and same CPA config.

Observed effects:
- prompt cache reads drop to 0 or much lower than expected
- some Claude Code requests fail at the CPA compatibility layer
- downgrading Claude Code to 2.1.111 restores expected cache-read behavior

## Environment
- CLIProxyAPI / cliproxyapi: 6.9.10 (brew package), error logs show Version: 6.9.24 in request dump
- cc-switch in front of CPA
- Claude Code routed through:
  - ANTHROPIC_BASE_URL=http://127.0.0.1:15721
  - cc-switch -> http://localhost:8317/v1/chat/completions
- Provider/model path remained the same across tests

## Version comparison
Same machine, same proxy chain, same provider path:
- Claude Code 2.1.110: cache_read_input_tokens = 20480
- Claude Code 2.1.111: cache_read_input_tokens = 34944
- Claude Code 2.1.112: cache_read_input_tokens = 0
- Claude Code 2.1.107: cache_read_input_tokens = 0 in a minimal single-turn test, but 2.1.110/2.1.111 clearly recover on the same tool-using repro

The important boundary appears to be:
- 2.1.110 / 2.1.111: works
- 2.1.112: regresses

## Evidence from CPA logs
Captured from CPA error log for a Claude Code 2.1.112 request to `/v1/chat/completions`:

Headers include:
- `User-Agent: claude-cli/2.1.112 (external, cli)`
- `Anthropic-Beta: claude-code-20250219,interleaved-thinking-2025-05-14,redact-thinking-2026-02-12,context-management-2025-06-27,prompt-caching-scope-2026-01-05,advisor-tool-2026-03-01`

Request body includes Anthropic-specific fields such as:
- `prompt_cache_key`
- `cache_control`
- `tool_choice`

One failing response from CPA/upstream was:
```json
{
  "error": {
    "message": "Unknown parameter: 'tool_choice.name'.",
    "type": "invalid_request_error",
    "param": "tool_choice.name",
    "code": "unknown_parameter"
  }
}
```

[...]

PR #2854, opened 2026-04-17
feat(config): scaffold OAuth cloaking levers on CloakConfig

### LINKED ISSUE #2803: <think> and <thinking> content is being ingested as reasoning content when using Claude
Using fake reasoning CoT for Claude (with API reasoning turned off) in SillyTavern or Lumiverse does not work any more, starting with CLIProxyAPI version v6.9.20.

In v6.9.19 and before, instructions such as "Always think and reason! Start with `<think>` on its own line before any thinking." correctly triggered reasoning process according to the custom CoT. Starting with v6.9.20 and afterwards, the CoT gets ignored completely and the model skips straight to the main response. This affects both Claude Opus and Sonnet 4.6, using both `<think></think>` and `<thinking></thinking>` prefix/suffix pairs.

ne0ark:
## Scope

Adds four new fields to `CloakConfig` that allow fine-grained opt-out of OAuth cloaking behaviors. No call-site changes are made in this PR — the fields are scaffolding only. Behavioral gating follows in PRs 1.2–1.4.

**New YAML keys (all default to legacy behavior):**

| Key | Type | Default (nil/omitted = legacy) | Description |
|---|---|---|---|
| `oauth-sanitize-system-prompt` | `*bool` | `nil` (true) | Gate for system-prompt sanitization on OAuth tokens |
| `oauth-remap-tool-names` | `*bool` | `nil` (true) | Gate for tool-name remapping to Claude Code canonical names |
| `oauth-inject-billing-header` | `*bool` | `nil` (true) | Gate for x-anthropic-billing-account header injection |
| `oauth-disable-header` | `string` | `""` (disabled) | Shared secret for per-request header opt-out (PR 1.5) |

`nil` (omitted from YAML) is treated as `true` by callers — this preserves the exact legacy behavior when the config key is absent.

## Dependencies

None. This is the first PR in the chain.

## Tests Added

- `internal/config/config_cloak_levers_test.go`
  - `TestCloakConfigOAuthLeverDefaults` — verifies zero-value struct has nil pointers and empty string
  - `TestCloakConfigOAuthLeverYAMLRoundTrip` — verifies marshal/unmarshal round-trip with correct YAML tag names; also verifies `omitempty` suppresses nil/empty fields

## AGENTS.md Checklist

- [x] No modifications to `internal/translator/**`
- [x] No `log.Fatal` in changed files
- [x] All comments in English
- [x] `gofmt -l` returns no output on changed files
- [x] New fields use `omitempty` YAML tags
- [x] Default behavior unchanged (`nil` == legacy `true`)
- [x] Diff ≤ 150 LOC (154 total, including 89 LOC of tests)
- [x] `go build -o test-output ./cmd/server` exits 0
- [x] Baseline regression suite passes

## Closes / References

Closes #2803. References #2853.

This is PR 1/5 in the OAuth cloaking levers chain (Gap 1, Milestone A).

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -200,6 +200,12 @@ nonstream-keepalive-interval: 0
 #         - "API"
 #         - "proxy"
 #       cache-user-id: true          # optional: default is false; set true to reuse cached user_id per API key instead of generating a random one each request
+#       # OAuth cloaking levers — all default to true (legacy behavior). Set to false to opt out.
+#       oauth-sanitize-system-prompt: true  # when false, client system prompts are forwarded as-is instead of being collapsed to a stub
+#       oauth-remap-tool-names: true        # when false, tool names are NOT remapped to Claude Code canonical names
+#       oauth-inject-billing-header: true   # when false, the x-anthropic-billing-account block is NOT prepended to the system prompt
+#       # oauth-disable-header: ""          # optional shared secret; when set, clients can disable individual levers at request time
+#                                           # via X-Cliproxy-Cloak-Opt-Out + X-Cliproxy-Cloak-Token headers. Empty = disabled.
 #     experimental-cch-signing: false # optional: default is false; when true, sign the final /v1/messages body using the current Claude Code cch algorithm
 #                                     # keep this disabled unless you explicitly need the behavior, so upstream seed changes fall back to legacy proxy behavior
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -361,6 +361,31 @@ type CloakConfig struct {
 	// CacheUserID controls whether Claude user_id values are cached per API key.
 	// When false, a fresh random user_id is generated for every request.
 	CacheUserID *bool `yaml:"cache-user-id,omitempty" json:"cache-user-id,omitempty"`
+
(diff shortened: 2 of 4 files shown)


ne0ark:
Force-pushed to correct commit author metadata. No code changes; only the author email was rewritten to the GitHub no-reply address. The 3 CI gates should re-run.

luispater (approved):
Summary
- Scaffolds four new `CloakConfig` fields to allow fine-grained opt-out of OAuth cloaking behaviors while keeping legacy defaults via `nil`/empty values.
- Updates config change diff output for the new fields (without leaking the shared-secret value).
- Adds YAML marshal/unmarshal round-trip coverage.

Non-blocking
- In `internal/config/config_cloak_levers_test.go`, consider using `strings.Contains` instead of hand-rolled substring helpers.
- In `internal/watcher/diff/config_diff.go`, `equalPBool` is documented as comparing “effective value”, but it treats `nil` and `true` as different; either adjust wording or align behavior with the “nil == legacy true” semantics (if that’s what you want diff output to reflect).

Tests
- Not run locally (review via `gh` only).
- CI observed: `agents-md-guard` (`close-when-agents-md-changed`) passed.

PR #2859, opened 2026-04-17
fix(translator): use string content for assistant messages in OpenAI format

ArcaneOrion:
## Problem
  The Claude→OpenAI translator was outputting assistant message `content` as an array (`[{"type":"text","text":"..."}]`), but the OpenAI Chat Completions API spec requires assistant content
  to be a **string**.

  This caused some upstream providers (notably New API / One API based relay stations) to reject requests mid-stream with `HTTP/2 INTERNAL_ERROR`, resulting in `unexpected EOF` or `stream
  error` disconnections visible as `500` errors with "Headers were already written" warnings in the proxy log.

  ## Fix
  For assistant messages, extract text from content items and concatenate into a single string. Non-assistant roles (user, etc.) keep the array format since OpenAI allows array content for
  user messages.

  ## Testing
  - All 11 existing tests in `internal/translator/openai/claude/` pass
  - All translator package tests pass (0 failures)

### DIFF
--- a/internal/translator/openai/claude/openai_claude_request.go
+++ b/internal/translator/openai/claude/openai_claude_request.go
@@ -218,13 +218,17 @@ func ConvertClaudeRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 					if hasContent || hasReasoning || hasToolCalls {
 						msgJSON := []byte(`{"role":"assistant"}`)
 
-						// Add content (as array if we have items, empty string if reasoning-only)
+						// OpenAI spec: assistant content must be string (not array).
+						// Concatenate text parts; skip non-text items (images not valid in assistant content).
 						if hasContent {
-							contentArrayJSON := []byte(`[]`)
+							var textParts []string
 							for _, contentItem := range contentItems {
-								contentArrayJSON, _ = sjson.SetRawBytes(contentArrayJSON, "-1", contentItem)
+								item := gjson.ParseBytes(contentItem)
+								if item.Get("type").String() == "text" {
+									textParts = append(textParts, item.Get("text").String())
+								}
 							}
-							msgJSON, _ = sjson.SetRawBytes(msgJSON, "content", contentArrayJSON)
+							msgJSON, _ = sjson.SetBytes(msgJSON, "content", strings.Join(textParts, "\n"))
 						} else {
 							// Ensure content field exists for OpenAI compatibility
 							msgJSON, _ = sjson.SetBytes(msgJSON, "content", "")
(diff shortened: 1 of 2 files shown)


luispater (commented):
### Summary
This PR updates the Claude→OpenAI translator to emit assistant `content` as a **string** (by concatenating text blocks) instead of an array, which should improve compatibility with OpenAI Chat Completions–style upstream relays that reject assistant content arrays.

### Findings
- No blocking issues found in the patch itself.
- Non-blocking: please double-check the choice of `"\n"` as the join separator for multiple text blocks; it may introduce new newlines compared to the original Claude block boundaries. A small unit test for multiple adjacent text blocks (no tools) would lock in the expected behavior.

### CI / Checks
- `pr-test-build` passed.
- `translator-path-guard` is failing (expected for translator-only changes, but may still be merge-gating depending on repo policy).

### Test plan
- Not run locally (reviewed via `gh` diff/checks only; no PR checkout).
- Suggested: `go test ./...` (or at least the translator package tests) before merge.

luispater (changes requested):
## Summary

This PR fixes the Claude-to-OpenAI translator to emit assistant message `content` as a string (concatenating text parts) instead of an array, aligning with the OpenAI Chat Completions API specification. The fix is well-motivated and the implementation is clean.

## Blocking

- **Test logic change in `TestConvertClaudeRequestToOpenAI_ThinkingToReasoningContent`**: The PR removes the guard `if tt.wantHasContent && tt.wantContentText != ""` and now unconditionally asserts `foundText != tt.wantContentText`. For test cases where `wantContentText` is `""` and `wantHasContent` is `false`, this will now compare an empty `foundText` against `""` and pass vacuously -- which is fine. However, the structural change means any future test case with `wantHasContent: false` but a non-empty `wantContentText` (a contradictory setup) would silently fail differently. Please confirm this is intentional and consider keeping the guard or adding a comment explaining the new semantics.

- **Brace formatting in `TestConvertClaudeRequestToOpenAI_AssistantTextToolUseTextOrder`**: The diff shows what appears to be an extra closing brace at inconsistent indentation after the new assertions. The build passes so it compiles, but please verify with `gofmt -d` that the formatting is clean.

## Non-blocking

- The `"\n"` join separator for multiple text blocks is a reasonable default but introduces newlines that weren't present in the original array representation. A brief comment explaining this choice (or a test case with 3+ text blocks) would help future maintainers.
- The `switch` in the thinking test now handles both array and string content types, which is good forward-compatibility for mixed role testing.

## CI

- `pr-test-build`: passed
- `translator-path-guard`: failed (expected for translator-only changes; ignored per policy)

## Test plan

- Existing tests pass per PR description (11 tests in the package).
- Suggested: run `gofmt -d` on the changed test file to confirm formatting, and consider adding a test case with 3+ text content blocks to lock in the join behavior.

PR #2880, opened 2026-04-17
feat(registry): enrich model metadata with OpenRouter context_length

dwillitzer:
## Summary

Adds a background enrichment job that populates `context_length` metadata for registered models by fetching OpenRouter's public models catalog (`https://openrouter.ai/api/v1/models`). Runs once on startup and then refreshes every 24 hours.

## Why

Clients that surface model capabilities (e.g. "supports 1M context") currently rely on hand-maintained entries in `internal/registry/models/models.json`. OpenRouter already publishes authoritative `context_length` values for every routed model; consuming that upstream eliminates the maintenance burden for any model OpenRouter fronts and keeps the catalog accurate as providers bump windows.

## Behavior

- **Startup fetch** populates the in-memory enrichment store; if the fetch fails (network, upstream outage, etc.) the registry falls back to the static JSON with no service impact.
- **24-hour periodic refresh** keeps the cache fresh without hot-pathing the OpenRouter API.
- **Management endpoint** exposes enriched metadata so the UI and clients can display accurate context windows.
- `model_registry.RefreshModels()` now writes through to the live registration so the enriched count is observable and returned.

## Timeout pattern — note to @luispater

Uses `http.Client{Timeout: 30s}` + `context.WithTimeout` — structurally identical to the existing precedent at `internal/misc/antigravity_version.go` (commits `3774b56e`, `8d5e470e`), which was accepted upstream for the same class of periodic metadata fetch. If the repo's "small set of documented exceptions" covers antigravity-version but not model-enrichment, I'm happy to tighten or rewrite — just point at the approved shape.

## Scope

This PR is intentionally scoped to OpenRouter enrichment only. The temporal-anti-drift work that was previously bundled in the same branch is split out to #2574.

## Test plan

- [x] `go build ./...` — compiles cleanly
- [x] Startup enrichment succeeds against live OpenRouter endpoint in local testing
- [x] Registry falls back to static JSON when OpenRouter is unreachable (simulated via DNS block)
- [ ] Maintainer review on timeout pattern

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -498,6 +498,7 @@ func main() {
 				misc.StartAntigravityVersionUpdater(context.Background())
 				if !localModel {
 					registry.StartModelsUpdater(context.Background())
+					registry.StartOpenRouterEnrichment(context.Background())
 				}
 				hook := tui.NewLogHook(2000)
 				hook.SetFormatter(&logging.LogFormatter{})
@@ -574,6 +575,7 @@ func main() {
 			misc.StartAntigravityVersionUpdater(context.Background())
 			if !localModel {
 				registry.StartModelsUpdater(context.Background())
+				registry.StartOpenRouterEnrichment(context.Background())
 			}
 			cmd.StartService(cfg, configFilePath, password)
 		}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -497,6 +497,7 @@ func main() {
 				misc.StartAntigravityVersionUpdater(context.Background())
 				if !localModel {
 					registry.StartModelsUpdater(context.Background())
+					registry.StartOpenRouterEnrichment(context.Background())
 				}
 				hook := tui.NewLogHook(2000)
 				hook.SetFormatter(&logging.LogFormatter{})
@@ -573,6 +574,7 @@ func main() {
 			misc.StartAntigravityVersionUpdater(context.Background())
 			if !localModel {
 				registry.StartModelsUpdater(context.Background())
+				registry.StartOpenRouterEnrichment(context.Background())
 			}
 			cmd.StartService(cfg, configFilePath, password)
 		}
--- a/internal/api/handlers/management/model_definitions.go
+++ b/internal/api/handlers/management/model_definitions.go
@@ -3,6 +3,7 @@ package management
 import (
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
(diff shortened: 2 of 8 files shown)


dwillitzer:
Addressed all P1 and critical-severity findings in `86e1b0b4`:

- **[P1, Codex] Wire `StartOpenRouterEnrichment` at startup.** `cmd/server/main.go` now calls it in both the standalone-TUI path and the main service path, gated by `!localModel` (matching `StartModelsUpdater`).
- **[P1, Codex] Register management routes.** `GET /v0/management/models/health` and `POST /v0/management/models/refresh` are now wired in `internal/api/server.go`.
- **[P1, Codex] Invalidate available-models cache.** `SetModelContextLength` now calls `invalidateAvailableModelsCacheLocked()` when the length actually changes.
- **[high, Gemini] `RefreshModels` handler.** Zero enriched is a legitimate outcome (all already enriched) and no longer returns 400.
- **[critical, Gemini] Tightened fuzzy matching.** Exact-match only — either the full OpenRouter ID or the base name after the last slash. Removed the substring/prefix heuristic that would have attached `gpt-4`'s context length to `gpt-4o`, etc.
- **[medium, Gemini] Package-level `enrichmentClient`** so TCP keep-alive is reused; switched to `json.NewDecoder` for streaming decode instead of buffering the full body.

Remaining Gemini medium items I'd like to defer or discuss:
- **P2, Codex: raw vs normalized keys in enrichment store.** The store is indexed by OpenRouter raw IDs for debugging visibility; lookups normalize through the same slash-strip logic as the enrichment pass, so functionally it works. Happy to switch to local-registry IDs if preferred.
- **Medium, Gemini: O(N·M) `LookupStaticModelInfo` in `RegisterClient`.** Real N and M are both small (tens of models, a handful of metadata fields) so this hasn't shown up on profiles, but I can add a map index if you'd like.

Re-review when you have time.

luispater (changes requested):
Summary:
This PR expands session-affinity ID extraction to additional request headers and to the `prompt_cache_key` body field to improve client compatibility (notably Codex-like clients).

Blocking:
- `X-Client-Request-Id` as a session-affinity key is risky: it is often per-request, which would both defeat stickiness and potentially grow `SessionCache` unbounded within the TTL window (there is no size cap). It also currently takes precedence over more session-shaped identifiers (and would override `prompt_cache_key` when both exist). Please either (a) remove it from session-affinity, or (b) make it lowest-priority / conditional based on a verified “stable per session” signal.
- Header name mismatch: existing Codex-related code paths in this repo use underscore headers (`Session_id`, `Conversation_id`). This PR adds hyphenated variants (`Session-Id`, `Conversation-Id`) while the comments mention `session_id` / `conversation_id`. Please align the accepted header keys and the comments (ideally support both underscore + hyphen variants if the goal is broad client compatibility).

Tests:
- Please extend `sdk/cliproxy/auth/selector_test.go` to cover:
  - `prompt_cache_key` extraction (and precedence vs `conversation_id`).
  - Each new header key you intend to support.
  - Priority behavior when multiple identifiers are present (e.g., ensure explicit session/conversation IDs win over request IDs, and ensure `prompt_cache_key` isn’t shadowed unintentionally).

Test plan:
- Not run locally (diff-only review; no checkout per constraints). Recommend `go test ./...` and specifically validating the updated `TestExtractSessionID_*` coverage.

dwillitzer:
Addressed all three blocking items from @luispater's review in `f38558a9`:

### 1. X-Client-Request-Id removed from session affinity

Removed entirely — it is per-request, which defeats stickiness and would grow `SessionCache` unbounded within the TTL window. Added an explicit comment in the code explaining the exclusion. Verified with `TestExtractSessionID_XClientRequestIdExcluded`.

### 2. Header name mismatch fixed

Now accepts both underscore (Codex convention) and hyphenated (HTTP canonical) variants:

| Priority | Headers accepted |
|----------|-----------------|
| 2 (session) | `X-Session-ID`, `Session-Id`, `Session_id` |
| 3 (conversation) | `Conversation-Id`, `Conversation_id` |

This aligns with the existing Codex executor code paths that set `Session_id` and `Conversation_id` (underscore form), while also supporting standard HTTP header conventions. Comments in code and docstrings updated to reflect the full priority chain.

### 3. Tests added

17 new tests in `sdk/cliproxy/auth/selector_test.go` covering:

- **`prompt_cache_key` extraction** — `TestExtractSessionID_PromptCacheKey`
- **`prompt_cache_key` precedence** — conversation_id wins (`TestExtractSessionID_PromptCacheKeyPrecedenceVsConversationID`), empty conversation_id doesn't shadow it (`TestExtractSessionID_PromptCacheKeyNotShadowedByEmpty`), it wins over message hash (`TestExtractSessionID_PromptCacheKeyWinsOverMessageHash`)
- **Each header variant** — `TestExtractSessionID_SessionHeaderHyphenated`, `TestExtractSessionID_SessionHeaderUnderscore`, `TestExtractSessionID_ConversationHeaderHyphenated`, `TestExtractSessionID_ConversationHeaderUnderscore`
- **Priority ordering** — session headers win over conversation headers, conversation headers win over body fields, Claude Code wins over everything, `X-Client-Request-Id` is excluded
- **Table-driven priority ladder** — `TestExtractSessionID_AllPriorityLevels` covers all 7 priority levels

Full priority chain (documented in code):
1. `metadata.user_id` (Claude Code `_session_{uuid}` format)
2. Session headers: `X-Session-ID` > `Session-Id` > `Session_id`
3. Conversation headers: `Conversation-Id` > `Conversation_id`
4. `metadata.user_id` (non-Claude Code format)
5. `conversation_id` body field
6. `prompt_cache_key` body field
7. Message hash fallback

All 123 tests in `sdk/cliproxy/auth` pass: `go test ./sdk/cliproxy/auth/... -v -count=1`

PR #2885, opened 2026-04-18
feat(auth): Add OAuth quota group suspension

38lunin021189-creator:
## Summary

Add configurable OAuth quota-group suspension for OAuth-backed providers (`antigravity`, `gemini`, `gemini-cli`).

This introduces a generic quota-group layer that can persist manual and automatic suspension state, route model-scoped 429 quota exhaustion into group-level cooldowns, and keep transient capacity failures model-scoped. The main motivation is to make provider behavior match real OAuth quota semantics for accounts that share quota across model families. In the current implementation a 429 usually freezes only the exact model that tripped it, which means a burned Claude family can still be retried through adjacent Claude models until they fail individually.

### What this change adds

- Config-driven quota groups (`oauth-quota-groups`) with provider + wildcard pattern matching.
- Selector enforcement of group-level suspension (both manual and automatic cooldown).
- Persistence of suspension state in `config.yaml` under `oauth-account-quota-group-state`, preserving comments.
- Management endpoints for admin-controlled suspend / lift / clear-cooldown.
- Focused tests including the Antigravity case where Claude can be suspended while Gemini Flash remains available.

## History

This was originally submitted against the downstream Plus repository as router-for-me/CLIProxyAPIPlus#524 and closed by the maintainer with the note that the Plus project only accepts third-party-provider PRs; core `AuthManager` changes belong in this mainline repo. The companion dashboard side has already been merged as itsmylife44/cliproxyapi-dashboard#192 and shipped in dashboard `v0.1.77`.

## Code review feedback from the previous submission

The automated review on the closed PR flagged three concerns. This resubmission addresses all of them in a follow-up commit (`fix(quota-groups): address PR #524 code review feedback`):

1. **Thread-safety when mutating shared config.** The previous version loaded the shared `*internalconfig.Config` pointer from `atomic.Value` and then mutated the `OAuthAccountQuotaGroupState` slice header in place. Any reader that still held the previous pointer could observe a torn slice header.
   - Now `cloneRuntimeConfigForQuotaGroups()` returns a shallow copy of the current snapshot with an independently owned state slice. All three mutation sites (`setOAuthQuotaGroupAutoStateLocked`, `clearOAuthQuotaGroupAutoStateLocked`, `ClearExpiredOAuthQuotaGroupAutoStates`) now mutate the copy and publish it via `SetConfig`. No in-place mutation of the shared snapshot remains.

2. **File corruption from concurrent writes of `config.yaml`.** The previous version used `os.Create` + `Write`, which truncates the live file before the new bytes are written and leaves the file in an inconsistent state on crash or concurrent writer.
   - `SaveConfigPreserveComments` now renders the merged YAML fully in memory first, then calls the new `writeFileAtomic` helper: `CreateTemp` sibling → `Write` → `Sync` → `Chmod` → `Rename`. Rename is atomic on POSIX and best-effort atomic on Windows.
   - A package-level `oauthQuotaGroupPersistMu` serializes quota-group-initiated writes so the merge/render step is not performed concurrently on the same file even though the final rename is atomic.

3. **Synchronous expired-state cleanup on the hot path.** The previous version called `ClearExpiredOAuthQuotaGroupAutoStates` at the top of `Execute`, `ExecuteCount`, and `ExecuteStream`. That call took the global `Manager` mutex and, on any expiry, synchronously rewrote `config.yaml` — inside the request path.
   - All three hot-path calls are removed. The lock-free `oauthQuotaGroupBlock` read already ignores entries whose `AutoSuspendedUntil` is in the past, so the request path stays correct without cleanup.
   - A new `StartOAuthQuotaGroupCleanup` / `StopOAuthQuotaGroupCleanup` pair runs the reaper in a single background goroutine (30s ticker + 1s primer), tracked by `Manager.quotaGroupCleanupCancel` so only one loop is alive at a time. It is auto-wired into the existing `StartAutoRefresh` / `StopAutoRefresh` lifecycle so hosts get it for free.

## Verification

- `go build ./...` — OK on `golang:1.26-alpine`.
- `go test -race -count=1 ./sdk/cliproxy/auth/... ./internal/config/...` — OK, no data races reported.
- Existing `conductor_oauth_alias_suspension_test.go`, `antigravity_quota_groups_behavior_test.go`, `oauth_quota_groups_test.go`, `oauth_quota_groups_manager` (via `auth` package tests), and `internal/config` tests pass.

## Commits

1. `feat(auth): Add OAuth quota group suspension` — the feature itself (ported from the closed PR).
2. `fix(quota-groups): address PR #524 code review feedback` — the three review fixes above.

## Related

- Dashboard counterpart (already merged): itsmylife44/cliproxyapi-dashboard#192
- Previous closed PR: router-for-me/CLIProxyAPIPlus#524

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -360,6 +360,59 @@ nonstream-keepalive-interval: 0
 #   kimi:
 #     - "kimi-k2-thinking"
 
+# OAuth quota groups for Google-backed shared quota families.
+# These groups are used for per-account + per-quota-family suspension logic.
+# oauth-quota-groups:
+#   - id: "claude_45"
+#     label: "Claude"
+#     providers: ["antigravity", "gemini", "gemini-cli"]
+#     patterns:
+#       - "claude-opus-4-6*"
+#       - "claude-opus-4-5*"
+#       - "claude-sonnet-4-6*"
+#       - "claude-sonnet-4-5*"
+#       - "gpt-oss-120b-medium*"
+#     priority: 300
+#     enabled: true
+#   - id: "g3_pro"
+#     label: "Gemini Pro"
+#     providers: ["antigravity", "gemini", "gemini-cli"]
+#     patterns:
+#       - "gemini-3.1-pro-high*"
+#       - "gemini-3.1-pro-low*"
+#       - "gemini-3-pro-high*"
+#       - "gemini-3-pro-low*"
+#       - "gemini-3-pro-image*"
+#     priority: 200
+#     enabled: true
+#   - id: "g3_flash"
+#     label: "Gemini Flash"
+#     providers: ["antigravity", "gemini", "gemini-cli"]
+#     patterns:
+#       - "gemini-3-flash*"
+#       - "gemini-3.1-flash*"
+#       - "gemini-3-flash-image*"
+#       - "gemini-3.1-flash-image*"
+#       - "gemini-3-flash-lite*"
+#       - "gemini-3.1-flash-lite*"
+#     priority: 100
+#     enabled: true
+
+# CPA-owned persisted state for account + quota-group manual/auto suspension.
+# This section is normally maintained by the management API.
+# oauth-account-quota-group-state:
+#   - auth_id: "auth-123"
+#     group_id: "claude_45"
+#     manual_suspended: true
+#     manual_reason: "maintenance"
+#     auto_suspended_until: 2026-04-14T12:00:00Z
+#     auto_reason: "quota_exhausted"
+#     source_model: "claude-opus-4-6-thinking"
+#     source_provider: "antigravity"
+#     reset_time_source: "retry_after"
+#     updated_at: 2026-04-14T10:00:00Z
+#     updated_by: "management:manual"
+
 # Optional payload configuration
 # payload:
 #   default: # Default rules only set parameters when they are missing in the payload.
--- a/internal/api/handlers/management/handler.go
+++ b/internal/api/handlers/management/handler.go
@@ -65,6 +65,10 @@ func NewHandler(cfg *config.Config, configFilePath string, manager *coreauth.Man
 		allowRemoteOverride: envSecret != "",
 		envSecret:           envSecret,
 	}
+	if manager != nil {
+		manager.SetConfig(cfg)
+		manager.SetConfigFilePath(configFilePath)
+	}
 	h.startAttemptCleanup()
 	return h
 }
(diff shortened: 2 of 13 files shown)


38lunin021189-creator:
Addressed the latest quota-group review items in commit e035c2ae.

What changed:
- publish runtime OAuth quota snapshots while still holding the manager lock, so concurrent cooldown updates do not drop each other
- apply manual suspend and auto-clear changes to live routing immediately instead of waiting for an external config reload
- update aggregated auth availability against the freshly published quota snapshot
- add regression coverage for manual publish, auto-clear publish, and quota-group cooldown lifecycle

Verification:
- docker run --rm -v "${PWD}:/workspace" -w /workspace golang:1.26 go test ./sdk/cliproxy/auth
- docker run --rm -v "${PWD}:/workspace" -w /workspace golang:1.26 go build -o /tmp/test-output ./cmd/server

luispater (changes requested):
### Summary
This PR adds configurable OAuth quota-group suspension for OAuth-backed providers (`antigravity`, `gemini`, `gemini-cli`), including:
- Config-driven quota groups (`oauth-quota-groups`) with default bootstrap groups + normalization.
- Persisted per-auth + per-group manual/auto suspension state (`oauth-account-quota-group-state`) in `config.yaml`.
- Selector-side enforcement to block suspended groups, and group-level 429 cooldown routing (instead of per-model freezing).
- Management API endpoints to view/update quota groups and suspension state.
- Background cleanup loop to reap expired auto suspensions (moved off the hot path).
- Atomic config writes to avoid truncation/corruption and preserve restrictive perms.
- Focused tests (including the Antigravity Claude-vs-Gemini behavior).

Overall the quota-group architecture and the “publish via cloned runtime config + atomic persistence” direction look solid and well-tested.

### Blocking: upstream attempt timeout
`sdk/cliproxy/auth/conductor.go` introduces `upstreamAttemptTimeout = 25s` and uses it to cancel *each* upstream attempt for:
- `executor.Execute(...)` (non-stream)
- `executor.CountTokens(...)`
- `executor.ExecuteStream(...)` (bootstrap)

This is a major behavioral change for non-stream requests: legitimate responses that take >25s will now be canceled and surfaced as a synthetic `upstream_timeout` (HTTP 504), and it appears to conflict with this repo’s timeout policy (timeouts only allowed in limited, explicitly-listed places).

**Request:** please remove this timeout, or gate it behind explicit config (default disabled), and/or scope it narrowly (e.g., stream bootstrap and possibly `CountTokens`, but not full non-stream `Execute`).

### Non-blocking suggestions
- Add `h.authManager == nil` guards to the new quota-group PATCH handlers for consistency/safety.
- Make `ClearOAuthQuotaGroupAutoState` / `ClearAllOAuthQuotaGroupAutoStates` return `nil` (or a “changed” flag) when no state changes, so management calls don’t rewrite `config.yaml` unnecessarily.
- Consider tightening `writeConfigFileAtomically`’s “delete target then retry rename” branch to only the error cases it actually intends to handle.

### Suggested test plan (not run locally)
- `go test -race -count=1 ./sdk/cliproxy/auth/... ./internal/config/...`
- `go test ./...`

PR #2892, opened 2026-04-18
feat(api): integrate auth index into key retrieval endpoints for Gemi…

LTbinglingfeng:
…ni, Claude, Codex, OpenAI, and Vertex


LTbinglingfeng:
Pushed follow-up commit c26936e2 to address the review feedback:

- Concurrency: protect config/authManager pointer access with h.mu (keysWithAuthIndex now lock; provider key PUT/PATCH/DELETE paths also lock and use persistLocked; SetConfig/SetAuthManager now lock).
- Perf: removed per-request full ConfigSynthesizer.Synthesize() for all providers.
- Robust mapping: auth-index is now derived by computing the same stable auth IDs (StableIDGenerator.Next with the same kind/parts) and then mapping to the live authManager index by ID (no cross-file order coupling).
- Side effects: live index map is built from authManager.List() clones; EnsureIndex only touches those clones.
- Added unit tests for auth-index resolution + omission when auth is not yet in the live manager.

PR #2895, opened 2026-04-18
fix(conductor): skip auth suspension for count_tokens 404s

ArcaneOrion:
## Summary
  - Non-Anthropic upstreams return 404 for `/v1/messages/count_tokens` because they don't
  support this Anthropic-specific endpoint
  - The conductor treated these 404s as "model not found" and suspended auths for 12 hours,
  causing cascading 503 errors for ALL requests on that model
  - Skip `MarkResult` for 404 errors in `executeCountMixedOnce` so count_tokens failures
  don't trigger auth suspension

  ## Test plan
  - [x] `TestManagerExecuteCount_404DoesNotSuspendAuth` — verifies count_tokens 404 does not
  suspend auth and subsequent Execute still succeeds
  - [x] Full `sdk/cliproxy/auth/...` test suite passes with no regressions

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -1393,7 +1393,18 @@ func (m *Manager) executeCountMixedOnce(ctx context.Context, providers []string,
 				if ra := retryAfterFromError(errExec); ra != nil {
 					result.RetryAfter = ra
 				}
-				m.MarkResult(execCtx, result)
+				// count_tokens 404s from non-Anthropic upstreams indicate an
+				// unsupported endpoint, not a missing model. Recording them via
+				// MarkResult would incorrectly suspend the auth for 12 hours and
+				// block ALL subsequent requests for the same model.
+				if result.Error.HTTPStatus != http.StatusNotFound {
+					m.MarkResult(execCtx, result)
+				} else {
+					logEntryWithRequestID(execCtx).Debugf(
+						"skipping MarkResult for count_tokens 404 on auth=%s model=%s: %s",
+						auth.ID, upstreamModel, errExec.Error(),
+					)
+				}
 				if isRequestInvalidError(errExec) {
 					return cliproxyexecutor.Response{}, errExec
 				}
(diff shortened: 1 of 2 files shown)


luispater (approved):
### Summary
This prevents `/v1/messages/count_tokens` 404s (common for non-Anthropic upstreams that don’t implement the Anthropic-specific endpoint) from being treated as “model not found” and triggering a 12-hour auth/model suspension that can cascade into 503s for normal traffic on the same model.

### What changed
- In `executeCountMixedOnce`, skip `MarkResult` when `CountTokens` returns HTTP 404, and emit a debug log instead.
- Add a regression test `TestManagerExecuteCount_404DoesNotSuspendAuth` to verify a count_tokens 404 does **not** suspend/cooldown the auth and that a subsequent `Execute` still succeeds.

### Notes (non-blocking)
- Skipping `MarkResult` means `hook.OnResult` won’t see these 404s; if you rely on hooks for metrics/monitoring, consider recording count_tokens 404s via a separate path that doesn’t affect auth state.
- The debug log includes `errExec.Error()`; please ensure CountTokens error strings cannot contain sensitive data.

### Test plan
- `go test ./sdk/cliproxy/auth/...`
- `go test ./...`

luispater (approved):
## Summary

Targeted fix for a cascading-failure bug: non-Anthropic upstreams return 404 for `/v1/messages/count_tokens`, and `MarkResult` treated this as "model not found", suspending the auth for 12 hours. The fix skips `MarkResult` specifically for 404 errors in `executeCountMixedOnce`.

## Key findings

- No blocking issues. Minimal, correct, well-tested.
- (non-blocking) `countTokenCalls` tracked but never asserted in the test.

## Verdict

Approve. CI passes, no merge conflicts.

---

observerss:
I reproduced this exact production failure on **v7.2.86** (commit `81d70f5d`): a generic `count_tokens` route 404 was followed by 503s for ordinary messages, and the configured model alias disappeared until an unchanged config save/hot reload.

Sanitized timeline and environment details are in #4410.

I opened #4411 as an alternative implementation that incorporates both review comments on this PR:

- it only treats generic route-level responses such as `404 page not found`, `{"detail":"Not Found"}`, and `Cannot POST /v1/messages/count_tokens` as availability-neutral;
- structured model-not-found 404s still retain the normal 12-hour model cooldown;
- failures still flow through request counters, `OnResult` hooks, and error event publishing.

Validation on the current `dev` branch:

```text
go test ./sdk/cliproxy/auth -count=1
go test ./... -count=1
go build -o /tmp/cli-proxy-api-fable5-fix ./cmd/server
```

All passed.

PR #2896, opened 2026-04-18
fix(amp): smart-mode tool name fixes + deep-mode response repair

edlsh:
## Summary

This consolidates the Amp smart-mode and deep-mode tool fixes into one PR.

- Fixes Claude OAuth tool-name reverse remapping so responses only map tool names back when the request actually forward-renamed that tool for the same request.
- Keeps the OAuth tool-name transform flow centralized and pre-allocates the reverse map.
- Preserves Amp's lowercase `glob` tool casing in proxied responses.
- Repairs OpenAI Responses streaming when upstream sends `response.completed` with an empty `response.output` by reconstructing it from prior `response.output_item.done` events.

## Amp Coverage

Smart mode:
- Includes the three Claude OAuth tool-name remap commits.
- Includes the Amp lowercase `glob` tool casing fix that was previously in #2766.
- Addresses the original "Amp CLI completely broken" OAuth tool remapping regression.

Deep mode:
- Includes the OpenAI Responses stream repair from #2766.
- Validated locally on April 25, 2026 against Amp deep mode / oracle on `gpt-5.4` high reasoning: before the fix, reasoning appeared but the final answer was blank; after running the fixed build, the final answer appeared correctly.

Closes #2766.

## Testing

- `go build ./internal/api/modules/amp/... ./sdk/api/handlers/openai/... ./internal/runtime/executor/...`
- `go test ./internal/api/modules/amp/... ./sdk/api/handlers/openai/... ./internal/runtime/executor/...`
- `go test -run "TestForwardResponsesStream|TestNormalizeAmpToolNames|TestRewriteStreamChunk" -v ./internal/api/modules/amp/... ./sdk/api/handlers/openai/...`

Note: `TestMultiSourceSecret_CacheEmptyResult` in `internal/api/modules/amp/secret_test.go` is a pre-existing failure on the parent commit as well.

### DIFF
--- a/internal/api/modules/amp/response_rewriter.go
+++ b/internal/api/modules/amp/response_rewriter.go
@@ -123,6 +123,52 @@ func (rw *ResponseRewriter) Flush() {
 
 var modelFieldPaths = []string{"message.model", "model", "modelVersion", "response.model", "response.modelVersion"}
 
+// ampCanonicalToolNames maps tool names to the exact casing expected by the
+// Amp mode tool whitelist (case-sensitive match).
+var ampCanonicalToolNames = map[string]string{
+	"bash":  "Bash",
+	"read":  "Read",
+	"grep":  "Grep",
+	"glob":  "glob",
+	"task":  "Task",
+	"check": "Check",
+}
+
+// normalizeAmpToolNames fixes tool_use block names to match Amp's canonical casing.
+// Some upstream models return lowercase tool names (e.g. "bash" instead of "Bash")
+// which causes Amp's case-sensitive mode whitelist to reject them.
+func normalizeAmpToolNames(data []byte) []byte {
+	// Non-streaming: content[].name in tool_use blocks
+	for index, block := range gjson.GetBytes(data, "content").Array() {
+		if block.Get("type").String() != "tool_use" {
+			continue
+		}
+		name := block.Get("name").String()
+		if canonical, ok := ampCanonicalToolNames[strings.ToLower(name)]; ok && name != canonical {
+			path := fmt.Sprintf("content.%d.name", index)
+			var err error
+			data, err = sjson.SetBytes(data, path, canonical)
+			if err != nil {
+				log.Warnf("Amp ResponseRewriter: failed to normalize tool name %q to %q: %v", name, canonical, err)
+			}
+		}
+	}
+
+	// Streaming: content_block.name in content_block_start events
+	if gjson.GetBytes(data, "content_block.type").String() == "tool_use" {
+		name := gjson.GetBytes(data, "content_block.name").String()
+		if canonical, ok := ampCanonicalToolNames[strings.ToLower(name)]; ok && name != canonical {
+			var err error
+			data, err = sjson.SetBytes(data, "content_block.name", canonical)
+			if err != nil {
+				log.Warnf("Amp ResponseRewriter: failed to normalize streaming tool name %q to %q: %v", name, canonical, err)
+			}
+		}
+	}
+
+	return data
+}
+
 // ensureAmpSignature injects empty signature fields into tool_use/thinking blocks
 // in API responses so that the Amp TUI does not crash on P.signature.length.
 func ensureAmpSignature(data []byte) []byte {
@@ -179,6 +225,7 @@ func (rw *ResponseRewriter) suppressAmpThinking(data []byte) []byte {
 
 func (rw *ResponseRewriter) rewriteModelInResponse(data []byte) []byte {
 	data = ensureAmpSignature(data)
+	data = normalizeAmpToolNames(data)
 	data = rw.suppressAmpThinking(data)
 	if len(data) == 0 {
 		return data
@@ -277,6 +324,9 @@ func (rw *ResponseRewriter) rewriteStreamChunk(chunk []byte) []byte {
 func (rw *ResponseRewriter) rewriteStreamEvent(data []byte) []byte {
 	// Inject empty signature where needed
 	data = ensureAmpSignature(data)
+
+	// Normalize tool names to canonical casing
+	data = normalizeAmpToolNames(data)
 
 	// Rewrite model name
 	if rw.originalModel != "" {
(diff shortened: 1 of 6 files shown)


edlsh:
## ⚠️ Severity: Amp CLI is completely broken without this fix

Want to flag the impact clearly for anyone triaging this PR:

**This is not a cosmetic or edge-case bug. With any Claude OAuth account in the pool, Amp CLI is 100% unusable end-to-end.**

Every single Amp request sends at least one lowercase tool (`glob`, `skill`, `todo_write`, `task_list`, …) alongside the TitleCase ones (`Bash`, `Read`, `Grep`, `Edit`, `Task`, `Write`, …). That means the `renamed=true` flag fires on literally every request, which means the global reverse map unconditionally lowercases every `Bash`/`Read`/`Grep`/`Edit`/`Task`/`Write` in the response. Amp's tool registry has these names in TitleCase only, so:

- `Bash` → `bash` → `tool "bash" is not allowed for smart mode` → **no commands run**
- `Read` → `read` → rejected → **no file reads**
- `Edit`/`Write` → `edit`/`write` → rejected → **no file edits**
- `Grep` → `grep` → rejected → **no searches**
- `Task` → `task` → rejected → **no subagents**

In practice the agent loop stalls on the very first turn. Users see "Done." with no work performed, or silent tool failures, depending on the request. There's no workaround short of disabling all Claude OAuth accounts in the config, which defeats the whole point of routing Amp through CLIProxyAPI.

This is the root cause of a cluster of "Amp doesn't work" reports that have been attributed to cloaking, cache-control, or model routing — none of which are the issue. The forward rename is correct; only the reverse pass is wrong.

Repro in under a minute on `origin/main`:

```
amp -m smart --execute "run: echo hi"
# → hangs or exits with no tool execution
```

With this patch applied, Amp works on the first try. Regression tests (`TestRemapOAuthToolNames_MixedCase_OnlyRenamedToolsReversed`, `TestReverseRemapOAuthToolNamesFromStreamLine_HonorsPerRequestMap`) lock in the mixed-case invariant so this can't regress silently.

Would appreciate a prioritized review — every day this sits unmerged, every Amp user on an OAuth account is dead in the water.

0xkaisa:
@luispater can you check this PR? Thanks.

edlsh:
Quick verification update: this is still reproducible on current upstream `dev` (`36cc762f`). A focused mixed-case repro still turns an original TitleCase `Bash` response tool into `bash` when the same request also includes a lowercase-renamed tool like `glob`.

The PR branch (`ed320c83`) passes the targeted regression tests:

```text
TestRemapOAuthToolNames_MixedCase_OnlyRenamedToolsReversed
TestReverseRemapOAuthToolNamesFromStreamLine_HonorsPerRequestMap
```

So the branch is still fixing the current upstream behavior, not a stale issue.

luispater (approved):
### Summary
Solid fix for the mixed-case OAuth tool-name anti-fingerprinting remap: reverse mapping is now per-request, so only tools that were actually forward-renamed get restored. This prevents TitleCase tools like `Bash` from being incorrectly lowercased when a different tool in the same request (e.g. `glob`) triggers a rename.

### Key findings
- ✅ `remapOAuthToolNames` now produces a per-request reverse map; reverse functions only touch entries present in that map.
- ✅ Call sites are updated across non-stream, stream (both passthrough and translated paths), and `count_tokens`.
- ✅ Good regression tests for the Amp mixed-case scenario, including the SSE line path.

### Test plan
- Not run locally (review only).
- PR description reports `go test ./internal/runtime/executor/...`; CI `build` is green.

ben-vargas:
@luispater quick update: I’ve been working with @edlsh over X/Twitter DMs and testing this PR locally against Amp CLI.

I pulled the latest PR branch, rebuilt `cli-proxy-api`, restarted the proxy, and re-ran Amp tests. The latest version fixes both issues I was seeing:

- Amp smart-mode tool-call failures caused by case-sensitive tool names like `Bash` / `Read` being rewritten incorrectly.
- Deep-mode / `gpt-5.4` Responses failures where streamed tool calls were present, but the final `response.completed.response.output` could arrive empty.

I also checked the CLIProxyAPI logs directly. With this PR, the repaired completed event includes the expected output items/function calls, and Amp continues the turn correctly.

The latest SSE framing follow-up also looks good to me: repaired multiline payloads are emitted with `data:` on each line, with regression coverage for multiline completed frames and indexed/unindexed output items.

Focused tests pass locally: `go test ./sdk/api/handlers/openai ./internal/runtime/executor ./internal/api/modules/amp`

I did see a few normal tool execution errors from the model guessing nonexistent paths, but not proxy formatting errors. The remaining full-suite `internal/registry` failure around `gpt-5.5` free-tier metadata appears unrelated.

From my local testing and log inspection, this PR looks good for the Amp smart-mode and deep-mode failures.

Amoodaa:
> Deep-mode / gpt-5.4 Responses failures where streamed tool calls were present, but the final response.completed.response.output could arrive empty.
this was fixed in latest release and gpt-5.5 is now working if i use it through the amp.internal.model flags

0xkaisa:
Still error, how can we fix it @ben-vargas @Amoodaa @edlsh

ben-vargas:
@0xkaisa - I don't know why this hasn't been merged/fixed, I'm maintaining a fork branch here and keeping it up to date - this is what I'm using: https://github.com/ben-vargas/ai-cli-proxy-api/tree/fix-amp-patch

That branch is on v6.10.5 with Amp fixes in this PR.

luispater:
> @0xkaisa - I don't know why this hasn't been merged/fixed, I'm maintaining a fork branch here and keeping it up to date - this is what I'm using: https://github.com/ben-vargas/ai-cli-proxy-api/tree/fix-amp-patch
> 
> That branch is on v6.10.5 with Amp fixes in this PR.

Sorry for the wait, we've been swamped with issues, PRs, and other chores lately.

As a maintainer, if everything looks good to you, I'll merge this right away and release a new version.

luispater (approved):

luispater:
@ben-vargas BTW, due to well-known reasons, the core team has lost all Claude subscriptions.

This has significantly impacted the review progress of all Claude related PRs.

We sincerely apologize for the inconvenience.

ben-vargas:
Thanks @luispater - Wasn't sure if this was being ignored on purpose or just delayed in the workload... obviously the latter, so no worries and thanks for the attention!

PR #2899, opened 2026-04-19
Expands session affinity ID extraction sources

can1357:
Broadens the set of identifiers used to maintain session affinity, improving compatibility with a wider range of clients (e.g., OpenAI Codex).

- Checks additional headers (`X-Client-Request-Id`, `Session-Id`, `Conversation-Id`) alongside `X-Session-ID` when extracting a session identifier
- Adds support for the `prompt_cache_key` body field as a session source, which Codex stamps with a session ID

### DIFF
--- a/internal/registry/model_definitions.go
+++ b/internal/registry/model_definitions.go
@@ -6,7 +6,10 @@ import (
 	"strings"
 )
 
-const codexBuiltinImageModelID = "gpt-image-2"
+const (
+	codexBuiltinImageModelID = "gpt-image-2"
+	codexBuiltinSparkModelID = "gpt-5.3-codex-spark"
+)
 
 // staticModelsJSON mirrors the top-level structure of models.json.
 type staticModelsJSON struct {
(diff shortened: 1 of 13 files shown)


luispater (approved):
## Summary
This change makes the management auth upload path reuse the existing auth-file synthesizer so uploaded OAuth JSON files populate derived runtime attributes (notably `priority`) consistently with the watcher synthesis behavior. It also adds a focused regression test to ensure `priority` is preserved on upload.

## Blocking
- None.

## Non-blocking
- The new test uses a real-looking Gmail address; consider switching to a placeholder to avoid embedding PII.
- Minor nit: `time.Now()` is called multiple times inside `buildAuthFromFileData`; using a single `now` would make the flow a bit cleaner.
- `synthesizer.SynthesizeAuthFile` may return multiple auth entries (e.g., gemini-cli multi-project virtual auths). `buildAuthFromFileData` currently takes only the first entry—worth documenting if intentional.

## Test plan
- Not run locally (no checkout per request). Suggested:

luispater (changes requested):
## Blocking

## Recommendation

PR #2900, opened 2026-04-19
feat(utls): add chatgpt.com to Cloudflare TLS fingerprint bypass

Aminatorex:
## Summary

Around 2026-04-10 Cloudflare started enforcing **managed challenges** on `chatgpt.com`, returning 403 HTML challenge pages to Go's default TLS client. ChatGPT-account Codex routing (the OAuth backend at `chatgpt.com/backend-api/codex`) stopped working — every request 403'd with `auth_unavailable: no auth available`. This was not an IP block: `curl` from multiple geographic locations all hit the same wall, and JWTs were valid. The signal was purely TLS-layer (JA3 fingerprint).

The proxy already had `refraction-networking/utls` Chrome fingerprinting wired up for `api.anthropic.com` (added previously for the same Cloudflare class of problem). This PR extends that same protection to `chatgpt.com`.

## Changes

1. **`internal/runtime/executor/helps/utls_client.go`** — rename the host map from `anthropicHosts` to `utlsProtectedHosts` to reflect that it's no longer Anthropic-specific, and add `chatgpt.com` to it.
2. **`internal/runtime/executor/codex_executor.go`** — switch the four `NewProxyAwareHTTPClient` call sites (`HttpRequest`, `Execute`, `executeCompact`, `ExecuteStream`) to `NewUtlsHTTPClient`, the existing helper that uses the utls Chrome fingerprint for protected hosts and falls back to the standard transport for everything else.

Net change: **+9 / -7** across two files. No new dependencies. No behavior change for hosts outside the protected list.

## Verification

Tested after local deploy on 2026-04-10 (`cli-proxy-api:v6.9.22-cftls` image built from this patch + v6.9.22):

- Non-stream `gpt-5-codex` via `/v1/chat/completions`: **200 OK**, ~1.16 s
- Streamed `gpt-5-codex` via `/v1/chat/completions`: **200 OK**, clean SSE, ~1.87 s
- High-reasoning non-stream request: **200 OK** with reasoning + message output

Image has been running in production on two hosts (local laptop and a Contabo VPS) for 9 days serving Factory Droid traffic without regressions. Re-verified today (2026-04-19) with same image: `gpt-5.3-codex` returns expected output through the BYOK chain.

## Test plan

- [ ] Build from this branch and confirm it compiles cleanly (`go build ./...`)
- [ ] Manual: hit `chatgpt.com` from a host where Cloudflare's managed challenge is currently triggering (without this patch you should see 403 challenge HTML; with this patch you should see the normal API response)
- [ ] Regression: confirm `api.anthropic.com` traffic still uses utls (existing protection unchanged)
- [ ] Regression: confirm a non-protected HTTPS host (e.g., `api.openai.com`) still routes through the standard transport

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -75,7 +75,7 @@ func (e *CodexExecutor) HttpRequest(ctx context.Context, auth *cliproxyauth.Auth
 	if err := e.PrepareRequest(httpReq, auth); err != nil {
 		return nil, err
 	}
-	httpClient := helps.NewProxyAwareHTTPClient(ctx, e.cfg, auth, 0)
+	httpClient := helps.NewUtlsHTTPClient(e.cfg, auth, 0)
 	return httpClient.Do(httpReq)
 }
 
@@ -141,7 +141,7 @@ func (e *CodexExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, re
 		AuthType:  authType,
 		AuthValue: authValue,
 	})
-	httpClient := helps.NewProxyAwareHTTPClient(ctx, e.cfg, auth, 0)
+	httpClient := helps.NewUtlsHTTPClient(e.cfg, auth, 0)
 	httpResp, err := httpClient.Do(httpReq)
 	if err != nil {
 		helps.RecordAPIResponseError(ctx, e.cfg, err)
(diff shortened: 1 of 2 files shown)


moonshardx:
The problem this PR addressed has resurfaced: ChatGPT-account Codex (`chatgpt.com/backend-api/codex`) is again being blocked by Cloudflare's managed challenge (HTTP 403, surfacing as `auth_unavailable: no auth available`). It reproduces across multiple egress IPs with valid JWTs, so it's the same TLS/JA3-fingerprint signal described here, not an IP or auth issue.

On current `main`, `chatgpt.com` still isn't covered: `utls_client.go`'s protected-host map only has `api.anthropic.com`, and `codex_executor.go` still uses `NewProxyAwareHTTPClient` at all four sites.

Filed #3626 to track this and re-request the fix. Would it be possible to revisit merging this (or an equivalent), since the utls mechanism already exists and is validated for Anthropic? Thanks for the great work on this project.

PR #2905, opened 2026-04-19
feat(access): per-api-key model allowlist with glob matching

mtuckerb:
## Summary

Adds enforcement for per-client-key model ACLs at the request middleware layer. The companion dashboard already stores `allowedModels` per UserApiKey and attempts to push them to `PUT /v0/management/api-keys`, but until now there was nowhere upstream for them to land — the structured payload was rejected and dashboards silently fell back to the legacy plain-array form, so no key was ever actually restricted.

This change adds:

**Config**
- `APIKeyPolicy{Key, AllowedModels}` and `SDKConfig.APIKeyPolicies`
- `SDKConfig.APIKeyDefaultPolicy` (`"allow-all"` | `"deny-all"`; defaults to `allow-all` so existing deployments are unaffected)
- `IsModelAllowedForKey` supports `path.Match` globs and tolerates a single leading `<prefix>/` segment so policies stay portable across prefixed credentials

**Management API**
- `PutAPIKeys` accepts both the legacy string list and the structured `[{key, allowedModels}]` form, with `allowedModels` / `allowed-models` / `allowed_models` all recognized so existing dashboard payloads work unchanged
- `GetAPIKeys` exposes `api-key-policies` when any are configured

**Middleware**
- New `ModelACLMiddleware` extracts the targeted model from the JSON body (`model` field) or the Gemini-style `/v1beta/models/<model>:<action>` path, then enforces the per-key policy
- The body is read once and restored via `io.NopCloser` so downstream handlers still see the original request bytes
- Wired into both `/v1` and `/v1beta` groups directly after `AuthMiddleware`
- Bypasses any request that has no identified api key (preserving the legacy "no auth provider configured => allow" behavior) and any shape with no extractable model (listings, websocket upgrades) so discovery endpoints keep working

## Test plan

- [x] Unit tests for `IsModelAllowedForKey` covering allow-all, deny-all, empty allowed list semantics, exact matches, glob matches, and the prefixed-credential case
- [x] Middleware tests covering the allow / deny / unknown-key / Gemini path / list-endpoint / body-pass-through paths
- [x] Parser tests for `PutAPIKeys` covering the legacy plain list, the wrapped items form, and the structured form in all three case styles (`allowedModels` / `allowed-models` / `allowed_models`)

## Backwards compatibility

Default policy is `allow-all`; deployments that do not opt in by setting `APIKeyPolicies` or flipping `APIKeyDefaultPolicy` to `deny-all` see no behavior change. Existing dashboards that `PUT` a plain string list continue to work — the new structured form is additive.

### DIFF
--- a/internal/api/handlers/management/api_keys_payload_test.go
+++ b/internal/api/handlers/management/api_keys_payload_test.go
@@ -1,0 +1,120 @@
+package management
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestParseAPIKeysPayload(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name         string
+		body         string
+		wantKeys     []string
+		wantPolicies int // count only — full equality covered below for one case
+		wantOK       bool
+	}{
+		{
+			name:     "plain string array",
+			body:     `["sk-aaa","sk-bbb"]`,
+			wantKeys: []string{"sk-aaa", "sk-bbb"},
+			wantOK:   true,
+		},
+		{
+			name:     "plain string array dedupes and trims",
+			body:     `["sk-aaa","  sk-aaa  ","sk-bbb"]`,
+			wantKeys: []string{"sk-aaa", "sk-bbb"},
+			wantOK:   true,
+		},
+		{
+			name:     "wrapped items array",
+			body:     `{"items":["sk-aaa","sk-bbb"]}`,
+			wantKeys: []string{"sk-aaa", "sk-bbb"},
+			wantOK:   true,
+		},
+		{
+			name:         "structured array camelCase",
+			body:         `[{"key":"sk-narrow","allowedModels":["gpt-4o*"]},{"key":"sk-open"}]`,
+			wantKeys:     []string{"sk-narrow", "sk-open"},
+			wantPolicies: 1,
+			wantOK:       true,
+		},
+		{
+			name:         "structured array kebab-case",
+			body:         `[{"key":"sk-narrow","allowed-models":["gpt-4o*","claude-3-*"]}]`,
+			wantKeys:     []string{"sk-narrow"},
+			wantPolicies: 1,
+			wantOK:       true,
+		},
+		{
+			name:         "structured array snake_case",
+			body:         `[{"key":"sk-narrow","allowed_models":["gpt-4o*"]}]`,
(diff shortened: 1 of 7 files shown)


luispater (changes requested):
### Summary
Adds per-API-key model allowlists (config + management API) and enforces them via `ModelACLMiddleware` on `/v1` + `/v1beta`, with DoS hardening (bounded body inspection) and an atomic cached policy index. Test coverage is substantial and generally well-targeted.

### Blocking
- **Bypass via `/backend-api/codex` aliases:** The ACL middleware is wired into `/v1` and `/v1beta` only. The `/backend-api/codex` group (chatgpt_base_url compatible) also uses `AuthMiddleware(...)` but does not use `ModelACLMiddleware(...)`, so restricted keys can reach `/responses` (including websocket upgrades) without allowlist enforcement.
  - Suggested fix: add `codexDirect.Use(ModelACLMiddleware(cfgFn))` immediately after `codexDirect.Use(AuthMiddleware(...))` (and consider parity for any other AuthMiddleware-wrapped groups).
  - Suggested regression guard: a focused route-setup/integration test proving the alias path is protected when policies are configured.

- **Websocket restriction edge case under deny-all default:** `keyHasModelRestriction()` returns `false` when a matching policy row exists but `AllowedModels` is empty, even if `api-key-default-policy=deny-all`. That makes the websocket-upgrade path potentially more permissive than the JSON-body path in this configuration.
  - Suggested fix: treat “matching policy row with empty AllowedModels” as restricted when the default is deny-all, consistent with `IsModelAllowedForKey` semantics; add a test for this edge case.

### Non-blocking
- Management API response shape: PUT accepts `allowedModels`/`allowed-models`/`allowed_models`, but GET will emit `allowed-models` due to struct tags. Consider documenting this or emitting camelCase to match dashboard payload expectations.
- `parseAPIKeysPayload` clears via `[]` but rejects `{"items":[]}`; consider allowing wrapped-empty if any clients clear through the wrapped form.
- `SetAPIKeyPolicies` copies the policy slice but not nested `AllowedModels` slices; consider deep-copying if there’s any chance of post-call mutation of the input.

### Test plan
- `go test ./...`
- Add/extend tests to cover `/backend-api/codex` alias enforcement and the deny-all + empty-AllowedModels websocket case.

PR #2908, opened 2026-04-19
fix(codex): preserve prompt_cache_retention on Codex paths

### LINKED ISSUE #2373: Codex prompt caching misses on repeated chat-completions requests

ne0ark:
## Scope

Preserve the `prompt_cache_retention` field in Codex request bodies instead of stripping it before forwarding to the upstream API.

## Problem

All three Codex execution paths (Execute, ExecuteStream, and CodexWebsocketsExecutor) unconditionally delete `prompt_cache_retention` from the translated request body via `sjson.DeleteBytes`. This prevents the upstream Codex Responses API from honoring the client's cache retention preference.

## Fix

Remove the 4 `sjson.DeleteBytes(body, "prompt_cache_retention")` calls:
- `codex_executor.go` line 116 (ExecuteStream path)
- `codex_executor.go` line 355 (Execute path — streaming)
- `codex_executor.go` line 458 (Execute path — non-streaming with tokenizer)
- `codex_websockets_executor.go` line 191 (WebSocket path)

## Dependencies

None. Orthogonal to F-2.2 (continuity key stabilization).

## Tests Added

- `TestCodexExecutor_PreservesPromptCacheRetention` — verifies `prompt_cache_retention` survives through `cacheHelper`
- `TestCodexWebsocketsExecutor_PreservesPromptCacheRetention` — verifies `prompt_cache_retention` survives through `applyCodexPromptCacheHeaders`

## AGENTS.md Checklist

- [x] No `log.Fatal` in new code
- [x] English-only comments
- [x] New test file follows existing naming convention
- [x] Helper files stay under `internal/runtime/executor/helps/` (no new helpers needed)
- [x] Zero modifications to `internal/translator/**`
- [x] `gofmt -l` on changed files returns empty
- [x] Existing tests remain green

## Closes/References

Closes #2373

---

**Note:** This PR is opened from a fork. The org `pull_request`-triggered workflows (`pr-test-build`, `translator-path-guard`) require one-time maintainer approval. Local verification: `go test ./...` exits 0, `go build` succeeds, `git diff --name-only origin/dev...HEAD -- internal/translator/**` is empty, `gofmt -l` on changed files is empty.

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -113,7 +113,6 @@ func (e *CodexExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, re
 	body, _ = sjson.SetBytes(body, "model", baseModel)
 	body, _ = sjson.SetBytes(body, "stream", true)
 	body, _ = sjson.DeleteBytes(body, "previous_response_id")
-	body, _ = sjson.DeleteBytes(body, "prompt_cache_retention")
 	body, _ = sjson.DeleteBytes(body, "safety_identifier")
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body = normalizeCodexInstructions(body)
@@ -352,7 +351,6 @@ func (e *CodexExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.Au
 	requestedModel := helps.PayloadRequestedModel(opts, req.Model)
 	body = helps.ApplyPayloadConfigWithRoot(e.cfg, baseModel, to.String(), "", body, originalTranslated, requestedModel)
 	body, _ = sjson.DeleteBytes(body, "previous_response_id")
-	body, _ = sjson.DeleteBytes(body, "prompt_cache_retention")
 	body, _ = sjson.DeleteBytes(body, "safety_identifier")
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body, _ = sjson.SetBytes(body, "model", baseModel)
@@ -455,7 +453,6 @@ func (e *CodexExecutor) CountTokens(ctx context.Context, auth *cliproxyauth.Auth
 
 	body, _ = sjson.SetBytes(body, "model", baseModel)
 	body, _ = sjson.DeleteBytes(body, "previous_response_id")
-	body, _ = sjson.DeleteBytes(body, "prompt_cache_retention")
 	body, _ = sjson.DeleteBytes(body, "safety_identifier")
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body, _ = sjson.SetBytes(body, "stream", false)
--- a/internal/runtime/executor/codex_executor_cache_test.go
+++ b/internal/runtime/executor/codex_executor_cache_test.go
@@ -11,6 +11,7 @@ import (
 	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/executor"
 	sdktranslator "github.com/router-for-me/CLIProxyAPI/v6/sdk/translator"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 )
 
 func TestCodexExecutorCacheHelper_OpenAIChatCompletions_StablePromptCacheKeyFromAPIKey(t *testing.T) {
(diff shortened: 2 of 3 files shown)


luispater:
Did you fully test the code after the AI edits?

Are you certain that `prompt_cache_retention` is a supported parameter in Codex?

PR #2914, opened 2026-04-19
feat(access): per-api-key model allowlist with glob matching (v2, review fixes)

mtuckerb:
## Summary

Re-submission of #2911 against \`dev\` (auto-retarget CI closed the original).
Includes the three fixes requested in the review.

The dashboard already stores \`allowedModels\` per UserApiKey and tries to push them to \`PUT /v0/management/api-keys\`, but upstream had nowhere to land them — so no key was ever actually restricted. This PR wires enforcement at the request-middleware layer.

### Feature (original commit)

- \`APIKeyPolicy{Key, AllowedModels}\` on \`SDKConfig\`; \`APIKeyDefaultPolicy\` (\`allow-all\` default, \`deny-all\` opt-in) so existing deployments are unaffected.
- \`IsModelAllowedForKey\` uses \`path.Match\` globs and tolerates a single leading \`<prefix>/\` segment so policies stay portable across prefixed credentials.
- \`PutAPIKeys\` accepts both legacy string list and structured \`[{key, allowedModels}]\`; \`GetAPIKeys\` surfaces \`api-key-policies\`.
- \`ModelACLMiddleware\` extracts the target model from the JSON body or the Gemini-style \`/v1beta/models/<model>:<action>\` path and enforces the per-key policy; wired into both \`/v1\` and \`/v1beta\` groups after \`AuthMiddleware\`.

### Review fixes (second commit)

1. **DoS hardening in ModelACLMiddleware.** Body read is now capped at 10 MiB via \`io.LimitReader\` (and short-circuited on \`Content-Length\`); oversize requests get HTTP 413 instead of silently buffering. Non-size read errors fail closed with 400 rather than bypassing the policy check.
2. **O(1) policy lookup.** \`APIKeyPolicies\` is backed by an \`atomic.Pointer[map[string]*APIKeyPolicy]\` cache, lazily built via CompareAndSwap, safe under concurrent reads. Adds \`SetAPIKeyPolicies\` (atomic replace + invalidate) and \`InvalidatePolicyIndex\`.
3. **Policy rotation/deletion sync.** \`PatchAPIKeys\` keeps \`APIKeyPolicies\` in sync on rename (via \`old\`/\`new\`) and rotation (via \`index\`/\`value\`); \`DeleteAPIKeys\` drops matching policy rows so a revoked credential cannot leave its allowlist behind; \`PutAPIKeys\` routes through \`SetAPIKeyPolicies\` for atomic cache refresh.

### Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./...\` green on \`dev\` + this branch
- [x] New middleware tests: \`OversizedBodyRejectedWith413\`, \`OversizedBodyRejectedViaContentLength\`
- [x] New cache tests: setter-rebuilds, in-place-edit+invalidate, nil-safe, concurrent reads, defensive-copy, empty-clears
- [x] New handler tests: rename-via-old-new, rename-via-index-value, rename-with-existing-target, append-new, delete-by-value, delete-by-index, last-row-clears, PUT-rebuilds-policies

Replaces #2911.


luispater (changes requested):
### Summary
Adds per-API-key model allowlisting (glob matching) with management API support for structured `allowedModels` payloads, and enforces it via a new `ModelACLMiddleware` on `/v1` and `/v1beta`.

### Blocking
- **Alias bypass:** The middleware is wired on `/v1` and `/v1beta` only. The `/backend-api/codex/*` alias group also uses `AuthMiddleware` but is not protected, so restricted keys can bypass allowlists by calling the alias endpoints.
- **Body buffering on non-JSON:** `extractRequestedModel` reads and buffers the full request body for any POST/PUT/PATCH to look for `"model"`. When ACLs are enabled, this will also buffer non-JSON/multipart requests like `POST /v1/images/edits`, which can be large. Suggest gating body reads to JSON content-types (or a small allowlist of JSON routes), and/or extracting `model` from multipart form fields for image edit requests.
- **CI:** `retarget` is failing because a separate PR (#2911) from the same head branch cannot be retargeted to `dev` (duplicate base/head). This likely needs the duplicate PR cleaned up so the check goes green.

### Non-blocking
- `GET /v0/management/api-keys` returns `api-key-policies` using `allowed-models` (kebab-case). If the dashboard expects `allowedModels`, consider adding an alias or documenting the response shape.
- Minor nit: the test case name `"empty plain array fails"` expects success (empty clears keys).

### Test plan
- CI `build` ✅
- No local checkout/tests run (review-only via `gh` as requested).

sccgaochengyi:
We actively need this feature. Per-key model allowlisting is exactly what we've been looking for to enforce user-tier access controls (e.g. restricting intern/contractor keys to deepseek-*/glm-*/qwen* while granting full-time employees broader access).

The implementation looks solid — the middleware-based enforcement, glob matching, atomic policy cache, and backward-compatible payload formats all align well with how we'd want to consume this.

That said, I'd echo the two blocking items from @luispater's review:

1. **Codex route bypass** — wiring ModelACLMiddleware onto /backend-api/codex/* is critical; without it the allowlist is trivially circumvented.
2. **/v1/models filtering** — returning the full model list to a restricted key is a poor UX (users see models they can't use). Ideally the models endpoint would filter based on the calling key's policy.

Both seem like small deltas on top of this PR. Hoping we can get these addressed and land this soon.

mtuckerb:
Addressed the remaining reviewer-facing gap around model-list UX. Restricted API keys now get policy-filtered model listings for both OpenAI/Claude-shaped /v1/models and Gemini-shaped /v1beta/models, while unrestricted keys retain the full catalog and deny-all/no-policy keys receive an empty filtered list.

Also added focused coverage for:
- restricted keys only seeing allowed models
- unrestricted keys seeing all models
- deny-all unknown keys seeing no models
- /v1/models response filtering behavior

Verification:
- go test ./internal/api
- go test ./...
- GitHub checks are green (build, close-when-agents-md-changed, ensure-no-translator-changes).

mtuckerb:
Addressed the two new bot-flagged gaps from the latest pass (commit 9369170b).

P1 — Content-Type bypass (codex, model_acl.go):
Several model-executing routes (ChatCompletions, Completions, Responses, ClaudeMessages) parse the body as JSON regardless of Content-Type, so the previous default-allow branch let a restricted key send Content-Type: text/plain with a JSON body containing a disallowed model and skip enforcement. The Content-Type switch now only branches off the JSON path for content types we explicitly know are not JSON (multipart/form-data). Everything else — including missing, unparseable, or misleading Content-Type values — runs the same peek+gjson inspection as application/json. Bodies without a JSON model field still pass through (gjson returns ok=false), and the existing 10 MiB cap continues to bound per-request memory.

P2 — Silent duplicate-key dedupe in PUT /api-keys (codex, config_lists.go):
parseAPIKeysPayload was silently skipping duplicate structured rows, which let payloads like `[{"key":"sk-x"},{"key":"sk-x","allowedModels":["gpt-4o*"]}]` widen access by dropping the restricted row. Duplicates are now rejected (HTTP 400). Legacy plain-list inputs still go through dedupeKeyList because they have no per-row metadata to lose.

Tests:
- TestModelACLMiddleware_NonJSONContentTypeWithJSONBodyStillEnforces — Content-Type: text/plain + JSON body with disallowed model → 403.
- TestModelACLMiddleware_OctetStreamWithoutModelAllowed — opaque octet-stream body without a JSON model field → 200.
- TestModelACLMiddleware_NonJSONContentTypeAllowedThrough — kept the test name (intent is right) but rewrote the body to a small 4 KiB opaque buffer; the prior 10 MiB-payload-returns-200 assertion was the exact bypass this round closes.
- TestParseAPIKeysPayload_StructuredDuplicateKeyRejected — duplicate structured rows → ok=false.

Verification:
- go test ./internal/api/...
- go test ./...
- GitHub checks green (build, close-when-agents-md-changed, ensure-no-translator-changes).

mtuckerb:
Round-4 fixes pushed at commit `9369170b9fac499c8846ebba42cccf3072a94371` addressing the two new codex findings against the prior commit:

**P1 — Content-Type bypass of model ACL** (model_acl.go)
The previous switch had a `default: return "", false, nil` arm for unknown content types. Several model-executing handlers (`ChatCompletions`, `Completions`, `Responses`, `ClaudeMessages`) parse the body via `c.GetRawData()` and decode JSON regardless of `Content-Type`, so a restricted key could send `Content-Type: text/plain` with a JSON body containing a disallowed model and bypass enforcement. The dispatch now branches off the JSON path only for `multipart/form-data`; everything else — including missing, unparseable, or deliberately misleading Content-Type values — runs the same peek + gjson inspection as `application/json`. Bodies that genuinely contain no JSON `model` field still pass through (gjson returns ok=false), and the existing 10 MiB cap continues to bound per-request memory.

**P2 — Silent duplicate-key dedupe in `PUT /api-keys` structured payload** (config_lists.go)
`parseAPIKeysPayload` previously skipped duplicate keys in the structured form, which silently dropped `allowedModels` from a later duplicate row and let the key fall back to the deployment default policy (usually allow-all). Duplicates now return `ok=false` → HTTP 400, so callers fix their payload instead of getting an invisible permission widening. The legacy plain-string form continues to dedupe via `dedupeKeyList` because it carries no per-row metadata to lose.

**Tests**
- `TestModelACLMiddleware_NonJSONContentTypeWithJSONBodyStillEnforces`: `text/plain` + JSON body with disallowed model → 403.
- `TestModelACLMiddleware_OctetStreamWithoutModelAllowed`: opaque octet-stream body without a JSON `model` field still returns 200.
- `TestModelACLMiddleware_NonJSONContentTypeAllowedThrough` updated to the new contract — small opaque body returns 200; the prior "large opaque payload returns 200" assertion was the exact bypass this round closes.
- `TestParseAPIKeysPayload_StructuredDuplicateKeyRejected`: duplicate structured rows are rejected.

Verification: `go test ./...` passes; PR checks (build, close-when-agents-md-changed, ensure-no-translator-changes) are green.

mtuckerb:
Red-team self-review of `9369170b` shipped at `e0e8d1f9`. No new blocking findings; added five adversarial test cases that the round-4 unit tests didn't cover.

**Coverage added:**
- `TestModelACLMiddleware_EmptyContentTypeWithDisallowedJSONStillEnforces`: no `Content-Type` header + JSON body with disallowed model → 403. Catches a regression of `parseRequestMediaType("") → ""` ever short-circuiting back to fail-open.
- `TestModelACLMiddleware_UnparseableContentTypeStillEnforces`: malformed Content-Type that `mime.ParseMediaType` rejects also lands in the default branch and still enforces.
- `TestModelACLMiddleware_MixedCaseContentTypeStillDispatches`: `APPLICATION/JSON` survives `ToLower`. A future refactor that drops the lowercase step surfaces here instead of as a silent ACL bypass.
- `TestParseAPIKeysPayload_StructuredWrappedDuplicateKeyRejected`: duplicate keys in the `{items:[...]}` wrapped structured form get rejected — same loop as the bare array form, but pin the wrapped form too because both shapes go through it.
- `TestParseAPIKeysPayload_StructuredDuplicateKeyAfterTrimRejected`: whitespace-padded duplicates are also rejected; trim-then-dedupe is correctly ordered.

Pre-existing concerns I noticed but they're not in scope for this PR — flagging for awareness:
- No global `MaxBytesReader` on the multipart code path. The 1 MiB `modelACLMultipartMemoryCap` is the in-memory portion of the form parse; `mime/multipart` will happily disk-spool larger file fields. Predates this PR.
- `extractModelFromMultipartBody` allow-throughs on a malformed multipart body. Intentional — the route handler should reject — but worth keeping in mind if the multipart route surface grows.

Tests pass against the new commit; PR checks are green.

mtuckerb:
ready for another review!

mtuckerb:
@luispater anything else you need from me on this ?

mtuckerb:
Quick status update: I re-checked the latest inline review state and found three newer codex comments after my prior red-team note that I hadn't accounted for yet:

- image endpoints can default to `gpt-image-2` when `model` is omitted, so the ACL needs to treat that implicit default as the requested model;
- malformed `multipart/form-data` currently fail-opens in extraction and needs to reject/fail safe;
- Amp provider aliases under `/api/provider/...` need the same model ACL middleware as `/v1`, `/v1beta`, and `/backend-api/codex`.

I'm treating those as the remaining blockers and will push another fix/test pass before asking for re-review again.

mtuckerb:
Status update — head is now `f1d4e740`.

**What landed since the last visible activity here:**
- Caught up with `upstream/main` (197 commits through `167edfec`) including the v6→v7 module bump
- Merged in the round-4 ACL followups I added to the branch (`1d8305ce`, `9369170b`, `e0e8d1f9`)

**Status of your Apr 25 review:**
- ✅ Alias bypass — `ModelACLMiddleware` is now wired on `/backend-api/codex/*` alongside `/v1` and `/v1beta` (commit `9f805087`)
- ✅ Body buffering on non-JSON — `extractRequestedModel` dispatches on `Content-Type`: multipart routes the dedicated form-parse path (1 MiB memory cap), and the round-4 followup (`9369170b`) made everything else go through JSON peek inspection so `Content-Type: text/plain` with a JSON body can't bypass the check
- ✅ `allowedModels` JSON tag — output is camelCase; the parser still accepts all three case styles on input
- ✅ Test name nit — renamed to "empty plain array clears keys"

**Still open (not asking for re-review yet):**
- Three items I flagged in my May 12 comment remain: implicit `gpt-image-2` default on image endpoints, multipart fail-open on malformed bodies, and ACL coverage for Amp provider aliases under `/api/provider/...`
- A new P2 from Codex today on `f1d4e740` against `server.go`: the `client_version` branch of `unifiedModelsHandler` returns `openaiHandler.OpenAIModels(c)` directly, bypassing `filterModelsForAPIKey`. Will address with the other listing-consistency work

I'll push the remaining fixes as another atomic commit + tests, then ping for re-review. Wanted to flag the rebase + give you a clean picture of where things stand in case you wanted to peek before then.

PR #2923, opened 2026-04-20
Fix auth upload priority propagation

xj999:
## Summary
- reuse the auth file synthesizer in the management upload path so uploaded OAuth files populate runtime auth attributes consistently
- preserve top-level auth-file fields such as `priority` during runtime auth rebuilds
- add a regression test covering uploaded auth files with a top-level `priority`

## Test Plan
- `docker compose build cli-proxy-api`

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -31,6 +31,7 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/misc"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/registry"
 	"github.com/router-for-me/CLIProxyAPI/v6/internal/util"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/watcher/synthesizer"
 	sdkAuth "github.com/router-for-me/CLIProxyAPI/v6/sdk/auth"
 	coreauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
Summary
- Adds an `insecure` flag to `POST /v0/management/api-call` to optionally skip TLS certificate verification, and threads it into `apiCallTransport`.
- Improves observability for APICall failures by logging at `Error` and returning the underlying `Do` error string.

Blocking
- Please document the new request field `insecure` in the APICall godoc (Request JSON section + curl examples). The comment currently lists `auth_index`, `method`, `url`, `header`, `data`, but not `insecure`.
- Please add a focused unit test for the insecure path (e.g., `apiCallTransport(..., true)` returns a `*http.Transport` with `TLSClientConfig != nil` and `InsecureSkipVerify == true`, and the `false` path keeps it unset). This is security-sensitive behavior and worth locking down.

Non-blocking / Suggestions
- Consider cloning `TLSClientConfig` before mutating it when it’s already non-nil (defensive against shared configs from `proxyutil.BuildHTTPTransport`).
- Returning `errDo.Error()` to the client could leak internal details; if that’s a concern, consider keeping `error: "request failed"` and adding a separate sanitized `details` field (or only include details when a debug flag is enabled).
- PR metadata: the `auto-retarget-main-pr-to-dev / retarget` check is failing; may need a rerun or base-branch adjustment depending on your workflow.

Test plan (not run locally)
- `go test ./...`
- `go test ./internal/api/handlers/management -run TestAPICallTransport`

PR #2932, opened 2026-04-21
.1616750161180714:2fb665accdd718b4eb14a70501355dff_69e6d82a59f68fc20e58829f.69e6ddf559f68fc20e5882e1.69e6ddf590f0ee6098651891:Trae CN.T(2026/4/21 10:16:21)

hxf582125:
修改API调用失败时的日志级别以更准确反映问题严重性，同时在返回给客户端的错误信息中包含具体错误原因，便于问题排查

### DIFF
--- a/internal/api/handlers/management/api_tools.go
+++ b/internal/api/handlers/management/api_tools.go
@@ -193,8 +193,8 @@ func (h *Handler) APICall(c *gin.Context) {
 
 	resp, errDo := httpClient.Do(req)
 	if errDo != nil {
-		log.WithError(errDo).Debug("management APICall request failed")
-		c.JSON(http.StatusBadGateway, gin.H{"error": "request failed"})
+		log.WithError(errDo).Error("management APICall request failed")
+		c.JSON(http.StatusBadGateway, gin.H{"error": "request failed: " + errDo.Error()})
 		return
 	}
 	defer func() {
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
### Summary
This PR updates the management `POST /v0/management/api-call` endpoint to:
- Promote request-failure logging from `Debug` to `Error`
- Include the underlying `http.Client.Do` error string in the JSON response on transport failure
- Add an `insecure` request flag to skip TLS certificate verification for that outbound request

### Blocking
- **Security/observability:** logging `errDo` at `Error` and echoing `errDo.Error()` back to the caller can leak sensitive details (full URL strings, internal proxy hosts) and may be overly noisy for an admin “curl-like” endpoint. Suggest using a stable `error` code plus a separate `message` field, and consider redacting sensitive parts and/or keeping the log level at `Warn`/`Debug`.
- **`insecure` footgun:** `InsecureSkipVerify` is inherently high-risk. Please document `insecure` in the endpoint’s request JSON docs (the handler comment), add an explicit warning/audit log when it’s enabled, and consider a server-side gate/allowlist (or config) so production deployments can disable it.

### Non-blocking
- Add tests for `apiCallTransport(auth, insecure=true)` to verify it sets `TLSClientConfig.InsecureSkipVerify` for both proxy and direct transports, and remains unset when `insecure=false`.
- CI: `retarget` is failing; the log indicates it tried to retarget PR #2933 and failed because a dev PR already exists for the same head branch—likely a duplicate PR / automation artifact.

### Test plan
- Not run locally (reviewed via `gh` diff + checks only). Suggested: `go test ./...`.

PR #2939, opened 2026-04-21
fix(codex): backfill streaming response output

stringer07:
## Summary
Patch the Codex HTTP streaming executor so the final response.output is reconstructed from prior response.output_item.done events when upstream leaves the completed output empty.

This aligns ExecuteStream with the existing non-stream fallback and fixes Amp deep mode clearing the final reply after streaming.

## Testing
- go test ./internal/runtime/executor -run focused codex stream fallback tests
- go test ./internal/runtime/executor
- go test ./internal/translator/codex/...
- go build -o /tmp/cliproxyapi-server ./cmd/server
- manual isolated Amp deep verification against a test CPA instance

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -59,9 +59,6 @@ func patchCodexCompletedOutput(eventData []byte, outputItemsByIndex map[int64][]
 		return eventData
 	}
 
-	completedDataPatched := eventData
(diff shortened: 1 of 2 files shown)


stringer07:
Addressed in b6781d69. The helper now builds the output array once and applies a single SetRawBytes call.

luispater (approved):

PR #2941, opened 2026-04-21
fix(gemini): preserve native Gemini/Vertex compatibility in mixed proxy chains

MaribelHearm:
## Summary
- map OpenAI Responses `developer` messages into Gemini `systemInstruction`
- strip relay-only `includeServerSideToolInvocations` flags before forwarding Gemini / Vertex requests upstream
- document the recommended Google-style provider routes for native Gemini / Vertex chains in all README variants

## Why
This helps mixed chains such as `Vertex -> CLIProxyAPI -> Gemini-compatible client` stay closer to the native Gemini protocol surface and avoids forwarding fields/roles that Google / Vertex may reject with `400` errors.

## Notes
- includes regression tests for the request sanitizer and developer-role mapping
- local Go tests could not be executed in this environment because the `go` binary is not installed in the current shell


MaribelHearm:
Closing this PR in favor of issue #2942.

The local patch works for my use case, but this PR touches `internal/translator/**` and currently fails the repository path guard for external PRs. I’m keeping the fix locally and leaving the upstream compatibility discussion in the issue instead.

PR #2962, opened 2026-04-22
feat(codex): inject image_generation tool + route aliases for Codex CLI image generation

MoYeRanqianzhi:
## Problem

When Codex CLI connects through CPA, image generation is unavailable. The root cause is a **client-side auth gate** in Codex:

```rust
// codex-rs/core/src/session/turn_context.rs
fn image_generation_tool_auth_allowed(auth_manager: Option<&AuthManager>) -> bool {
    matches!(auth_manager.and_then(AuthManager::auth_mode), Some(AuthMode::Chatgpt))
}
```

Only `AuthMode::Chatgpt` (OAuth login) enables the built-in `image_generation` tool. When using API key auth via CPA, the tool is never included in requests — the upstream never receives it, so image generation silently fails.

Additionally, Codex CLI with OAuth auth sends requests to `{chatgpt_base_url}/codex/responses`, but CPA only exposes `/v1/responses`. This path mismatch prevents using the `chatgpt_base_url` config to redirect Codex CLI traffic through CPA while keeping `AuthMode::Chatgpt`.

The recent `e935196d` (GPT-Image-2 via `/v1/images/generations`) solves the **standalone Images API** use case but does **not** address the Codex CLI integration path — Codex CLI uses the Responses API (`/responses`), not the Images API.

## Solution

Two targeted changes, no modifications to Codex CLI source code:

### 1. Tool injection in Codex executor (`codex_executor.go`)

`ensureImageGenerationTool()` appends `{"type":"image_generation","output_format":"png"}` to the `tools` array if not already present. Applied to all three execution paths: `Execute`, `executeCompact`, `ExecuteStream`.

The function is idempotent — if the client already includes `image_generation`, the existing tool config is preserved.

### 2. Route aliases for Codex CLI direct access (`server.go`)

Added `/backend-api/codex/responses` routes mapping to the same OpenAI Responses API handlers as `/v1/responses`. This allows Codex CLI to connect via `chatgpt_base_url` config while keeping `AuthMode::Chatgpt` (image_generation tool enabled client-side).

## Verified

Successfully tested end-to-end: Codex CLI (v0.122.0, gpt-5.4 xhigh) -> CPA -> upstream, `$imagegen` skill generating images through the reverse proxy.

> **Note**: This PR was developed with the assistance of Claude Opus 4.6 (Anthropic), including root cause analysis of the Codex auth gating mechanism, implementation, and test coverage.

## Test plan

- [x] `TestEnsureImageGenerationTool_NoTools` — creates tools array when missing
- [x] `TestEnsureImageGenerationTool_ExistingToolsWithoutImageGen` — appends to existing tools
- [x] `TestEnsureImageGenerationTool_AlreadyPresent` — preserves existing config
- [x] `TestEnsureImageGenerationTool_EmptyToolsArray` — handles empty array
- [x] `TestEnsureImageGenerationTool_WebSearchAndImageGen` — coexists with other built-in tools
- [x] All 94 existing executor tests pass (no regressions)
- [x] Manual E2E: Codex CLI `$imagegen` through CPA proxy

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -353,6 +353,15 @@ func (s *Server) setupRoutes() {
 		v1.POST("/responses/compact", openaiResponsesHandlers.Compact)
 	}
 
+	// Codex CLI direct route aliases (chatgpt_base_url compatible)
+	codexDirect := s.engine.Group("/backend-api/codex")
+	codexDirect.Use(AuthMiddleware(s.accessManager))
+	{
+		codexDirect.GET("/responses", openaiResponsesHandlers.ResponsesWebsocket)
+		codexDirect.POST("/responses", openaiResponsesHandlers.Responses)
+		codexDirect.POST("/responses/compact", openaiResponsesHandlers.Compact)
+	}
+
 	// Gemini compatible API routes
 	v1beta := s.engine.Group("/v1beta")
 	v1beta.Use(AuthMiddleware(s.accessManager))
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -180,6 +180,7 @@ func (e *CodexExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, re
 	body, _ = sjson.DeleteBytes(body, "safety_identifier")
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body = normalizeCodexInstructions(body)
+	body = ensureImageGenerationTool(body)
 
 	url := strings.TrimSuffix(baseURL, "/") + "/responses"
 	httpReq, err := e.cacheHelper(ctx, from, url, req, body)
@@ -326,6 +327,7 @@ func (e *CodexExecutor) executeCompact(ctx context.Context, auth *cliproxyauth.A
 	body, _ = sjson.SetBytes(body, "model", baseModel)
 	body, _ = sjson.DeleteBytes(body, "stream")
 	body = normalizeCodexInstructions(body)
+	body = ensureImageGenerationTool(body)
 
 	url := strings.TrimSuffix(baseURL, "/") + "/responses/compact"
 	httpReq, err := e.cacheHelper(ctx, from, url, req, body)
@@ -420,6 +422,7 @@ func (e *CodexExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.Au
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body, _ = sjson.SetBytes(body, "model", baseModel)
 	body = normalizeCodexInstructions(body)
+	body = ensureImageGenerationTool(body)
 
 	url := strings.TrimSuffix(baseURL, "/") + "/responses"
 	httpReq, err := e.cacheHelper(ctx, from, url, req, body)
@@ -818,6 +821,24 @@ func normalizeCodexInstructions(body []byte) []byte {
 	if !instructions.Exists() || instructions.Type == gjson.Null {
 		body, _ = sjson.SetBytes(body, "instructions", "")
 	}
+	return body
+}
+
+var imageGenToolJSON = []byte(`{"type":"image_generation","output_format":"png"}`)
+var imageGenToolArrayJSON = []byte(`[{"type":"image_generation","output_format":"png"}]`)
+
+func ensureImageGenerationTool(body []byte) []byte {
+	tools := gjson.GetBytes(body, "tools")
+	if !tools.Exists() || !tools.IsArray() {
+		body, _ = sjson.SetRawBytes(body, "tools", imageGenToolArrayJSON)
+		return body
+	}
+	for _, t := range tools.Array() {
+		if t.Get("type").String() == "image_generation" {
+			return body
+		}
+	}
+	body, _ = sjson.SetRawBytes(body, "tools.-1", imageGenToolJSON)
 	return body
 }
 
--- a/internal/runtime/executor/codex_executor_imagegen_test.go
+++ b/internal/runtime/executor/codex_executor_imagegen_test.go
@@ -1,0 +1,89 @@
+package executor
+
+import (
+	"testing"
+
+	"github.com/tidwall/gjson"
+)
+
+func TestEnsureImageGenerationTool_NoTools(t *testing.T) {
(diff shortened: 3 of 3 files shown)


MoYeRanqianzhi:
Regarding the `CountTokens` suggestion from @gemini-code-assist:

After analyzing the code, this change is **unnecessary and would be a no-op**:

1. **`CountTokens` is a local estimation** — it never sends requests to upstream. Actual usage is reported by the upstream `response.completed` event via `helps.ParseCodexUsage()`.

2. **The `image_generation` built-in tool has no `name`, `description`, or `parameters` fields.** `countCodexInputTokens()` (line 624-642) only extracts these three fields from each tool for token counting. Since `{"type":"image_generation","output_format":"png"}` contains none of them, injecting it would add exactly **0 tokens** to the count — the result is identical with or without the call.

While the suggestion looks consistent on the surface, the actual token count would not change. Skipping the injection here avoids unnecessary work in the counting path.

kvokka:
Brilliant work, thank you!

gaozuo:
We really do need this tool call!

gaoren002:
我将对sub提交一系列改动，重点参考了本PR

luispater (approved):

pnd280:
Please provide an option to disable this tool injection. It doesnt work in any other harnesses but Codex. I keep getting empty responses.

kvokka:
```bash
curl -v -H "Authorization: Bearer $MY_CLI_LLM_PROXY_API_KEY" -H 'Content-Type: application/json' http://host.docker.internal:8317/v1/images/generations -d '{
    "model": "gpt-image-2",
    "prompt": "simple blueprint of a mouse",
    "n": 1,
    "size": "1024x1024"
  }'
  ```
  
  works for me

pnd280:
> ```shell
> curl -v -H "Authorization: Bearer $MY_CLI_LLM_PROXY_API_KEY" -H 'Content-Type: application/json' http://host.docker.internal:8317/v1/images/generations -d '{
>     "model": "gpt-image-2",
>     "prompt": "simple blueprint of a mouse",
>     "n": 1,
>     "size": "1024x1024"
>   }'
> ```
> 
> works for me

Sorry for not being clear in my previous comment, directly hit the API works just fine, but what I meant was both github copilot chat and claude code (CLI) were failed to use the tool with no errors or warnings whatsoever. 
image

Edit: if this tool was only meant to be injected into Codex CLI, why is it still available in other harnesses? 
image

gaoren002:
> > ```shell
> > curl -v -H "Authorization: Bearer $MY_CLI_LLM_PROXY_API_KEY" -H 'Content-Type: application/json' http://host.docker.internal:8317/v1/images/generations -d '{
> >     "model": "gpt-image-2",
> >     "prompt": "simple blueprint of a mouse",
> >     "n": 1,
> >     "size": "1024x1024"
> >   }'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > works for me
> 
> Sorry for not being clear in my previous comment, directly hit the API works just fine, but what I meant was both github copilot chat and claude code (CLI) were failed to use the tool with no errors or warnings whatsoever. image
> 
> Edit: if this tool was only meant to be injected into Codex CLI, why is it still available in other harnesses? image

Actually only Codex support image generation in harness , so the injection in other harness won't take effect

pnd280:
> > > ```shell
> > > curl -v -H "Authorization: Bearer $MY_CLI_LLM_PROXY_API_KEY" -H 'Content-Type: application/json' http://host.docker.internal:8317/v1/images/generations -d '{
> > >     "model": "gpt-image-2",
> > >     "prompt": "simple blueprint of a mouse",
> > >     "n": 1,
> > >     "size": "1024x1024"
> > >   }'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > works for me
> > 
> > 
> > Sorry for not being clear in my previous comment, directly hit the API works just fine, but what I meant was both github copilot chat and claude code (CLI) were failed to use the tool with no errors or warnings whatsoever. image
> > Edit: if this tool was only meant to be injected into Codex CLI, why is it still available in other harnesses? image
> 
> Actually only Codex support image generation in harness , so the injection in other harness won't take effect

What do you mean by "wont take effect"? Both claude code and copilot literally list `image_gen.imagegen` as a callable tool. Either via explicit or implicit, if the llm calls the tool, it will break the turn.

kvokka:
Then you should share that in the issues for the tools that you are using, since this PR works for me in Codex, and, after some fixing, in [OpenClaw](https://github.com/openclaw/openclaw/issues/71114); WIP on PR for that

MoYeRanqianzhi:
To clarify the scope of this PR and address @pnd280's concern:

## What this PR solves

This PR specifically targets **Codex CLI image generation through CPA**. The problem: Codex CLI gates the built-in `image_generation` tool behind `AuthMode::Chatgpt` (OAuth only). When connecting through CPA, the auth mode is `ApiKey`, so the tool is never included in requests — image generation silently fails.

The solution has two parts:
1. **Tool injection** (`ensureImageGenerationTool` in `codex_executor.go`) — ensures `image_generation` is present in the tools array sent to the Codex upstream
2. **Route aliases** (`/backend-api/codex/responses`) — allows Codex CLI to set `chatgpt_base_url` in `~/.codex/config.toml` pointing to CPA, keeping `AuthMode::Chatgpt` on the client side (which enables the tool locally) while routing requests through CPA

## Why it affects other harnesses

@pnd280 raises a valid point. The injection lives in `CodexExecutor`, which handles **all** requests routed to Codex upstream — not just Codex CLI. When Claude Code or GitHub Copilot sends requests through CPA to a Codex model, the `image_generation` tool gets injected too. Those clients see the tool as callable, the LLM may attempt to invoke it, but the client has no handler for `image_generation_call` responses → broken turn.

This is a real side effect. The `image_generation` tool is a **Codex-native built-in** — only the Codex upstream and Codex CLI understand it. Other clients (Claude Code, Copilot, etc.) do not.

## Possible improvements

A config toggle or User-Agent–based scoping (only inject when the client is `codex-tui/*`) would limit the injection to Codex CLI. This could be addressed in a follow-up — happy to hear the maintainers' preferred approach.

PR #2990, opened 2026-04-23
fix(openai-compat): fix invalid reasoning_effort caused by model alias lookup mismatch

Ethan-W20:
## Problem

When an `openai-compatibility` provider is configured with a model alias and a restricted `thinking.levels` capability (e.g. `[low, medium, high]`), a Claude client request with `thinking.budget_tokens > 24576` fails with an upstream HTTP 400.

The Claude → OpenAI translator maps large thinking budgets to `reasoning_effort="xhigh"`, but the user-declared `thinking.levels` is silently ignored along the way, so `xhigh` reaches the upstream unchanged:

```
API Error: 400 {"error":{"message":"... 'input': 'xhigh', 'ctx': {'expected': \"'none', 'low', 'medium' or 'high'\"}}...","type":"Bad Request"}}
```

## Root Cause

The conductor (`sdk/cliproxy/auth/conductor.go`) rewrites `req.Model` from the client-visible alias to the upstream name before dispatching to executors. `thinking.ApplyThinking` in `openai_compat_executor.go` was then looking up the registry with that rewritten upstream name — but the model was registered under the alias (see `sdk/cliproxy/service.go`). The lookup returned `nil`, routing through the user-defined fallback path that skips validation/clamping, letting `xhigh` pass straight through.

## Fix

Use `helps.PayloadRequestedModel(opts, req.Model)` — which returns the client-visible model name stored in request metadata (populated by `sdk/api/handlers/handlers.go`) — as the model argument to `ApplyThinking`. This mirrors the existing pattern already used for `ApplyPayloadConfigWithRoot` in the same executor (`helps/payload_helpers.go` explicitly documents `requestedModel` as *"the client-visible model name before alias resolution"*).

Three call sites updated: `Execute`, `ExecuteStream`, `CountTokens`. `PayloadRequestedModel` falls back to `req.Model` when metadata is absent, so SDK embedders and other execution paths are unaffected; no new config option is introduced.

## Test

`TestOpenAICompatExecutor_AliasResolvedThinkingLookup` simulates the real-world scenario (Claude client → OpenAI-compat upstream with alias and `thinking.levels`). Verified:

- ✅ Passes against this fix
- ✅ Fails against pre-fix code (verified via `git stash` revert) with `reasoning_effort="xhigh"` appearing in the forwarded body
a4b42058-8eda-474d-95c1-5dc8d94b5c53


luispater (approved):
### Summary
Fixes OpenAI-compat thinking application to use the client-requested model (alias) for registry lookups, preventing invalid `reasoning_effort` values (e.g. `xhigh`) from leaking to upstreams when `thinking.levels` is restricted. Adds a focused regression test for the Claude → OpenAI-compat alias path.

### Key points
- Aligns `thinking.ApplyThinking(...)` with the existing `requestedModel := helps.PayloadRequestedModel(...)` pattern already used for `ApplyPayloadConfigWithRoot`.
- Updates all relevant call sites (`Execute`, `ExecuteStream`, `CountTokens`) with a safe fallback to `req.Model` when metadata is absent.
- Regression test asserts clamping behavior (`xhigh` → `high`) when the model is registered under the alias with `Thinking.Levels` configured.

### Test plan
- Suggested: `go test ./...`
- Focused: `go test ./internal/runtime/executor -run TestOpenAICompatExecutor_AliasResolvedThinkingLookup`

PR #3010, opened 2026-04-24
feat(codex): implement Codex ↔ OpenAI Chat Completions protocol conversion

SmallNew2003:
## Summary
This PR implements automatic conversion between Codex Responses API and OpenAI Chat Completions API, enabling Codex clients to use providers like Kimi Coding that only support Chat Completions API.

## Changes
- Added `Protocol` field to `CodexKey` config (values: `codex`, `openai-chat`)
- Created new translator package: `internal/translator/openai/codex/`
  - `ConvertCodexRequestToOpenAI`: Converts Responses API requests to Chat Completions format
    - Maps input array to messages with role/content
    - Handles tools with function name normalization (`.` → `_`)
    - Supports multimodal content, function calls, tool choices
  - `ConvertOpenAIResponseToCodex`: Converts Chat Completions streaming responses to Responses format
  - `ConvertOpenAIResponseToCodexNonStream`: Converts non-streaming responses
  - Function name normalization with reverse mapping
- Registered translator for bidirectional Codex ↔ OpenAI conversion
- Updated `CodexExecutor` to route requests based on Protocol field:
  - `Execute`: uses `/chat/completions` when `protocol=openai-chat`
  - `ExecuteStream`: same routing logic
  - `CountTokens`: uses correct translator format
- Added comprehensive unit tests
- Fixed config structure: restored `Cloak` and `ExperimentalCCHSigning` fields to `ClaudeKey`

## Testing
- All unit tests pass
- Build verified in Docker with Go 1.26

## Related Frontend PR
- router-for-me/Cli-Proxy-API-Management-Center#233

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/refraction-networking/utls v1.8.2
 	github.com/sirupsen/logrus v1.9.3
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
+	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.18.0
 	github.com/tidwall/sjson v1.2.5
 	github.com/tiktoken-go/tokenizer v0.7.0
@@ -29,6 +30,11 @@ require (
 	golang.org/x/sync v0.18.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	gopkg.in/yaml.v3 v3.0.1
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
 )
 
 require (
@@ -94,6 +100,6 @@ require (
 	golang.org/x/arch v0.8.0 // indirect
 	golang.org/x/sys v0.38.0 // indirect
 	golang.org/x/text v0.31.0 // indirect
-	google.golang.org/protobuf v1.34.1 // indirect
+	google.golang.org/protobuf v1.34.1
 	gopkg.in/ini.v1 v1.67.0 // indirect
 )
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -393,10 +393,12 @@ type ClaudeKey struct {
 	// ExcludedModels lists model IDs that should be excluded for this provider.
 	ExcludedModels []string `yaml:"excluded-models,omitempty" json:"excluded-models,omitempty"`
 
-	// Protocol specifies the API protocol to use for this Codex provider.
-	// "codex" (default): native Responses API support.
-	// "openai-chat": use Chat Completions API with automatic conversion.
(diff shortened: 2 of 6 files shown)


luispater (changes requested):
### Summary
This PR adds a `protocol` option to `CodexKey` and tries to let the Codex executor call either the native Codex Responses endpoint (`/responses`) or OpenAI Chat Completions (`/chat/completions`) with automatic conversion.

### Blocking issues

1) **Non-stream `Execute` is broken for `protocol: openai-chat`**
- `internal/runtime/executor/codex_executor.go` keeps forcing `stream: true` and then scans for a terminal `{"type":"response.completed"}` event.
- Chat Completions streaming chunks don’t contain `type=response.completed`, so this path will never hit the completion branch and will fall through to the 408 error (`stream closed before response.completed`).
- Fix direction: for `protocol=openai-chat`, send a non-stream request (`stream:false`) to `/chat/completions` (similar to `OpenAICompatExecutor.Execute`) and translate the non-stream JSON response back; or implement a proper aggregation of Chat Completions SSE into a final response before calling `TranslateNonStream`.

2) **Token counting is incorrect for `protocol: openai-chat`**
- `CountTokens` switches `to` to `"openai"` but still uses Codex token counting (`countCodexInputTokens`), which expects `instructions/input/tools` and won’t count `messages`/`tools.function.*` correctly.
- Fix direction: keep `to="codex"` for counting, or add a Chat Completions token counter (like `OpenAICompatExecutor.CountTokens`).

3) **`cacheHelper` may inject invalid fields into Chat Completions payload**
- `cacheHelper` is called with `from=opts.SourceFormat` (typically `openai-response`) and may set `prompt_cache_key` on the translated Chat Completions JSON.
- Some upstreams are strict and may reject unknown top-level fields.
- Fix direction: gate this behavior on `!useOpenAIChat` (or make it depend on `to`/endpoint).

4) **New translator `internal/translator/openai/codex` has a streaming state bug and likely isn’t wired into the intended path**
- `ConvertOpenAIResponseToCodex` sets `*param` to `&map[string]any{...}` but type-asserts `(*param).(map[string]any)`, which will panic if invoked.
- Also, the Codex executor’s source format is usually `openai-response`, so the new `Codex ↔ OpenAI` translator won’t be selected; the existing `openai-response ↔ openai` translators will be used instead. That means the new dot-to-underscore normalization and role mapping won’t actually affect the `protocol=openai-chat` path.

### Non-blocking suggestions
- In `ConvertCodexRequestToOpenAI`, tool names are normalized but `function_call.name` is not—consider normalizing those too for consistency.
- Add a minimal config example/documentation for `codex.protocol`.

### Tests to add
- A unit/integration test covering `protocol=openai-chat` for non-stream `Execute` (ensuring it returns a valid non-stream response, not 408).
- A streaming unit test for `ConvertOpenAIResponseToCodex` to catch the `param` panic and validate emitted events.

PR #3025, opened 2026-04-24
Add Vertex prompt caching for Claude-compatible Gemini requests

RunhuaHuang:
## Summary

- add Vertex native `cachedContents` proxy routes for `/v1beta/cachedContents` and the Amp Google provider alias
- add automatic Vertex prompt cache creation/reuse for Claude-compatible requests translated to Gemini
- report Vertex cache reads back to Claude-compatible clients as `cache_read_input_tokens`, with `input_tokens` excluding cached prompt tokens
- keep the new stream usage parser and Claude cache accounting scoped to Vertex so the regular Gemini path remains unchanged

## Root Cause

Claude-compatible Agent requests were translated to Gemini/Vertex generateContent calls, but Anthropic `cache_control` hints were not bridged to Vertex native context caching. Separately, Vertex stream responses can carry empty or wrapped `usageMetadata`; the previous shared Gemini parser could publish an empty usage record before the terminal usage chunk. Finally, Gemini `cachedContentTokenCount` was not mapped back to Claude `cache_read_input_tokens`, so Agent clients could not see cache hits even when Vertex returned them.

## Implementation Notes

- For Vertex service-account requests, stable prompt fields (`system_instruction`, `tools`, `toolConfig`) are used to create a Vertex cachedContent resource with a 1 hour TTL.
- The generated request then references `cachedContent` and removes the cached prompt fields from the live request to avoid duplicating context.
- Cache reuse is keyed by the actual Vertex cache create body plus project/location/model/auth, avoiding repeated cache creation for identical stable prompts.
- The `/v1beta/cachedContents` proxy is Vertex-only and normalizes `models/<model>` into the full Vertex publisher model resource.
- Claude-compatible usage mapping is gated by a Vertex context marker: `input_tokens = promptTokenCount - cachedContentTokenCount`, `cache_read_input_tokens = cachedContentTokenCount`.

## Validation

- `go test ./internal/translator/gemini/claude ./internal/runtime/executor/helps ./internal/runtime/executor ./internal/api ./internal/api/modules/amp ./sdk/api/handlers/gemini`

Manual Vertex validation against a live Vertex service-account credential:

```text
5 identical /v1/messages?beta=true Agent requests
cachedContents before: 8
cachedContents after: 9
new Vertex cache: projects/710749625912/locations/global/cachedContents/6999339392896598016
new cache totalTokenCount: 40902

round1 usage: input_tokens=20, output_tokens=588, cache_read_input_tokens=40902
round2 usage: input_tokens=20, output_tokens=212, cache_read_input_tokens=40902
round3 usage: input_tokens=20, output_tokens=109, cache_read_input_tokens=40902
round4 usage: input_tokens=20, output_tokens=223, cache_read_input_tokens=40902
round5 usage: input_tokens=20, output_tokens=183, cache_read_input_tokens=40902
```

This confirms the cache resource is created by Vertex, later requests reuse the same resource without creating additional caches, and the Claude-compatible Agent response receives the cache-read token count returned by Vertex.

### DIFF
--- a/internal/api/modules/amp/routes.go
+++ b/internal/api/modules/amp/routes.go
@@ -330,5 +330,7 @@ func (m *AmpModule) registerProviderAliases(engine *gin.Engine, baseHandler *han
 		v1betaAmp.GET("/models", geminiHandlers.GeminiModels)
 		v1betaAmp.POST("/models/*action", fallbackHandler.WrapHandler(geminiHandlers.GeminiHandler))
 		v1betaAmp.GET("/models/*action", geminiHandlers.GeminiGetHandler)
+		v1betaAmp.Any("/cachedContents", fallbackHandler.WrapHandler(geminiHandlers.GeminiCachedContentsHandler))
+		v1betaAmp.Any("/cachedContents/*name", fallbackHandler.WrapHandler(geminiHandlers.GeminiCachedContentsHandler))
 	}
 }
--- a/internal/api/modules/amp/routes_test.go
+++ b/internal/api/modules/amp/routes_test.go
@@ -211,6 +211,8 @@ func TestRegisterProviderAliases_V1BetaRoutes(t *testing.T) {
 	}{
 		{"/api/provider/google/v1beta/models", http.MethodGet},
 		{"/api/provider/google/v1beta/models/generateContent", http.MethodPost},
+		{"/api/provider/google/v1beta/cachedContents", http.MethodPost},
+		{"/api/provider/google/v1beta/cachedContents/test-cache", http.MethodGet},
 	}
 
 	for _, tc := range v1betaPaths {
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -369,6 +369,8 @@ func (s *Server) setupRoutes() {
 		v1beta.GET("/models", geminiHandlers.GeminiModels)
 		v1beta.POST("/models/*action", geminiHandlers.GeminiHandler)
 		v1beta.GET("/models/*action", geminiHandlers.GeminiGetHandler)
+		v1beta.Any("/cachedContents", geminiHandlers.GeminiCachedContentsHandler)
+		v1beta.Any("/cachedContents/*name", geminiHandlers.GeminiCachedContentsHandler)
 	}
 
 	// Root endpoint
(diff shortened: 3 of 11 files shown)


luispater (changes requested):
## Summary
This PR adds Vertex-native `cachedContents` proxy routes (`/v1beta/cachedContents...` and Amp google alias), plus automatic Vertex prompt cache creation/reuse for Claude-compatible requests translated to Gemini/Vertex. It also improves Vertex stream usage parsing and maps `cachedContentTokenCount` back to Claude usage as `cache_read_input_tokens` (with `input_tokens` excluding cached prompt tokens).

## Blocking
- The new cachedContents proxy cannot reliably handle the *full* Vertex resource name returned by Vertex (e.g. `projects/<project>/locations/<loc>/cachedContents/<id>`) when the client uses that value in the URL path.
  - Today, `geminiCachedContentNameFromPath()` always prefixes `cachedContents/`, so a request like:
    - `GET /v1beta/cachedContents/projects/.../locations/.../cachedContents/...`
    becomes `cachedContents/projects/...` internally, and `geminiCachedContentsURL()` produces an invalid upstream URL.
  - Suggested fix: treat `projects/...` names as already-canonical (do not add the `cachedContents/` prefix). Add a regression test covering:
    1) Vertex create response returns `name: "projects/.../locations/.../cachedContents/..."`  
    2) A follow-up GET/DELETE using that exact `projects/...` name proxies to `.../v1/projects/.../locations/.../cachedContents/...`.

## Non-blocking
- When `cacheName` starts with `projects/`, consider selecting a Vertex auth whose `metadata.project_id` + `metadata.location` match that name (helps when multiple Vertex creds exist).
- `vertexPromptCaches` is a process-wide `sync.Map` with TTL checks but no singleflight/periodic cleanup; consider deduping concurrent creations and preventing long-term growth from expired entries.

## Test plan
- CI `pr-test-build` is green. Please add/adjust tests for full resource-name paths as noted above.
- Suggested local: `go test ./internal/api ./sdk/api/handlers/gemini ./internal/runtime/executor/... ./internal/translator/gemini/claude`.

_Note: `translator-path-guard` failing is expected given translator changes (ignored for this review recommendation as requested)._

PR #3037, opened 2026-04-25
fix: apply tool call reordering to HTTP /v1/responses path

lmikey9550:
## Summary

Apply the same `repairResponsesToolCallsArray` logic to the HTTP `/v1/responses` path that already exists for WebSocket.

MiniMax upstream rejects requests where `function_call_output` items appear before their corresponding `function_call` items in the `input` array (2013 error). The WebSocket handler already fixes this via `repairResponsesWebsocketToolCalls()`, but the HTTP path was missing the fix entirely.

### `sdk/api/handlers/openai/openai_responses_handlers.go`
- `Responses()`: call `repairResponsesWebsocketToolCalls()` at the top before dispatching to streaming/non-streaming handlers
- `Compact()`: same fix applied before executing the compact request

### `sdk/api/handlers/openai/openai_responses_websocket_toolcall_repair.go`
- Add `httpResponsesSessionKey()` to derive a deterministic session key from `model`+`previous_response_id` for per-session tool call cache lookups in HTTP context

## Behavior

After this change, the HTTP `/v1/responses` path behaves identically to WebSocket: the `input` array is reordered so each `function_call_output` immediately follows its `function_call` before the request is forwarded to upstream.

This fixes the 2013 error when using MiniMax models via the HTTP path through cc-switch or direct API calls.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -257,7 +257,7 @@ func (h *OpenAIResponsesAPIHandler) Responses(c *gin.Context) {
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
 	// This fixes 2013 errors from MiniMax上游 which rejects out-of-order function_call / function_call_output.
-	rpcSessionKey := httpResponsesSessionKey(rawJSON)
+	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
 	rawJSON = repairResponsesWebsocketToolCalls(rpcSessionKey, rawJSON)
 
 	// Check if the client requested a streaming response.
@@ -299,7 +299,7 @@ func (h *OpenAIResponsesAPIHandler) Compact(c *gin.Context) {
 	}
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
-	rpcSessionKey := httpResponsesSessionKey(rawJSON)
+	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
 	rawJSON = repairResponsesWebsocketToolCalls(rpcSessionKey, rawJSON)
 
 	c.Header("Content-Type", "application/json")
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -256,7 +256,7 @@ func (h *OpenAIResponsesAPIHandler) Responses(c *gin.Context) {
 	}
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
-	// This fixes 2013 errors from MiniMax上游 which rejects out-of-order function_call / function_call_output.
+	// This fixes 2013 errors from MiniMax upstream which rejects out-of-order function_call / function_call_output.
 	// If no session key is available (first-turn, no prevRespID, no header), use the no-cache
 	// variant to still reorder without polluting a shared cache namespace.
 	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -257,8 +257,14 @@ func (h *OpenAIResponsesAPIHandler) Responses(c *gin.Context) {
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
 	// This fixes 2013 errors from MiniMax上游 which rejects out-of-order function_call / function_call_output.
+	// If no session key is available (first-turn, no prevRespID, no header), use the no-cache
+	// variant to still reorder without polluting a shared cache namespace.
 	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
-	rawJSON = repairResponsesWebsocketToolCalls(rpcSessionKey, rawJSON)
+	if rpcSessionKey != "" {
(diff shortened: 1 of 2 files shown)


lmikey9550:
Thanks for the thorough review. Addressing each point:

**P1 - Session key collision**: Fixed in commit 90ce8654. Added `repairResponsesHTTPNoCache` which calls `reorderResponsesToolCallsNoCache` (a dedicated no-cache reorder) when no session key is available. This performs true reordering without any shared state, preventing both 2013 errors and cross-user cache pollution.

**P1 - Output-before-call reordering**: Fixed in commit 1c84cce7. The previous `repairResponsesHTTPNoCache` reused `repairResponsesToolCallsArray` which only filters, preserving original order. The new `reorderResponsesToolCallsNoCache` explicitly repositions outputs to immediately follow their call regardless of original order, handling [output, call], [call, output], multiple outputs per call, and orphaned outputs.

**P2 - HTTP cache key per-turn growth**: Noted. This is a secondary concern about cache lifecycle management on the HTTP path. The current design (per-request session key from model+prevRespID) creates one-off sessions that expire via TTL cleanup. A proper fix would require either (a) explicit cache invalidation when HTTP responses complete, or (b) scoped identity from upstream auth context. Happy to address this in a follow-up if maintainers consider it necessary for merge.

lmikey9550:
Thanks for the follow-up.

**P1 - X-Client-Request-Id client-controlled cache poisoning**: Noted. The  path is there for consistency with WebSocket, but you're right that a client can pick any value for  and collide with another client's cache if s match. For cc-switch this is not a concern (it doesn't send this header), and for your use case the  path is the primary multi-turn mechanism. I'm open to restricting the header path to known-safe values or removing it entirely for the HTTP fallback, but wanted to confirm with maintainers whether the header path is intentional for HTTP. Will await maintainer guidance on whether to narrow this.

**P2 - Per-request cache session growth**: Acknowledged. The HTTP path does not have the retain/release pattern that WebSocket has, so each distinct  creates a session that lives until TTL expiry. Under high-traffic multi-tenant deployments with many distinct previous_response_id values, this could cause  to scan an ever-growing map. This is a valid secondary concern and a follow-up could either (a) use an LRU/NRD-based cache instead of per-session maps, or (b) add HTTP-specific session lifecycle handling. I want to keep this PR focused on the 2013 fix; we can address cache growth separately.

luispater (commented):
Summary:
This PR applies the existing WebSocket tool-call repair to the HTTP `/v1/responses` and `/v1/responses/compact` handlers, and adds an HTTP session-key derivation plus a no-cache fallback for first-turn requests. This should reduce MiniMax “2013” upstream rejections caused by mismatched / out-of-order `function_call` and `function_call_output` items.

Key findings:
- Good scope: the HTTP path now mirrors the WebSocket pre-forward “repair” step.
- The session-key logic correctly prefers request/session identity headers (via the same logic as WebSocket) and falls back to `model + previous_response_id` hashing.
- The no-cache fallback when there is no stable identity (no headers + no `previous_response_id`) is a good safety measure to avoid cross-user cache pollution.

Non-blocking nits / suggestions:
- Please keep comments English-only: change `MiniMax上游` to `MiniMax upstream`.
- `reorderResponsesToolCallsNoCache()` appends orphaned outputs by iterating a map, which is nondeterministic; consider preserving original order for any remaining orphan outputs to keep forwarded payloads stable.
- (Optional) For consistency with `repairResponsesToolCallsArray()`, consider dropping `function_call` items without `call_id` in the no-cache reorder path.

Test plan:
- Not run locally (per request). CI checks appear green on this PR.

lmikey9550:
Thanks for the review. All suggestions addressed in commit 42598010:

- ✅ Changed `MiniMax上游` → `MiniMax upstream` (English-only comments)
- ✅ `reorderResponsesToolCallsNoCache`: now drops tool calls without `call_id` (consistent with `repairResponsesToolCallsArray`), and orphan outputs are now appended in sorted `call_id` order for stable output
- Build succeeds, all changes pushed

luispater (approved):
## Summary

This PR extends the existing WebSocket tool-call reordering logic to the HTTP `/v1/responses` and `/v1/responses/compact` handlers. It fixes MiniMax upstream 2013 errors caused by `function_call_output` items appearing before their corresponding `function_call` items in the `input` array.

## What changed

1. `Responses()` and `Compact()` in `openai_responses_handlers.go` now call the repair logic before dispatching to upstream.
2. `httpResponsesSessionKey()` derives a session key from request headers (preferred) or `model + previous_response_id` (fallback), returning empty when neither is available to avoid cross-user cache pollution.
3. `repairResponsesHTTPNoCache()` + `reorderResponsesToolCallsNoCache()` provide a stateless reorder path for first-turn requests where no stable session identity exists.

## Assessment

The implementation is sound:
- The session key derivation correctly mirrors WebSocket behavior and avoids the cross-tenant cache collision identified during iterative review.
- The no-cache fallback still performs reordering (fixing the 2013 error) without shared state (preventing data leakage).
- Edge cases are handled: empty payloads, missing `input` arrays, orphaned outputs, and tool calls without `call_id`.
- CI checks pass (build green).

## Non-blocking suggestions

1. **Tests**: No unit tests cover the new functions (`httpResponsesSessionKey`, `repairResponsesHTTPNoCache`, `reorderResponsesToolCallsNoCache`). Adding table-driven tests for key scenarios (first-turn no headers, multi-turn with prevRespID, orphaned outputs, empty input) would strengthen confidence.
2. **Commit hygiene**: The 6-commit iterative history could be squashed to 1-2 logical commits before merge.
3. **Observability**: Consider a debug-level log when tool calls are dropped due to missing `call_id`, to aid troubleshooting.

## Verdict

Approve. The change is well-scoped, addresses a real upstream compatibility issue, handles security concerns (cross-user cache pollution), and all CI checks pass.

---

PR #3042, opened 2026-04-25
fix(usage): defer retried auth failures

gaozuo:
## Summary

This PR aligns usage reporting with the final logical request outcome across auth/model retries.

Executors can publish usage for each provider attempt, while the auth conductor may retry another auth/model after an intermediate failure and eventually return success to the client. Before this change, those retried internal failures could be recorded as user-visible failed usage rows even when the logical request succeeded.

The fix is request-scoped and provider-agnostic:

- Retrying callers install a request-scoped deferred-failure context.
- `UsageReporter` defers failed usage records when that context is present.
- If a later attempt succeeds, the conductor discards the deferred failure.
- If all attempts fail, the conductor flushes the latest deferred failure.

This keeps usage accounting aligned with the final request outcome while preserving retry/auth failure telemetry in the conductor, hooks, and logs.

## Implementation notes

- `auth/conductor.go` does not import or depend on `usage`.
- `sdk/cliproxy/usage/manager.go` is unchanged.
- No business-layer package, database logic, provider-specific rule, or billing logic is involved.
- The deferral helper lives in `sdk/cliproxy/executor/context.go`, next to other executor request context helpers.
- The helper does not use `sync.Mutex`; it uses a small atomic state machine to safely coordinate stream goroutines and conductor finalization.

## Why

A restored local stack with a mixed Codex auth pool reproduced the issue with real requests:

- 5 real `/v1/responses` requests to `gpt-5.4`
- Client responses: all HTTP 200
- Usage rows before the fix: 16 total
  - 11 failed rows: `provider=codex`, `model=gpt-5.4`, `failed=true`, `error_status_code=500`, `total_tokens=0`
  - 5 successful rows for the same logical requests
- App logs for the same window showed intermediate retry/auth failures:
  - `rate limited: too many requests provider=codex model=gpt-5.4`

Those failed rows represented internal retry attempts, not final client-visible request failures.

## Validation

Unit tests:

```bash
go test ./sdk/cliproxy/executor ./sdk/cliproxy/usage ./sdk/cliproxy/auth ./internal/runtime/executor/helps
go test -race ./sdk/cliproxy/executor ./sdk/cliproxy/usage ./sdk/cliproxy/auth ./internal/runtime/executor/helps
```

Added coverage:

- Deferred failure helper has no effect without a deferral scope.
- Multiple deferred failures flush only the latest failure.
- Discard suppresses deferred failure publishing.
- Closed deferral scopes no longer defer failures.
- Concurrent defer/flush publishes at most once.
- Non-stream conductor retry success publishes only successful usage.
- Non-stream conductor all-fail publishes one failed usage.
- Non-stream canceled retry wait flushes the last failed usage.
- Stream conductor retry success publishes only successful usage.

Real local stack validation after the latest changes:

- Same restored local database and mixed auth pool: 22 auth entries, including 21 Codex auths.
- 5 real `/v1/responses` requests to `gpt-5.4`.
- Client responses: all HTTP 200.
- New usage rows after the fix: 5 total.
- Failed usage rows after the fix: 0.
- Failed 500 usage rows after the fix: 0.
- Logs still showed intermediate `rate limited: too many requests provider=codex model=gpt-5.4`, confirming auth retry telemetry remains visible while usage accounting no longer records retried internal failures as final failed usage.

### DIFF
--- a/internal/runtime/executor/helps/usage_helpers.go
+++ b/internal/runtime/executor/helps/usage_helpers.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/executor"
 	"github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/usage"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
@@ -72,7 +73,13 @@ func (r *UsageReporter) publishWithOutcome(ctx context.Context, detail usage.Det
 		}
 	}
 	r.once.Do(func() {
-		usage.PublishRecord(ctx, r.buildRecord(detail, failed))
+		record := r.buildRecord(detail, failed)
+		if failed && cliproxyexecutor.DeferFailure(ctx, func(ctx context.Context) {
+			usage.PublishRecord(ctx, record)
+		}) {
+			return
+		}
+		usage.PublishRecord(ctx, record)
 	})
 }
 
--- a/internal/runtime/executor/helps/usage_helpers_test.go
+++ b/internal/runtime/executor/helps/usage_helpers_test.go
@@ -1,11 +1,59 @@
 package helps
 
 import (
+	"context"
+	"fmt"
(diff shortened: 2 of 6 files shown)


gaozuo:
@claude review

gaozuo:
Closing this larger PR for now. I plan to follow up with a smaller, focused retry/usage fix based on the latest dev branch.

PR #3054, opened 2026-04-26
feat: add first-class DeepSeek V4 provider

kimpro12:
## Summary

Adds first-class DeepSeek provider support to CLIProxyAPI.

## Changes

- Add `deepseek-api-key` configuration support.
- Add DeepSeek auth synthesis and provider registration.
- Add dedicated DeepSeek executor using the OpenAI-compatible Chat Completions API.
- Add DeepSeek thinking support with `thinking.type` and `reasoning_effort`.
- Add static DeepSeek V4 model definitions.
- Add management API endpoints for DeepSeek API keys.
- Add model refresh compatibility so DeepSeek models survive remote catalog refreshes while upstream model catalogs catch up.

## Validation

- `gofmt -w .`
- `go test ./...`


luispater:
If a provider does not have its own CLI client, it is not qualified to be a first-class provider.

PR #3062, opened 2026-04-26
feat: estimate token usage when upstream response lacks usage field

LucasInsight:
## Summary

When LLM API responses omit the `usage` field, the proxy previously recorded zero tokens for all fields. This PR introduces tiktoken-based token estimation that fills in `InputTokens`, `OutputTokens`, and `TotalTokens` from the actual request/response content when usage is missing.

### Core: Estimation logic
- `EstimateUsage(model, requestBody, responseBody)` in `token_helpers.go`: Uses existing `tiktoken-go/tokenizer` to estimate tokens when upstream usage is absent
- `ExtractInputText()`: Extracts text content from request bodies across OpenAI, Claude, Gemini, Codex formats
- `ExtractOutputText()`: Extracts text content from response bodies (non-streaming and streaming) across all provider formats
- `countTextTokens()`: Helper that tokenizes a plain string with a model-appropriate codec

### Core: UsageReporter integration
- `UsageReporter.captureRequestBody()` / `captureResponseChunk()`: Captures raw payloads during logging, enabling estimation at publish time
- `UsageReporter.completeDetail()`: When `detail` is all-zero (no usage from upstream), automatically invokes `EstimateUsage()` to fill in values
- `publishWithOutcome()` now calls `completeDetail()` before publishing, so all executors benefit automatically
- `EnsurePublished()` now delegates to `publishWithOutcome()`, ensuring streaming paths also get estimation

### Executors: EnsurePublished coverage
Added `reporter.EnsurePublished(ctx)` to all streaming executor paths that previously only called `Publish` on usage chunks (which may never arrive):
- `aistudio_executor.go`
- `claude_executor.go`
- `codex_executor.go`
- `codex_websockets_executor.go`
- `gemini_cli_executor.go`
- `gemini_executor.go`
- `gemini_vertex_executor.go`
- `kimi_executor.go`

### Tests
- `TestEstimateUsageCountsRequestAndResponse`: Verifies both input and output tokens are estimated
- `TestExtractOutputTextProviderFormats`: Covers OpenAI, Claude, Gemini, Codex response formats
- `TestEstimateUsageEmptyResponseOutput`: Empty response content yields OutputTokens=0
- `TestUsageReporterDoesNotEstimateWhenUsageExists`: Real usage is never overwritten

## Design decisions

- **No marking**: Estimated values are not tagged as `estimated`; they fill `usage.Detail` transparently
- **Centralized**: Estimation happens in `publishWithOutcome()`, so every executor benefits without per-executor changes
- **Lazy capture**: Request/response bodies are captured as a side-effect of existing logging calls (zero overhead when logging is disabled)
- **Fallback only**: Estimation only triggers when `TotalTokens == 0 && InputTokens == 0 && OutputTokens == 0`; real usage is always preferred
- **Empty content**: If response content is empty (abnormal response), `OutputTokens` stays 0 per requirement

## Testing

```
go test ./internal/runtime/executor/helps/... — PASS
```

### DIFF
--- a/internal/runtime/executor/aistudio_executor.go
+++ b/internal/runtime/executor/aistudio_executor.go
@@ -308,6 +308,7 @@ func (e *AIStudioExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth
 					break
 				}
 			case wsrelay.MessageTypeStreamEnd:
+				reporter.EnsurePublished(ctx)
 				return false
 			case wsrelay.MessageTypeHTTPResp:
 				if !metadataLogged && event.Status > 0 {
@@ -322,6 +323,7 @@ func (e *AIStudioExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth
 					out <- cliproxyexecutor.StreamChunk{Payload: ensureColonSpacedJSON(lines[i])}
 				}
 				reporter.Publish(ctx, helps.ParseGeminiUsage(event.Payload))
+				reporter.EnsurePublished(ctx)
 				return false
 			case wsrelay.MessageTypeError:
 				helps.RecordAPIResponseError(ctx, e.cfg, event.Err)
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -488,6 +488,8 @@ func (e *ClaudeExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.A
 				helps.RecordAPIResponseError(ctx, e.cfg, errScan)
 				reporter.PublishFailure(ctx)
 				out <- cliproxyexecutor.StreamChunk{Err: errScan}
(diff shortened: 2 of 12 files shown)


luispater (changes requested):
### Summary
This PR improves usage accounting by estimating token usage when upstream responses omit `usage`, and ensuring a usage record is emitted for streaming paths even when no usage chunk arrives.

### Blocking
- Unbounded buffering risk: `UsageReporter` captures full request/response bodies (`responseBuf`) with no size cap, which can grow without limit on long streams and create OOM/DoS risk. Please add a hard cap and define behavior when truncated (skip estimation or accept undercounting).

### Non-blocking
- Consider flagging estimated vs upstream-reported usage to reduce ambiguity.

### Test plan
- Suggested: `go test ./internal/runtime/executor/helps/...`.

PR #3092, opened 2026-04-27
Bridge responses clients to chat-only providers

MagicFun1241:
Some OpenAI-compatible providers expose /chat/completions but return a plain 404 for /responses. The responses handler now accepts chat completion fallback output and converts it back to Responses API shape, while the auth manager retries only OpenAI-compatible API-key auths for the observed endpoint-missing body.

Constraint: NVIDIA integrate returns HTTP 404 with body '404 page not found' for /v1/responses while /v1/chat/completions is available.

Tested: go test ./sdk/api/handlers/openai ./sdk/cliproxy/auth
Tested: go build -o test-output ./cmd/server

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_compact_test.go
+++ b/sdk/api/handlers/openai/openai_responses_compact_test.go
@@ -20,6 +20,7 @@ type compactCaptureExecutor struct {
 	alt          string
 	sourceFormat string
 	calls        int
+	err          error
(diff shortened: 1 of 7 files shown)


luispater:
You can use `codex-api-key` to complete this task. The `openai-compatibility` is only used to support the Open Chat Completions endpoint.

PR #3101, opened 2026-04-27
fix(cliproxy): rotate auth on Anthropic 'out of extra usage' 400

can1357:
## What

When an OAuth Claude account runs out of its extra-usage budget, Anthropic returns:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"You're out of extra usage. Add more at claude.ai/settings/usage and keep going."}}
```

Right now the conductor treats anything tagged `invalid_request_error` as a request-shape problem and bails out instead of trying the next auth, so the user just sees the 400 even when they have other healthy accounts loaded.

## Change

Treat that specific message as a quota signal rather than a malformed-request signal, scoped narrowly to the "out of extra usage" / `claude.ai/settings/usage` wording so we don't accidentally swallow real 400s.

In `sdk/cliproxy/auth/conductor.go`:

- New `isOutOfExtraUsageMessage` / `isOutOfExtraUsageError` / `isOutOfExtraUsageResultError` helpers.
- `isRequestInvalidError` returns `false` for this case so the retry loop falls through to the next eligible auth.
- `applyAuthFailureState` and the per-model branch in `MarkResult` coerce the status to 429 when this hits, so the offending auth gets `Quota.Exceeded = true` plus the normal backoff cooldown and won't be re-picked until it recovers.

Net effect: hitting this 400 now rotates accounts the same way a 429 does.

## Verification

- `go build ./...`
- `go test ./sdk/cliproxy/auth/ -count=1`

Both clean.

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -117,6 +117,13 @@ type StoppableSelector interface {
 	Stop()
 }
 
+// AuthInvalidator is implemented by selectors that keep per-auth session
+// bindings. It lets the manager drop affinity for an auth that failed this
+// request without marking the auth itself unhealthy.
+type AuthInvalidator interface {
+	InvalidateAuth(authID string)
+}
+
 // Hook captures lifecycle callbacks for observing auth changes.
 type Hook interface {
 	// OnAuthRegistered fires when a new auth is registered.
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
## Summary
This PR makes Anthropic's "out of extra usage" (HTTP 400 `invalid_request_error`) retryable at the credential layer so the conductor can fall through to other auths.

## Blocking
- The PR description says this 400 is coerced to 429 and pushed through the existing quota cooldown path (`Quota.Exceeded=true` + backoff) via `applyAuthFailureState` / `MarkResult`. The current diff does **not** implement 429 coercion or any quota/cooldown state; it only clears session affinity (`InvalidateAuth`) and skips persistence/hook via `clearAffinityOnly`. Please either implement the described cooldown behavior, or update the PR description and confirm the intended behavior.
- Clearing session affinity alone does not prevent repeated selection on new requests (especially with `routing.strategy=fill-first`, which will keep trying the same auth first). If the goal is “rotate like 429 and avoid re-picking until recovery”, this should likely enter the 429 quota handling path (auth-level and model-level) or at least set a non-zero cooldown window.

## Non-blocking
- The message match predicate is somewhat broad (`out of extra usage` OR `claude.ai/settings/usage`). If you want this narrowly scoped, consider requiring both substrings (or parsing the JSON to confirm `invalid_request_error` plus the expected wording).
- Consider still calling `hook.OnResult` on the affinity-only path for observability, even if you intentionally avoid persisting auth/model error state.

## Test plan
- Not run locally.

PR #3104, opened 2026-04-28
fix(executor): ignore empty OpenAI-compatible streamed usage chunks

### LINKED ISSUE #3076: Streaming token usage always zero with OpenAI-compatible providers (DeepSeek)
[x] No, it's another issue.

---

**Describe the bug**

When using DeepSeek (or other OpenAI-compatible providers) in streaming mode, the usage statistics always show 0 tokens for all fields (input/output/reasoning/cached/total tokens). Latency is also incorrect — the record appears within ~1 second of the request, even while the stream is still in progress. Non-streaming mode works fine.

**Root cause**: When CPA injects `stream_options.include_usage: true` into the request, the upstream provider includes a `usage` object in **every** streaming chunk (with zero/null values in non-terminal chunks, and real values only in the final chunk alongside `finish_reason: "stop"`).

In `ParseOpenAIStreamUsage()`, the `usage` field is extracted from **any** chunk without checking whether it's the terminal chunk. The first chunk with `usage` (all zero) triggers `UsageReporter.Publish()`, and since `UsageReporter` uses `sync.Once`, the real token counts from the final chunk are silently discarded.

This affects both `openai_compat_executor.go` and `kimi_executor.go` (both call `ParseOpenAIStreamUsage`).

**Evidence** — DeepSeek API raw responses:

Non-streaming (usage at top level, correct):
```json
{"id":"...","object":"chat.completion","usage":{"prompt_tokens":10,"completion_tokens":64,"total_tokens":74,...}}
```

Streaming — intermediate chunk (no `usage` OR `usage: null`):
```json
data: {"choices":[{"delta":{"content":"..."},"finish_reason":null}]}
```

Streaming — final chunk (usage **only** here, alongside `finish_reason: "stop"`):
```json
data: {"choices":[{"delta":{"content":""},"finish_reason":"stop"}],"usage":{"prompt_tokens":10,"completion_tokens":39,"total_tokens":49,...}}
```

The fix is to check `choices.0.finish_reason` in `ParseOpenAIStreamUsage()` before accepting `usage`, so only terminal chunks contribute to token counts. This mirrors the existing `FilterSSEUsageMetadata()` logic used for Gemini providers.

---

**CLI Type**
openai-compatibility

**Model Name**
deepseek-v4-pro (applies to any OpenAI-compatible provider supporting `stream_options.include_usage`)

**LLM Client**
claude code

**Expected behavior**

Streaming usage statistics should show actual token counts (matching the final chunk's `usage` values), and latency should reflect the full streaming duration.

---

**OS Type**
- OS: Docker amd64

MuziIsabel:
## Summary
Fixes #3076.

This patch fixes streamed token accounting for OpenAI-compatible providers by ignoring empty `usage` payloads until a chunk contains real token counts.

## Root cause
When `stream_options.include_usage` is enabled, some providers emit either:
- `"usage": null`, or
- a zero-valued `usage` object

on intermediate chunks before the terminal usage arrives.

`ParseOpenAIStreamUsage()` treated those intermediate chunks as valid usage records. Because `UsageReporter` publishes through `sync.Once`, the first empty record won and the final real usage was discarded.

## Fix
- ignore streamed usage when `usage` is not an object
- ignore streamed usage when the parsed usage detail is entirely zero-valued
- reuse the shared OpenAI-style usage parser for stream usage
- add regression tests for:
  - `usage: null`
  - zero-valued `usage` object
  - terminal chunk with real usage

## Notes
- I reproduced this against DeepSeek via CLIProxyAPI locally.
- Go is not installed in my current Windows environment, so I could not run `go test` locally before opening this PR. The added unit tests are intended for CI verification.


MuziIsabel:
Addressed the Codex review note in 60d89e2. Kimi streamed requests now call \
eporter.EnsurePublished(ctx)\ at stream end, matching the OpenAI-compatible executor fallback so request accounting is preserved even when no non-zero usage chunk is emitted.

luispater (approved):
### Summary
Fixes streamed usage accounting for OpenAI-compatible streams by ignoring `usage: null` and all-zero `usage` objects until a real usage payload arrives, preventing an empty record from “winning” the `sync.Once` publish. Also ensures Kimi streaming still emits a request-count record even if no usage is ever parsed.

### Blocking
- None.

### Non-blocking / suggestions
- Add a stream-usage test case using `{"usage":{"input_tokens":...,"output_tokens":...}}` to cover the Responses-style field names in streaming.
- Note: `UsageReporter` still publishes the first non-zero stream usage chunk due to `sync.Once`; this PR fixes the “null/zero usage arrives before terminal usage” case, but it doesn’t change behavior for providers that emit incremental non-zero usage on earlier chunks.

### Test plan

luispater (approved):
## Summary

Two incremental improvements to streamed usage accounting: a `hasNonZeroTokenUsage` guard rejecting zero-valued usage objects, and a `reporter.EnsurePublished(ctx)` fallback in the Kimi executor matching established patterns.

## Non-blocking

- Zero-value guard assumes valid completed requests never have all-zero tokens. Reasonable but worth a comment.
- One test largely duplicates an existing test. Consider consolidating.

## Verdict

Approve. CI passes, no merge conflicts.

---

PR #3105, opened 2026-04-28
fix: remove client-controlled header from HTTP session key (P1 security)

lmikey9550:
## Summary

Remove the client-controlled header path (X-Client-Request-Id) from  to prevent cache poisoning attacks.

## Security Issue

The original PR introduced a session key derivation that preferred  header over the  hash. Since this header value is client-controlled, a malicious user could set it to collide with another user's cache entry, enabling cache poisoning / data leakage between tenants.

## Fix

- Removed the  call from 
- The function now only uses  hash
- When  is empty, returns empty string to trigger  (no shared cache, no cross-user pollution)

This mirrors the documented security intent: first-turn requests without a stable identity go through the no-cache path automatically, and the hash path provides reasonable isolation without trusting client headers.

## Testing

- Build succeeds
- No functional changes to the actual repair logic (only session key derivation affected)
- Existing tests for tool call reordering pass unchanged

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -257,7 +257,7 @@ func (h *OpenAIResponsesAPIHandler) Responses(c *gin.Context) {
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
 	// This fixes 2013 errors from MiniMax上游 which rejects out-of-order function_call / function_call_output.
-	rpcSessionKey := httpResponsesSessionKey(rawJSON)
+	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
 	rawJSON = repairResponsesWebsocketToolCalls(rpcSessionKey, rawJSON)
 
 	// Check if the client requested a streaming response.
@@ -299,7 +299,7 @@ func (h *OpenAIResponsesAPIHandler) Compact(c *gin.Context) {
 	}
 
 	// Repair tool call ordering before sending to upstream (mirrors WebSocket behavior).
-	rpcSessionKey := httpResponsesSessionKey(rawJSON)
+	rpcSessionKey := httpResponsesSessionKey(c.Request, rawJSON)
 	rawJSON = repairResponsesWebsocketToolCalls(rpcSessionKey, rawJSON)
 
 	c.Header("Content-Type", "application/json")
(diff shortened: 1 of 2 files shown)


lmikey9550:
Latest commit e26fa203 addresses both P1 and P2:

**P1 fixed**: `httpResponsesSessionKey` now only uses `model + previous_response_id` hash. No `websocketDownstreamSessionKey` call, no client-controlled headers. prevRespID is server-generated globally unique.

**P2 fixed**: HTTP path now uses `httpResponsesRepair()` with a dedicated `httpSessionCache` (isolated LRU, max 512 sessions, 2min TTL). No writes to WebSocket global caches.

Please re-scan from commit e26fa203.

luispater (changes requested):
## Summary

This PR:
- Hooks tool-call repair into the HTTP `/v1/responses` and `/v1/responses/compact` handlers before forwarding upstream.
- Removes client-controlled header usage from the HTTP session key derivation (good hardening vs cache poisoning).
- Introduces a dedicated bounded HTTP repair cache (TTL + max sessions) to avoid mixing HTTP and WebSocket caches.

## Blocking

1) `repairResponsesHTTPSingle()` does not currently guarantee correct ordering and can duplicate outputs.

In the second pass it does:
- For output items: appends them directly to `filtered`
- For call items: appends the call, then appends the cached output again (when `outputPresent[call_id]` is true)

That means for an input like `[function_call_output, function_call]` the output remains before the call, and a second copy may be appended after the call. This undermines the stated goal (“ensure outputs immediately follow calls”) and could still trigger upstream validation failures.

Suggested fix direction:
- Mirror the no-cache reorder strategy: skip output items in the main pass and only append outputs immediately after their matching call.
- Use the cache only to fill *missing* outputs (not to append a second copy of an already-present output).

2) Tests

No new tests cover the new HTTP repair behaviors. Given this is labeled P1 security/correctness, please add focused unit tests for:
- `[output, call]` => `[call, output]`
- multiple outputs per call_id
- call items without `call_id` are dropped (or explicitly handled)
- orphan outputs behavior is deterministic

## Non-blocking

- `httpResponsesSessionKey(req *http.Request, payload []byte)` doesn’t use `req`; consider removing the parameter.
- `callPresent` in `repairResponsesHTTPSingle()` is currently unused; remove or use it.

## Test plan

- Suggested locally: `go test ./...` plus focused unit tests for the HTTP repair functions in `sdk/api/handlers/openai`.

PR #3137, opened 2026-04-29
feat: add DeepSeek direct executor for additional LLM provider support

berthojoris:
### Summary
This PR adds DeepSeek as an additional direct executor/provider for CLIProxyAPI.

### Why
This allows CLIProxyAPI users to route requests to DeepSeek through the same unified proxy flow as other supported LLM providers.

### Changes
- Add DeepSeek direct executor
- Add OpenAI-compatible request/response handling
- Add PoW/helper handling
- Register DeepSeek model/provider support
- Add safe config example and ignore local config.json

### Security
- No real credentials are included
- config.json is ignored
- config.example.json is provided as a template


luispater (changes requested):
### Summary
This PR adds a new `deepseek` provider executor that talks to `chat.deepseek.com` directly (login/session + PoW), translates OpenAI-style requests into DeepSeek prompt format, and translates DeepSeek SSE back into OpenAI-compatible Chat Completions responses (including tool calls).

Key files:
- `internal/runtime/executor/deepseek_direct_executor.go`
- `internal/runtime/executor/deepseek_direct_client.go`
- `internal/runtime/executor/deepseek_direct_compat.go`
- `internal/runtime/executor/deepseek_direct_pow.go`
- `internal/registry/model_definitions.go`
- `sdk/cliproxy/service.go`

### Blocking
- GitHub reports this PR is not mergeable into `dev` right now (`mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`).
- Also, `dev` has moved to module path `github.com/router-for-me/CLIProxyAPI/v7`, while this PR targets `.../v6`. Please rebase onto current `dev`, resolve conflicts, and update import paths + registration wiring accordingly.

### Non-blocking / Suggestions
- `internal/runtime/executor/deepseek_proxy_executor.go` is an empty stub; consider removing it to reduce noise.
- `resolveDeepSeekModel` includes `deepseek-v4-vision*`, but `GetDeepSeekModels()` doesn’t list those models. Either add the model definitions or remove the resolver branches so users don’t discover “hidden” unsupported variants.
- Consider clarifying expected DeepSeek auth inputs (token vs email/mobile+password) somewhere user-facing; `config.example.json` exists but isn’t clearly tied into the main config flow.

### Test plan (local)
- `go test ./internal/runtime/executor -run DeepSeek`

Note: `go test ./...` currently fails in `internal/registry` with `TestCodexFreeModelsExcludeGPT55` (this appears to already be failing on the PR merge-base as well, not introduced by this PR).

PR #3141, opened 2026-04-29
fix(codex): restore minimal continuity for prompt cache reuse

VooDisss:
Fixes prompt-cache continuity regressions for Codex requests by restoring the smallest executor-level behavior needed to keep repeated requests on one stable upstream cache identity.

## Summary

This PR restores minimal Codex continuity handling on top of the current `origin/main` tip so repeated Codex requests can reuse upstream prompt cache again.

The key change is intentionally small:

- resolve one stable continuity key for Codex requests
- apply that same key to outgoing `prompt_cache_key`
- apply that same key to outgoing `session_id`
- keep HTTP and websocket Codex paths aligned

No auth-affinity routing, scheduler changes, or `internal/translator` changes are included.

## Why this is needed

I previously opened PR #2374 to restore Codex cache continuity. That approach was later reverted by @luispater in PR #2400.

After that revert, repeated Codex requests again stopped sharing a stable continuity identity, which caused cache hits to disappear in my real OpenCode -> CLIProxyAPI -> Codex usage.

The regression came from losing the executor-level continuity behavior that ensured repeated requests reused one stable cache/session identity instead of falling back to narrower or inconsistent request identity handling.

## Before this change

In the regressed state:

- repeated Codex requests could lose a stable continuity identity
- OpenAI-style Codex requests could fall back too narrowly when deriving continuity
- HTTP and websocket paths did not preserve the same effective continuity model
- prompt cache reads stopped occurring again in repeated OpenCode sessions

## After this change

CLIProxyAPI restores one minimal continuity rule for Codex executors:

- prefer explicit incoming `prompt_cache_key`
- otherwise use execution session metadata if available
- otherwise use a stable client-principal-derived hash
- otherwise use a stable auth-ID-derived hash

That chosen continuity key is then applied to both:

- request header `session_id`

This keeps repeated Codex requests on one stable upstream continuity identity without changing routing policy.

## Scope

This PR is intentionally narrow.

Included:

- `internal/runtime/executor/codex_continuity.go`
- `internal/runtime/executor/codex_executor.go`

Not included:

- auth-affinity or sticky-routing changes
- scheduler or conductor changes
- `internal/translator` changes
- broader cache policy redesign

## Manual validation

This change was manually validated in OpenCode sessions.

Confirmed behavior:

- `gpt-5.4` works
- `gpt-5.4-mini` works
- repeated OpenCode requests through CLIProxyAPI regain prompt cache hits

I am only explicitly claiming manual validation with OpenCode, but the fix is implemented as a generic Codex continuity improvement rather than an OpenCode-specific special case.

## Relationship to previous PRs

- prior attempt: #2374
- revert of that attempt: #2400

This PR is the minimal reapplication of the executor-level continuity behavior needed to restore prompt cache reuse while avoiding the broader review surface that the earlier attempt accumulated.

### DIFF
--- a/internal/runtime/executor/codex_continuity.go
+++ b/internal/runtime/executor/codex_continuity.go
@@ -1,0 +1,97 @@
+package executor
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	"github.com/router-for-me/CLIProxyAPI/v6/internal/runtime/executor/helps"
+	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/auth"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v6/sdk/cliproxy/executor"
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
+)
+
+type codexContinuity struct {
+	Key string
+}
+
+func ginContextFrom(ctx context.Context) *gin.Context {
+	if ctx == nil {
+		return nil
+	}
+	ginCtx, _ := ctx.Value("gin").(*gin.Context)
+	return ginCtx
+}
+
+func principalString(raw any) string {
+	switch v := raw.(type) {
+	case string:
+		return strings.TrimSpace(v)
+	case fmt.Stringer:
+		return strings.TrimSpace(v.String())
+	default:
+		return strings.TrimSpace(fmt.Sprintf("%v", raw))
+	}
+}
+
+func continuityMetadataString(meta map[string]any, key string) string {
+	if len(meta) == 0 {
+		return ""
+	}
+	raw, ok := meta[key]
+	if !ok || raw == nil {
+		return ""
+	}
+	switch v := raw.(type) {
+	case string:
+		return strings.TrimSpace(v)
+	case []byte:
+		return strings.TrimSpace(string(v))
+	default:
+		return ""
+	}
+}
+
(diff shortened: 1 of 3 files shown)


luispater (changes requested):
## Summary

This PR introduces a small, executor-level “continuity key” for Codex requests to restore prompt cache reuse across repeated requests. It resolves one stable continuity key (preferring an explicit `prompt_cache_key`, then `execution_session_id`, then a deterministic hash of the client principal / auth ID) and applies it to:
- request body: `prompt_cache_key`
- request headers: `session_id`
It also aligns HTTP and WebSocket Codex paths to use the same continuity behavior.

## Blocking

- **Unit tests will not compile:** `CodexExecutor.cacheHelper(...)` signature is changed, but `internal/runtime/executor/codex_executor_cache_test.go` is not updated accordingly. Please update the test to call the new signature (pass `auth` and `opts`) and ensure `go test ./...` passes.

## Non-blocking / Suggestions

- **WebSocket header behavior change:** WebSocket previously set `Conversation_id`; this PR switches to only `session_id`. If upstream still relies on `Conversation_id` for continuity, consider keeping it (or add a regression test to confirm it’s no longer needed).
- **Header/body consistency risk:** `misc.EnsureHeader` prioritizes downstream `ginHeaders` over the already-set target header. If a client sends `Session_id`, it may override the continuity-derived `session_id`, potentially de-syncing `prompt_cache_key` and `session_id`. Consider either:
  - making continuity headers authoritative (do not override once set), or
  - explicitly incorporating downstream `Session_id` into continuity resolution so header/body stay aligned.
- Add focused unit coverage for `resolveCodexContinuity` precedence rules.

## Test plan

- `go test ./...`
- (optional) targeted: `go test ./internal/runtime/executor -run TestCodexExecutorCacheHelper`

PR #3150, opened 2026-04-30
feat: add trusted proxies config

kkqy:
A common scenario is deploying CPA using Docker and then using nginx as a reverse proxy. In this case, the IP obtained by CPA is the internal IP of 172.16.0.0/12 instead of the real client IP. Therefore, it is recommended to add trusted proxies config.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -5,6 +5,13 @@ host: ""
 # Server port
 port: 8317
 
+# Trusted proxies used to resolve the real client IP from forwarded headers.
+# When omitted, Gin's default trusted proxy behavior is left unchanged.
+# Configure only local loopback proxies by default; add your reverse proxy/container subnet explicitly if needed.
+trusted-proxies:
+  - "127.0.0.0/8"
+  - "::1"
+
 # TLS settings for HTTPS. When enabled, the server listens with the provided certificate and key.
 tls:
   enable: false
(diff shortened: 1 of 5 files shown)


luispater (changes requested):
## Summary
This PR introduces a `trusted-proxies` config to control Gin trusted proxies so `ClientIP()` can resolve the real client IP from forwarded headers in common reverse-proxy/container deployments. The example config defaults to loopback-only proxies, while omitting the key preserves Gin’s existing default behavior for backwards compatibility.

## Blocking
- The PR currently has merge conflicts against `dev` (`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`). Please rebase/merge `dev` into your branch and resolve conflicts.

## Non-blocking notes
- Implementation looks reasonable:
  - `Config.TrustedProxies []string` with `SanitizeTrustedProxies()` to trim and drop empty entries.
  - `server.NewServer()` applies `engine.SetTrustedProxies(cfg.TrustedProxies)` only when the field is non-nil, and warns (non-fatal) on invalid entries.
- Tests added are helpful and cover the key behaviors (omitted vs empty vs configured).

## Test plan (local)
- ✅ `go test ./internal/api ./internal/config`
- ⚠️ `go test ./...` fails in `internal/registry` and `internal/runtime/executor` on the PR head; the same failures also reproduce on the PR base commit, so they don’t appear to be caused by this change.

PR #3154, opened 2026-04-30
fix(openai/responses): 新版codex支持 deepseak 等

OmMaBaMiHong:
… DeepSeek compat

Two fixes for OpenAI Responses → Chat Completions translation when used with DeepSeek models (deepseek-v4-flash):

1. Tool call grouping: consecutive function_call items in the input array are now grouped into a single assistant message with multiple tool_calls, fixing "insufficient tool messages following tool_calls message" errors.

2. Thinking mode disabled: DeepSeek's deepseek-v4-flash defaults to thinking mode, which returns reasoning_content and requires it to be echoed back. Send thinking: {type: "disabled"} instead of reasoning_effort to avoid this requirement entirely.

Also strips reasoning_content from streaming and non-streaming responses since the proxy does not support echoing it back.


luispater (changes requested):
### Summary
This PR aims to improve OpenAI Responses → Chat Completions conversion and reasoning-content round-trip behavior.

### Blocking
- `flushToolGroup()` can emit an assistant message with empty `tool_calls: []` and then emit `role:"tool"` messages when only `function_call_output` items are present, which is likely invalid for strict Chat Completions implementations.
- The converter injects `thinking: {"type":"disabled"}` unconditionally when `reasoning` is absent or has empty effort; this is likely to hard-fail on providers that don’t support `thinking`, and the current comment about “retry layer handling” does not match `sdk/cliproxy/auth/conductor.go` behavior.
- If `docs/reasoning-content-roundtrip.md` is included (or similar docs added), new Markdown docs should be in English per repo guidance.

### Non-blocking
- Please add focused unit tests for tool-output-only inputs and for streaming `reasoning_content` interleaved with `content`/`tool_calls`.

### Test plan
- Suggested: `go test ./internal/translator/openai/openai/responses` and `go build -o test-output ./cmd/server && rm -f test-output`.

PR #3164, opened 2026-04-30
fix(openai): improve Claude compatibility translation

sunyui:
## Summary
- Fix OpenAI streaming tool call translation when later deltas omit or repeat function names.
- Normalize Anthropic tool schemas for OpenAI-compatible providers by adding missing object properties.
- Preserve cache_control metadata and skip Claude Code billing headers in OpenAI-compatible request translation.

## Test plan
- [x] GOROOT=/usr/local/Cellar/go/1.26.2/libexec GOPROXY=https://goproxy.cn,direct go test ./internal/translator/openai/claude

### DIFF
--- a/internal/translator/openai/claude/openai_claude_request.go
+++ b/internal/translator/openai/claude/openai_claude_request.go
@@ -13,6 +13,10 @@ import (
 	"github.com/tidwall/sjson"
 )
 
+func isClaudeCodeBillingHeaderText(text string) bool {
+	return strings.HasPrefix(text, "x-anthropic-billing-header:")
+}
+
 // ConvertClaudeRequestToOpenAI parses and transforms an Anthropic API request into OpenAI Chat Completions API format.
 // It extracts the model name, system instruction, message contents, and tool declarations
 // from the raw JSON request and returns them in the format expected by the OpenAI API.
@@ -103,7 +107,7 @@ func ConvertClaudeRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 	hasSystemContent := false
 	if system := root.Get("system"); system.Exists() {
 		if system.Type == gjson.String {
-			if system.String() != "" {
+			if system.String() != "" && !isClaudeCodeBillingHeaderText(system.String()) {
 				oldSystem := []byte(`{"type":"text","text":""}`)
 				oldSystem, _ = sjson.SetBytes(oldSystem, "text", system.String())
 				systemMsgJSON, _ = sjson.SetRawBytes(systemMsgJSON, "content.-1", oldSystem)
@@ -113,6 +117,9 @@ func ConvertClaudeRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 			if system.IsArray() {
 				systemResults := system.Array()
 				for i := 0; i < len(systemResults); i++ {
+					if systemResults[i].Get("type").String() == "text" && isClaudeCodeBillingHeaderText(systemResults[i].Get("text").String()) {
+						continue
+					}
 					if contentItem, ok := convertClaudeContentPart(systemResults[i]); ok {
 						systemMsgJSON, _ = sjson.SetRawBytes(systemMsgJSON, "content.-1", []byte(contentItem))
 						hasSystemContent = true
@@ -188,6 +195,9 @@ func ConvertClaudeRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 							toolResultJSON, _ = sjson.SetRawBytes(toolResultJSON, "content", []byte(toolResultContent))
 						} else {
 							toolResultJSON, _ = sjson.SetBytes(toolResultJSON, "content", toolResultContent)
+						}
+						if cacheControl := part.Get("cache_control"); cacheControl.Exists() {
+							toolResultJSON, _ = sjson.SetRawBytes(toolResultJSON, "cache_control", []byte(cacheControl.Raw))
 						}
 						toolResults = append(toolResults, toolResultJSON)
 					}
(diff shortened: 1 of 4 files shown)


luispater (changes requested):
### Summary
This PR improves the Claude ↔ OpenAI translator in a few targeted ways:

- **Request (Claude → OpenAI chat.completions)**:
  - Skips Claude Code billing header system blocks (`x-anthropic-billing-header:`).
  - Preserves `cache_control` metadata on content parts, tools, and tool_result messages.
  - Normalizes tool `input_schema` to ensure every `{"type":"object"}` has a `properties` object (including nested objects/array items), which helps with stricter schema validators.

- **Streaming response (OpenAI → Claude)**:
  - Fixes tool-call streaming translation when later deltas omit `function.name` (empty string) or repeat the name, ensuring only a single `tool_use` content block is emitted while arguments are accumulated correctly.

New/updated tests cover schema normalization, cache_control preservation, billing-header skipping, and the streaming tool-call edge cases.

### Blocking
- **Potential upstream incompatibility**: The translated OpenAI payload now includes `cache_control` fields in places the upstream may not accept (e.g. `messages[].content[].cache_control`, `tools[].cache_control`, `role=tool` message `cache_control`). If the OpenAI target is strict, this can cause 400s. Please confirm the intended upstream(s) accept this extension, or gate/strip `cache_control` when targeting strict OpenAI endpoints.

### Non-blocking
- `isClaudeCodeBillingHeaderText` might want `strings.TrimSpace(text)` before prefix matching, just to be robust to any leading whitespace.

### Test plan
- ✅ `go test ./internal/translator/openai/claude`

### Notes on CI
- `translator-path-guard` fails because this PR touches `internal/translator/...` (ignored per instructions).

cyclic-elevator:
Hi @sunyui — thanks for this, it cleared up a class of OpenRouter-via-Azure failures for us. While running Claude Code through CLIProxyAPI to OpenRouter with this branch applied, I ran into a sibling case that the current `normalizeObjectSchemaProperties` doesn't quite cover, and I think it's worth folding in here since the fix is one extra branch in the same function. Would love your thoughts.

## Use case

Downstream is Claude Code; upstream is OpenRouter routing to an OpenAI-API-compatible Azure deployment (the same chain this PR is targeting). The agent registers MCP tools whose Anthropic `input_schema` includes `anyOf` branches with bare `{"type": "array"}` — a common pattern for "nullable list" fields:

```json
{
  "name": "mcp__plugin_gus_gus_server__create_gus_work_item",
  "input_schema": {
    "type": "object",
    "properties": {
      "acceptance_criteria": {
        "anyOf": [
          { "type": "array" },
          { "type": "null" }
        ]
      }
    }
  }
}
```

This schema is legal Anthropic input and is accepted by Claude's first-party API, but after translation it reaches the upstream as-is and the strict OpenAI-compatible validator (Azure) returns:

```
Invalid schema for function 'mcp__plugin_gus_gus_server__create_gus_work_item':
In context=('properties', 'acceptance_criteria', 'anyOf', '0'),
array schema items is not an object.
```

So the same family of bug this PR fixes for `type: "object"` without `properties` also exists symmetrically for `type: "array"` without `items`. The current PR's nested-tools test (`TestConvertClaudeRequestToOpenAI_ToolSchemaAddsMissingObjectProperties`) happens to side-step this because the `items` it uses is already an object schema (`{"type": "object"}`).

## Rationale

- Both shapes are valid JSON Schema and valid Anthropic tool input; only the strict OpenAI/Azure validator rejects them.
- Real Claude Code agents emit this shape: MCP tools generated from Pydantic/Zod with `Optional[List[...]]` typing routinely produce `{"anyOf": [{"type": "array"}, {"type": "null"}]}` without an `items` field.
- It's the same OpenAI-compat strictness rule this PR is already remediating — handling the array case here keeps the fix coherent rather than splitting it across two PRs.
- The fix is symmetric and stays inside the helper you already added; no new helper, no new imports, no behavior change for valid schemas.

## Suggested minimal extension

```go
func normalizeObjectSchemaProperties(schema any) any {
	switch value := schema.(type) {
	case map[string]any:
		switch value["type"] {
		case "object":
			if _, ok := value["properties"]; !ok {
				value["properties"] = map[string]any{}
			}
		case "array":
			if _, ok := value["items"].(map[string]any); !ok {
				value["items"] = map[string]any{}
			}
		}
		for key, child := range value {
			value[key] = normalizeObjectSchemaProperties(child)
		}
		return value
	case []any:
		for i, child := range value {
			value[i] = normalizeObjectSchemaProperties(child)
		}
		return value
	default:
		return schema
	}
}
```

Empty `items: {}` is accepted by Azure/OpenAI — they only require `items` to be an object schema, not a non-empty one.

## Suggested test

```go
func TestConvertClaudeRequestToOpenAI_ToolSchemaAddsMissingArrayItems(t *testing.T) {
	inputJSON := `{
		"model": "claude-3-opus",
		"tools": [{
			"name": "create_work_item",
			"description": "",
			"input_schema": {
				"type": "object",
				"properties": {
					"acceptance_criteria": {
						"anyOf": [
							{"type": "array"},
							{"type": "null"}
						]
					}
				}
			}
		}],
		"messages": [{"role": "user", "content": "hi"}]
	}`

	result := ConvertClaudeRequestToOpenAI("test-model", []byte(inputJSON), false)
	branch := gjson.GetBytes(result, "tools.0.function.parameters.properties.acceptance_criteria.anyOf.0")

	if got := branch.Get("type").String(); got != "array" {
		t.Fatalf("anyOf[0] type = %q, want array", got)
	}
	if !branch.Get("items").Exists() || !branch.Get("items").IsObject() {
		t.Fatalf("expected object-valued items on array schema, got %s", branch.Raw)
	}
}
```

Happy to either roll this into your branch with co-authorship, or send a tiny follow-up PR after this one merges — whichever you prefer.

PR #3179, opened 2026-05-01
feat: add usage backup image

Yuxcoo:
## Summary
- add a stats Docker image entrypoint that preserves in-memory usage statistics across container restarts
- support PostgreSQL-compatible storage, including Supabase via `DATABASE_URL` / `PGSTORE_DSN`
- support S3-compatible object storage, including Cloudflare R2, through the existing `OBJECTSTORE_*` environment variables
- document configuration, backend selection, and local-only fallback behavior

## Details
The wrapper starts CLIProxyAPI normally, imports a previously saved Management API usage snapshot on startup, periodically exports the current snapshot, and exports once more during graceful shutdown.

Supported backup modes:
- `postgres`, `postgresql`, or `pgsql` for PostgreSQL/Supabase
- `s3`, `objectstore`, or `object-storage` for S3-compatible buckets
- `local`, `none`, or `disabled` for local file-only snapshots

When `STATS_BACKEND` is not set, the wrapper auto-detects PostgreSQL first, then object storage, then falls back to local-only snapshots under `/CLIProxyAPI/data/usage_backup.json`.

## Compatibility
- The normal Dockerfile and default runtime behavior are unchanged.
- The stats image is opt-in via `Dockerfile.stats`.
- Usage import/export requires `MANAGEMENT_PASSWORD` to match `remote-management.secret-key`; without it the service still starts and logs that backup import/export is disabled.
- The new shell entrypoint is kept LF-only via `.gitattributes` so it remains executable in Linux containers when edited from Windows.

## Tests
- `git diff --check`
- `bash -n unified-stats-entrypoint.sh`

A local `go build` / Docker build was not run in this workspace because Go and Docker are not available in PATH here.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+unified-stats-entrypoint.sh text eol=lf
--- a/Dockerfile.stats
+++ b/Dockerfile.stats
@@ -1,0 +1,35 @@
+FROM golang:1.26-alpine AS builder
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+ARG VERSION=dev
+ARG COMMIT=none
+ARG BUILD_DATE=unknown
+
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w -X 'main.Version=${VERSION}' -X 'main.Commit=${COMMIT}' -X 'main.BuildDate=${BUILD_DATE}'" -o ./CLIProxyAPI ./cmd/server/
+
+FROM alpine:3.22.0
+
+RUN apk add --no-cache postgresql-client rclone tzdata
+
+RUN mkdir /CLIProxyAPI
+
+COPY --from=builder /app/CLIProxyAPI /CLIProxyAPI/CLIProxyAPI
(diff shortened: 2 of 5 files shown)


Yuxcoo:
Thanks for the automated review. I pushed `ebc53516` to address the actionable findings:

- Added `curl` to the stats runtime image dependencies.
- Changed PostgreSQL sync so the usage JSON is written through a temporary SQL file instead of being passed through `psql -v`, avoiding command-line length limits.
- Restricted the generated rclone config file to `0600` before writing object storage credentials.
- Reworked usage export/import HTTP calls to capture status codes with `curl -o ... -w '%{http_code}'` instead of parsing a combined body/status stream.

I kept `golang:1.26-alpine` and `alpine:3.22.0` because the repository's existing `Dockerfile` already uses those same base images; the stats Dockerfile is intentionally aligned with the current project Docker baseline.

Local checks rerun:
- `git diff --check`
- `bash -n unified-stats-entrypoint.sh`

Yuxcoo:
https://github.com/router-for-me/CLIProxyAPI/commit/18bb9c315fced2c428f57b4a0e66b06183c46c06 chore: remove usage tracking and logging functionality

PR #3213, opened 2026-05-04
fix: preserve parallel tool-call grouping for OpenAI Responses

Ranshen1209:
## Summary

This fixes a bug in the OpenAI Responses -> Chat Completions request conversion path that breaks parallel tool calling for Kimi-compatible upstreams.

## Root cause

OpenAI Responses can emit multiple consecutive `function_call` items in a single assistant turn when `parallel_tool_calls` is enabled.

Before this patch, `ConvertOpenAIResponsesRequestToOpenAIChatCompletions` converted each `function_call` item into its own assistant message containing a single `tool_calls` entry.

That rewrote a single assistant turn like:

- assistant: `function_call(read_a:0)`
- assistant: `function_call(read_b:1)`
- tool: `function_call_output(read_a:0)`
- tool: `function_call_output(read_b:1)`

into Chat Completions messages shaped like:

- assistant with `tool_calls=[read_a:0]`
- assistant with `tool_calls=[read_b:1]`
- tool result for `read_a:0`
- tool result for `read_b:1`

Some upstreams, including Kimi on the Responses compatibility path, validate tool results strictly against the immediately preceding assistant tool-call turn. Because the parallel calls were split across multiple assistant messages, the first tool result no longer matched the expected transcript shape and upstream rejected it with errors like:

- `an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'`
- `The following tool_call_ids did not have response messages: read_a:0`

## Fix

- Buffer consecutive Responses `function_call` items.
- Flush them as a single Chat Completions assistant message with multiple `tool_calls` entries.
- Flush buffered tool calls before any normal message or any `function_call_output` item.
- Keep single-tool behavior unchanged.

## Tests

Added request-conversion regression tests covering:

- single function call + output stays unchanged
- parallel function calls are grouped into one assistant `tool_calls` message
- corresponding tool outputs keep their original `tool_call_id` / `call_id` mapping, including `name:index`-style ids such as `read_a:0`

## Validation

Ran:

- `go test -run 'TestConvertOpenAIResponsesRequestToOpenAIChatCompletions|TestNormalizeKimiToolMessageLinks' ./internal/translator/openai/openai/responses ./internal/runtime/executor`
- `go build -o /tmp/cli-proxy-api-test ./cmd/server && rm /tmp/cli-proxy-api-test`

Note: `go test ./internal/runtime/executor` currently includes unrelated pre-existing `antigravity` failures in this environment, so validation was scoped to the affected Responses/Kimi paths.

### DIFF
--- a/internal/translator/openai/openai/responses/openai_openai-responses_request.go
+++ b/internal/translator/openai/openai/responses/openai_openai-responses_request.go
@@ -1,6 +1,7 @@
 package responses
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/tidwall/gjson"
@@ -32,6 +33,22 @@ func ConvertOpenAIResponsesRequestToOpenAIChatCompletions(modelName string, inpu
 	out := []byte(`{"model":"","messages":[],"stream":false}`)
 
 	root := gjson.ParseBytes(rawJSON)
+	appendMessage := func(message []byte) {
+		out, _ = sjson.SetRawBytes(out, "messages.-1", message)
+	}
+
+	flushAssistantToolCalls := func(toolCalls []string) {
+		if len(toolCalls) == 0 {
+			return
+		}
+
+		assistantMessage := []byte(`{"role":"assistant","tool_calls":[]}`)
+		for idx, toolCall := range toolCalls {
+			path := "tool_calls." + strconv.Itoa(idx)
+			assistantMessage, _ = sjson.SetRawBytes(assistantMessage, path, []byte(toolCall))
+		}
+		appendMessage(assistantMessage)
+	}
 
 	// Set model name
 	out, _ = sjson.SetBytes(out, "model", modelName)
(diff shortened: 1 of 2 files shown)


luispater (approved):
### Summary
This PR fixes an OpenAI Responses -> Chat Completions conversion bug by buffering consecutive `function_call` items and flushing them as **one** assistant message with multiple `tool_calls`. This preserves the “single assistant turn” transcript shape required by strict OpenAI-compatible upstreams (e.g., Kimi), preventing tool-call-id validation errors.

### Key findings
- ✅ Correctness: Consecutive Responses `function_call` items are grouped into a single Chat Completions assistant `tool_calls` message.
- ✅ Ordering: Buffered tool calls are flushed before any normal message and before any `function_call_output`, so tool outputs immediately follow the relevant tool-call turn.
- ✅ Regression coverage: Added tests for both parallel tool-call grouping and single-call behavior.

### Non-blocking suggestions
- Consider storing pending tool calls as `[][]byte` instead of `[]string` to avoid extra conversions (purely a micro-cleanup).

### Test plan (suggested)
- `go test ./internal/translator/openai/openai/responses`
- (Optional) `go test ./...` if the broader suite is stable in CI.

luispater (changes requested):
## Blocking

## Recommendation

PR #3222, opened 2026-05-05
fix(management): bound log API responses

Reshwell:
## Summary
- default `/v0/management/logs` to the latest 200 lines instead of returning every log line
- clamp requested log limits to 5000 lines
- read initial log views from newest files backward and skip old rotated files for incremental reads
- add regression tests for default, clamped, incremental, and invalid limit behavior

## Tests
- `go test ./internal/api/...`

### DIFF
--- a/internal/api/handlers/management/logs.go
+++ b/internal/api/handlers/management/logs.go
@@ -60,7 +60,7 @@ func (h *Handler) GetLogs(c *gin.Context) {
 		return
 	}
 
-	limit, errLimit := parseLimit(c.Query("limit"))
+	limit, limitProvided, errLimit := parseLimit(c.Query("limit"))
 	if errLimit != nil {
 		c.JSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("invalid limit: %v", errLimit)})
 		return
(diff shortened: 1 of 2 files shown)


luispater (approved):
- Summary:
  - Bounds `/v0/management/logs` responses by defaulting to the latest 200 lines and clamping `limit` to 5000.
  - Improves incremental reads by skipping clearly-old rotated files and preserving “full delta” behavior when `after` is set but `limit` is omitted.
  - Stabilizes `latest-timestamp` for tail views even when the last returned lines are continuations without timestamps.

- Key findings:
  - Blocking: none.
  - Non-blocking:
    - `line-count` semantics differ between tail vs incremental paths (returned vs scanned).

- Test plan:
  - go test ./internal/api/handlers/management
  - go test ./internal/api/...

luispater (changes requested):
## Summary

This PR bounds the `/v0/management/logs` endpoint: defaults to 200 lines, clamps to 5000 max, reads from newest files backward for initial views, and skips old rotated files for incremental reads.

## Blocking

- **Import version mismatch**: `logs_test.go` imports `v6/internal/config`, but `dev` moved to `v7`. The PR will fail to compile after merge. Please rebase onto current `dev`.

## Non-blocking

- Breaking behavior change (200 lines default instead of all) is intentional but worth a changelog entry.
- `line-count` semantics differ between tail and incremental paths.
- Double `TrimRight` is harmless but redundant.

## Tests

Good coverage. CI passes on the PR branch.

PR #3223, opened 2026-05-05
fix: recover image generation stream disconnects

ddkwing:
## Summary
- recover completed image output items when upstream closes before `response.completed`
- retry recoverable image stream disconnects according to `streaming.bootstrap-retries`
- add regression tests for output-item recovery and retry behavior

## Verification
- `go build -o /tmp/cli-proxy-api-release-check ./cmd/server`

## Notes
- `go test ./... -count=1` still has unrelated existing failures in `internal/registry` (`TestCodexFreeModelsExcludeGPT55`) and `internal/runtime/executor` antigravity credits tests.

### DIFF
--- a/sdk/api/handlers/openai/openai_images_handlers.go
+++ b/sdk/api/handlers/openai/openai_images_handlers.go
@@ -591,9 +591,10 @@ func (h *OpenAIAPIHandler) collectImagesFromResponses(c *gin.Context, responsesR
 	if mainModel == "" {
 		mainModel = defaultImagesMainModel
 	}
-	dataChan, upstreamHeaders, errChan := h.ExecuteStreamWithAuthManager(cliCtx, "openai-response", mainModel, responsesReq, "")
-
-	out, errMsg := collectImagesFromResponsesStream(cliCtx, dataChan, errChan, responseFormat)
+	maxAttempts := 1 + handlers.StreamingBootstrapRetries(h.Cfg)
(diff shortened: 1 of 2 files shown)


luispater (approved):
## Summary
This improves non-streaming `/v1/images/*` by:
- Recovering image results from `response.output_item.done` when the upstream closes before `response.completed`.
- Retrying recoverable “stream disconnected before completion” cases, bounded by `streaming.bootstrap-retries`.
- Adding focused regression tests for both recovery and retry.

## Non-blocking notes
- The retry decision currently depends on string-matching `"stream disconnected before completion"`; consider switching to a sentinel error/type to avoid fragile coupling.
- Recovery returns a successful images response even without `response.completed` (so `usage` is missing and, if upstream could disconnect mid-way in multi-image, partial results might be returned). If that’s the intended behavior, looks good.

## Test plan
- `go test ./sdk/api/handlers/openai -run TestCollectImagesFromResponses -v`

luispater (changes requested):
## Summary

This PR adds resilience to image generation streaming by recovering completed output items when the upstream closes before sending `response.completed`, and retrying recoverable disconnects. The implementation is clean and well-tested.

## Blocking

- **Merge conflicts**: The PR currently has conflicts with the `dev` base branch. Please rebase to resolve before this can be merged.

## Non-blocking

- The retry predicate matches on a literal string produced by `buildRecoveredResponse`. This coupling is fragile -- consider a sentinel error or dedicated error type.
- Good extraction of `extractImageResultFromOutputItem` to eliminate duplication.

## Tests

Two new tests cover recovery and retry scenarios. CI passes.

PR #3227, opened 2026-05-05
test: add server regression for /v1/responses/compact route

dos1989:
## Summary
- add a server-level regression test for `POST /v1/responses/compact`
- verify route wiring + handler dispatch using `stream: true` rejection path
- assert `400` with `Streaming not supported for compact responses`

## Testing
- go test ./internal/api -run TestServerRegistersResponsesCompactRoute

### DIFF
--- a/internal/api/server_test.go
+++ b/internal/api/server_test.go
@@ -147,6 +147,24 @@ func TestManagementUsageRequiresManagementAuthAndPopsArray(t *testing.T) {
 	}
 }
 
+func TestServerRegistersResponsesCompactRoute(t *testing.T) {
+	server := newTestServer(t)
+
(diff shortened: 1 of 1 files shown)


luispater (approved):
- Summary: Adds a server-level regression test to ensure `/v1/responses/compact` and the Codex alias `/backend-api/codex/responses/compact` are registered, and that `stream=true` is rejected with the expected 400 error message.

- Key findings:
  - Non-blocking: Covers both routes via subtests and exercises the early `stream=true` rejection path, avoiding any upstream dependency.
  - Optional: Consider decoding the JSON error response and asserting `error.message` directly instead of `strings.Contains` on the raw body (more precise).

- Test plan:
  - Reviewed diff only (no local test run).
  - CI (informational): build ✅

luispater (changes requested):
## Blocking

## Recommendation

PR #3234, opened 2026-05-06
[codex] fix claude embedded image data urls

ultrano:
## Summary
- Extract large embedded `data:image/...;base64,...` URLs from OpenAI Responses user text before forwarding to Claude.
- Convert extracted image data into Claude `image` content blocks while preserving surrounding text.
- Apply the same extraction to `function_call_output` content and keep short data URL literals unchanged.

## Root Cause
Codex-style image prompts can arrive as base64 data URLs embedded in text content instead of structured `input_image` parts. The Claude Responses translator forwarded those data URLs as text, which made Claude prompt/token sizes grow abnormally.

## Validation
- `go test ./internal/translator/claude/openai/responses`
- `go build -o test-output.exe ./cmd/server`

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_request.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_request.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math/big"
+	"regexp"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


luispater (changes requested):
## Summary
This PR addresses a real pain point: Codex-style prompts can embed large `data:image/...;base64,...` blobs inside user text, and forwarding those as plain text to Claude inflates prompt size. The approach (extract → emit Claude `image` content blocks, preserve surrounding text) makes sense, and the added tests/validation are a good start.

## Request changes (blocking)
1) **Avoid `strings.ToLower(text)` allocation on the hot path**
   - `claudeTextPartsWithEmbeddedImages()` currently does `strings.Contains(strings.ToLower(text), "data:image/")`, which copies the entire prompt even when there is no embedded image.
   - Suggestion: replace the precheck with a non-alloc path (e.g. `embeddedImageDataURLPattern.MatchString(text)` or another cheap scan) and only do heavier work when there is a real match.

2) **Restrict / normalize extracted `media_type` before emitting Claude image blocks**
   - Today extraction triggers for any `image/*`. If a user includes something like `data:image/svg+xml;base64,...`, this change will convert it into an `image` block and could make the Claude request invalid (whereas leaving it as literal text would have worked).
   - Suggestion: allowlist the Claude-supported image types and normalize common aliases (at least `image/png`, `image/jpeg` with `image/jpg -> image/jpeg`, `image/gif`, `image/webp`). If the type is not supported, skip extraction and keep it as text.

## Nice-to-haves (non-blocking)
- Consider making the markdown wrapper detector tolerate optional titles/whitespace so it can strip more variants of `alt`.
- Add tests for:
  - Multiple embedded images in one input text.
  - Unsupported `media_type` staying as literal text (no extraction).

## Test plan
- Not run locally (no checkout).
- PR author notes: `go test ./internal/translator/claude/openai/responses` and `go build -o test-output.exe ./cmd/server`.

PR #3265, opened 2026-05-07
fix(codex): keep responses websocket across recoverable upstream disconnects

zzl360:
## Summary
- avoid closing downstream /v1/responses/ws connections for retryable Codex upstream websocket resets
- reset the upstream disconnect channel when a session ID is reused after a terminal disconnect
- remove the downstream handler subscription that closed clients on any upstream disconnect signal

## Context
This fixes regressions from fb08b92402187cd87f888d371ee5d6f5107f6d36 where recoverable upstream websocket reconnects could terminate downstream websocket clients, and reused session IDs could subscribe to an already-closed disconnect channel.

## Tests
- go test ./internal/runtime/executor -run 'TestCodexWebsocketsUpstreamDisconnectChan|TestCodexWebsockets.*' -count=1
- go test ./sdk/api/handlers/openai -run 'TestResponsesWebsocket|TestWebsocketUpstream' -count=1
- git diff --check

### DIFF
--- a/internal/runtime/executor/codex_websockets_executor.go
+++ b/internal/runtime/executor/codex_websockets_executor.go
@@ -77,8 +77,9 @@ type codexWebsocketSession struct {
 
 	readerConn *websocket.Conn
 
-	upstreamDisconnectOnce sync.Once
-	upstreamDisconnectCh   chan error
+	upstreamDisconnectMu     sync.Mutex
+	upstreamDisconnectCh     chan error
+	upstreamDisconnectClosed bool
 }
 
 func NewCodexWebsocketsExecutor(cfg *config.Config) *CodexWebsocketsExecutor {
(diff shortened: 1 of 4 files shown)


zzl360:
Addressed the terminal-close concern in the latest push. `closeExecutionSession` now signals and closes the upstream disconnect channel before closing the underlying Codex websocket session, so subscribers do not remain attached to stale session state. Added `TestCodexWebsocketsCloseExecutionSessionSignalsDisconnect` and reran the Codex websocket/openai handler tests plus the server build check.

luispater (commented):
## Summary
This PR updates Codex Responses WebSocket session handling so the downstream `/v1/responses/ws` connection can remain open across recoverable upstream disconnects. It refactors disconnect signaling in `CodexWebsocketsExecutor`, removes the downstream auto-close behavior on Codex upstream disconnect notifications, and adds tests to cover disconnect-channel reset + terminal session close signaling.

## What changed
  - Replace one-shot disconnect signaling with a resettable disconnect channel (`upstreamDisconnectChan()`).
  - Gate disconnect notifications via `notifyDisconnect` so retryable invalidations (e.g. `send_error` retry path) don’t force downstream teardown.
  - Ensure `CloseExecutionSession` signals and closes the disconnect channel (prevents stale subscribers).
- `sdk/api/handlers/openai/openai_responses_websocket.go`
  - Remove the goroutine that closed the downstream websocket when the Codex upstream disconnect channel closed.
- Tests updated accordingly:
  - `internal/runtime/executor/codex_websockets_executor_test.go`
  - `sdk/api/handlers/openai/openai_responses_websocket_test.go`

## Key findings
- The disconnect-channel lifecycle looks sound and is now covered by tests (reset after close, and signaling on terminal session close).
- The behavior change is intentional: downstream websockets are no longer forcibly closed on Codex upstream disconnect events.

## Test plan
- Not run locally.
- CI `build` is green (checks `translator-path-guard` and `pr-test-build` ignored per request).

## Follow-ups (non-blocking)
- Add an integration-style test that asserts `/v1/responses/ws` stays open and can serve a subsequent request after an upstream disconnect/reconnect, since the prior “closes on disconnect” behavior was removed.

luispater (approved):
## Summary

Clean fix for a regression where recoverable upstream websocket reconnects were terminating downstream `/v1/responses/ws` clients. The approach is sound: replace the one-shot `sync.Once` disconnect signal with a resettable mutex-guarded channel, gate disconnect notifications behind a `notifyDisconnect` flag, and remove the downstream handler that indiscriminately closed client connections on any upstream disconnect.

## Key findings

- (non-blocking) In `readUpstreamLoop`, the `notifyDisconnect := ch == nil` check and the subsequent `invalidateUpstreamConn` call are not atomic with respect to the active handler lifecycle. This is safe because the handler already received the error via the channel, but a brief code comment explaining the intent would help future readers.
- (non-blocking) The test `TestCodexWebsocketsCloseExecutionSessionSignalsDisconnect` relies on the literal string `"session_closed"` matching inside the formatted error. Consider extracting this as a constant to avoid silent breakage if the reason string changes.

## Tests

The PR adds three well-targeted unit tests covering: retryable invalidation suppression, channel reset after close, and terminal session close signaling. The removed downstream disconnect test is correctly obsoleted by the design change. CI passes.

## Verdict

Approve. The concurrency model is correct, the regression is properly addressed, and test coverage is adequate.

---

PR #3266, opened 2026-05-07
fix(openai): handle null usage field in streaming token calculation

Ovear:
In streaming mode, the usage field may be null. Calling reporter.Publish in this case can incorrectly consume the sync.Once guard and prevent later valid token reports.


Ovear:
Duplicated: #3221
Already fixed. Closed.

PR #3268, opened 2026-05-07
fix: fallback disabled Claude accounts

CutieDeng:
## Scenario

Claude OAuth accounts can remain present in the credential manager even after the upstream organization or account has been disabled. In that state, Anthropic returns a `400 invalid_request_error` response with a message like:
`This organization has been disabled.`

## Problem

The proxy previously treated this response as a request-shape error. That stopped credential failover immediately, even when other Claude accounts were still available.
For Responses WebSocket sessions, the issue was worse because the session could stay pinned to the disabled account. Repeated `continue` requests would keep hitting the same unusable credential instead of rotating to another account.

## Solution

This PR changes disabled-credential handling so the proxy can recover automatically:

- Detects disabled organization/account/credential/user messages in `400 invalid_request_error` responses.
- Treats those errors as credential availability failures instead of invalid request errors.
- Marks the affected auth as temporarily unavailable at the auth level, so future scheduling skips it across models.
- Releases pinned Responses WebSocket auths when the same disabled-credential error is returned.
- Adds regression tests for non-streaming, streaming, websocket pinned-auth failover, and auth-level scheduler blocking.

## Verification

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket.go
@@ -943,17 +943,42 @@ func shouldReleaseResponsesWebsocketPinnedAuth(errMsg *interfaces.ErrorMessage) 
 		return false
 	}
 	status := errMsg.StatusCode
+	errText := ""
 	if status <= 0 && errMsg.Error != nil {
 		if se, ok := errMsg.Error.(interface{ StatusCode() int }); ok && se != nil {
 			status = se.StatusCode()
 		}
 	}
+	if errMsg.Error != nil {
+		errText = errMsg.Error.Error()
+	}
 	switch status {
 	case http.StatusUnauthorized, http.StatusPaymentRequired, http.StatusForbidden, http.StatusTooManyRequests:
 		return true
+	case http.StatusBadRequest:
+		return isResponsesWebsocketCredentialDisabledError(errText)
 	default:
 		return false
 	}
+}
+
+func isResponsesWebsocketCredentialDisabledError(message string) bool {
+	lower := strings.ToLower(strings.TrimSpace(message))
+	if lower == "" {
+		return false
+	}
+	patterns := [...]string{
+		"organization has been disabled",
+		"account has been disabled",
+		"credential has been disabled",
+		"user has been disabled",
+	}
+	for _, pattern := range patterns {
+		if strings.Contains(lower, pattern) {
+			return true
+		}
+	}
+	return false
 }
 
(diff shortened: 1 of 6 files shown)


luispater (approved):
Summary
This PR treats Anthropic `400 invalid_request_error` "...has been disabled" responses as credential availability failures (instead of request-shape errors), so requests can rotate to other usable Claude OAuth auths. For `/v1/responses/ws`, it also releases pinned auths on the same signature to unblock continued websocket requests.

What changed
- `sdk/cliproxy/auth/conductor.go`
  - `isRequestInvalidError`: excludes the disabled-credential signature so the manager can fail over.
  - `Manager.MarkResult`: marks the model unavailable and also marks the whole auth temporarily unavailable (`auth.Unavailable` + `auth.NextRetryAfter`) so scheduling skips it across models.
- `sdk/cliproxy/auth/selector.go`
  - `isAuthBlockedForModel`: respects auth-level unavailability even when picking with a specific model.
- `sdk/api/handlers/openai/openai_responses_websocket.go`
  - `shouldReleaseResponsesWebsocketPinnedAuth`: releases pinned websocket auths on the disabled-credential 400 signature.
- Adds regression coverage for non-streaming, streaming, websocket pinned-auth failover, and auth-level scheduler blocking.

Key findings
- Blocking: none.

Test plan
- CI: green.
- Suggested local: `go test ./sdk/cliproxy/auth` and `go test ./sdk/api/handlers/openai`.

luispater (approved):
## Summary

This PR correctly handles disabled Claude organization/account errors (HTTP 400 with "has been disabled" messages) by treating them as credential availability failures rather than invalid request errors. This enables automatic failover to other accounts.

## What changed

1. **Conductor**: New `case 400` in `MarkResult` detects disabled-credential messages, marks the model state unavailable with a 30-minute cooldown, and marks the auth itself as unavailable.
2. **Selector**: New auth-level blocking check that blocks any model request when the auth is marked unavailable with a future retry time.
3. **WebSocket handler**: Extends pinned-auth release to cover 400 disabled-credential errors.
4. **`isRequestInvalidError`**: Excludes credential-disabled messages from being classified as invalid requests, enabling retry/failover.

## Non-blocking suggestions

- Pattern-matching logic is duplicated between `isResponsesWebsocketCredentialDisabledError` and `isCredentialDisabledErrorMessage`. Consider extracting to a shared utility.

## Tests

Four new regression tests cover all key scenarios. CI passes. No merge conflicts.

## Verdict

Approve.

---

PR #3279, opened 2026-05-08
fix(openai-compatible): ignore null stream usage chunks

seakee:
## Summary
- Ignore stream usage chunks unless usage is an OpenAI-style object with known token fields.
- Reuse the shared OpenAI-style usage parser for stream usage.
- Add regression coverage for usage:null and token-less usage objects before the final usage chunk.

## Tests
- go test ./internal/runtime/executor/helps ./internal/runtime/executor ./internal/redisqueue ./internal/api/handlers/management ./test -run 'TestParseOpenAI|TestOpenAICompat|TestUsageQueue|TestGetUsageQueue|TestGeminiExecutorRecordsSuccessfulZeroUsageInQueue'
- go build -o /private/tmp/cpa-cli-proxy-api-test-output ./cmd/server

### DIFF
--- a/internal/runtime/executor/helps/usage_helpers.go
+++ b/internal/runtime/executor/helps/usage_helpers.go
@@ -301,27 +301,29 @@ func parseOpenAIStyleUsageNode(usageNode gjson.Result) usage.Detail {
 	return detail
 }
 
+func isOpenAIStyleUsageNode(usageNode gjson.Result) bool {
+	return usageNode.IsObject() &&
+		(usageNode.Get("prompt_tokens").Exists() ||
+			usageNode.Get("input_tokens").Exists() ||
+			usageNode.Get("completion_tokens").Exists() ||
+			usageNode.Get("output_tokens").Exists() ||
+			usageNode.Get("total_tokens").Exists() ||
+			usageNode.Get("prompt_tokens_details.cached_tokens").Exists() ||
+			usageNode.Get("input_tokens_details.cached_tokens").Exists() ||
+			usageNode.Get("completion_tokens_details.reasoning_tokens").Exists() ||
+			usageNode.Get("output_tokens_details.reasoning_tokens").Exists())
+}
+
 func ParseOpenAIStreamUsage(line []byte) (usage.Detail, bool) {
 	payload := jsonPayload(line)
 	if len(payload) == 0 || !gjson.ValidBytes(payload) {
 		return usage.Detail{}, false
 	}
 	usageNode := gjson.GetBytes(payload, "usage")
-	if !usageNode.Exists() {
+	if !isOpenAIStyleUsageNode(usageNode) {
 		return usage.Detail{}, false
 	}
-	detail := usage.Detail{
-		InputTokens:  usageNode.Get("prompt_tokens").Int(),
-		OutputTokens: usageNode.Get("completion_tokens").Int(),
-		TotalTokens:  usageNode.Get("total_tokens").Int(),
-	}
-	if cached := usageNode.Get("prompt_tokens_details.cached_tokens"); cached.Exists() {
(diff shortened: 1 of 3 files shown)


Xie-free:
I independently reproduced this issue on an OpenAI-compatible streaming provider (MiMo / MiniMax-compatible route) and verified that this patch fixes it in production.

Repro details:
- CLIProxyAPI release affected: v6.10.9
- Endpoint: `POST /v1/chat/completions`
- Request uses `stream: true` and `stream_options: {"include_usage": true}`
- SSE stream contains many intermediate chunks with explicit `"usage": null`
- The final SSE chunk contains real usage, for example:
  - `prompt_tokens: 259`
  - `completion_tokens: 30`
  - `total_tokens: 289`
  - `prompt_tokens_details.cached_tokens: 192`
  - `completion_tokens_details.reasoning_tokens: 29`

Before the fix, CPA-Manager could see the request but stored zero usage for this provider/model because the first `usage: null` chunk was treated as a valid zero-token usage event, and the later real usage chunk was ignored by the usage reporter.

I applied the equivalent local patch to ignore explicit JSON null usage chunks in `ParseOpenAIStreamUsage` and rebuilt/deployed CLIProxyAPI. After restart, the same MiMo streaming request was stored correctly by CPA-Manager:

```text
provider: mimo
model: mimo-v2.5-pro
input_tokens: 259
output_tokens: 30
total_tokens: 289
cached_tokens: 192
reasoning_tokens: 29
```

Focused regression tests for the parser also passed locally:

```text
TestParseOpenAIUsageChatCompletions PASS
TestParseOpenAIUsageResponses PASS
TestParseOpenAIStreamUsage_NullUsageReturnsFalse PASS
TestParseOpenAIStreamUsage_EmptyUsageObjectReturnsFalse PASS
TestParseOpenAIStreamUsage_MissingUsageFieldReturnsFalse PASS
TestParseOpenAIStreamUsage_RealUsageReturnsTrue PASS
```

So the root cause/fix direction here is confirmed: explicit `usage: null` chunks must not trigger usage reporting; only the final non-empty usage object should be reported.

luispater (changes requested):
### Summary
Adds regression coverage for OpenAI-compatible streaming usage handling:
- Ignore keepalive chunks where `usage` is `null` or an object without recognized token fields.
- Ensure the final usage is still published exactly once (no zero-token record from `usage:null` chunks).

### Blocking
This PR branch is behind the current `dev` branch, which has already moved to module `github.com/router-for-me/CLIProxyAPI/v7`.
- PR base/head (`baseRefOid=09ac8a1165149bdd2649bde9b87dd812120b23dc`, `headRefOid=f0466e9adb86556ba22dd697a5639a3b4708cc31`) are still `module .../v6`.
- Current `dev` (`041ccf01952300f6b0daa6a18dc8c69c60784351`) is `module .../v7`.

Please rebase/merge latest `dev`, update imports to `/v7/...`, and rerun tests/build on top of current `dev` to confirm this merges cleanly and stays green.

### Notes
- `TestParseOpenAIStreamUsageIgnoresUsageWithoutTokenFields` is a good guardrail for `usage:{}` / token-less usage objects.
- `TestOpenAICompatExecutorStreamPublishesFinalUsageAfterNullUsageChunks` follows the existing `redisqueue` test approach (enable+clear queue, poll with a deadline) and validates only the final usage record is enqueued.

### Test plan
Not run locally (review only). After rebasing onto current `dev`, please run at least:
- `go test ./internal/runtime/executor/helps ./internal/runtime/executor -run 'TestParseOpenAIStreamUsage|TestOpenAICompatExecutorStreamPublishesFinalUsageAfterNullUsageChunks'`

luispater (changes requested):
## Blocking

## Recommendation

PR #3283, opened 2026-05-08
Fix Responses WebSocket tool output context repair

wuchulonly:
## Summary
- Insert cached `function_call` items before matching `function_call_output` items even when `previous_response_id` is present.
- Prevent upstream 400 errors like `No tool call found for function call output with call_id ...` for delayed/background tool outputs on Responses WebSocket sessions.
- Add regression coverage for `previous_response_id` incremental tool output repair.

## Root Cause
The WebSocket repair path treated any request with `previous_response_id` as safe to forward orphan `function_call_output` items. For delayed/background tool outputs, `previous_response_id` can point at a newer response that does not contain the original `function_call`, while the proxy has already cached that call from earlier response output events. Upstream then validates the incremental input and rejects the orphan output.

## Fix
For `function_call_output` / `custom_tool_call_output` items, the repair logic now first checks the current payload and cached tool-call context. If a matching cached call exists, it inserts that call before the output. Only when no cached call exists does it preserve the previous `previous_response_id` passthrough behavior.

## Tests
- `go test ./sdk/api/handlers/openai -run 'TestRepairResponsesWebsocketToolCalls|TestNormalizeSubsequentRequestIncrementalInputStillMerges'`

## Notes
- `go test ./...` still has unrelated pre-existing failures in `internal/registry` (`TestCodexFreeModelsExcludeGPT55`) and `internal/runtime/executor` Antigravity credits URL tests.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket_test.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket_test.go
@@ -662,6 +662,34 @@ func TestRepairResponsesWebsocketToolCallsInsertsCachedCallForOrphanOutput(t *te
 	}
 }
 
+func TestRepairResponsesWebsocketToolCallsInsertsCachedCallForPreviousResponseOutput(t *testing.T) {
+	outputCache := newWebsocketToolOutputCache(time.Minute, 10)
+	callCache := newWebsocketToolOutputCache(time.Minute, 10)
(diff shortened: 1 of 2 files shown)


luispater (approved):

WatermelonSumer:
Updated to v7.0.1, but the same problem still exists.

PR #3291, opened 2026-05-08
fix: allow disabled OAuth auths to keep auto-refreshing

Hades-ming:
## Summary

This change allows disabled OAuth auth files to continue participating in the auto-refresh flow.

Disabled auths are still excluded from request routing and scheduling. The change only separates traffic eligibility from OAuth credential refresh.

## Why

Some operators temporarily disable auth files for quota or maintenance reasons. If disabled OAuth auths stop refreshing, their tokens may expire before they are re-enabled.


luispater:
已经在v7.0.0中实现该功能。

PR #3293, opened 2026-05-08
fix(codex/claude): preserve service_tier in Claude-to-Codex request translation

### LINKED ISSUE #3276: bug: Codex Claude translator drops service_tier priority
## Description

The Codex Claude-compatible translator currently rebuilds the upstream Codex Responses request from a fresh template and does not preserve top-level `service_tier` from the incoming Claude request.

That blocks clients which sit in front of `/api/provider/codex/v1/messages` and translate a user-facing fast-mode alias into a Codex request tier. Current OpenAI Codex uses `service_tier: "priority"` as the request value for Fast mode, while some clients may still provide the legacy/config spelling `"fast"`.

## Reproduction payload

```json
{
  "model": "gpt-5.4",
  "service_tier": "priority",
  "messages": [{ "role": "user", "content": "Reply with OK" }]
}
```

Expected translated Codex request contains:

```json
{
  "model": "gpt-5.4",
  "service_tier": "priority"
}
```

Actual translated request omits `service_tier`.

## Rationale

`/v1/responses` already preserves `service_tier: "priority"`, but the Claude-compatible path is the path used by Claude Code-compatible clients against a pinned Codex provider route. Dropping the field means Fast mode cannot work through that path even when the client selected it correctly.

## Intended implementation

In `internal/translator/codex/claude/codex_claude_request.go`, after setting `reasoning.summary`, copy a safe normalized service tier into the template:

```go
if serviceTier := normalizeCodexServiceTier(rootResult.Get("service_tier")); serviceTier != "" {
    template, _ = sjson.SetBytes(template, "service_tier", serviceTier)
}
```

Suggested helper:

```go
func normalizeCodexServiceTier(result gjson.Result) string {
    if !result.Exists() || result.Type != gjson.String {
        return ""
    }

    switch strings.ToLower(strings.TrimSpace(result.String())) {
    case "fast", "priority":
        return "priority"
    default:
        return ""
    }
}
```

This keeps the change narrow: it only passes the Codex-supported fast/priority request tier and continues omitting unsupported values such as `default`.

## Suggested tests

Add translator coverage for:
- `service_tier: "priority"` passes through as `"priority"`
- `service_tier: "fast"` normalizes to `"priority"`
- unsupported tier values are omitted

lihan3238:
## Summary

Fixes the Claude-compatible translator dropping the service_tier field when translating to Codex format. This prevents Fast mode from working through the Claude-compatible path.

## Fix

- Add normalizeCodexServiceTier helper that maps both legacy/fast and priority to the Codex-supported priority value
- Pass service_tier through after reasoning.summary in ConvertClaudeRequestToCodex
- Only applies when service_tier is fast or priority, unsupported values are omitted as before

## Testing

- Requests without service_tier: behavior unchanged (field omitted)
- Requests with service_tier priority: now correctly passed through
- Requests with service_tier fast: normalized to priority and passed through
- Requests with other values: omitted as before

Fixes #3276

### DIFF
--- a/internal/translator/codex/claude/codex_claude_request.go
+++ b/internal/translator/codex/claude/codex_claude_request.go
@@ -34,6 +34,18 @@ import (
 //
 // Returns:
 //   - []byte: The transformed request data in internal client format
+func normalizeCodexServiceTier(result gjson.Result) string {
+	if !result.Exists() || result.Type != gjson.String {
+		return ""
+	}
+	switch strings.ToLower(strings.TrimSpace(result.String())) {
+	case "fast", "priority":
+		return "priority"
+	default:
+		return ""
+	}
+}
+
 func ConvertClaudeRequestToCodex(modelName string, inputRawJSON []byte, _ bool) []byte {
 	rawJSON := inputRawJSON
 
@@ -321,6 +333,9 @@ func ConvertClaudeRequestToCodex(modelName string, inputRawJSON []byte, _ bool) 
 	}
 	template, _ = sjson.SetBytes(template, "reasoning.effort", reasoningEffort)
 	template, _ = sjson.SetBytes(template, "reasoning.summary", "auto")
+	if serviceTier := normalizeCodexServiceTier(rootResult.Get("service_tier")); serviceTier != "" {
+		template, _ = sjson.SetBytes(template, "service_tier", serviceTier)
+	}
 	template, _ = sjson.SetBytes(template, "stream", true)
 	template, _ = sjson.SetBytes(template, "store", false)
 	template, _ = sjson.SetBytes(template, "include", []string{"reasoning.encrypted_content"})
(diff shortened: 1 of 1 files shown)


lihan3238:
Thanks for the review! I've addressed the doc association issue by moving the helper function to the end of the file, after all the main functions.

Regarding the auto/default values: the issue (#3276) specifically states that only fast/priority should be passed through, and other values like default should be omitted. This matches the behavior of the OpenAI Responses path which already preserves only priority. Adding auto/default passthrough would change the semantics, so I'd prefer to keep it as-is unless a maintainer wants different behavior.

lihan3238:
Thanks for the review feedback. I see the CI is failing because of the `translator-path-guard` workflow, which restricts changes to `internal/translator/**`. I'll close this PR and create an issue instead so the maintenance team can handle the fix.

lihan3238:
Closing in favor of #3294 — the translator-path-guard prevents external PRs from modifying this path. The issue describes the bug and suggested fix for the maintenance team.

PR #3303, opened 2026-05-09
fix(api): prevent idle TCP connections from blocking the accept loop

### LINKED ISSUE #3267: Management/API server can stop responding when an idle TCP connection blocks protocol sniffing
  [ ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
  [x] No, it's another issue.

  Not a request payload issue. This appears to be a listener/protocol multiplexer issue before normal HTTP request handling.

  **Describe the bug**
  CLIProxyAPI can stop responding to management/API HTTP requests when an idle TCP connection blocks protocol sniffing in the shared listener.

  In my case, the management page loaded, but after entering the management password the UI showed:

  ```text
  Connection timed out, server not responding
  ```

  Restarting cli-proxy-api made it work again temporarily, then the same problem returned.

  The process was still alive and systemd reported the service as running, but HTTP requests connected at TCP level and then received no response.

  CLI Type
  N/A. This is not specific to a CLI account type. It affects the CLIProxyAPI management/API listener itself before model/provider routing.

  Model Name
  N/A

  LLM Client
  N/A. Reproduced with direct curl / raw TCP connection against the CLIProxyAPI listener.

  Request Information
  Affected version:

  CLIProxyAPI Version: v6.10.9
  Commit: 785b00c3127eea6aa207f1207ead8a2aa93690a3

  Service bind address:

  127.0.0.1:8318

  When the service was in the bad state, this request connected but received no response:

  curl -v --max-time 5 http://127.0.0.1:8318/

  Output:

  *   Trying 127.0.0.1:8318...
  * Connected to 127.0.0.1 (127.0.0.1) port 8318
  > GET / HTTP/1.1
  > Host: 127.0.0.1:8318
  > User-Agent: curl/8.5.0
  > Accept: */*
  >
  * Operation timed out after 5002 milliseconds with 0 bytes received
  * Closing connection

  systemctl status cli-proxy-api.service still showed the process as active/running.

  A Go SIGQUIT goroutine dump showed the process blocked in the protocol multiplexer accept path:

  goroutine 12 [IO wait]:
  github.com/router-for-me/CLIProxyAPI/v6/internal/api.(*Server).acceptMuxConnections(...)
      /tmp/cliproxyapi-v6.10.9-build/internal/api/protocol_multiplexer.go:77
  created by github.com/router-for-me/CLIProxyAPI/v6/internal/api.(*Server).Start
      /tmp/cliproxyapi-v6.10.9-build/internal/api/server.go:857

  From reading internal/api/protocol_multiplexer.go, acceptMuxConnections accepts one connection and then performs protocol sniffing inline before returning to listener.Accept().

  That includes:

  - TLS handshake sniffing
  - reader.Peek(1) for non-TLS protocol detection

  If a client opens a TCP connection and sends no bytes, the single accept loop can block on that connection and stop accepting later HTTP requests. The process remains alive, but new management/API requests
  queue up and eventually time out.

  A minimal reproduction should be possible by opening an idle TCP connection and then sending a normal HTTP request from another shell.

  Terminal 1:

  bash -c 'exec 3<>/dev/tcp/127.0.0.1/8318; sleep 60'

  Terminal 2:

  curl -v --max-time 5 http://127.0.0.1:8318/

  Expected: the HTTP request should still be accepted and served.

  Actual on the affected build: the HTTP request can time out because the accept loop is blocked sniffing the idle connection.

  Expected behavior
  An idle or slow TCP client should not be able to block the whole management/API listener.

  The server should continue accepting and serving other HTTP requests even if one accepted connection sends no bytes.

  A possible fix is to move per-connection protocol sniffing out of the accept loop.

  The accept loop should accept connections quickly and dispatch each accepted connection to a separate goroutine, for example:

  for {
      conn, errAccept := listener.Accept()
      if errAccept != nil {
          return errAccept
      }
      if conn == nil {
          continue
      go s.routeMuxConnection(conn, httpListener)
  }

  Then routeMuxConnection can perform the existing TLS/Redis/HTTP detection for that single connection.

  It should also set a short deadline while sniffing the protocol, for example:

  const muxProtocolSniffTimeout = 5 * time.Second

  _ = conn.SetReadDeadline(time.Now().Add(muxProtocolSniffTimeout))
  prefix, errPeek := reader.Peek(1)
  _ = conn.SetReadDeadline(time.Time{})

  For TLS connections, apply a similar deadline around tlsConn.Handshake().

  If the sniff/handshake times out or fails, close only that connection. Do not block the main accept loop.

  A regression test could be:

  TestProtocolMultiplexer_IdleConnectionDoesNotBlockHTTP

  Test idea:

  1. Start the shared listener and HTTP server.
  2. Open one TCP connection and send no bytes.
  3. While that idle connection is still open, issue a normal HTTP request.
  4. Assert that the HTTP request returns 200 OK within a short timeout.

  I locally tested this approach by routing accepted connections in a goroutine and adding a short protocol sniff timeout. After that, the service continued responding even with 20 deliberately idle TCP
  connections held open:

  /                    -> 200 OK
  /v0/management/debug -> 200 OK
  listener backlog     -> 0

  I can open a PR with the small multiplexer change and regression test if maintainers agree with this direction.

  Screenshots
  N/A

  OS Type

  - OS: Linux
  - Version: Ubuntu/Linux systemd host

  Additional context
  This looks different from upstream provider timeout issues. The failure happens before normal request handling: the TCP listener accepts a connection, then the accept loop blocks while trying to sniff that
  connection’s protocol.

  The management page may still be cached/loaded in the browser, but API calls made by the management UI can time out because new HTTP requests are not being accepted and routed.

lihan3238:
## Summary

- Move per-connection protocol detection (TLS handshake, reader.Peek) out of the accept loop and into a per-connection goroutine
- Add regression test TestAcceptMuxNotBlockedByIdleConnection

## Problem

acceptMuxConnections performs protocol sniffing inline: after accepting a connection, it calls tlsConn.Handshake() and reader.Peek(1) before returning to listener.Accept(). If a client opens a TCP connection and never sends any bytes, Peek(1) blocks indefinitely, preventing all subsequent connections from being accepted. This makes the entire management/API server unresponsive while the process remains alive.

Reproduction (from #3267):

Terminal 1: open idle TCP connection
  bash -c "exec 3<>/dev/tcp/127.0.0.1/8317; sleep 60"

Terminal 2: HTTP request times out
  curl -v --max-time 5 http://127.0.0.1/8317/

## Fix

Extract the protocol detection logic into routeMuxConnection() and call it via a goroutine. The accept loop now only calls listener.Accept() and immediately dispatches each connection, so a single slow/idle connection cannot block the loop.

The extracted method uses return instead of the original continue since it runs in its own goroutine.

## Test Plan

- [x] TestAcceptMuxNotBlockedByIdleConnection -- opens an idle TCP connection, then sends a real HTTP request and verifies it succeeds within 3 seconds
- [x] All existing tests in internal/api/... pass
- [x] go build ./internal/api/... compiles cleanly

Closes #3267

### DIFF
--- a/internal/api/protocol_multiplexer.go
+++ b/internal/api/protocol_multiplexer.go
@@ -115,6 +115,7 @@ func (s *Server) routeMuxConnection(conn net.Conn, httpListener *muxListener) {
 			}
 			return
 		}
+		_ = conn.SetReadDeadline(time.Time{})
 		s.handleRedisConnection(conn, reader)
 		return
 	}
--- a/internal/api/protocol_multiplexer.go
+++ b/internal/api/protocol_multiplexer.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


lihan3238:
Good catch. Added a 10s read deadline in `routeMuxConnection` that is cleared once the connection is successfully routed to its handler. Idle connections will now time out and release the goroutine + fd. Pushed in 4ee690b.

lihan3238:
Good catch, thanks @chatgpt-codex-connector! Added deadline clear before entering  in the latest push (c5596e0).

luispater (approved):
Summary:
This change keeps the shared listener responsive by moving TLS/protocol sniffing out of the accept loop and into per-connection routing. The added sniff deadline also addresses the follow-up resource-leak risk for clients that connect and never send bytes.

Key findings:
- No blocking findings.

Test plan:
- Reviewed the diff for `internal/api/protocol_multiplexer.go` and `internal/api/protocol_multiplexer_test.go`.
- Verified PR checks are passing: `pr-test-build`, `translator-path-guard`, and `agents-md-guard`.
- Locally ran `go build -o test-output ./cmd/server && rm test-output` on PR HEAD `c5596e09`.

luispater (approved):

PR #3313, opened 2026-05-10
fix(usagereport): fix connection leak in resp client and management client

likunkun:
## Problem

After `cliproxy-usage-collector` ran for several days, it accumulated **7,368 leaked TCP connections** to CLIProxyAPI (observed in production), exhausting server file descriptors and preventing new connections from being accepted — including TLS handshakes from external clients.

Two root-cause bugs:

### Bug 1 — `resp/client.go`: goroutine hangs after TCP connect

`Pop()` used `net.Dialer{Timeout: 10s}` for the dial phase but set **no deadline** on the connection after it was established. If the server was slow or unresponsive, `bufio.Reader.ReadString()` would block indefinitely, leaking both the goroutine and the socket.

**Fix:** derive a deadline from the caller's `context.Context` (30s fallback) and apply it via `conn.SetDeadline` immediately after connecting.

### Bug 2 — `management/client.go`: new `http.Transport` on every call

`FetchAuthFiles` and `httpClient()` each constructed a **new `http.Transport`** on every invocation. Go's HTTP/1.1 keep-alive returns idle connections to the transport's pool — but since each transport was a local variable, it became GC-eligible immediately. Under sustained polling the pool fills faster than GC drains it, leaving thousands of `ESTABLISHED` connections.

**Fix:** add `NewClient()` constructor that builds one shared `*http.Client` (and thus one `Transport`) for the lifetime of the `Client`. `httpClient(timeout)` returns a **shallow copy** with the requested timeout but the same underlying transport, so connections are reused.

## Changes

| File | Change |
|------|--------|
| `internal/usagereport/resp/client.go` | Add `conn.SetDeadline` with context-aware 30s fallback |
| `internal/usagereport/management/client.go` | Add `NewClient()` constructor + shared transport; fix `httpClient()` shallow-copy pattern |
| `internal/usagereport/resp/client_test.go` | `TestPopTimesOutWhenServerDoesNotRespond`: server holds connection open, verifies context deadline is respected |
| `internal/usagereport/management/client_test.go` | `TestNewClientReusesConnection`: 3 sequential calls produce exactly 1 new TCP connection |

## Test plan

- [ ] `go test ./internal/usagereport/resp/...` — `TestPopTimesOutWhenServerDoesNotRespond` passes
- [ ] `go test ./internal/usagereport/management/...` — `TestNewClientReusesConnection` passes
- [ ] Deploy to production and monitor open file descriptors on `cliproxy-usage-collector` over 24h — count should remain stable

### DIFF
--- a/internal/usagereport/management/client.go
+++ b/internal/usagereport/management/client.go
@@ -1,0 +1,383 @@
+package management
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+	"time"
+)
+
+type Client struct {
+	BaseURL       string
+	ManagementKey string
+	TLSSkipVerify bool
+	hc            *http.Client // shared across all calls; nil means lazy-create (no reuse)
+}
+
+// NewClient builds a Client with a single shared http.Client so that
+// keep-alive connections are reused across repeated calls instead of
+// leaking one TCP connection per request.
+func NewClient(baseURL, managementKey string, tlsSkipVerify bool) Client {
+	var transport http.RoundTripper
+	if tlsSkipVerify {
+		transport = &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}} //nolint:gosec
+	} else {
+		transport = http.DefaultTransport
+	}
+	return Client{
+		BaseURL:       baseURL,
+		ManagementKey: managementKey,
+		TLSSkipVerify: tlsSkipVerify,
+		hc:            &http.Client{Transport: transport},
+	}
+}
+
+type AuthFile struct {
+	AuthIndex               string
+	Provider                string
+	Email                   string
+	Account                 string
+	AccountID               string
+	PlanType                string
+	SubscriptionActiveStart time.Time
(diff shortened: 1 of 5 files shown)


luispater:
Closing this PR because the added `internal/usagereport` client code has no runtime meaning in CPA as submitted.

The evidence is:

- The PR only adds `internal/usagereport/management` and `internal/usagereport/resp` packages plus their tests.
- No CPA runtime code imports these packages. A PR-head grep for `internal/usagereport` imports returns no matches.
- `go list -deps ./cmd/server` on the PR head does not include `github.com/router-for-me/CLIProxyAPI/v7/internal/usagereport/...`, so the server dependency graph never reaches this code.
- The only call sites for `FetchAuthFiles`, `FetchAPIKeys`, `FetchCodexQuotaWindows`, `resp.New`, and `Pop` are inside the newly added package tests.
- `go test ./internal/usagereport/...` passing only proves the new isolated packages compile and pass their own tests; it does not prove any CLIProxyAPI runtime connection leak is fixed.

If the leak is in `cliproxy-usage-collector`, the fix needs to be applied in the collector codebase or this PR must include the actual CPA integration point that uses these clients. In its current form, merging this PR would add dead code to CPA without changing production behavior.

PR #3328, opened 2026-05-11
feat: per-request passthrough-system flag (preserve caller system prompt under OAuth)

phj6688:
Adds an opt-in mechanism to preserve the caller's forwarded `system` content under Claude OAuth mode instead of having `sanitizeForwardedSystemPrompt` replace it with a generic 3-line stub. The canonical Claude Code system field, `cch` signing, and fake user_id are all unchanged — only the sanitize step becomes opt-out-able.

## Three control surfaces (any one activates it)

1. Request header `X-CPA-Passthrough-System: true|1` (highest priority — useful for routers/gateways with per-request control)
2. Auth attribute `cloak_passthrough_system: "true"` (per-credential default via the SDK's auth.Attributes)
3. ClaudeKey config field `passthrough-system: true` (static config-driven)

Default behavior: unchanged. Existing call sites and tests untouched (the 2-arg wrapper `checkSystemInstructionsWithMode` is preserved; only the inner `WithSigningMode` gained a parameter).

## Diff

- `internal/config/config.go`: +9 LOC — new `PassthroughSystem bool` on `CloakConfig`
- `internal/runtime/executor/claude_executor.go`: +47 LOC — header context reader, attribute reader (5th return value from `getCloakConfigFromAuth`), signature threading, merge logic in `applyCloaking`

~56 LOC total.

## Motivation

Self-hosted small-team setups frequently want to share one Claude Pro/Max subscription across multiple apps via CLIProxyAPI: a daily-driver Claude Code session AND an in-house specialist app (medical info, legal aid, support triage). Today only the daily-driver case works correctly because the specialist app's system prompt is deleted in `sanitizeForwardedSystemPrompt`. This PR closes that gap without adding a second subscription or a separate API key.

## Validation

Built v6.9.26-passthrough from this branch. Ran 6 probes against the live Anthropic endpoint through the patched binary:

| Test | Header | Result |
|------|--------|--------|
| knee-pain + MedChat system msg | none | refuses ("I'm not able to help with medical concerns") — upstream behavior preserved |
| same | `X-CPA-Passthrough-System: true` | asks 4 intake questions (age, location, duration, modifiers) — caller persona honored |
| chest pain emergency + MedChat | passthrough on | "Call 911 immediately. Classic warning signs of a heart attack." |
| PubMed meta-analysis query + MedChat | passthrough on | detailed retrieval-style summary |
| Python dataclass coding query + MedChat | passthrough on | correct Python answer (passthrough doesn't break coding) |
| same | none | same answer, Claude Code persona path — regression-clean |

Zero OAuth flagging across the run.

## Test plan

- [x] Manual probes of all four behavior buckets against live Anthropic
- [x] Regression: coding queries still work with and without the header
- [x] End-to-end via a multi-hop proxy chain (nexus-router → CLIProxyAPI → Anthropic) on a real medical-info product surface
- [ ] Reviewers may want to run the existing `internal/runtime/executor` test suite — wrapper signatures preserved, only inner `*WithSigningMode` gained a parameter

Happy to add inline unit tests for the new code path on request.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -345,6 +345,15 @@ type CloakConfig struct {
 	// CacheUserID controls whether Claude user_id values are cached per API key.
 	// When false, a fresh random user_id is generated for every request.
 	CacheUserID *bool `yaml:"cache-user-id,omitempty" json:"cache-user-id,omitempty"`
+
+	// PassthroughSystem, when true, suppresses the OAuth-mode sanitization that
+	// replaces the caller's forwarded system prompt with a generic stub. The
+	// canonical Claude Code system field (billing/agent/static blocks) is still
+	// injected so the request body remains recognizable to Anthropic; only the
+	// caller-supplied system message survives verbatim into the first user
+	// message. Can also be enabled per-request via the X-CPA-Passthrough-System
+	// header (value "true"/"1").
+	PassthroughSystem bool `yaml:"passthrough-system,omitempty" json:"passthrough-system,omitempty"`
 }
 
 // ClaudeKey represents the configuration for a Claude API key,
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -1006,7 +1006,7 @@ func claudeCreds(a *cliproxyauth.Auth) (apiKey, baseURL string) {
 }
 
 func checkSystemInstructions(payload []byte) []byte {
-	return checkSystemInstructionsWithSigningMode(payload, false, false, false, "2.1.63", "", "")
+	return checkSystemInstructionsWithSigningMode(payload, false, false, false, false, "2.1.63", "", "")
 }
 
 func isClaudeOAuthToken(apiKey string) bool {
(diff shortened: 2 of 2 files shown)


luispater (changes requested):
Summary
- Nice addition: an opt-in path to preserve caller-provided `system` content under Claude OAuth cloaking, while keeping the canonical Claude Code system blocks and signing behavior unchanged by default.

Blocking
- The `X-CPA-Passthrough-System` header currently enables passthrough purely based on caller input, even when the server operator did not explicitly allow this behavior via config/auth. This effectively allows any client to bypass `sanitizeForwardedSystemPrompt` in OAuth mode as long as they can set the header.
  - Suggestion: gate header activation behind an explicit server-side allow (e.g., only honor the header when either `cloak.passthrough-system: true` is set for the credential, or `auth.Attributes["cloak_passthrough_system"] == "true"`, or introduce a dedicated `allow-passthrough-system-header` flag defaulting to false).
- Please add unit tests for the new branch:
  - OAuth mode + passthrough disabled => sanitize is applied.
  - OAuth mode + passthrough enabled => sanitize is skipped and original forwarded system content is preserved.
  - Precedence across surfaces (header/auth/config), plus header parsing for `true` / `1`.

Non-blocking
- Please document the new knobs in `config.example.yaml` (cloak section) and in docs, including the auth attribute `cloak_passthrough_system` and the request header `X-CPA-Passthrough-System`.

Test plan
- `go test ./...` (or at least the existing `internal/runtime/executor` tests).
- Add/verify new unit tests for passthrough behavior and precedence.

PR #3333, opened 2026-05-11
Add xAI image generation compatibility

lifeiscontent:
## Summary
- route configured OpenAI-compatible image models through /v1/images/generations
- add xAI-scoped size to aspect_ratio/resolution request adaptation
- document xAI Grok image generation in the OpenAI compatibility config example

## Verification
- go test -count=1 -run 'Test(OpenAICompatExecutorImagesGeneration|XAIImageSizeMapping|ImagesGenerationsRoutesOpenAICompatModel)' ./internal/runtime/executor ./sdk/api/handlers/openai
- go test -count=1 ./sdk/api/handlers/openai ./internal/config
- go vet ./...
- go build -o cli-proxy-api ./cmd/server
- live HTTP smoke against local CLIProxyAPI: POST /v1/images/generations model=grok-imagine-image-quality size=1024x1024 returned data[0].url from imgen.x.ai


luispater (changes requested):
## Summary
This PR adds OpenAI-compatibility routing for image generation/editing via `/v1/images/generations` and `/v1/images/edits`, plus xAI-specific request adaptation (mapping `size` -> `aspect_ratio`/`resolution`) and accompanying unit tests.

## Blocking
- **Silent parameter dropping on OpenAI-compat image edits**: when `model` is detected as `openai-compatibility`, the handler rebuilds a minimal JSON body via `buildOpenAICompatImageEditRequest(model, prompt, images, size)` and forwards it. This currently omits `mask` and other edit params (`quality`, `background`, `output_format`, `input_fidelity`, `moderation`, `output_compression`, `partial_images`, etc.), meaning clients can send these fields and get no error while the request semantics change.
  - Please either:
    - Reject unsupported fields for OpenAI-compat image edits with a clear 400 (recommended if upstream truly doesn't support them), or
    - Forward supported fields through to the OpenAI-compat upstream.
- **Payload config / model normalization skipped for image passthrough**: `OpenAICompatExecutor.Execute` bypasses `TranslateRequest(...)` and `ApplyPayloadConfigWithRoot(...)` for `images/generations` and `images/edits`. If users rely on payload rules (or on the translator fallback that normalizes `model`), this can cause surprising behavior. If this is intentional, please document why and add coverage; otherwise consider re-applying these steps in the images path (while still skipping `thinking.ApplyThinking(...)`).

## Non-blocking
- Consider making the "passthrough response" behavior consistent for both `images/generations` and `images/edits` (today it early-returns only for generations).
- Please double-check the intended mapping for non-standard square sizes (the current logic promotes `>1024 && <=2048` squares to `2k`).

## Test plan
- Not run locally.
- Suggest running the focused tests added/updated in this PR (executor + OpenAI image handler tests), plus a quick `go test ./...` sanity pass.

Issue #3245, opened 2026-05-06
Amp Claude OAuth cloaking breaks tool names and drops forwarded system prompt


burugo:
## Background

I use AmpCode with CLIProxyAPI to call Claude through a Claude OAuth auth file / OAuth model path.

In this mode, the request is cloaked as Claude Code before being sent upstream.

## Problem: Amp system prompt is not preserved in `messages[0]`

When using a Claude API key config, the original Amp system prompt is prepended into `messages[0]`.

When using a Claude OAuth auth file / OAuth model path, the original Amp system prompt is replaced by a sanitized generic reminder:

`Use the available tools when needed to help with software engineering tasks...`

This causes Amp-specific behavior instructions to be lost.

The critical issue is that Amp sends additional system blocks after the first prompt, including AGENTS.md project guidance, loaded/available skill descriptions, workspace context, and cache-control-marked context. Claude OAuth cloaking currently sanitizes the combined forwarded system text, so these later blocks are dropped as well. Losing skill descriptions is especially harmful because the model no longer knows which Amp skills exist or when to use them. A safer fix is to sanitize only the first forwarded system block while preserving and appending system[1:] into the first user message.

PR #3335, opened 2026-05-11
fix(claude): preserve Amp system context in Claude OAuth cloaking

burugo:
## Summary

- Preserve additional forwarded system blocks during Claude OAuth cloaking
- Keep Amp-provided project guidance, AGENTS.md content, and skill descriptions in the first user message
- Add regression test `TestCheckSystemInstructionsWithSigningMode_OAuthPreservesAdditionalSystemBlocks`

## Problem

When Amp requests are routed through the Claude OAuth path, CLIProxyAPI applies Claude Code-style cloaking before sending the request upstream.

Amp sends its system context as multiple `system[]` blocks, not as a single prompt. The first block contains Amp's main agent prompt, while later blocks can include important runtime context such as:

- AGENTS.md project guidance
- available / loaded skill descriptions
- workspace and environment context
- cache-control-marked context blocks

Before this change, OAuth cloaking collected all forwarded system blocks into one string and passed the entire combined text through `sanitizeForwardedSystemPrompt()`. That function intentionally replaces input with a short generic reminder:

`Use the available tools when needed to help with software engineering tasks...`

As a result, not only Amp's first agent prompt but also the later system blocks were dropped. This caused Amp to lose important project instructions and skill descriptions, which can significantly affect tool use and task behavior.

Reproduction context from #3245:

Amp + Claude OAuth cloaking can drop forwarded system prompt context. In practice, the generated upstream request keeps only the sanitized generic reminder and loses additional Amp system context.

## Fix

Keep the existing OAuth cloaking behavior for the first forwarded system block, but limit sanitization to that first block only.

The updated flow is:

1. Collect forwarded `system[]` text blocks as before.
2. In OAuth mode, sanitize only `userSystemParts[0]`.
3. Preserve `userSystemParts[1:]` unchanged.
4. Join all parts and prepend them to the first user message using the existing `<system-reminder>` path.

This keeps the Claude OAuth cloaking safety behavior for the primary third-party agent prompt while preserving Amp's additional project guidance and skill descriptions.

## Test Plan

- [x] `TestCheckSystemInstructionsWithSigningMode_OAuthPreservesAdditionalSystemBlocks` -- verifies OAuth cloaking sanitizes the first forwarded system block but preserves later AGENTS.md guidance and skill description blocks
- [x] `go test -run 'TestCheckSystemInstructions' ./internal/runtime/executor`

Refs #3245

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -1689,10 +1689,10 @@ func checkSystemInstructionsWithSigningMode(payload []byte, strictMode bool, exp
 		}
 
 		if len(userSystemParts) > 0 {
-			combined := strings.Join(userSystemParts, "\n\n")
 			if oauthMode {
-				combined = sanitizeForwardedSystemPrompt(combined)
+				userSystemParts[0] = sanitizeForwardedSystemPrompt(userSystemParts[0])
 			}
+			combined := strings.Join(userSystemParts, "\n\n")
 			if strings.TrimSpace(combined) != "" {
 				payload = prependToFirstUserMessage(payload, combined)
 			}
--- a/internal/runtime/executor/claude_executor_test.go
+++ b/internal/runtime/executor/claude_executor_test.go
@@ -1923,6 +1923,36 @@ func TestCheckSystemInstructionsWithMode_ArraySystemStillWorks(t *testing.T) {
 	}
 }
 
+func TestCheckSystemInstructionsWithSigningMode_OAuthPreservesAdditionalSystemBlocks(t *testing.T) {
+	payload := []byte(`{
+		"system":[
+			{"type":"text","text":"Original Amp agent prompt that should be sanitized."},
+			{"type":"text","text":"AGENTS.md guidance should remain."},
+			{"type":"text","text":"Available skills: behavior-driven-development should remain.","cache_control":{"type":"ephemeral"}}
+		],
+		"messages":[{"role":"user","content":[{"type":"text","text":"hi"}]}]
+	}`)
+
+	out := checkSystemInstructionsWithSigningMode(payload, false, false, true, "2.1.63", "", "")
+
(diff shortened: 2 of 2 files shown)


luispater (approved):
Summary:
This PR fixes Claude OAuth cloaking for Amp requests by sanitizing only the first forwarded `system[]` text block (Amp’s primary agent prompt) while preserving additional forwarded system blocks (e.g., AGENTS.md guidance and skill descriptions) when they are prepended into the first user message.

Blocking:
- None.

Non-blocking:
- Consider adding an edge-case test where the first `system[]` element is not a `type:"text"` block (or is empty). Today the code sanitizes the first non-empty collected text block; it’s likely fine, but worth locking down with a test.

Test plan:
- `go test -run 'TestCheckSystemInstructions' ./internal/runtime/executor`
- `go test -run 'TestCheckSystemInstructionsWithSigningMode_OAuthPreservesAdditionalSystemBlocks' ./internal/runtime/executor`

luispater (approved):
## Summary

Clean, well-scoped bug fix. The change correctly limits OAuth sanitization to only the first forwarded system block while preserving additional Amp context (AGENTS.md, skills, etc.) that was previously being dropped.

## Key findings

- **Non-blocking**: The test assertion `sanitizeForwardedSystemPrompt("Original Amp agent prompt...")` works but is slightly misleading — since that function always returns the same static string regardless of input, the assertion is really just checking that the static sanitized text is present. A brief comment or using the literal expected string would make intent clearer.

## Correctness

- Index `[0]` access is safe because it is guarded by `len(userSystemParts) > 0`.
- Non-OAuth path is unaffected.
- Single-element edge case works correctly (element gets sanitized, then trivially joined).
- The security-relevant sanitization of the primary third-party agent prompt is preserved.

## Test coverage

The added regression test covers the main scenario well: multi-block system array in OAuth mode, verifying first block sanitized, subsequent blocks preserved, and user content intact.

## Recommendation

Approve — the fix is minimal, correct, and well-tested.

---

PR #3345, opened 2026-05-12
fix: add Claude OAuth token refresh and Anthropic header injection in /api-call

calonye:
## Summary

- Add Claude OAuth token refresh in `resolveTokenForAuth` — the `claude` case was missing, so stale tokens caused 401 on `/api-call` routed requests
- Auto-inject `x-api-key` header and `anthropic-beta: oauth-2025-04-20` for `api.anthropic.com` destinations — Anthropic's Messages API requires these instead of `Authorization: Bearer`

## Root Cause

Two issues in `internal/api/handlers/management/api_tools.go`:

1. **Missing Claude OAuth refresh** — `resolveTokenForAuth` had dedicated OAuth token refresh for `gemini-cli` and `antigravity`, but `claude` fell through to `tokenValueForAuth` which returned the token as-is without refresh
2. **Missing Anthropic header injection** — The generic `/api-call` handler does `$TOKEN$` substitution but applies no Anthropic-specific processing. The Claude executor handles this correctly but `/api-call` bypasses it

## Changes

- Add `claude` case in `resolveTokenForAuth` calling `claudeauth.NewClaudeAuthWithProxyURL().RefreshTokensWithRetry(ctx, refreshToken, 3)`
- In `APICall`, after `$TOKEN$` substitution, detect `api.anthropic.com` and inject `x-api-key` + `anthropic-beta` headers
- Add `refreshClaudeOAuthAccessToken` and `isAnthropicAPI` helper methods

## Safety

- Changes are additive (new case in existing switch, header injection guarded by URL detection)
- Reuses existing `claudeauth` package (already imported in codebase)
- No change to non-Anthropic provider behavior
- Build + vet pass cleanly


luispater (changes requested):
## Summary

This PR improves the management `/v0/management/api-call` proxy path by:
- Adding a `claude` branch in `resolveTokenForAuth` so Claude OAuth tokens are refreshed (via `internal/auth/claude`) instead of using potentially-stale values.
- Injecting `x-api-key` and `anthropic-beta: oauth-2025-04-20` when the destination host is `api.anthropic.com`.

## Blocking

1) Dual auth headers risk in `APICall`
- The new Anthropic injection adds `x-api-key` but does **not** remove an existing `Authorization` header (commonly `Authorization: Bearer $TOKEN$` per the endpoint docs/examples).
- This can cause requests to carry both `Authorization` and `x-api-key`, which is exactly what the PR description says Anthropic requires to avoid (“instead of Authorization: Bearer”), and it’s inconsistent with existing `internal/runtime/executor/claude_executor.go` behavior (it explicitly deletes `Authorization` when sending `x-api-key` to `api.anthropic.com`).

**Suggested adjustment**
- When `isAnthropicAPI(parsedURL)` and you decide to inject `x-api-key`, also delete `Authorization` (or at least delete it when it matches `Bearer <token>` case-insensitively).

## Non-blocking / suggestions

- `isAnthropicAPI` should consider `u.Hostname()` instead of `u.Host` so `api.anthropic.com:443` matches.
- Consider adding a small unit test covering:
  - `api.anthropic.com` -> inject `x-api-key` + beta header
  - `Authorization` removed when `x-api-key` injected
- Minor readability: `antigravityTokenNeedsRefresh` is reused for Claude; a provider-neutral wrapper/name would be clearer.

## Test plan

Not run locally. Please run at least:
(and/or `go test ./...`).

calonye:
Thanks for the review @luispater! Addressed both points:

1. **Authorization header removal** (blocking): Added `delete(reqHeaders, "Authorization")` after x-api-key injection to prevent dual auth headers, consistent with `claude_executor.go` behavior.
2. **Hostname() instead of Host** (non-blocking): Changed `isAnthropicAPI` to use `u.Hostname()` so `api.anthropic.com:443` correctly matches.

Ready for re-review.

PR #3353, opened 2026-05-12
fix(auth): cool down EOF transport failures

CitrusIce:
## Summary
- classify bare transport EOF and unexpected EOF failures as transient upstream errors so auth cooldown logic can suspend the failing credential briefly
- let lower-priority fallback credentials take over after 78code-style transport failures instead of immediately reselecting the same high-priority auth
- add a regression test that proves a higher-priority auth yielding unexpected EOF cools down and falls through to the next credential

## Testing
- go test ./sdk/cliproxy/auth -run 'TestManager_(TransientTransportError_CoolsDownHigherPriorityAuth|ModelSupportBadRequest_FallsBackAndSuspendsAuth|ShouldRetryAfterError_RespectsAuthRequestRetryOverride)'

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -1379,6 +1379,8 @@ func (m *Manager) executeMixedOnce(ctx context.Context, providers []string, req 
 				result.Error = &Error{Message: errExec.Error()}
 				if se, ok := errors.AsType[cliproxyexecutor.StatusError](errExec); ok && se != nil {
 					result.Error.HTTPStatus = se.StatusCode()
+				} else if status := classifyTransientTransportStatus(errExec); status != 0 {
+					result.Error.HTTPStatus = status
 				}
 				if ra := retryAfterFromError(errExec); ra != nil {
 					result.RetryAfter = ra
@@ -1467,6 +1469,8 @@ func (m *Manager) executeCountMixedOnce(ctx context.Context, providers []string,
 				result.Error = &Error{Message: errExec.Error()}
 				if se, ok := errors.AsType[cliproxyexecutor.StatusError](errExec); ok && se != nil {
 					result.Error.HTTPStatus = se.StatusCode()
+				} else if status := classifyTransientTransportStatus(errExec); status != 0 {
+					result.Error.HTTPStatus = status
 				}
 				if ra := retryAfterFromError(errExec); ra != nil {
 					result.RetryAfter = ra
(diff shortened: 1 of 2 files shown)


luispater (approved):
Thanks for the PR — this looks good to me.

## Summary
- Treats bare transport EOF / unexpected EOF as transient upstream failures (502) so the auth cooldown logic can temporarily suspend the failing credential and fall back to lower-priority auths.
- Refactors repeated execution error status mapping into a shared helper (`buildExecutionError`) and reuses it across Execute/CountTokens/ExecuteStream paths.
- Adds regression coverage for both non-stream and stream fallback behavior.

## Key findings
- Blocking: none.
- Non-blocking: consider letting `statusCodeFromError()` fall through to the transient transport classifier when `StatusCode()==0`, to avoid missing EOFs wrapped in a status-carrying error type that doesn't set a non-zero status.

## Test plan
- `go test ./sdk/cliproxy/auth -run 'TestManager_(TransientTransportError_CoolsDownHigherPriorityAuth|ExecuteStream_TransientTransportError_CoolsDownHigherPriorityAuth|ModelSupportBadRequestFallsBackAndSuspendsAuth|ShouldRetryAfterError_RespectsAuthRequestRetryOverride)$'`

luispater (approved):
## Summary

This PR classifies bare transport EOF and unexpected EOF failures as transient upstream errors (HTTP 502 Bad Gateway), enabling the auth cooldown logic to temporarily suspend the failing credential and fall through to lower-priority alternatives. A second commit refactors the repeated error-to-status mapping into a shared `buildExecutionError` helper, reducing duplication across Execute, CountTokens, ExecuteStream, and AntigravityCredits paths.

## Key findings

**Blocking:** None.

**Non-blocking:**
- The string-based EOF detection (`strings.Contains(lower, "unexpected eof")`) is a pragmatic fallback for errors that don't properly wrap `io.EOF`/`io.ErrUnexpectedEOF`. Low risk of false positives in practice, but worth a comment explaining why both `errors.Is` and string matching are needed.
- If an error implements `StatusCode()` returning 0, `statusCodeFromError` returns 0 without falling through to `classifyTransientTransportStatus`. This is an unlikely edge case but could be hardened in the future by checking `sc.StatusCode() != 0` before returning.

## Test plan

The PR includes two regression tests covering both non-streaming and streaming paths. Both verify that a higher-priority auth yielding "unexpected EOF" is cooled down and subsequent requests fall through to the next credential.

## Verdict

Clean, well-scoped fix with good test coverage and a sensible refactoring follow-up. Approve.

---

PR #3355, opened 2026-05-12
fix(openai): add missing `data: [DONE]` SSE terminal marker in /v1/re…

LennyFace24:
## Summary
- /v1/responses streaming endpoint was missing the `data: [DONE]` SSE terminal marker, causing OpenAI SDK clients (`openai>=2.6`) to throw `JSONDecodeError` after `response.completed`
- the `WriteDone` callback only wrote a `\n`; changed to emit `data: [DONE]\n\n`, matching the existing /v1/chat/completions behavior in `openai_handlers.go:680`

## Testing
- `go build ./sdk/api/handlers/openai/` passes


luispater:
OpenAI Responses do not require the `[DONE]`.

PR #3357, opened 2026-05-12
fix(cloak): preserve tool-use contracts in forwarded system prompt

NomenAK:
## Summary

`sanitizeForwardedSystemPrompt` replaces ANY non-empty input with a fixed 3-line generic reminder. That's effective against Anthropic's third-party-agent classifier but destroys legitimate BYOK client contracts.

Reproducible example: when Capy forwards a system prompt containing
> Visible messages to the human flow through one tool: \`message_user\`. Plain assistant text outside of \`message_user\` is treated as a reasoning trace and is NOT shown as a chat message. **You MUST call message_user for any user-facing reply.**

the entire contract — including the MUST language and the \`<communication>\` block — is replaced by:
> Use the available tools when needed to help with software engineering tasks.
> Keep responses concise and focused on the user's request.
> Prefer acting on the user's task over describing product-specific workflows.

Claude then defaults to plain text instead of the tool call, and Capy's UI renders the response in the wrong area.

Verified by direct replay (2026-05-12, claude-opus-4-7 via OAuth cloak):
- sanitize = current 3-line nuke → \`stop_reason: end_turn\`, plain text \"Bonjour !\" (3/3 trials)
- sanitize = no-op (full preserve) → \`stop_reason: tool_use\`, \`message_user({text: \"Salut Anton.\", isFinal: true})\` (3/3 trials)

## Approach

Replace the wholesale replacement with three selective strips that target the surface markers Anthropic's classifier is known to flag:

1. **URLs** (\`https?://...\`)
2. **Environment-banner XML blocks**: \`<env>\`, \`<environment>\`, \`<env_context>\`, \`<banner>\`, \`<workspace_status>\`, \`<available_skills>\`, \`<available_mcp_servers>\`, \`<available_integrations>\`
3. **Branded agent-introduction lines**: \"You are <Brand>\", \"powered/built/made by <Brand>\" for a short list of known third-party CLIs (Capy, OpenCode, Cursor, Windsurf, Devin, Cody, Continue.dev)

Everything else passes through verbatim — including \`<communication>\` blocks, MUST-language tool contracts, and general task instructions.

**Safety net**: if the strips reduce the text to under 80 chars (input was nearly pure branding/banners), fall back to the original 3-line generic reminder so Claude still has some task framing.

The cloak's primary defenses (CC sentinel \`system[1]\`, billing header \`system[0]\`, static Claude Code prompts \`system[2]\`, fake user_id, uTLS spoofing, CCH signing) are untouched.

## Test plan

- [x] \`TestSanitizeForwardedSystemPrompt_PreservesToolUseContract\` — \`<communication>\` + MUST language survives
- [x] \`TestSanitizeForwardedSystemPrompt_StripsThirdPartyMarkers\` — URLs, branded intro lines, \`<env_context>\` removed
- [x] \`TestSanitizeForwardedSystemPrompt_FallbackOnPureBranding\` — input entirely branding → falls back to generic reminder
- [x] \`TestSanitizeForwardedSystemPrompt_EmptyInputReturnsEmpty\` — empty / whitespace-only → empty
- [x] All existing \`TestCheckSystemInstructions*\` tests still green
- [x] Live replay against Anthropic via Claude OAuth cloak: 3/3 trials → \`message_user\` tool call (was: 3/3 plain text)

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


luispater (changes requested):
### Summary
This PR improves Claude OAuth cloaking behavior in two ways:
1) When `metadata.user_id` is present but not already Claude-Code-shaped, it derives a deterministic CC-shaped `user_id` from the inbound seed (instead of overwriting with a per-call random ID), improving cache attribution across calls.
2) It replaces `sanitizeForwardedSystemPrompt` “always nuke to 3-line reminder” behavior with selective stripping (URLs, env-banner XML blocks, branded agent intro lines) so legitimate forwarded tool-use contracts (e.g. “MUST call message_user”, `<communication>...</communication>`) survive.

### Blocking
- **Env-block stripping can over-delete across multiple same-tag blocks.**
  `compileEnvBlockPatterns()` currently builds patterns like:
  - `(?si)<env_context\b[^>]*>.*?</env_context>\s*`
  Because `.*` is greedy and Go regexp picks the leftmost-longest match, if a forwarded prompt contains multiple `<env_context>...</env_context>` blocks, the match can span from the first opening tag to the last closing tag and remove *everything in between*, including `<communication>` tool contracts (the exact thing this PR is trying to preserve).
  **Suggested fix:** make the inner match non-greedy (e.g. `.*?`) and add a regression test with:
  - `<env_context>...</env_context>`
  - `<communication>...MUST call message_user...</communication>`
  - `<env_context>...</env_context>`
  and assert the contract remains after sanitization.

### Non-blocking notes
- URL stripping via `https?://\S+` may swallow trailing punctuation/`)` in some real-world strings. If this causes surprising output, consider narrowing the URL character class.
- The deterministic `user_id` derivation is a good change (stable identity across calls for BYOK clients) and the new tests around stability / CC-shape / integration branches look solid.

### Test plan (what I ran locally)
- `go test ./internal/runtime/executor/helps -run TestDeterministicFakeUserID`
- `go test ./internal/runtime/executor -run InjectFakeUserID`
- `go test ./internal/runtime/executor -run SanitizeForwardedSystemPrompt`

(Also note: `go test ./internal/runtime/executor/...` currently fails on `dev` due to existing antigravity credits tests; failures appear unrelated to this PR.)

PR #3363, opened 2026-05-12
sync: merge v7.0.2 and harden Vertex prompt caching

RunhuaHuang:
## Summary
- merge upstream `v7.0.2` into the fork main branch
- keep the local Vertex prompt caching work and make it usable by caching stable prompt prefixes into Vertex cachedContents
- add one-shot fallback for stale or invalid `cachedContent` failures by evicting the local cache entry and retrying without cache

## Why
The previous Vertex cache path only cached a narrow subset of prompt state and could fail hard when a locally cached entry pointed at a remote cachedContent resource that had been deleted, expired, or become inaccessible. This change keeps the fork-specific Vertex cache behavior while making it operationally safe.

## Validation
- `go test ./...`

## Notes
This PR is opened from the fork `main` branch after syncing with upstream `v7.0.2`.


luispater (changes requested):
## Summary
This PR merges upstream `v7.0.2` and significantly improves Vertex prompt caching for Claude-compatible flows by:
- Creating/storing Vertex `cachedContents` for stable prompt prefixes (incl. system/tools/toolConfig + prefix contents).
- Applying `cachedContent` on subsequent requests and adjusting usage accounting to surface `cache_read_input_tokens`.
- Adding a one-shot fallback for cache-related 400/403/404 by evicting the local cache entry and retrying without cache.
It also introduces fork-specific “local run/update ergonomics” changes and new documentation/scripts.

## Blocking (must fix before merge)
1) **Do not commit generated binaries**
- The PR adds a compiled binary `cliproxy` (`Binary files /dev/null and b/cliproxy differ`). Please remove it from the PR (and consider adding it to `.gitignore` if needed locally).

2) **Fork-local helper scripts should not land in the main repo as-is**
- Four macOS `.command` scripts are added (`1.编译`, `2.启动服务`, `3.停止服务`, `4.重启服务`). These are platform-specific and assume a local workflow that also depends on the (now-committed) `./cliproxy` binary.
- If you want to keep them, please move them under a clearly-scoped location (e.g. `docs/` or `scripts/`), make them optional examples, and ensure they never require committing build artifacts.

3) **Auto-opening the management dashboard on every start needs opt-in**
- `internal/cmd/run.go` now spawns a goroutine that opens `.../management.html` after startup.
- This should be behind an explicit flag/config (or at least disabled by default) to avoid surprising behavior in headless/remote/server deployments.

4) **Remove or sanitize `mergerule.md`**
- `mergerule.md` includes a hard-coded local path (`cd '/Users/runhua/Runhua MBP/...'`) and fork-only workflow details.
- Please either remove it from the upstream PR or rewrite it as repo-agnostic documentation and consolidate with `docs/runhua-merge-playbook.md` to avoid duplication.

## Non-blocking
- The Vertex caching implementation itself looks well thought out (singleflight + TTL/backoff + targeted retry) and has good tests (`gemini_vertex_cache_test.go`, usage parsing tests, Claude translator tests, server proxy test for `/v1beta/cachedContents`).
- Minor style: `GeminiCachedContentsHandler` uses `defer upstreamResp.Body.Close()` without checking close errors.

## Suggested test plan
- `go test ./internal/translator/gemini/claude`
- `go test ./internal/api ./internal/api/modules/amp ./sdk/api/handlers/gemini`

## Notes on checks
- `translator-path-guard` failing is expected here because translator changes are part of the Vertex cache work (intentionally ignored per review instructions).

PR #3364, opened 2026-05-12
Add tls-skip-verify config option for Antigravity behind corp proxies

aeroxy:
Adds a `tls-skip-verify` config option that disables TLS certificate verification for outbound Antigravity connections (OAuth token exchange and API requests). Fixes the order of proxy/TLS configuration so that tls-skip-verify is layered on top of proxy transport rather than being overwritten by it.

Changes:
- internal/config/sdk_config.go: add TLSSkipVerify field
- internal/auth/antigravity/auth.go: apply proxy first, then TLS skip verify
- internal/runtime/executor/antigravity_executor.go: skipTLSVerifyTransport helper
- config.example.yaml: document the new option

## 1. High-Level Summary (TL;DR)
*   **Impact:** Medium - Introduces a critical configuration flag that allows disabling TLS certificate validation for outbound requests, primarily targeting users behind corporate proxies.
*   **Key Changes:**
    *   Introduced `tls-skip-verify` parameter in the SDK configuration.
    *   Updated Antigravity Auth initialization (`NewAntigravityAuth`) to respect the new TLS setting while maintaining proxy compatibility.
    *   Modified the Antigravity Executor (`newAntigravityHTTPClient`) to apply `InsecureSkipVerify` by cloning the existing HTTP transport.

## 2. Visual Overview (Code & Logic Map)

```mermaid
graph TD
    subgraph "Configuration (internal/config)"
        Config["SDKConfig"]
        TLSSkipVerify["TLSSkipVerify"]
        Config --> TLSSkipVerify
    end

    subgraph "Antigravity Auth (internal/auth/antigravity)"
        AuthInit["NewAntigravityAuth()"]
        AuthTransport["Set InsecureSkipVerify"]
        AuthInit --> |"Reads"| TLSSkipVerify
        AuthInit --> |"Modifies Transport"| AuthTransport
    end

    subgraph "Antigravity Executor (internal/runtime/executor)"
        ExecInit["newAntigravityHTTPClient()"]
        SkipTLSFunc["skipTLSVerifyTransport()"]
        ExecInit --> |"Reads"| TLSSkipVerify
        ExecInit --> |"Calls"| SkipTLSFunc
        SkipTLSFunc --> |"Clones & Modifies Transport"| ExecTransport["Set InsecureSkipVerify"]
    end

    classDef configNode fill:#e1f5fe,color:#01579b,stroke:#0288d1;
    classDef authNode fill:#e8f5e9,color:#1b5e20,stroke:#388e3c;
    classDef execNode fill:#fff3e0,color:#e65100,stroke:#f57c00;
    
    class Config,TLSSkipVerify configNode;
    class AuthInit,AuthTransport authNode;
    class ExecInit,SkipTLSFunc,ExecTransport execNode;
```

### Configuration Updates
Added a new configuration option to bypass TLS checks, aimed at environments intercepting TLS traffic.

| Key | Type | Description |
| :--- | :--- | :--- |
| `tls-skip-verify` | `bool` | Disables TLS certificate verification for outbound connections (e.g., Antigravity OAuth). Added to `config.example.yaml` (default commented out as `false`) and `SDKConfig` struct. |

### Auth Module (`internal/auth/antigravity/auth.go`)
*   **What Changed:** Modified the `NewAntigravityAuth` function to check for `cfg.TLSSkipVerify`. If enabled, it attempts to cast the proxy-configured `http.Client.Transport` to `*http.Transport`, clones it to preserve existing settings (like the proxy URL), and explicitly sets `InsecureSkipVerify = true`. If casting fails, it falls back to creating a new `http.Transport` with the insecure setting.

### Executor Module (`internal/runtime/executor/antigravity_executor.go`)
*   **What Changed:** Integrated TLS verification skipping into the `newAntigravityHTTPClient` function.
*   **Logic Refactoring:** Extracted the transport manipulation logic into a new dedicated helper function `skipTLSVerifyTransport(client *http.Client)`. This ensures that existing transports (e.g., those configured for proxies) are safely cloned and updated with `InsecureSkipVerify: true` without losing their original configuration.

## 4. Impact & Risk Assessment
*   **Security Risk:** ⚠️ Setting `tls-skip-verify: true` entirely removes TLS certificate validation, exposing outbound requests to potential Man-in-the-Middle (MitM) attacks. This should only be used in trusted environments (e.g., behind known corporate proxies).
*   **Breaking Changes:** None. The default behavior remains unchanged (strict TLS verification) since the config defaults to `false` implicitly.
*   **Testing Verified:**
    *   Verify that outbound requests to Antigravity endpoints succeed when intercepted by a local proxy (like Charles and Proxyman) when `tls-skip-verify` is `true`.
    *   Verify that setting `tls-skip-verify: true` simultaneously with a custom `proxy-url` works correctly, ensuring the proxy transport isn't overwritten but properly cloned.
    *   Ensure requests fail appropriately with certificate errors when intercepted by a proxy and `tls-skip-verify` is `false`.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -82,6 +82,11 @@ redis-usage-queue-retention-seconds: 60
 # Per-entry proxy-url also supports "direct" or "none" to bypass both the global proxy-url and environment proxies explicitly.
 proxy-url: ""
 
+# When true, disables TLS certificate verification for outbound connections (e.g. Antigravity OAuth).
+# Useful behind corporate proxies that intercept TLS traffic with their own certificates.
+# WARNING: This removes certificate validation — use only in trusted environments.
+# tls-skip-verify: false
+
 # When true, unprefixed model requests only use credentials without a prefix (except when prefix == model name).
 force-model-prefix: false
 
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -28,7 +28,8 @@ func (h *Handler) GetConfig(c *gin.Context) {
 		c.JSON(200, gin.H{})
 		return
 	}
-	c.JSON(200, new(*h.cfg))
+	cfg := *h.cfg
+	c.JSON(200, &cfg)
 }
 
 type releaseInfo struct {
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -127,7 +127,8 @@ func (m *AmpModule) Register(ctx modules.Context) error {
 		m.modelMapper = NewModelMapper(settings.ModelMappings)
 
 		// Store initial config for partial reload comparison
-		m.lastConfig = new(settings)
+		settingsCopy := settings
+		m.lastConfig = &settingsCopy
 
 		// Initialize localhost restriction setting (hot-reloadable)
 		m.setRestrictToLocalhost(settings.RestrictManagementToLocalhost)
--- a/internal/auth/antigravity/auth.go
+++ b/internal/auth/antigravity/auth.go
@@ -54,11 +54,11 @@ func NewAntigravityAuth(cfg *config.Config, httpClient *http.Client) *Antigravit
 			}
 			clone.TLSClientConfig.InsecureSkipVerify = true
 			client.Transport = clone
-		} else {
-			client.Transport = &http.Transport{
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-			}
-		}
+        } else if defaultTransport, ok := http.DefaultTransport.(*http.Transport); ok {
+            clone := defaultTransport.Clone()
+            clone.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+            client.Transport = clone
+        }
 	}
 	return &AntigravityAuth{
 		httpClient: client,
--- a/internal/auth/antigravity/auth.go
+++ b/internal/auth/antigravity/auth.go
@@ -3,6 +3,7 @@ package antigravity
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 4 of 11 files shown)


luispater (changes requested):
### Summary
This PR introduces a `tls-skip-verify` config option intended to disable TLS certificate verification for outbound Antigravity connections (OAuth + API), and adjusts the proxy/TLS layering order so the TLS setting isn’t overwritten by proxy transport setup.

### Blocking
- **Transport fallback breaks defaults + pooling (executor):**
  In `internal/runtime/executor/antigravity_executor.go`, `skipTLSVerifyTransport()` creates a brand-new `&http.Transport{TLSClientConfig: ...}` when `client.Transport == nil`. Because this runs *before* the existing `client.Transport == nil { client.Transport = antigravityTransport; return }` branch, enabling `tls-skip-verify` prevents use of the shared `antigravityTransport` singleton and creates a fresh transport/pool per call. It also loses `http.DefaultTransport` defaults (env proxy support, dialer timeouts/keepalive, idle conn settings, etc.).

  **Suggested fix:** keep the singleton behavior. For example, decide the base transport first (use `antigravityTransport` when nil), and for `tls-skip-verify` use a dedicated singleton insecure variant (e.g. `antigravityTransportInsecure`) rather than allocating a new transport per request.

- **Transport fallback breaks defaults (auth):**
  In `internal/auth/antigravity/auth.go`, when `tls-skip-verify` is enabled but no proxy transport exists (i.e. `util.SetProxy` leaves `Transport == nil`), the code falls back to `&http.Transport{TLSClientConfig: ...}`. This discards `http.DefaultTransport` behavior (including env proxies) and its tuned defaults.

  **Suggested fix:** when there is no existing `*http.Transport` to clone, start from a clone of `http.DefaultTransport` (or another appropriately configured baseline) and only toggle `InsecureSkipVerify` on that, rather than constructing an empty transport.

### Non-blocking / follow-ups
- `skipTLSVerifyTransport` only applies to `*http.Transport`; if the transport is a different `http.RoundTripper`, the config silently won’t apply. Consider logging a warning in that case.
- The `SDKConfig.TLSSkipVerify` comment mentions “other outbound HTTP clients” but the change only affects Antigravity right now.
- Consider logging a prominent warning when `tls-skip-verify` is enabled since it disables certificate validation (MitM risk).

### Test plan
- Not run locally.
- CI status observed as passing, but per request I did not use `translator-path-guard` / `pr-test-build` outcomes to determine the recommendation.

aeroxy:
@luispater should consider merging this w/o waitin for fixes on another PR

PR #3367, opened 2026-05-12
fix(api): emit "data: [DONE]" SSE terminator for /v1/responses streaming

### LINKED ISSUE #3352: Bug: /v1/responses streaming missing `data: [DONE]` terminal marker causes OpenAI SDK JSONDecodeError
## Bug Description

When CLIProxyAPI proxies `/v1/responses` streaming requests, the SSE stream does not emit a `data: [DONE]` terminal marker at the end. The OpenAI Python SDK (`openai>=2.6`) uses `sse.data.startswith("[DONE]")` as the stream termination signal (source: `openai/_streaming.py:63`). Without this marker, the SDK continues reading after `response.completed`, encounters the empty line written by `WriteDone`, parses it as an SSE event with empty `data`, and throws:

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

This affects **all upstream paths** that serve `/v1/responses` streaming — including:
- Codex/OpenAI Responses upstream passthrough
- OpenAI Chat Completions → Responses translation
- Anthropic/Claude → Responses translation

## Root Cause

In `sdk/api/handlers/openai/openai_responses_handlers.go`, the `WriteDone` callback in `forwardResponsesStream` (line ~570) only writes a newline character instead of the required `data: [DONE]` SSE event:

```go
WriteDone: func() {
    framer.Flush(c.Writer)
    _, _ = c.Writer.Write([]byte("\n"))
},
```

The OpenAI Responses API SSE specification requires `data: [DONE]` as the final event. When the upstream provider closes the connection cleanly, the OpenAI SDK *can* tolerate the absence — but when CLIProxyAPI injects an extra `\n` via `WriteDone`, the SDK interprets it as a new SSE event with an empty `data` field, triggering the JSONDecodeError.

## Reproduction

1. Configure CLIProxyAPI with any upstream provider that supports `/v1/responses`
2. Send a streaming request via the OpenAI Python SDK:
```python
from openai import OpenAI

client = OpenAI(api_key="<key>", base_url="http://<cliproxy-host>:8317/v1")

# Both paths fail after response.completed:
with client.responses.stream(model="<model>", input="ping") as stream:
    for event in stream:
        pass
    final = stream.get_final_response()  # 💥 JSONDecodeError

# Also fails:
resp = client.responses.create(model="<model>", input="ping", stream=True)
for chunk in resp:
    pass  # 💥 JSONDecodeError after response.completed
```

3. Observe the error after all streaming events (including `response.completed`) are received successfully.

## Expected Behavior

CLIProxyAPI should emit `data: [DONE]\n\n` after forwarding the last upstream SSE event, matching the OpenAI Responses API streaming specification.

## Suggested Fix

```go
WriteDone: func() {
    framer.Flush(c.Writer)
    _, _ = c.Writer.Write([]byte("data: [DONE]\n\n"))
},
```

This one-line change ensures all downstream OpenAI SDK clients correctly detect stream termination.

## Additional Note: Upstream `:HTTP_STATUS` SSE Comment Lines

Some upstream providers (e.g. Alibaba Cloud DashScope) inject `:HTTP_STATUS/200` comment lines between `event:` and `data:` lines in each SSE event block:

```
event:response.created
:HTTP_STATUS/200          ← non-standard comment splits the event
data: {...}
```

While these are valid SSE comments (per the SSE spec, `:`-prefixed lines are comments to be ignored), and the OpenAI Python SDK's `SSEDecoder` correctly ignores them when calling DashScope directly, they become problematic when proxied through CLIProxyAPI:

- The `responsesSSEFramer` reassembles split frames correctly (fixed in #2460), but it **does not strip** `:HTTP_STATUS` comment lines from the repaired frames
- This means downstream clients receive non-standard SSE format even though CLIProxyAPI is in a position to normalize it

Suggested: `responsesSSEFramer.WriteChunk()` should strip `:HTTP_STATUS/xxx` comment lines before processing, ensuring downstream clients receive clean, standard-compliant SSE regardless of upstream provider behavior.

## Environment

- CLIProxyAPI: v7.0.2
- Client: openai-python v2.31.0
- Upstream: Various providers serving `/v1/responses`

lihan3238:
## Summary

Fixes #3352. The `/v1/responses` streaming handler emitted a bare newline in `WriteDone` instead of the spec-required `data: [DONE]\n\n` terminator. OpenAI Python SDK 2.6+ uses `data: [DONE]` as the stream termination signal (see `openai/_streaming.py:63` — `sse.data.startswith("[DONE]")`). Without it, the SDK continues reading after `response.completed`, encounters the empty SSE event from the bare `\n`, and raises `JSONDecodeError`.

## Change

`sdk/api/handlers/openai/openai_responses_handlers.go` (1 line):

```diff
 WriteDone: func() {
     framer.Flush(c.Writer)
-    _, _ = c.Writer.Write([]byte("\n"))
+    _, _ = c.Writer.Write([]byte("data: [DONE]\n\n"))
 },
```

This matches the existing `/v1/chat/completions` handler at `openai_handlers.go:681` which already emits `data: [DONE]\n\n`. The two handlers are now consistent.

## Test

`sdk/api/handlers/openai/openai_responses_handlers_stream_test.go` — 8 existing assertions updated to expect the new terminator. Each `forwardResponsesStream` test now verifies that the body ends with a `data: [DONE]` event.

Local: `go test -race -count=1 ./sdk/api/handlers/openai/...` passes.

## Notes

- Path-guard safe: change is in `sdk/api/handlers/openai/`, not in `internal/translator/**`.
- Pre-existing race-condition failures in `internal/watcher/` (`TestPersistConfigAndAuthAsyncInvokePersister`, `TestScheduleConfigReloadDebounces`) reproduce on clean `main` — unrelated to this PR.
- The issue body also mentions a secondary fix (`:HTTP_STATUS` comment-line stripping in `responsesSSEFramer.WriteChunk`). Left out of this PR to keep the scope to the single SDK-blocking bug; happy to follow up in a separate PR if you'd like.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -520,7 +520,7 @@ func (h *OpenAIResponsesAPIHandler) handleStreamingResponse(c *gin.Context, rawJ
 				// Stream closed without data? Send headers and done.
 				setSSEHeaders()
 				handlers.WriteUpstreamHeaders(c.Writer.Header(), upstreamHeaders)
-				_, _ = c.Writer.Write([]byte("\n"))
+				_, _ = c.Writer.Write([]byte("data: [DONE]\n\n"))
 				flusher.Flush()
 				cliCancel(nil)
 				return
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -567,7 +567,7 @@ func (h *OpenAIResponsesAPIHandler) forwardResponsesStream(c *gin.Context, flush
 		},
 		WriteDone: func() {
 			framer.Flush(c.Writer)
-			_, _ = c.Writer.Write([]byte("\n"))
+			_, _ = c.Writer.Write([]byte("data: [DONE]\n\n"))
 		},
 	})
 }
--- a/sdk/api/handlers/openai/openai_responses_handlers_stream_test.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers_stream_test.go
@@ -45,8 +45,8 @@ func TestForwardResponsesStreamSeparatesDataOnlySSEChunks(t *testing.T) {
 	h.forwardResponsesStream(c, flusher, func(error) {}, data, errs, nil)
 	body := recorder.Body.String()
 	parts := strings.Split(strings.TrimSpace(body), "\n\n")
-	if len(parts) != 2 {
-		t.Fatalf("expected 2 SSE events, got %d. Body: %q", len(parts), body)
+	if len(parts) != 3 {
+		t.Fatalf("expected 3 SSE events (2 data + [DONE] terminator), got %d. Body: %q", len(parts), body)
 	}
 
 	expectedPart1 := "data: {\"type\":\"response.output_item.done\",\"item\":{\"type\":\"function_call\",\"arguments\":\"{}\"}}"
@@ -57,6 +57,10 @@ func TestForwardResponsesStreamSeparatesDataOnlySSEChunks(t *testing.T) {
 	expectedPart2 := "data: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp-1\",\"output\":[{\"type\":\"function_call\",\"arguments\":\"{}\"}]}}"
 	if parts[1] != expectedPart2 {
 		t.Errorf("unexpected second event.\nGot: %q\nWant: %q", parts[1], expectedPart2)
+	}
+
+	if parts[2] != "data: [DONE]" {
+		t.Errorf("expected final part to be SSE [DONE] terminator, got %q", parts[2])
 	}
 }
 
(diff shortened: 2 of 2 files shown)


lihan3238:
Good catch — addressed in 8661d4d. The empty-stream fast-path at `handleStreamingResponse` line 523 had the same bare-newline bug as the standard streaming path; both now emit `data: [DONE]\n\n`. This also brings the responses handler in line with `openai_handlers.go:501` (chat/completions empty-stream fast-path), so all three streaming exit points are consistent.

Local tests still pass (`go test -race ./sdk/api/handlers/openai/...`).

luispater (approved):
## Summary

This PR fixes `/v1/responses` streaming to emit the SSE terminator `data: [DONE]\n\n` on normal completion, and also on the empty-stream fast-path. This matches the existing `/v1/chat/completions` behavior and prevents clients from interpreting a bare newline as an empty SSE event.

## Findings

- **Blocking:** none.
- **Non-blocking:** consider adding a small unit test that covers the `handleStreamingResponse` empty-stream fast-path to lock in the `[DONE]` behavior.

## Test plan

- Ran: `go test -race -count=1 ./sdk/api/handlers/openai/...`

luispater:
OpenAI Responses do not require the [DONE].

lihan3238:
Thank you for the clarification — you're right, the Responses API uses `response.completed` as the terminal event, not the chat/completions-style `data: [DONE]`. I conflated the two streaming protocols and over-relied on the issue author's spec citation without independently verifying it against the actual Responses spec.

Apologies for the noise; lesson taken on board.

PR #3371, opened 2026-05-13
Expose model details in /v1/models

dos1989:
## Summary
- Return full registered model metadata from OpenAI-compatible `/v1/models` instead of filtering to `id/object/created/owned_by` only.
- Include token limits, supported modalities/generation methods, supported parameters, and thinking metadata.
- Add derived `thinking.supported_levels` and `thinking.level_budgets` so clients can choose reasoning effort from the model list.
- Add regression tests for registry metadata and `/v1/models` response.

## Tests
- `go test ./internal/registry -run 'TestGetAvailableModelsOpenAIIncludesModelDetailsAndThinking|TestGetAvailableModelsReturnsClonedSupportedParameters'`\n- `go test ./sdk/api/handlers/openai -run TestOpenAIModelsIncludesModelDetails`\n\n## Note\n- Full `go test ./internal/registry ./sdk/api/handlers/openai` currently hits an unrelated upstream failure: `TestCodexFreeModelsExcludeGPT55` expects free tier to exclude `gpt-5.5`.

### DIFF
--- a/internal/registry/model_registry.go
+++ b/internal/registry/model_registry.go
@@ -917,7 +917,7 @@ func cloneThinkingSupportMap(thinking *ThinkingSupport) map[string]any {
 	return result
 }
 
-func buildThinkingLevelDetails(thinking *ThinkingSupport) ([]string, map[string]int) {
+func buildThinkingLevelDetails(thinking *ThinkingSupport) ([]string, map[string]any) {
 	if thinking == nil {
 		return nil, nil
 	}
@@ -956,7 +956,7 @@ func buildThinkingLevelDetails(thinking *ThinkingSupport) ([]string, map[string]
 		}
 	}
 
-	levelBudgets := make(map[string]int)
+	levelBudgets := make(map[string]any)
 	for _, level := range levels {
 		for _, entry := range standardThinkingLevelBudgets {
 			if level == entry.level {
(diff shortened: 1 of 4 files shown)


luispater (approved):
## Summary

This PR enriches the OpenAI-compatible `/v1/models` endpoint to return full model metadata (token limits, modalities, generation methods, thinking support with derived `supported_levels` and `level_budgets`) instead of the previous minimal 4-field response. It also adds solid regression tests at both the registry and handler levels, including clone-safety assertions.

## Key findings

**Non-blocking:**

1. **Duplicated level-to-budget mapping** -- `standardThinkingLevelBudgets` (model_registry.go:872) replicates the values from `internal/thinking/convert.go:levelToBudgetMap`. A future change to one without the other would cause inconsistency. Consider referencing the canonical map or adding a cross-reference comment.

2. **Additive API surface change** -- The response payload grows substantially. Existing clients reading only `id`/`object`/`created`/`owned_by` are unaffected, but integrators that validate or snapshot the full response shape may need updating.

## What looks good

- Clone safety for thinking metadata is properly tested (mutate-and-re-fetch pattern).
- The handler-level test uses `defer UnregisterClient` for proper test isolation against the global registry.
- The second commit corrects field naming to `snake_case`, consistent with the OpenAI handler convention.
- `buildThinkingLevelDetails` correctly handles edge cases (nil thinking, empty levels, budget range filtering).
- The change aligns `OpenAIModels` with `OpenAIResponsesModels`, which already returned the full map.

## Test coverage

- Registry-level: `TestGetAvailableModelsOpenAIIncludesModelDetailsAndThinking` covers field presence, thinking derivation, and clone safety.
- Handler-level: `TestOpenAIModelsIncludesModelDetails` covers the full HTTP round-trip including JSON serialization.
- Both tests exercise the thinking level/budget derivation logic with explicit assertions.

LGTM -- approve.

---

PR #3372, opened 2026-05-13
Use lazy oauth for antigravity

aeroxy:
## 1. High-Level Summary (TL;DR)
*   **Impact:** Medium
*   **Key Changes:**
    *   ✨ **Lazy Auth Refresh:** Migrated the `antigravity` provider from a background auto-refresh loop to an on-demand, lazy refresh mechanism during auth selection.
    *   🚀 **Singleflight Integration:** Added `singleflight` to the Auth Manager to coalesce concurrent lazy refresh requests and prevent token refresh storms.
    *   ⚙️ **New Configuration:** Introduced `antigravity-use-default-project-id` to skip fetching the project ID and default to the base URL directly.
    *   🐛 **Concurrency & Bug Fixes:** Added a mutex lock to `GetConfig()` to fix data races, and prevented `NewAntigravityAuth()` from unconditionally overwriting an existing HTTP client transport.

## 2. Visual Overview (Code & Logic Map)

```mermaid
graph TD
    classDef default fill:#f3e5f5,color:#7b1fa2,stroke:#7b1fa2;
    classDef condition fill:#fff3e0,color:#e65100,stroke:#e65100;
    classDef action fill:#c8e6c9,color:#1a5e20,stroke:#1a5e20;
    classDef skip fill:#ffcdd2,color:#b71c1c,stroke:#b71c1c;

    subgraph "Auth Manager (conductor.go)"
        A("pickNext()"):::action --> B{"shouldRefresh()"}:::condition
        B -- "Yes" --> C("lazyRefreshAuth()"):::action
        B -- "No" --> D("Return Auth"):::action
        C --> E("singleflight.Do(id)"):::action
        E --> F("refreshAuth()"):::action
        F --> D
    end

    subgraph "Background Loop (auto_refresh_loop.go)"
        G("handleDueAuth()"):::action --> H{"Provider == 'antigravity'?"}:::condition
        H -- "Yes" --> I("Skip Refresh (Lazy)"):::skip
        H -- "No" --> J("Proceed Auto Refresh"):::action
    end
```

### ⚙️ Configuration Updates
Introduced a new flag to bypass project ID fetching when none is available.
| Key | Type | Default | Description |
| :--- | :--- | :--- | :--- |
| `antigravity-use-default-project-id` | `bool` | `false` | Skips fetching project ID via `loadCodeAssist` and uses the default base URL directly. Useful for environments lacking a project ID. |

### 🔐 Auth Manager & Conductor (`conductor.go`, `auto_refresh_loop.go`)
*   **Lazy Token Refresh:** Added `refreshAuthIfNeeded()` which is now invoked within `pickNext`, `pickNextLegacy`, `pickNextMixed`, and `pickNextMixedLegacy`.
*   **Concurrency Control:** Introduced `lazyRefreshGroup singleflight.Group` to ensure that multiple concurrent threads requesting the same expired auth token only trigger a single underlying HTTP refresh call.
*   **Background Loop Exclusion:** Hardcoded a bypass in `handleDueAuth()` so that `antigravity` auths are skipped by the background auto-refresh loop (Source: `auto_refresh_loop.go`).

### 🚀 Antigravity Executor (`antigravity_executor.go`)
*   **In-Place Refresh:** Changed `e.refreshToken(refreshCtx, auth.Clone())` to `e.refreshToken(refreshCtx, auth)`, updating the auth struct directly instead of a clone.
*   **Nil Safety:** Added widespread `e.cfg != nil` checks across multiple methods (`maybeRefreshAntigravityCreditsHint`, `Execute`, etc.) to prevent panics in unconfigured environments.
*   **Conditional Project ID:** Wrapped the `ensureAntigravityProjectID()` call with a check for the newly added `!e.cfg.AntigravityUseDefaultProjectID` flag.

### 🐛 HTTP Client & Handlers (`auth.go`, `config_basic.go`)
*   **Transport Safety:** Modified `NewAntigravityAuth()` to only clone and apply `http.DefaultTransport` (with `InsecureSkipVerify = true`) if `client.Transport == nil`. This preserves custom transports injected upstream.
*   **Data Race Fix:** Added `h.mu.Lock()` and `h.mu.Unlock()` around config struct dereferencing (`cfg := *h.cfg`) in the `GetConfig` management handler.

## 4. Impact & Risk Assessment
*   **⚠️ Breaking Changes:** None. This is a backward-compatible refactoring.
*   **🧪 Testing Suggestions:**
    *   **High Concurrency:** Flood the system with requests requiring `antigravity` tokens right after they expire to verify `singleflight` prevents multiple OAuth refresh calls.
    *   **Config Toggle:** Enable `antigravity-use-default-project-id: true` and verify that the system successfully bypasses project ID fetching without failing the auth phase.
    *   **Data Race Check:** Run tests with `-race` to ensure the `GetConfig` handler is now fully thread-safe.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -82,6 +82,11 @@ redis-usage-queue-retention-seconds: 60
 # Per-entry proxy-url also supports "direct" or "none" to bypass both the global proxy-url and environment proxies explicitly.
 proxy-url: ""
 
+# When true, disables TLS certificate verification for outbound connections (e.g. Antigravity OAuth).
+# Useful behind corporate proxies that intercept TLS traffic with their own certificates.
+# WARNING: This removes certificate validation — use only in trusted environments.
+# tls-skip-verify: false
+
 # When true, unprefixed model requests only use credentials without a prefix (except when prefix == model name).
 force-model-prefix: false
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -122,6 +122,10 @@ quota-exceeded:
   switch-preview-model: true # Whether to automatically switch to a preview model when a quota is exceeded
   antigravity-credits: true # Whether to use credits as last-resort fallback when all free-tier auths are exhausted for Claude models
 
+# Skips fetching project ID via loadCodeAssist and uses the default base URL directly.
+# Set to true when you don't have a project ID.
+# antigravity-use-default-project-id: false
+
 # Routing strategy for selecting credentials when multiple match.
 routing:
   strategy: "round-robin" # round-robin (default), fill-first
--- a/internal/api/handlers/management/api_key_usage.go
+++ b/internal/api/handlers/management/api_key_usage.go
@@ -48,9 +48,9 @@ func (h *Handler) GetAPIKeyUsage(c *gin.Context) {
 		return
 	}
 
-	h.mu.Lock()
+	h.mu.RLock()
 	manager := h.authManager
-	h.mu.Unlock()
+	h.mu.RUnlock()
 	if manager == nil {
 		c.JSON(http.StatusServiceUnavailable, gin.H{"error": "core auth manager unavailable"})
 		return
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -28,7 +28,8 @@ func (h *Handler) GetConfig(c *gin.Context) {
 		c.JSON(200, gin.H{})
 		return
 	}
-	c.JSON(200, new(*h.cfg))
+	cfg := *h.cfg
+	c.JSON(200, &cfg)
 }
 
 type releaseInfo struct {
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -28,7 +28,9 @@ func (h *Handler) GetConfig(c *gin.Context) {
 		c.JSON(200, gin.H{})
 		return
 	}
+	h.mu.Lock()
 	cfg := *h.cfg
+	h.mu.Unlock()
 	c.JSON(200, &cfg)
 }
 
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -28,9 +28,9 @@ func (h *Handler) GetConfig(c *gin.Context) {
 		c.JSON(200, gin.H{})
 		return
 	}
-	h.mu.Lock()
+	h.mu.RLock()
 	cfg := *h.cfg
-	h.mu.Unlock()
+	h.mu.RUnlock()
 	c.JSON(200, &cfg)
 }
 
--- a/internal/api/handlers/management/handler.go
+++ b/internal/api/handlers/management/handler.go
@@ -36,7 +36,7 @@ const attemptMaxIdleTime = 2 * time.Hour
 type Handler struct {
 	cfg                 *config.Config
 	configFilePath      string
-	mu                  sync.Mutex
+	mu                  sync.RWMutex
 	attemptsMu          sync.Mutex
 	failedAttempts      map[string]*attemptInfo // keyed by client IP
 	authManager         *coreauth.Manager
--- a/internal/api/modules/amp/amp.go
+++ b/internal/api/modules/amp/amp.go
@@ -127,7 +127,8 @@ func (m *AmpModule) Register(ctx modules.Context) error {
 		m.modelMapper = NewModelMapper(settings.ModelMappings)
 
 		// Store initial config for partial reload comparison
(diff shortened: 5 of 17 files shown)


luispater (commented):
## Summary

This PR migrates the antigravity provider from background auto-refresh to lazy on-demand OAuth refresh using `singleflight` for deduplication. It also adds `tls-skip-verify` config support, a new `antigravity-use-default-project-id` flag, and fixes several concurrency bugs (GetConfig data race, `new(value)` misuse, transport overwrite).

The architecture is sound: `refreshAuthIfNeeded` is called in `pickNext*` variants before cloning, `singleflight.Group` coalesces concurrent refresh attempts, and the background loop explicitly excludes antigravity via `nextRefreshCheckAt`. The mutex choreography in `pickNextLegacy`/`pickNextMixedLegacy` (release -> refresh -> re-acquire -> clone) is correct.

## Key Findings

**Blocking:**
- No tests are included for the new concurrency paths. This PR changes mutex acquisition order in hot-path auth selection and introduces singleflight coordination. A race-detector test (`go test -race`) and a unit test verifying singleflight coalesces concurrent refreshes are needed before merge.

**Non-blocking:**
- The nil-config behavior for `useCredits` and `maybeRefreshAntigravityCreditsHint` is inverted: `e.cfg == nil` now means "credits enabled" (previously "credits disabled"). Document this as intentional.
- `lazyRefreshAuth` discards the singleflight error silently. A debug log on failure would aid troubleshooting.
- 28 commits with back-and-forth changes should be squashed before merge.
- `GetConfig` uses exclusive lock for a read; consider `sync.RWMutex` if the handler sees traffic.

## Test Plan

The following should be verified before merge:
1. `go build -o test-output ./cmd/server && rm test-output` (compile check)
2. `go test -race ./sdk/cliproxy/auth/...` (race detector on conductor)
3. `go test -race ./internal/runtime/executor/...` (race detector on executor)
4. Manual: configure `antigravity-use-default-project-id: true` and verify auth works without project ID fetch
5. Manual: configure `tls-skip-verify: true` behind a corporate proxy and verify OAuth succeeds

This is an automated Claude Code review result and still requires manual verification by a human reviewer.

luispater (changes requested):
## Follow-up: Upgrading to Request Changes

After deeper analysis, two additional blocking issues were identified beyond the missing tests noted in the previous comment review.

## Blocking Issues

### 1. Data race: `Clone()` without lock in `pickNext` / `pickNextMixed`

In `pickNext` (scheduler fast path) and `pickNextMixed`, after `refreshAuthIfNeeded` returns, `selected.Clone()` is called without holding `m.mu.RLock()`:

```go
selected = m.refreshAuthIfNeeded(ctx, selected)
authCopy := selected.Clone()  // no lock held -- race with refreshAuth mutating the same *Auth
```

`refreshAuthIfNeeded` releases its internal `m.mu.RLock()` before returning, so the returned pointer is unprotected. The legacy variants (`pickNextLegacy`, `pickNextMixedLegacy`) correctly re-acquire the lock before cloning. The fast-path variants should do the same.

### 2. Inverted nil-guard logic in `antigravity_executor.go`

The nil-guard additions invert behavior when `e.cfg == nil`:

- `useCredits` lines: `antigravityCreditsRetryEnabled(e.cfg)` changed to `(e.cfg == nil || antigravityCreditsRetryEnabled(e.cfg))`. Since `antigravityCreditsRetryEnabled(nil)` returns `false`, the old expression disabled credits when config is nil. The new expression enables them (`true || false` = `true`). This is likely unintended.
- `maybeRefreshAntigravityCreditsHint`: `!antigravityCreditsRetryEnabled(e.cfg)` changed to `(e.cfg != nil && !antigravityCreditsRetryEnabled(e.cfg))`. When `e.cfg == nil`, the old code early-returned (skipped hint); the new code proceeds with the hint refresh without config.

Suggested fix: use `e.cfg != nil && antigravityCreditsRetryEnabled(e.cfg)` for the `useCredits` lines, and keep the original guard logic for `maybeRefreshAntigravityCreditsHint`.

### 3. No concurrency tests (from prior review)

This PR changes mutex acquisition order in hot-path auth selection and introduces singleflight coordination. A race-detector test and a unit test verifying singleflight coalesces concurrent refreshes are needed before merge.

## Non-blocking

- Unused `ctx` parameter in `lazyRefreshAuth` (replaced by `context.Background()` internally). Remove or document.
- 28 commits with several fix-on-fix and revert patterns. Squashing would improve readability.
- The `new(value)` refactoring is unrelated to lazy OAuth and could be split into a separate PR.
- `GetConfig` uses exclusive `Mutex.Lock()` for a read; consider upgrading to `RWMutex` if contention is a concern.

PR #3382, opened 2026-05-13
fix: scope antigravity credits fallback gate

sususu98:
## Summary
- skip the Antigravity credits fallback gate when the provider set does not include antigravity
- keep the existing credits config/error checks for Antigravity-capable routes
- add a regression test for Codex-only failures with unrelated Antigravity credentials present

## Tests
- go test ./sdk/cliproxy/auth

### DIFF
--- a/sdk/cliproxy/auth/antigravity_credits_test.go
+++ b/sdk/cliproxy/auth/antigravity_credits_test.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
 	internalconfig "github.com/router-for-me/CLIProxyAPI/v7/internal/config"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/registry"
 	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


luispater (approved):
## Summary

Clean, well-scoped fix that prevents non-antigravity routes (e.g. Codex-only) from entering the antigravity credits fallback gate. The refactoring extracts the provider-membership check into a standalone `hasAntigravityProvider` helper and short-circuits at the call sites in `Execute` and `ExecuteStream`.

## Key findings

**No blocking issues.**

Non-blocking:
- `hasAntigravityProvider` uses `strings.EqualFold` + `strings.TrimSpace` on already-normalized (lowered, trimmed) input from `normalizeProviders`. A plain `p == "antigravity"` would be equivalent and slightly more efficient, though the defensive style is consistent with the rest of the file.
- The `if len(providers) > 0` guard was removed from `shouldAttemptAntigravityCreditsFallback`. Both current call sites guarantee non-empty providers, so this is safe. A short doc comment noting the precondition would help future maintainers.

## Tests

The new `TestManagerExecuteStream_CodexOnlyDoesNotEnterAntigravityCreditsFallback` test is well-constructed: it registers both a codex executor and an antigravity executor, then verifies that a codex-only request never triggers the fallback log message. Good regression coverage.

## CI

All checks pass (build, agents-md-guard, translator-path-guard).

LGTM -- approve.

---

PR #3389, opened 2026-05-13
fix: Fix the incorrect toolcall format of the openai2codex

Hoshino-Yumetsuki:
Such as
```
Invalid schema for function 'background_cancel': None is not of type 'array'.
```

```
"message":"Invalid schema for function'get_changed_fils': None is not of
type
'array'.","type":"invalid_request_error","param":"tools[9].parameters", "code":"inv
alid_function_parameters"}
```


luispater (changes requested):
## Summary

This PR fixes a real Codex API rejection when tool schemas contain null values for fields that must be arrays (`required`, `enum`, `allOf`, `anyOf`, `oneOf`). The fix extracts a shared `NormalizeCodexToolSchema` utility into `internal/util/codex_schema.go` and integrates it into all four Codex translator paths (OpenAI chat-completions, OpenAI responses, Claude, Gemini).

The approach is sound: the iterative stack-based walker avoids stack overflow on deeply nested schemas, the sjson path escaping handles special characters and numeric keys correctly, and the package-level `codexPathReplacer` avoids repeated allocations.

## Blocking

- **Missing tests**: `internal/util/codex_schema.go` has no test file. The schema walker handles non-trivial edge cases (null array fields at various nesting depths, numeric object keys, special characters in property names). Please add unit tests covering at minimum:
  - A schema with a null `required` field at the top level
  - A schema with null `enum` nested inside `properties.someField`
  - A schema with a numeric property name (e.g. `"123"`)
  - A schema with special characters in property names (`.`, `*`, `?`, `:`, `\`)
  - An empty/null/invalid input
  - A deeply nested schema to validate the iterative approach works correctly

## Non-blocking

- The early return guard in `normalizeToolParameters` (Claude translator) for empty/null/invalid input is now redundant with the same check inside `NormalizeCodexToolSchema`. Consider removing it or adding a comment explaining why it is kept (e.g., to preserve the `string` return type early).
- The `:` prefix in `escapeCodexSchemaPathKey` for numeric keys is correct sjson behavior but non-obvious. A short comment like `// sjson treats bare numeric paths as array indices; prefix with ":" to force object key lookup` would help.
- PR title and commit messages are in Chinese. Project convention prefers English for commit messages.

PR #3390, opened 2026-05-13
fix(watcher): emit initial auth updates even when queue is not yet set

AdamsGH:
`prepareAuthUpdatesLocked` returned `nil` on the very first call if
`authQueue` was `nil` at that moment, dropping all `AuthUpdateActionAdd`
events for accounts discovered at startup.

The queue is wired up later via `setAuthUpdateQueue`, which starts the
dispatch loop, but by then there is nothing queued to send. Those
accounts never became visible to consumers until the next periodic
refresh.

The fix removes the early return: the updates slice is always built and
returned, and `dispatchAuthUpdates` already handles a `nil` queue
gracefully (it is a no-op).

### DIFF
--- a/internal/watcher/dispatcher.go
+++ b/internal/watcher/dispatcher.go
@@ -106,9 +106,6 @@ func (w *Watcher) prepareAuthUpdatesLocked(auths []*coreauth.Auth, force bool) [
 	}
 	if w.currentAuths == nil {
 		w.currentAuths = newState
-		if w.authQueue == nil {
-			return nil
-		}
 		updates := make([]AuthUpdate, 0, len(newState))
 		for id, auth := range newState {
 			updates = append(updates, AuthUpdate{Action: AuthUpdateActionAdd, ID: id, Auth: auth.Clone()})
--- a/internal/watcher/dispatcher.go
+++ b/internal/watcher/dispatcher.go
@@ -112,10 +112,6 @@ func (w *Watcher) prepareAuthUpdatesLocked(auths []*coreauth.Auth, force bool) [
 		}
 		return updates
 	}
-	if w.authQueue == nil {
-		w.currentAuths = newState
-		return nil
-	}
 	updates := make([]AuthUpdate, 0, len(newState)+len(w.currentAuths))
 	for id, auth := range newState {
 		if existing, ok := w.currentAuths[id]; !ok {
--- a/internal/watcher/dispatcher_test.go
+++ b/internal/watcher/dispatcher_test.go
@@ -1,0 +1,53 @@
+package watcher
+
+import (
+	"testing"
+
+	coreauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+)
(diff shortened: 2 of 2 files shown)


AdamsGH:
Good catch. Removed both nil-queue guards from `prepareAuthUpdatesLocked` (initial and subsequent calls). `dispatchAuthUpdates` already handles a nil queue gracefully, so the checks were redundant and dropped updates in both cases.

luispater (approved):
## Summary

Clean bug fix that removes two unnecessary nil-queue early returns in `prepareAuthUpdatesLocked`. The guards caused auth updates to be silently dropped when `authQueue` had not yet been wired up via `setAuthUpdateQueue`, which happens after the first `refreshAuthState` call at startup.

## Analysis

The fix is correct:

1. **First guard (commit 1):** When `currentAuths == nil` (first call) and `authQueue == nil`, the old code returned nil, dropping all initial Add events. After the fix, the Add updates are always built and returned. `dispatchAuthUpdates` already handles a nil queue gracefully (returns early at line 143-145), so no panic or channel-send-on-nil occurs.

2. **Second guard (commit 2):** On subsequent refreshes with `authQueue == nil`, the old code set `currentAuths = newState` but returned nil without computing the diff. This left consumers unaware of changes. After the fix, the diff is always computed and `currentAuths` is still updated at line 135. Again, `dispatchAuthUpdates` safely no-ops when the queue is nil.

The function is always called under `clientsMutex.Lock()`, so removing the `authQueue` reads inside it has no concurrency implications -- in fact it simplifies the locking story since `prepareAuthUpdatesLocked` no longer touches `authQueue` at all.

## Checklist

- **Correctness:** Fix is sound. Edge cases (empty auths, nil auth entries) are handled by the existing guards at lines 102-104.
- **Compatibility:** No API or behavioral change for consumers that already have the queue set up. Only fixes the startup race.
- **Performance:** Negligible -- builds an updates slice that may be immediately discarded by `dispatchAuthUpdates`. The slice is small (number of auth accounts).
- **Tests:** Existing tests pass (they set `authQueue` to a buffered channel). No new test covers the nil-queue scenario -- would be a nice addition but not blocking given the simplicity of the change.

## Non-blocking suggestion

Consider adding a unit test like:

```go
func TestPrepareAuthUpdatesLockedNilQueue(t *testing.T) {
    w := &Watcher{} // currentAuths nil, authQueue nil
    updates := w.prepareAuthUpdatesLocked([]*coreauth.Auth{{ID: "a", Provider: "p"}}, false)
    if len(updates) != 1 || updates[0].Action != AuthUpdateActionAdd || updates[0].ID != "a" {
        t.Fatalf("expected Add update even with nil queue, got %+v", updates)
    }
}
```

---

AdamsGH:
Added `dispatcher_test.go` with two tests covering the nil-queue scenario as suggested: one for the initial call (no existing state) and one for subsequent calls (verifies the diff is still computed and `currentAuths` is updated correctly).

PR #3391, opened 2026-05-13
feat(auth): add -claude-label flag for multi-org Claude accounts

AdamsGH:
Users with multiple Claude organizations need to log in to each one
independently. Without this, running `-claude-login` a second time
overwrites `claude-<email>.json`, invalidating the first session.

This adds a `Label` field to `LoginOptions` (both the `sdk/auth` and
`internal/cmd` layers) and a `-claude-label` CLI flag. When set, the
auth file is named `claude-<email>-<label>.json` instead of
`claude-<email>.json`, so multiple org tokens can coexist in one
auth-dir.

Usage:

```
# personal org (default, no change)
./CLIProxyAPIPlus -claude-login

# work org
./CLIProxyAPIPlus -claude-login -claude-label work
```

Both files live in the same `auth-dir` and are picked up independently
by the conductor.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -64,6 +64,7 @@ func main() {
 	var codexLogin bool
 	var codexDeviceLogin bool
 	var claudeLogin bool
+	var claudeLabel string
 	var noBrowser bool
 	var oauthCallbackPort int
 	var antigravityLogin bool
@@ -84,6 +85,7 @@ func main() {
 	flag.BoolVar(&codexLogin, "codex-login", false, "Login to Codex using OAuth")
 	flag.BoolVar(&codexDeviceLogin, "codex-device-login", false, "Login to Codex using device code flow")
 	flag.BoolVar(&claudeLogin, "claude-login", false, "Login to Claude using OAuth")
+	flag.StringVar(&claudeLabel, "claude-label", "", "Label suffix appended to the auth filename (claude-<email>-<label>.json); allows multiple Claude orgs to coexist in one auth-dir")
 	flag.BoolVar(&noBrowser, "no-browser", false, "Don't open browser automatically for OAuth")
 	flag.IntVar(&oauthCallbackPort, "oauth-callback-port", 0, "Override OAuth callback port (defaults to provider-specific port)")
 	flag.BoolVar(&antigravityLogin, "antigravity-login", false, "Login to Antigravity using OAuth")
@@ -516,6 +518,7 @@ func main() {
 	options := &cmd.LoginOptions{
 		NoBrowser:    noBrowser,
 		CallbackPort: oauthCallbackPort,
+		Label:        claudeLabel,
 	}
 
 	// Register the shared token store once so all components use the same persistence backend.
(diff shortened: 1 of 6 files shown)


AdamsGH:
Fixed. Added `ValidateLabel` in `sdk/auth/interfaces.go` that restricts the label to `[a-zA-Z0-9_-]` and returns an error for anything else. `DoClaudeLogin` calls it before starting the OAuth flow.

AdamsGH:
Added `ValidateLabel` call in `ClaudeAuthenticator.Login` itself, so the check happens at the SDK boundary regardless of how the caller constructs `LoginOptions`.

luispater (changes requested):
## Summary

This PR adds a `-claude-label` CLI flag that lets users with multiple Claude organizations save separate auth files (`claude-<email>-<label>.json`) without overwriting each other. The implementation is clean, well-scoped, and backward-compatible.

## Blocking

- **Missing tests**: `ValidateLabel` is an exported pure function used at two call sites. It needs unit tests covering at minimum: valid labels (`"work"`, `"org-2"`, `"my_team"`), invalid labels (`"a/b"`, `"foo bar"`, `"../etc"`), and the empty-string pass-through. The filename construction in `claude.go` (label vs no-label path) should also have a test.

## Non-blocking

- **Error reporting style**: In `DoClaudeLogin`, the validation failure uses `fmt.Println(err)` while the rest of the function uses `log.Error(...)` or `fmt.Printf(...)`. Consider using `log.Errorf("invalid claude-label: %v", err)` for consistency with structured logging.
- **No length cap**: The regex validates characters but not length. A 200-character label combined with a long email could exceed the 255-byte filesystem filename limit. A simple `len(label) > 32` guard in `ValidateLabel` would prevent this edge case.
- **Management API gap**: The web-based OAuth callback in `auth_files.go` also creates `claude-<email>.json` without label support. If multi-org is intended to work from the management UI too, this should be tracked as a follow-up.

## Verdict

Request changes for the missing test coverage. The feature logic itself is sound.

AdamsGH:
Addressed all review points:

**Blocking:**
- Added `TestValidateLabel` covering valid labels (`work`, `org-2`, `my_team`, alphanumerics), invalid labels (`a/b`, `foo bar`, `../etc`, special chars), empty string pass-through, and the length boundary (32 chars ok, 33 fails).
- Added `TestClaudeFilenameConstruction` covering the no-label and labeled filename paths.

**Non-blocking:**
- `DoClaudeLogin`: replaced `fmt.Println(err)` with `log.Errorf("invalid claude-label: %v", err)` for consistency with the rest of the function.
- `ValidateLabel`: added `len(label) > 32` guard before the regex check to prevent excessively long labels from hitting filesystem limits.

The management UI gap (OAuth callback in `auth_files.go`) is noted for a follow-up.

AdamsGH:
One more fix: the filename construction test was previously testing a local helper function rather than the real code in `claude.go`. Extracted `claudeFileName(email, label)` as a package-level function, wired `Login` to use it, and updated `TestClaudeFilenameConstruction` to call it directly.

PR #3392, opened 2026-05-13
feat(observability): expose upstream request/response data in gin context

AdamsGH:
`RecordAPIRequest` and `AppendAPIResponseChunk` previously wrote nothing
to the gin context when `RequestLog` was disabled. External plugins
(tracing, billing, observability sidecars) had no way to read the
upstream URL, the request content, or the response without forcing
`request_log: true` for all users.

This adds four context keys that are written unconditionally, regardless
of `RequestLog`:

| Key | Content |
|---|---|
| `upstream_url` | outbound upstream URL |
| `upstream_first_user_msg` | text of the first user message (max 2000 chars) |
| `upstream_response_text` | accumulated response text from streaming or full body (max 4000 chars) |
| `upstream_raw_usage` | `map[string]int64` of token counts from `usage` fields |

The keys are exported as named constants from the `helps` package so
plugins can reference them without hardcoding strings.

The existing `RequestLog`-gated logging path is unchanged.

### DIFF
--- a/internal/runtime/executor/helps/logging_helpers.go
+++ b/internal/runtime/executor/helps/logging_helpers.go
@@ -605,6 +605,37 @@ func extractFirstUserText(body []byte) string {
 		}
 	}
 
+	// Responses API (Codex): input as string or input[].content[].text
+	if input := gjson.GetBytes(body, "input"); input.Exists() {
+		var text string
+		if input.Type == gjson.String {
+			text = strings.TrimSpace(input.String())
+		} else if input.IsArray() {
+			input.ForEach(func(_, item gjson.Result) bool {
+				if role := item.Get("role").String(); role != "" && role != "user" {
+					return true
+				}
+				item.Get("content").ForEach(func(_, block gjson.Result) bool {
+					if block.Get("type").String() == "input_text" {
+						if t := strings.TrimSpace(block.Get("text").String()); t != "" {
+							text = t
+							return false
+						}
+					}
+					return true
+				})
+				return text == ""
+			})
+		}
+		if text != "" {
+			const maxLen = 2000
+			if len(text) > maxLen {
+				text = text[:maxLen] + "..."
+			}
+			return text
+		}
+	}
+
 	// Anthropic/OpenAI: messages[].content
 	messages := gjson.GetBytes(body, "messages")
 	if !messages.Exists() || !messages.IsArray() {
@@ -702,6 +733,10 @@ func accumulateResponseText(ginCtx *gin.Context, chunk []byte) {
 			if gjson.GetBytes(data, "type").String() != "response.output_text.delta" {
 				delta = ""
 			}
+		}
+		if delta == "" {
+			// Gemini streaming: response.candidates[0].content.parts[0].text
+			delta = gjson.GetBytes(data, "response.candidates.0.content.parts.0.text").String()
 		}
 		if delta == "" {
 			continue
--- a/internal/runtime/executor/helps/logging_helpers.go
+++ b/internal/runtime/executor/helps/logging_helpers.go
@@ -732,6 +732,21 @@ func accumulateResponseText(ginCtx *gin.Context, chunk []byte) {
 		if text == "" {
 			text = gjson.GetBytes(chunk, "response.candidates.0.content.parts.0.text").String()
 		}
+		// Codex Responses API: top-level output[].content[].text (compact/non-streaming)
+		if text == "" {
+			gjson.GetBytes(chunk, "output").ForEach(func(_, item gjson.Result) bool {
+				item.Get("content").ForEach(func(_, block gjson.Result) bool {
+					if block.Get("type").String() == "output_text" {
+						if t := strings.TrimSpace(block.Get("text").String()); t != "" {
+							text = t
+							return false
+						}
+					}
+					return true
+				})
+				return text == ""
+			})
+		}
 		// Codex Responses API: response.completed event - response.output[].content[].text
 		if text == "" {
 			gjson.GetBytes(chunk, "response.output").ForEach(func(_, item gjson.Result) bool {
--- a/internal/runtime/executor/helps/logging_helpers.go
+++ b/internal/runtime/executor/helps/logging_helpers.go
@@ -25,6 +25,12 @@ const (
 	apiResponseKey          = "API_RESPONSE"
 	apiWebsocketTimelineKey = "API_WEBSOCKET_TIMELINE"
 	creditsUsedKey          = "__antigravity_credits_used__"
+
+	// Context keys written regardless of RequestLog; available to plugins and middleware.
+	UpstreamURLKey          = "upstream_url"
+	UpstreamFirstUserMsgKey = "upstream_first_user_msg"
+	UpstreamResponseTextKey = "upstream_response_text"
+	UpstreamRawUsageKey     = "upstream_raw_usage"
 )
 
 // UpstreamRequestLog captures the outbound upstream request details for logging.
@@ -55,6 +61,18 @@ type upstreamAttempt struct {
 
 // RecordAPIRequest stores the upstream request metadata in Gin context for request logging.
 func RecordAPIRequest(ctx context.Context, cfg *config.Config, info UpstreamRequestLog) {
+	// Always-on: store the upstream URL and first user message for plugins.
+	if ginCtx := ginContextFrom(ctx); ginCtx != nil {
+		if info.URL != "" {
+			ginCtx.Set(UpstreamURLKey, info.URL)
+		}
+		if len(info.Body) > 0 {
+			if msg := extractFirstUserText(info.Body); msg != "" {
+				ginCtx.Set(UpstreamFirstUserMsgKey, msg)
+			}
+		}
+	}
+
 	if cfg == nil || !cfg.RequestLog {
 		return
 	}
@@ -153,6 +171,11 @@ func RecordAPIResponseError(ctx context.Context, cfg *config.Config, err error) 
 
 // AppendAPIResponseChunk appends an upstream response chunk to Gin context for request logging.
 func AppendAPIResponseChunk(ctx context.Context, cfg *config.Config, chunk []byte) {
+	// Always-on: accumulate response text for plugins.
+	if ginCtx := ginContextFrom(ctx); ginCtx != nil {
+		accumulateResponseText(ginCtx, chunk)
+	}
+
 	if cfg == nil || !cfg.RequestLog {
 		return
 	}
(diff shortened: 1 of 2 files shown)


AdamsGH:
Applied all four suggestions:

1. `RecordAPIRequest` now resets `UpstreamResponseTextKey` and `UpstreamRawUsageKey` at the start of each attempt so retries do not mix stale data with the final response.
2. Non-streaming fallback: added `choices.0.message.content` after the Anthropic paths.
3. SSE loop: read the current value once before the loop, write back only if something changed; delta extraction now uses `delta.text` directly (Anthropic) and falls back to `choices.0.delta.content` (OpenAI-compatible).
4. Usage: added `prompt_tokens` and `completion_tokens` to the tracked keys alongside the Anthropic-specific ones.

AdamsGH:
Added two more extraction paths: `candidates.0.content.parts.0.text` for Gemini/Vertex non-streaming, and `response.output_text.delta` for Codex Responses API streaming.

AdamsGH:
Added Gemini support in both extraction points:

- `extractFirstUserText`: now checks `contents[].parts[].text` (Gemini/Vertex format) before falling back to the Anthropic/OpenAI `messages[]` path.
- `accumulateUsage`: reads `usageMetadata.{promptTokenCount,candidatesTokenCount,totalTokenCount}` from the top-level response when present.

AdamsGH:
Two more format additions:

- `extractFirstUserText`: handles Responses API `input` field — string form and array form (`input[].content[].text` with `type=input_text`).
- Streaming SSE: extracts `response.candidates.0.content.parts.0.text` for Gemini streaming events.

AdamsGH:
`accumulateUsage` now checks `response.usage` first (Responses API `response.completed` event), then falls back to the top-level `usage` field. Gemini `usageMetadata` path is checked independently as before.

AdamsGH:
Added the remaining two paths:

- Non-streaming: `response.candidates.0.content.parts.0.text` for Antigravity-wrapped responses.
- Streaming: `candidates.0.content.parts.0.text` for native Gemini SSE (top-level, no `response.` wrapper) and `response.candidates.0.content.parts.0.text` for Antigravity-wrapped Gemini SSE.

AdamsGH:
Two more fixes:

- **Content block iteration**: replaced `content.#.text` (which gjson returns as a JSON array) with a `ForEach` walk that finds the first `type=text` block. Now works correctly when a thinking or tool-use block precedes the text block.
- **Stale `total_tokens`**: removed the `== 0` guard - `total_tokens` is now updated whenever the sum of component tokens exceeds the current stored value, so late-arriving `output_tokens` events are reflected correctly.

AdamsGH:
Added `response.output_item.done` handling: when the event type matches, walks `item.content[]` looking for a block with `type=output_text` and extracts its `text` field. This covers the fallback path the translator tests reference.

AdamsGH:
Added both Antigravity paths:

- `extractFirstUserText`: checks `request.contents` before the top-level `contents` path.
- `accumulateUsage`: checks `response.usageMetadata` alongside top-level `usageMetadata`.

AdamsGH:
Two more fixes:

- **Claude `message_start`**: `message.usage` is now tried as a fallback after `response.usage` and `usage`, so prompt/cache token counts from streaming `message_start` events are captured.
- **OpenAI nested details**: `prompt_tokens_details.cached_tokens` → `cache_read_input_tokens` and `output_tokens_details.reasoning_tokens` → `reasoning_tokens` are now read explicitly alongside the top-level fields.

AdamsGH:
Two more fixes:

- **Codex text duplication**: the SSE loop now tracks a `sawOutputTextDelta` flag. `response.output_item.done` text is only extracted when no `response.output_text.delta` events were seen first, so streams that include both don't produce duplicated text.
- **`total_tokens` inflation**: `reasoning_tokens` is a subset of `output_tokens`/`completion_tokens` and must not be added again in the derived total. Removed it from the derived sum; explicit `total_tokens` from the upstream payload is still preserved as a lower bound.

AdamsGH:
Added `response.output[].content[]` extraction for the `response.completed` event: walks output items and their content blocks looking for `type=output_text`, same shape the Codex executor reads at `codex_executor.go:263-274`.

AdamsGH:
Added top-level `output[].content[]` extraction alongside `response.output[]`, covering compact/non-streaming Responses objects where the output array sits at the root rather than under a `response.completed` wrapper.

AdamsGH:
Now checks both `output_tokens_details.reasoning_tokens` (Responses API) and `completion_tokens_details.reasoning_tokens` (chat completions) when populating `reasoning_tokens`.

AdamsGH:
`RecordAPIWebsocketRequest` now sets `upstream_url`, `upstream_first_user_msg`, and resets `upstream_raw_usage` before the `RequestLog` guard, matching the always-on behaviour of `RecordAPIRequest`.

luispater (changes requested):
## Summary

This PR adds four always-on gin context keys (`upstream_url`, `upstream_first_user_msg`, `upstream_response_text`, `upstream_raw_usage`) that are populated regardless of the `RequestLog` config flag, enabling external plugins to access upstream request/response data. The implementation covers Anthropic, OpenAI, Gemini, Codex Responses API, and Antigravity wrapper formats for both streaming and non-streaming paths.

## Blocking Issues

1. **No tests.** 417 lines of new parsing logic across multiple provider formats with zero test coverage. The `helps` package already has test files -- please add unit tests for `extractFirstUserText`, `accumulateResponseText`, and `accumulateUsage` covering at least one case per provider format and the edge cases (empty body, malformed JSON, truncation at maxLen boundaries).

2. **`total_tokens` double-counting.** The derived total formula sums both Anthropic-style (`input_tokens + output_tokens`) and OpenAI-style (`prompt_tokens + completion_tokens`) fields. When a provider reports both (or when the code maps one to the other via `usageMetadata`), the total is inflated. Use `max(input_tokens, prompt_tokens) + max(output_tokens, completion_tokens)` or guard against double-counting.

3. **String concatenation in hot loop.** The SSE streaming path does `current += delta` on every chunk, creating O(n^2) allocations for large responses. Use a `strings.Builder` stored in the gin context to accumulate text efficiently.

## Non-blocking

- No consumer of the new context keys exists in the codebase yet. Consider adding at least a middleware or test that reads them to validate the contract.
- `extractFirstUserText` handles 4 formats in one function -- consider splitting into named helpers for each format.
- The Anthropic streaming delta extraction no longer checks `delta.type == "text_delta"`, which could capture non-text deltas (e.g. `input_json_delta`).
- `RecordAPIWebsocketRequest` resets `UpstreamRawUsageKey` but not `UpstreamResponseTextKey` (inconsistent with `RecordAPIRequest`).
- Consider squashing the 17 fixup commits before merge for a cleaner history.

## Test Plan

The PR needs unit tests. Suggested coverage:
- `extractFirstUserText`: Anthropic messages body, OpenAI messages body, Gemini contents body, Responses API input (string + array), Antigravity wrapper, empty/malformed input, truncation at 2000 chars
- `accumulateResponseText`: non-streaming (each provider format), SSE streaming (Anthropic delta, OpenAI delta, Codex delta, Gemini delta), maxLen truncation, empty chunks
- `accumulateUsage`: each usage path (response.usage, usage, message.usage, usageMetadata, response.usageMetadata), setMax behavior, total_tokens derivation correctness

AdamsGH:
All three blocking issues addressed:

**1. Tests** - added `observability_helpers_test.go` with 39 subtests covering:
- `extractFirstUserText`: Anthropic/OpenAI messages (string + array content), Gemini `contents[]`, Antigravity `request.contents[]`, Responses API `input` (string + array), empty body, malformed JSON, truncation boundary
- `accumulateResponseText`: non-streaming (Anthropic, OpenAI, Gemini, Antigravity, Codex `output[]`, Codex `response.output[]`), SSE streaming (Anthropic delta, OpenAI delta, Codex `response.output_text.delta`, Gemini native + Antigravity, accumulation across chunks, `output_item.done` with and without prior delta), truncation
- `accumulateUsage`: Anthropic, OpenAI, Claude `message.usage`, Responses API `response.usage`, Gemini `usageMetadata`, Antigravity `response.usageMetadata`, nested details (cached + reasoning tokens), setMax behaviour, empty body

**2. `total_tokens` double-counting** - replaced `input+output+prompt+completion` with `max(input,prompt) + max(output,completion)` so providers that populate both field families don't inflate the total.

**3. O(n^2) string concatenation** - replaced `current += delta` with a `*strings.Builder` stored in the gin context under a private key, reused across chunk calls.

**Bonus fix found by the tests:** `sawOutputTextDelta` was a local variable reset on every `accumulateResponseText` call, so the `output_item.done` dedup guard didn't survive across separate chunk calls. Now persisted in the gin context alongside the builder.

AdamsGH:
Non-blocking items addressed:

- **Anthropic delta type guard**: now checks `delta.type == "text_delta"` (or absent) before reading `delta.text`, so `input_json_delta` and other non-text delta events are not captured.
- **WebSocket reset**: `RecordAPIWebsocketRequest` now resets `UpstreamResponseTextKey` alongside `UpstreamRawUsageKey`, consistent with `RecordAPIRequest`.
- **Consumer**: the 39-subtest test file added in the previous commit reads all four context keys, which validates the contract.
- **Squashing**: happy to squash before merge if that is the project preference - just say the word.

AdamsGH:
Two more fixes:

- **Retry accumulator reset**: `responseTextBuilderKey` and `sawOutputTextDeltaKey` are now promoted to package-level constants and reset in `RecordAPIRequest` alongside the public keys, so a retried attempt starts with a clean builder and no stale saw-delta flag.
- **Websocket SSE path**: `AppendAPIWebsocketResponse` now wraps the raw JSON frame in `data: ...\n\n` before calling `accumulateResponseText`, so Codex `response.output_text.delta` and `response.output_item.done` frames reach the SSE streaming path instead of being mishandled by the non-streaming JSON branch.

AdamsGH:
Two more usage fields:

- **Responses API cached tokens**: now checks `input_tokens_details.cached_tokens` alongside `prompt_tokens_details.cached_tokens` for `cache_read_input_tokens`.
- **Gemini thinking/cache**: both `usageMetadata` and `response.usageMetadata` blocks now also map `thoughtsTokenCount` -> `reasoning_tokens` and `cachedContentTokenCount` -> `cache_read_input_tokens`.

AdamsGH:
Two more fixes:

- **Gemini parts iteration**: replaced fixed `parts.0.text` lookups with a `geminiPartsText()` helper that iterates all parts and returns the first non-empty text field. Applied to both the non-streaming and SSE streaming paths (native and Antigravity).
- **snake_case `usage_metadata`**: the early-return guard and both accumulation blocks now check `usage_metadata` / `response.usage_metadata` alongside `usageMetadata` / `response.usageMetadata`.

AdamsGH:
`AppendAPIWebsocketResponse` now tries the non-streaming JSON path first, so `response.completed` frames with `response.output[]` are handled correctly. SSE wrapping is only applied when the non-streaming pass found no text, avoiding double-accumulation for completed frames while still handling `response.output_text.delta` and `response.output_item.done` events via the streaming branch.

AdamsGH:
Replaced the non-empty check with explicit routing by `type` field: `response.output_text.delta` and `response.output_item.done` frames always go through the SSE path so all delta frames in a multi-frame stream are captured; everything else goes through the non-streaming path for `response.completed` / `response.output[]` handling.

AdamsGH:
Moved `total_tokens` derivation to after both `usageMetadata` blocks so Gemini prompt/candidate counts are already in `current` when the total is computed. `reasoning_tokens` (`thoughtsTokenCount`) is included in the fallback only when `outputSide == 0`, which is the pure Gemini metadata case where thoughts are independent of candidates. For OpenAI responses `reasoning_tokens` is already a subset of `output_tokens`, so it is not added again.

AdamsGH:
Replaced the `outputSide == 0` heuristic with a `sawGeminiMeta` flag set when either `usageMetadata` or `response.usageMetadata` block is processed. When the flag is set, `reasoning_tokens` (`thoughtsTokenCount`) is included in the fallback total, matching `parseGeminiFamilyUsageDetail` in `usage_helpers.go`. For OpenAI responses the flag is never set so `reasoning_tokens` is not double-counted.

AdamsGH:
Content blocks with no `type` field are now treated as `input_text` in the Responses API `input[]` path, matching the converter behaviour.

PR #3393, opened 2026-05-13
feat(examples): Langfuse observability plugin

AdamsGH:
Adds a working example of a [Langfuse](https://langfuse.com) observability
plugin built on top of the `usage.Plugin` interface.

**Depends on #3392** (always-on gin context keys). Without that PR the
plugin still works but `input`, `output` and token counts will be empty
unless the provider populates them through `Record.Detail`.

## What is added

`examples/langfuse-plugin/` - a standalone binary that can be built and
run in place of the standard `CLIProxyAPIPlus`:

```
examples/langfuse-plugin/
  main.go              entry point, registers plugin and starts the service
  langfuse/
    client.go          thin HTTP client for Langfuse Ingestion API
    plugin.go          usage.Plugin implementation
  README.md
```

`sdk/cliproxy/usage/manager.go` - four exported constants for the gin
context keys written by the runtime (`CtxUpstreamURL`, `CtxFirstUserMsg`,
`CtxResponseText`, `CtxRawUsage`). These let plugin authors reference the
keys without depending on `internal/`.

## How it works

The plugin implements `HandleUsage(ctx, record)`. For each upstream
request it produces one Langfuse `generation-create` event:

| Field | Source |
|---|---|
| `traceId` | `__request_id__` from gin context, or fresh UUID |
| `model` | `record.Model` |
| `input` | `upstream_first_user_msg` gin key |
| `output` | `upstream_response_text` gin key |
| `usage` | `upstream_raw_usage` gin key, falls back to `record.Detail` |
| `metadata` | provider, auth_id, latency_ms, upstream_url |

The HTTP call is fire-and-forget in a goroutine.

## Usage

```sh
go build -o cpa-langfuse ./examples/langfuse-plugin

LANGFUSE_BASE_URL=https://cloud.langfuse.com \
LANGFUSE_PUBLIC_KEY=pk-lf-... \
LANGFUSE_SECRET_KEY=sk-lf-... \
./cpa-langfuse -config config.yaml
```

When the env vars are absent the binary starts normally with no Langfuse
traffic.

### DIFF
--- a/examples/langfuse-plugin/README.md
+++ b/examples/langfuse-plugin/README.md
@@ -1,0 +1,53 @@
+# Langfuse observability plugin
+
+This example shows how to build a CLIProxyAPI binary that forwards every
+upstream request to [Langfuse](https://langfuse.com) as a generation span.
+
+## How it works
+
+`sdk/cliproxy/usage` exposes a `Plugin` interface. The proxy runtime calls
+`HandleUsage` after each upstream request completes. This plugin packages the
+record into a Langfuse `generation-create` event and ships it in a background
+goroutine so the response path is not blocked.
+
+When an upstream gateway populates `X-Request-Id`, the span is attached to
+the matching parent trace. Without it a standalone trace is created per
+request.
+
+For richer input/output capture the proxy runtime also writes per-request
+context keys (see `sdk/cliproxy/usage` constants). These are populated when
+the runtime's observability path is active and contain the first user message,
+accumulated response text, and a token usage breakdown.
+
+## Build
+
+```sh
+go build -o cpa-langfuse ./examples/langfuse-plugin
+```
+
+## Run
+
+```sh
+LANGFUSE_BASE_URL=https://cloud.langfuse.com \
+LANGFUSE_PUBLIC_KEY=pk-lf-... \
+LANGFUSE_SECRET_KEY=sk-lf-... \
+./cpa-langfuse -config config.yaml
+```
+
+If the environment variables are not set the binary starts normally without
+sending any data to Langfuse.
+
+## What gets sent
+
+Each upstream request produces one `generation` event:
+
+| Field | Value |
+|---|---|
+| `traceId` | `X-Request-Id` from the inbound request, or a fresh UUID |
+| `name` | `cpa.upstream` |
+| `model` | upstream model name |
+| `input` | first user message text (when available) |
+| `output` | response text (when available) |
+| `usage` | input/output/cache/reasoning token counts |
+| `metadata` | provider, auth_id, latency_ms, upstream_url |
+| `level` | `ERROR` on upstream failure, `DEFAULT` otherwise |
--- a/examples/langfuse-plugin/langfuse/client.go
+++ b/examples/langfuse-plugin/langfuse/client.go
@@ -87,10 +87,27 @@ func (c *Client) ingest(ctx context.Context, events ...ingestionEvent) error {
 		return fmt.Errorf("langfuse: http: %w", err)
 	}
 	defer resp.Body.Close()
+	respBody := make([]byte, 2048)
+	n, _ := resp.Body.Read(respBody)
+	respBody = respBody[:n]
 	if resp.StatusCode >= 300 {
-		body := make([]byte, 512)
-		n, _ := resp.Body.Read(body)
-		return fmt.Errorf("langfuse: status %d: %s", resp.StatusCode, bytes.TrimSpace(body[:n]))
+		return fmt.Errorf("langfuse: status %d: %s", resp.StatusCode, bytes.TrimSpace(respBody))
+	}
+	// Langfuse returns 207 Multi-Status when the batch is accepted but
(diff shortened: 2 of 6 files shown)


AdamsGH:
Fixed two issues:
- `requestIDFrom`: added nil guard on `ctx` before calling `ctx.Value`
- `main`: replaced hardcoded `"config.yaml"` with a `-config` flag

The unbounded goroutines point is noted. For a first example the overhead is acceptable (one goroutine per request with an 8s timeout), but a worker pool would be the right approach if this moves into the main binary.

AdamsGH:
Addressed the remaining feedback:

- **Unique generation IDs**: generation `ID` is now `uuid.New()` per `HandleUsage` call, so multiple records emitted for the same request (image tool, etc.) get distinct IDs and do not overwrite each other in Langfuse. `TraceID` still uses the request ID to link them to the same trace.
- **Trailing slash in `NewClient`**: stripped at construction time so callers do not have to sanitize it.
- **Error response body**: `ingest` now reads up to 512 bytes of the response body on non-2xx replies and includes it in the error string.

The `-config` flag and `nil` ctx guard were already fixed in the previous round.

AdamsGH:
Addressed the new round of feedback:

- **Gin context race**: `HandleUsage` builds the full `GenerationBody` synchronously (including all `gc.Get` calls in `buildMeta`/`buildUsage`/input+output) before sending to the worker channel. The worker only receives the already-snapshotted struct, never holds a reference to `gc`.
- **HTTP 207 Multi-Status**: `ingest` now decodes the 207 body and returns an error if any event has a per-event error from Langfuse.
- **Cache tokens**: switched from `rec.Detail.CachedTokens` to `rec.Detail.CacheReadTokens` as the fallback for `cache_read_input_tokens`.
- **`log.Fatalf`**: replaced with a `run()` helper that returns errors; `main` logs and exits with code 1.
- **`CtxFirstUserMsg` not written in this repo**: noted - these keys are populated by PR #3392 (`logging_helpers.go`). The dependency is documented in the PR description. Without #3392 merged the fields will be empty, which is expected and stated in the README.

AdamsGH:
Removed both timeouts per the AGENTS.md policy (timeouts allowed only during credential acquisition): the `http.Client{Timeout: 10s}` and the `context.WithTimeout(8s)` wrapper in the worker loop. The worker channel cap (256) already provides the only back-pressure needed.

AdamsGH:
Found in live testing: Langfuse v3 does not auto-create a trace when it receives a `generation-create` for an unknown `traceId`. `SendGeneration` now sends a `trace-create` event in the same batch before the generation, so the generation always has a parent trace to attach to. Re-sending the same trace ID on subsequent calls is idempotent on the Langfuse side.

AdamsGH:
`godotenv.Load()` is now called at the top of `run()` before `NewFromEnv()`, so credentials in a `.env` file in the working directory are picked up automatically. Errors are silently ignored since the file is optional.

luispater (approved):
## Summary

Adds a Langfuse observability plugin example (`examples/langfuse-plugin/`) demonstrating the `usage.Plugin` interface, plus 4 exported context-key constants in `sdk/cliproxy/usage/manager.go`. The example is well-structured with a worker pool, graceful degradation when credentials are absent, and proper Langfuse v3 trace+generation batching.

## Key findings

**Non-blocking:**

1. **defer resp.Body.Close() style** (`langfuse/client.go`): Project convention requires wrapping defer errors in a closure. Low impact here since the body is read before return, but worth aligning for consistency.

2. **No worker shutdown path**: The `Plugin` struct spawns 4 workers but provides no `Close()` method. For an example binary this is fine, but a brief comment noting this is intentional would help readers.

3. **No HTTP timeout on telemetry calls**: Patch 7/9 removes all timeouts to comply with AGENTS.md. The policy seems aimed at upstream proxy connections, not auxiliary third-party calls. If Langfuse is unreachable, workers block indefinitely until the queue fills and events are dropped. Consider adding a comment explaining the rationale.

4. **No tests**: The SDK constants are additive and compile-checked by CI, but a simple test asserting the constant values match the runtime keys would guard against drift.

5. **Dependency on #3392**: The PR body notes this clearly. Merge order matters -- without #3392, the gin context keys won't be populated.

## CI

All checks pass. No translator changes involved.

## Verdict

Approve. The code is clean, iteratively hardened across 9 commits, and follows project conventions. The non-blocking items are suggestions for future improvement.

---

AdamsGH:
Addressed the non-blocking findings:

- `defer resp.Body.Close()` wrapped in a closure (`func() { _ = resp.Body.Close() }()`)
- Added comments to `Plugin` and `worker()` explaining the intentional no-`Close()` and no-timeout design (AGENTS.md policy + bounded queue as back-pressure)
- Added `TestContextKeyConstants` in `sdk/cliproxy/usage` to guard the four constant string values against accidental rename

AdamsGH:
Added blank-import of `sdk/translator/builtin` to register request/response translators at startup, matching `cmd/server/main.go` and `examples/translator/main.go`.

PR #3437, opened 2026-05-16
Strip Claude Code attribution from non-Claude upstream translations

madwiki:
## Summary
- Strip the Claude Code `x-anthropic-billing-header` attribution system block when translating Claude-format requests to non-Claude upstream formats.
- Apply the strip across OpenAI-compatible providers, Gemini, Gemini CLI, Antigravity, and Codex translations.
- Add shared attribution detection plus focused translator tests for each affected route.

## Problem
Claude Code prepends an attribution/billing system block such as `x-anthropic-billing-header: cc_version=...; cc_entrypoint=cli; cch=...;`. That block is meaningful for the official Anthropic Claude channel, but it becomes prompt pollution when CLIProxyAPI routes the Claude Code client to non-Anthropic upstream channels.

For example, `Claude Code -> CLIProxyAPI -> OpenRouter -> Gemini` currently translates the dynamic attribution block into an OpenAI-compatible system message. Because the block carries per-request fingerprint data, upstream prompt/gateway caching sees a different prompt prefix and can miss cache repeatedly, increasing cost for users routing Claude Code through OpenRouter or similar OpenAI-compatible providers.

## Safety
This change is channel/translator based, not model-name based. It only strips attribution while translating Claude-format requests to non-Claude upstream formats. The direct Claude Code -> official Anthropic Claude path still uses the Claude executor/`to=claude` path and remains unchanged, preserving Claude Code attribution/fingerprint behavior for users relying on the official Anthropic channel.

This also means Claude-family models served through non-Anthropic channels such as Antigravity or OpenRouter do not receive the Claude Code attribution block, which has no useful meaning there and can harm cache stability.

## Testing
- `go test ./internal/util ./internal/translator/openai/claude ./internal/translator/gemini/claude ./internal/translator/gemini-cli/claude ./internal/translator/antigravity/claude ./internal/translator/codex/claude`
- `go test ./...`

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -330,6 +330,7 @@ func (s *Server) setupRoutes() {
 	{
 		v1.GET("/models", s.unifiedModelsHandler(openaiHandlers, claudeCodeHandlers))
 		v1.POST("/chat/completions", openaiHandlers.ChatCompletions)
+		v1.POST("/audio/transcriptions", openaiHandlers.AudioTranscriptions)
 		v1.POST("/completions", openaiHandlers.Completions)
 		v1.POST("/messages", claudeCodeHandlers.ClaudeMessages)
 		v1.POST("/messages/count_tokens", claudeCodeHandlers.ClaudeCountTokens)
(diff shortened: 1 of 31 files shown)


madwiki:
Closing this PR for now because it was prepared from my fork's local main rather than resynced from the upstream repository before submission. I will rebase/resync against the upstream branch before any resubmission.

PR #3438, opened 2026-05-16
fix: strip Claude Code attribution from non-Anthropic translations

madwiki:
## Summary
- Add a shared Claude attribution detector for the Claude Code `x-anthropic-billing-header` system block.
- Strip that attribution block when translating Claude-format requests to non-Anthropic upstream formats: OpenAI-compatible, Gemini, Gemini CLI, Antigravity, and Codex.
- Add focused tests for each affected translation path.

## Problem
Claude Code can prepend a system block like `x-anthropic-billing-header: cc_version=...; cc_entrypoint=cli; cch=...;`. That block is useful for the official Anthropic Claude channel, but it is per-request dynamic metadata.

When a Claude Code client is routed through CLIProxyAPI to a non-Anthropic upstream channel, for example `Claude Code -> CLIProxyAPI -> OpenRouter -> Gemini`, the block is translated into the target provider prompt. Because `cch` and related attribution content can vary per request, upstream gateway/provider prompt caches see a changing prefix and can miss cache repeatedly.

## Approach
The fix is translation-layer based rather than model-name based. Claude-family models can be served by non-Anthropic channels such as OpenRouter or Antigravity, and those channels do not need Claude Code attribution metadata. Conversely, official Anthropic Claude traffic should keep the Claude Code attribution path intact.

This PR therefore only strips the attribution block while converting Claude-format requests into non-Anthropic target formats. The direct Anthropic Claude executor path (`to=claude`) is not changed, so Claude Code attribution/fingerprint behavior for the official Anthropic channel is preserved.

## Testing
- `go test ./internal/util ./internal/translator/openai/claude ./internal/translator/gemini/claude ./internal/translator/gemini-cli/claude ./internal/translator/antigravity/claude ./internal/translator/codex/claude`

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -101,7 +101,7 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 			systemTypePromptResult := systemPromptResult.Get("type")
 			if systemTypePromptResult.Type == gjson.String && systemTypePromptResult.String() == "text" {
 				systemPrompt := systemPromptResult.Get("text").String()
-				if strings.HasPrefix(systemPrompt, "x-anthropic-billing-header:") {
+				if util.IsClaudeCodeAttributionSystemText(systemPrompt) {
 					continue
 				}
 				partJSON := []byte(`{}`)
@@ -112,7 +112,7 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 				hasSystemInstruction = true
 			}
 		}
-	} else if systemResult.Type == gjson.String {
+	} else if systemResult.Type == gjson.String && !util.IsClaudeCodeAttributionSystemText(systemResult.String()) {
 		systemInstructionJSON = []byte(`{"role":"user","parts":[{"text":""}]}`)
 		systemInstructionJSON, _ = sjson.SetBytes(systemInstructionJSON, "parts.0.text", systemResult.String())
 		hasSystemInstruction = true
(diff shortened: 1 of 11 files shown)


luispater (approved):
## Summary

Clean, well-scoped PR that extracts a shared `IsClaudeCodeAttributionSystemText` detector into `internal/util` and applies it consistently across all non-Anthropic translation paths. The approach of stripping at the translation layer (rather than model-name based) is sound and preserves attribution for direct Anthropic traffic.

Build passes. Tests are focused and cover the primary (array-type system) path for four of five translators.

## Key findings

**Non-blocking:**

1. **Over-broad filter in `convertClaudeContentPart` (OpenAI translator):** This shared helper is called for user/assistant message content parts (line 161), not just system prompts. Adding the attribution check here means a user message starting with `x-anthropic-billing-header:` would be silently dropped. The practical risk is negligible (Claude Code only injects this as a system block), but architecturally it would be cleaner to limit the check to system-context callers, or add a comment explaining the design choice.

2. **Missing test for string-type system path:** All four translators add filtering for `"system": "<string>"` (not just `"system": [...]`), but no test exercises this path. One test case would close the gap.

3. **No Codex attribution test:** The other four translators each have a dedicated `_StripsClaudeCodeAttribution` test; the Codex translator does not.

4. **Minor behavioral expansion in Codex:** The old check was `strings.HasPrefix(text, "x-anthropic-billing-header: ")` (trailing space required). The new utility matches without trailing space and trims leading whitespace. This is correct but slightly broader.

## Test plan

The PR's suggested test command covers the main paths:
```
go test ./internal/util ./internal/translator/openai/claude ./internal/translator/gemini/claude ./internal/translator/gemini-cli/claude ./internal/translator/antigravity/claude ./internal/translator/codex/claude
```

## Verdict

Approve. The findings above are non-blocking improvements that can be addressed in a follow-up.

---

luispater (approved):

PR #3470, opened 2026-05-18
Fix Antigravity Gemini thought signatures

sususu98:
## Summary
- replace client-provided Gemini thoughtSignature values with skip_thought_signature_validator in the Antigravity Gemini path
- keep the behavior scoped to non-Claude models and align it with the existing Gemini CLI signature handling
- add coverage for functionCall and text parts carrying client signatures

## Validation
- go test ./internal/translator/antigravity/gemini

@hkfires could you help review?

### DIFF
--- a/internal/translator/antigravity/gemini/antigravity_gemini_request.go
+++ b/internal/translator/antigravity/gemini/antigravity_gemini_request.go
@@ -99,35 +99,19 @@ func ConvertGeminiRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 	}
 
 	// Gemini-specific handling for non-Claude models:
-	// - Add skip_thought_signature_validator to functionCall parts so upstream can bypass signature validation.
-	// - Also mark thinking parts with the same sentinel when present (we keep the parts; we only annotate them).
-	if !strings.Contains(modelName, "claude") {
+	// - Replace client-provided thoughtSignature values with the skip sentinel.
+	// - Add the same sentinel to functionCall and thinking parts so upstream can bypass signature validation.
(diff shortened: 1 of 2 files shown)


sususu98:
Reproduction notes from production error logs, sanitized.

Observed failure shape:

- Downstream endpoint: `POST /v1beta/models/gemini-3.1-pro-preview:streamGenerateContent?alt=sse`
- Upstream provider/path: Antigravity, `POST https://daily-cloudcode-pa.googleapis.com/v1internal:streamGenerateContent?alt=sse`
- Effective upstream model: `gemini-pro-agent` via alias `gemini-3.1-pro-preview`
- Request body shape: Gemini `contents` history with prior `role: "model"` parts carrying client-provided `thoughtSignature` values.
- One sampled failing request had `22` `thoughtSignature` fields on historical model text parts and no `functionCall` parts.
- Upstream response:

```json
{
  "error": {
    "code": 400,
    "message": "Corrupted thought signature.",
    "status": "INVALID_ARGUMENT"
  }
}
```

Local replay using the copied production log/auth material:

- Original logged request body: `400 INVALID_ARGUMENT`, `Corrupted thought signature.` in about 1.7s.
- Same request body after replacing every non-Claude `role=model` part `thoughtSignature` with `skip_thought_signature_validator`: `200` SSE response, first chunk received in about 6.2s.

A second sampled request with historical `functionCall` parts also reproduced the same original `400`; after the sentinel replacement it no longer returned the signature error within the same immediate window.

No credentials, API keys, proxy details, or full prompt bodies are included here; the important reproducer is the presence of client-provided Gemini `thoughtSignature` values in historical `role=model` parts forwarded through the Antigravity path.

PR #3533, opened 2026-05-24
fix: send streaming keepalives before first chunk

GrothKeiran:
## Summary
- send SSE headers/keep-alive heartbeats before the first upstream chunk to avoid Cloudflare first-byte 524s
- replace per-handler peek loops with shared `ForwardStream` header-commit/error handling
- update Claude, Gemini, OpenAI chat/completions/responses, and image streaming handlers
- enable example streaming keepalive config

## Tests
- `go test ./...`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -132,9 +132,9 @@ enable-gemini-cli-endpoint: false
 # When > 0, emit blank lines every N seconds for non-streaming responses to prevent idle timeouts.
 nonstream-keepalive-interval: 0
 # Streaming behavior (SSE keep-alives + safe bootstrap retries).
-# streaming:
-#   keepalive-seconds: 15   # Default: 0 (disabled). <= 0 disables keep-alives.
-#   bootstrap-retries: 1    # Default: 0 (disabled). Retries before first byte is sent.
+streaming:
+  keepalive-seconds: 15   # Default: 0 (disabled). <= 0 disables keep-alives.
+  bootstrap-retries: 1    # Default: 0 (disabled). Retries before first byte is sent.
 
 # Signature cache validation for thinking blocks (Antigravity/Claude).
 # When true (default), cached signatures are preferred and validated.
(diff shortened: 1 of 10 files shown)


luispater:
Thanks for the PR. The requested behavior can already be achieved by configuring `streaming.keepalive-seconds`, which sends SSE keep-alive heartbeats for streaming responses before upstream data arrives. Because this does not need a separate code change, I am closing this PR.

GrothKeiran:
Thanks for reviewing! However, I believe the existing `streaming.keepalive-seconds` config alone does **not** solve the Cloudflare 524 timeout issue. Here's why:

**Root cause:** Cloudflare's free tier enforces a ~100s first-byte timeout. The current streaming handlers all use a "peek first chunk" pattern — they wait for upstream data before writing SSE headers (`Content-Type: text/event-stream`, etc.). If the upstream provider is slow to respond, **zero bytes** are sent to the client during the wait, triggering HTTP 524.

**What `keepalive-seconds` does:** It sends `: keep-alive\n\n` SSE heartbeats **after** streaming has started (i.e., after headers are already committed). This helps prevent mid-stream timeouts but does nothing for the initial wait period before headers are sent.

**What this PR changes:** The `commitHeaders()` logic in `StreamForwardOptions` ensures SSE headers are written **before** any upstream data arrives — either on the first keep-alive tick or before the first data chunk. This is what actually prevents the 524: the client (and Cloudflare) receive headers within seconds, establishing the connection as active before upstream responds.

Without this change, even with `keepalive-seconds: 15` configured, the server still sends nothing to the client until the first upstream chunk — because `setSSEHeaders` is called inside the handler's peek loop, not before it.

In short:
- `keepalive-seconds` = keeps an **already-started** stream alive
- This PR = ensures the stream **starts** (headers sent) before upstream data arrives

Both are needed to fully prevent 524 errors behind Cloudflare. Happy to discuss further or rework the PR if there's a simpler approach I'm missing.

PR #3546, opened 2026-05-25
fix(auth): fail fast on unavailable session-affinity auth for /v1/responses

tianxiaochannel-oss88:
## Summary
- Fail fast with a retryable `session_auth_unavailable` error and HTTP 503 when `/v1/responses` session-affinity has a cached auth that is temporarily unavailable.
- Preserve the existing fallback/reselect behavior for non-Responses endpoints.
- Keep the original session binding intact so a restored pinned auth can continue the same stateful Responses session.

## Root cause
`/v1/responses` continuations can depend on encrypted/reasoning state tied to the auth that created the session. The previous cache-hit-unavailable path silently selected another auth and replaced the cache binding, which can break state continuity for stateful Responses workflows.

## Validation
- `go test ./sdk/cliproxy/auth ./sdk/api/handlers`

### DIFF
--- a/sdk/cliproxy/auth/selector.go
+++ b/sdk/cliproxy/auth/selector.go
@@ -504,6 +504,15 @@ func (s *SessionAffinitySelector) Pick(ctx context.Context, provider, model stri
 				return auth, nil
 			}
 		}
+		if requestPath, strict := sessionAffinityStrictLeasePath(opts); strict {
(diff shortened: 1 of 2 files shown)


luispater:
Thanks for the PR.

CLIProxyAPI is designed to complete requests whenever possible instead of stopping on recoverable routing or auth-selection issues. This change turns a previously recoverable session-affinity auth-unavailable case into a fail-fast 503 for `/v1/responses`, which breaks the existing design logic and fallback behavior.

Closing this PR for that reason.

PR #3564, opened 2026-05-26
Preserve reasoning_content for DeepSeek models

VacantHusky:
## Summary

This PR adds a new configuration option:

`openai-responses-preserve-reasoning-content`

When enabled, reasoning messages returned by the Responses API will no longer be removed.

## Motivation

Some DeepSeek reasoning models require the original `reasoning_content`
to be passed back in subsequent requests.

Previously, reasoning messages could be stripped during processing,
which caused requests to fail with errors similar to:

```json
{
  "error": {
    "message": "The `reasoning_content` in the thinking mode must be passed back to the API."
  }
}
```

## Changes

* Added `openai-responses-preserve-reasoning-content` config option
* Preserve reasoning messages in Responses API flow
* Fix compatibility issues with DeepSeek reasoning models

## Result

DeepSeek models using reasoning mode can now work correctly without
triggering missing `reasoning_content` errors.


luispater:
The `translator-path-guard` rule exists for a reason. Please submit translator-related requests as GitHub issues instead of changing `internal/translator/` directly in a pull request.

PR #3595, opened 2026-05-28
fix(amp): restore response tool casing from request

### LINKED ISSUE #3592: Claude Code tool calls fail because Anthropic response tool_use.name is lowercased (Glob -> glob)
## Summary

When using Claude Code through CLIProxyAPI's Anthropic Messages route with a Codex-backed model alias, Claude Code advertises built-in tools with TitleCase names such as `Glob`, but the streamed response returned by CLIProxyAPI contains a lowercase tool call name `glob`.

Claude Code tool dispatch is case-sensitive, so the local executor rejects the call with:

```text
No such tool available: glob
```

This appears to be a missing response-side tool name reverse mapping on the Anthropic `/v1/messages` streaming path.

## Environment

- CLIProxyAPI: `v7.1.24`
  - observed startup line:
    ```text
    CLIProxyAPI Version: v7.1.24, Commit: 94c1b25, BuiltAt: 2026-05-27T19:00:34Z
    ```
- Client: Claude Code `2.1.152`
- Route:
  ```text
  Claude Code
    -> local proxy
    -> new-api /v1/messages?beta=true
    -> CLIProxyAPI /api/provider/anthropic/v1/messages
    -> Codex-backed provider
  ```
- Model alias:
  ```text
  gpt-5.5
  ```
- new-api route info:
  ```text
  request_path: /v1/messages
  request_conversion: ["Claude Messages"]
  channel: Anthropic-compatible channel
  model: gpt-5.5
  stream_status: ok
  ```

## Related previous reports

This looks related to previous tool-name remapping issues, but the bug still reproduces on the current Claude Code + Anthropic Messages + Codex path:

- #1741
- #2656
- #2737
- #3149
- PR #2896

In particular, #2656 mentioned a very similar failure mode where the client-side tool name was `Read`, but the returned payload used lowercase `read`.

## Reproduction

Use Claude Code with a provider routed through CLIProxyAPI's Anthropic Messages endpoint:

```text
POST /api/provider/anthropic/v1/messages
```

Ask Claude Code to use the built-in Glob tool:

```text
Please use the Glob tool to find all *.py files under a local project directory.
Do not modify any files.
```

Claude Code initialization shows that the advertised tool is TitleCase:

```json
{
  "type": "system",
  "subtype": "init",
  "tools": [
    "Glob"
  ],
  "model": "claude-opus-4-7[1M]",
  "claude_code_version": "2.1.152"
}
```

The request forwarded into CLIProxyAPI also contains the correct tool schema:

```json
{
  "tools": [
    {
      "name": "Glob",
      "input_schema": {
        "type": "object",
        "required": ["pattern"],
        "properties": {
          "path": {
            "type": "string"
          },
          "pattern": {
            "type": "string"
          }
        }
      }
    }
  ]
}
```

However, the streamed response returned by CLIProxyAPI contains lowercase `glob`:

```text
event: content_block_start
data: {
  "type": "content_block_start",
  "index": 2,
  "content_block": {
    "type": "tool_use",
    "id": "call_1HhIILj3NcR4O4ubBJWQclKf",
    "name": "glob",
    "input": {},
    "signature": ""
  }
}
```

This happened repeatedly in the same session:

```json
{
  "type": "tool_use",
  "id": "call_1HhIILj3NcR4O4ubBJWQclKf",
  "name": "glob",
  "input": {
    "path": "<local project path>",
    "pattern": "**/*.py"
  }
}
```

```json
{
  "type": "tool_use",
  "id": "call_mB07yfiyRIOtu7MkGeVzxs1c",
  "name": "glob",
  "input": {
    "path": "<local project path>",
    "pattern": "**/*.py"
  }
}
```

```json
{
  "type": "tool_use",
  "id": "call_0GxhtnDKsN63xMC2JqAlOcnm",
  "name": "glob",
  "input": {
    "path": "<local project path>",
    "pattern": "**/*.py"
  }
}
```

Claude Code then fails locally:

```text
<tool_use_error>Error: No such tool available: glob</tool_use_error>
```

## Packet capture evidence

A packet capture was taken on the server-side plaintext HTTP links:

```text
tcp port 3000 or tcp port 8317
```

The important observation is:

### Incoming request to CLIProxyAPI

The request body entering CLIProxyAPI contained the original Claude Code tool name:

```json
"name":"Glob"
```

### Response from CLIProxyAPI

The SSE response from CLIProxyAPI back to new-api contained:

```json
"content_block": {
  "type": "tool_use",
  "id": "call_1HhIILj3NcR4O4ubBJWQclKf",
  "name": "glob",
  "input": {},
  "signature": ""
}
```

The same lowercase `glob` was then forwarded unchanged by new-api to the client.

So the lowercasing happens before or inside CLIProxyAPI's response path, not in Claude Code and not in new-api.

## Expected behavior

If the client request advertises:

```json
{"name": "Glob"}
```

then any upstream/sanitized/lowercase tool call should be restored before returning to the Anthropic client:

```json
{
  "type": "tool_use",
  "name": "Glob"
}
```

Claude Code should then be able to dispatch the built-in `Glob` tool.

## Actual behavior

CLIProxyAPI returns:

```json
{
  "type": "tool_use",
  "name": "glob"
}
```

Claude Code rejects it because tool dispatch is case-sensitive:

```text
No such tool available: glob
```

## Suspected root cause

The request-side conversion likely normalizes or lowercases tool names for the Codex/OpenAI-compatible upstream, but the streaming Anthropic response path does not restore the upstream tool name back to the original client-provided name.

This should probably use a per-request, collision-aware reverse map:

```text
upstream/sanitized name -> original client tool name
```

For example, when the incoming request contains:

```json
[
  {"name": "Glob"},
  {"name": "Read"}
]
```

then the response mapper should restore:

```text
glob -> Glob
read -> Read
```

But this must only happen for tools that were actually present in the current request, to avoid breaking clients whose tools are intentionally lowercase.

## Suggested fix direction

Build a per-request tool name mapping from the incoming `tools[]` schema.

Pseudo logic:

```go
// For each incoming client tool:
originalName := tool.Name
upstreamName := sanitizeOrNormalizeToolName(originalName)

// Store only non-conflicting mappings.
reverseMap[upstreamName] = originalName
```

Then, before streaming the response back to the Anthropic client, restore tool names in:

- `content_block_start.content_block.name`
- non-streaming `content[].name`
- any other response-side `tool_use.name` field

Example:

```go
if block.Type == "tool_use" {
    if original, ok := reverseToolNameMap[block.Name]; ok {
        block.Name = original
    }
}
```

Important: this should be per-request and collision-aware. It should not be a global hardcoded conversion like `glob -> Glob`, because some clients may legitimately define lowercase tool names.

### 1. Restore Claude Code TitleCase tools

Input client tools:

```json
[
  {"name": "Glob"},
  {"name": "Read"}
]
```

Upstream response:

```json
{"type": "tool_use", "name": "glob"}
```

Expected client response:

```json
{"type": "tool_use", "name": "Glob"}
```

### 2. Do not modify lowercase-native tools

Input client tools:

```json
[
  {"name": "glob"}
]
```

Upstream response:

```json
{"type": "tool_use", "name": "glob"}
```

Expected client response:

```json
{"type": "tool_use", "name": "glob"}
```

### 3. Collision handling

Input client tools:

```json
[
  {"name": "Glob"},
  {"name": "glob"}
]
```

The mapper should not blindly map `glob -> Glob`, because that would be ambiguous.

### 4. Streaming SSE path

Input upstream SSE:

```text
event: content_block_start
data: {"type":"content_block_start","content_block":{"type":"tool_use","name":"glob"}}
```

Expected downstream SSE:

```json
"name":"Glob"
```

## Additional note

The same captured response also included:

```json
"signature":""
```

inside the `tool_use` block. That may be a separate response-rewriting issue, because Anthropic `tool_use` blocks should not need a `signature` field. I am mentioning it here only as an observation; the main bug in this issue is the tool name lowercasing.

Progress-infinitely:
Fixes #3592

## Summary

Claude Code tool dispatch is case-sensitive and expects response `tool_use.name` values to match the tool names it advertised, e.g. `Glob` rather than `glob`. When routing Claude Code through the Amp provider alias path (`/api/provider/anthropic/v1/messages`), a Codex-backed/local provider can stream back a lowercase `glob` tool call, which Claude Code rejects with:

```text
No such tool available: glob
```

A global `glob -> Glob` normalization would be unsafe because Amp intentionally preserves lowercase `glob` for Amp CLI compatibility (`TestNormalizeAmpToolNames_GlobPreserved`, introduced by `fix(amp): preserve lowercase glob tool name`).

This PR makes Amp response tool-name normalization request-aware and collision-aware.

## Fix

`ResponseRewriter` now records the exact tool-name casing advertised by the current request from:

- `tools[].name`
- `tool_choice.name`

During response rewriting, it uses that request-local casing before falling back to the existing Amp canonical map.

Behavior:

- Claude Code request advertises `Glob` -> response `glob` is rewritten to `Glob`
- Lowercase-native request advertises `glob` -> response remains `glob`
- Ambiguous request advertises both `Glob` and `glob` -> that lowercased key is treated as conflicted and is not force-remapped
- No request tool context -> existing Amp default behavior is preserved, including lowercase `glob`

The fallback handler now constructs the response rewriter with the sanitized request body on local-provider and model-mapping paths, so streaming and non-streaming responses can use the per-request tool-name map.

## Tests

Added/kept coverage for both the reported bug and compatibility boundaries:

1. `TestNormalizeAmpToolNames_GlobPreserved` — default Amp behavior still keeps lowercase `glob`
2. `TestNormalizeAmpToolNames_RequestToolCasing_NonStreaming` — request-local `Glob` casing fixes non-streaming `tool_use.name`
3. `TestNormalizeAmpToolNames_RequestToolCasing_Streaming` — request-local `Glob` casing fixes streaming `content_block.name`
4. `TestResponseRewriter_RequestToolCasingFromBody` — verifies the rewriter derives casing from `tools[].name`
5. `TestResponseRewriter_LowercaseNativeRequestPreserved` — lowercase-native client tools are not TitleCased
6. `TestCollectRequestToolNames_CollisionIgnored` — conflicting `Glob/glob` and `Bash/bash` keys are not blindly remapped
7. `TestResponseRewriter_RequestToolCasingFromBody_Streaming` — direct streaming rewriter coverage
8. `TestFallbackHandler_RequestToolCasing_RewritesStreamingResponse` — covers `FallbackHandler.WrapHandler -> NewResponseRewriterForRequest -> SSE rewrite`

Validation:

```text
/usr/local/go/bin/go test ./internal/api/modules/amp
/usr/local/go/bin/go test ./...
/usr/local/go/bin/go build -o /tmp/cli-proxy-api-toolname-fix ./cmd/server
git diff --check
```

All pass.

## Evidence

The issue was confirmed via packet capture:

- Incoming request to CLIProxyAPI advertised `tools[].name = "Glob"`
- Outgoing SSE from CLIProxyAPI returned `content_block.name = "glob"`
- Claude Code failed with `No such tool available: glob`

The fix is scoped to the Amp response rewriter used by provider alias local/model-mapping paths and avoids changing the global lowercase-`glob` default required by Amp CLI.

### DIFF
--- a/internal/api/modules/amp/fallback_handlers.go
+++ b/internal/api/modules/amp/fallback_handlers.go
@@ -252,7 +252,7 @@ func (fh *FallbackHandler) WrapHandler(handler gin.HandlerFunc) gin.HandlerFunc 
 			// Log: Model was mapped to another model
 			log.Debugf("amp model mapping: request %s -> %s", normalizedModel, resolvedModel)
 			logAmpRouting(RouteTypeModelMapping, modelName, resolvedModel, providerName, requestPath)
-			rewriter := NewResponseRewriter(c.Writer, modelName)
+			rewriter := NewResponseRewriterForRequest(c.Writer, modelName, bodyBytes)
 			rewriter.suppressThinking = true
 			c.Writer = rewriter
 			// Filter Anthropic-Beta header only for local handling paths
@@ -267,7 +267,7 @@ func (fh *FallbackHandler) WrapHandler(handler gin.HandlerFunc) gin.HandlerFunc 
 			// Wrap with ResponseRewriter for local providers too, because upstream
 			// proxies (e.g. NewAPI) may return a different model name and lack
 			// Amp-required fields like thinking.signature.
-			rewriter := NewResponseRewriter(c.Writer, modelName)
+			rewriter := NewResponseRewriterForRequest(c.Writer, modelName, bodyBytes)
 			rewriter.suppressThinking = providerName != "claude"
 			c.Writer = rewriter
 			// Filter Anthropic-Beta header only for local handling paths
--- a/internal/api/modules/amp/fallback_handlers_test.go
+++ b/internal/api/modules/amp/fallback_handlers_test.go
@@ -13,6 +13,38 @@ import (
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/registry"
 )
 
+func TestFallbackHandler_RequestToolCasing_RewritesStreamingResponse(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+
+	reg := registry.GetGlobalRegistry()
+	reg.RegisterClient("test-client-amp-tool-casing", "codex", []*registry.ModelInfo{
+		{ID: "test/gpt-tool-casing", OwnedBy: "openai", Type: "codex"},
+	})
+	defer reg.UnregisterClient("test-client-amp-tool-casing")
+
+	fallback := NewFallbackHandlerWithMapper(func() *httputil.ReverseProxy { return nil }, nil, nil)
+	handler := func(c *gin.Context) {
+		c.Writer.Header().Set("Content-Type", "text/event-stream")
+		_, _ = c.Writer.Write([]byte("event: content_block_start\ndata: {\"type\":\"content_block_start\",\"index\":1,\"content_block\":{\"type\":\"tool_use\",\"name\":\"glob\",\"id\":\"toolu_01\",\"input\":{}}}\n\n"))
+	}
+
+	r := gin.New()
+	r.POST("/messages", fallback.WrapHandler(handler))
+
+	reqBody := []byte(`{"model":"test/gpt-tool-casing","tools":[{"name":"Glob","input_schema":{"type":"object"}}]}`)
+	req := httptest.NewRequest(http.MethodPost, "/messages", bytes.NewReader(reqBody))
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+	r.ServeHTTP(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("Expected status 200, got %d", w.Code)
+	}
+	if !bytes.Contains(w.Body.Bytes(), []byte(`"name":"Glob"`)) {
+		t.Fatalf("expected streaming response to restore glob->Glob, got %s", w.Body.String())
+	}
+}
+
 func TestFallbackHandler_ModelMapping_PreservesThinkingSuffixAndRewritesResponse(t *testing.T) {
 	gin.SetMode(gin.TestMode)
 
--- a/internal/api/modules/amp/response_rewriter.go
+++ b/internal/api/modules/amp/response_rewriter.go
@@ -22,6 +22,7 @@ type ResponseRewriter struct {
 	originalModel    string
 	isStreaming      bool
 	suppressThinking bool
+	requestToolNames map[string]string
 }
 
 // NewResponseRewriter creates a new response rewriter for model name substitution.
@@ -31,6 +32,12 @@ func NewResponseRewriter(w gin.ResponseWriter, originalModel string) *ResponseRe
 		body:           &bytes.Buffer{},
 		originalModel:  originalModel,
 	}
+}
+
+func NewResponseRewriterForRequest(w gin.ResponseWriter, originalModel string, requestBody []byte) *ResponseRewriter {
+	rw := NewResponseRewriter(w, originalModel)
+	rw.requestToolNames = collectRequestToolNames(requestBody)
+	return rw
 }
 
 const maxBufferedResponseBytes = 2 * 1024 * 1024 // 2MB safety cap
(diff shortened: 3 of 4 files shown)


Progress-infinitely:
Addressed Gemini Code Assist's suggestion by parsing the request JSON once with gjson.ParseBytes in collectRequestToolNames. Re-ran go test ./internal/api/modules/amp, go test ./..., go build ./cmd/server, and git diff --check successfully.

luispater (approved):

PR #3632, opened 2026-05-30
fix(executor): normalize single-underscore mcp_ tool prefix on OAuth

Dworf:
## Problem

Third-party clients that name MCP tools with a **single-underscore** prefix
(`mcp_<tool>`) get rejected on OAuth (Claude Code) traffic. Anthropic's overage
gate fingerprints that single-underscore prefix as a user-installed third-party
MCP extension and bills it to "extra usage". On accounts where extra usage is
disabled at the org level, the request fails with **HTTP 400**:

> Third-party apps now draw from your extra usage, not your plan limits. Add more
> at claude.ai/settings/usage and keep going.

Claude Code's own convention is the **double-underscore** form
(`mcp__<server>__<tool>`), which Anthropic accepts as first-party.

Reproduced with BoltAI (`ai-sdk/anthropic/...-boltai.2`), which emits
`mcp_<tool>` tool names, against an OAuth account with org-level extra usage
disabled (`Anthropic-Ratelimit-Unified-Overage-Disabled-Reason: org_level_disabled`).

## Fix

Add `oauthRenamedToolName()` in the Claude executor. It applies the existing
static rename map (`bash` -> `Bash`, ...) and then normalizes a single-underscore
`mcp_` prefix to the double-underscore `mcp__` form, while leaving names that are
already `mcp__...` untouched. It's wired into every OAuth tool-name rewrite site:
the `tools` array, `tool_choice`, `tool_use` / `tool_reference` message parts, and
nested content.

The existing reverse mapping restores the client's original `mcp_` name on the
response, so the client never sees the rewrite.

## Evidence (credentials/PII redacted)

Same BoltAI request, same model, **same account with overage still disabled** --
the only variable is the outbound prefix.

**Before** -- proxy forwards `mcp_*` upstream -> **400**:
```
Status: 400
Anthropic-Ratelimit-Unified-Overage-Disabled-Reason: org_level_disabled
 "message":"Third-party apps now draw from your extra usage, ..."}}
```

**After** -- proxy rewrites to `mcp__*` upstream -> **200** (overage still disabled):
```
Status: 200
Anthropic-Ratelimit-Unified-Overage-Disabled-Reason: org_level_disabled
Anthropic-Ratelimit-Unified-Overage-Status: rejected
Anthropic-Ratelimit-Unified-Status: allowed
```
Downstream response reverses the name back to the client's original
`mcp_get_youtube_transcript_...`, confirming the full round-trip.

## Test

`TestRemapOAuthToolNames_McpSingleUnderscore_NormalizedAndReversed` covers the
forward normalization and reverse restoration.

`go test ./internal/runtime/executor/...` passes; `gofmt` clean; server builds.

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -106,6 +106,34 @@ var oauthToolRenameMap = map[string]string{
 // then the global reverse map incorrectly rewrote every `Bash` in the
 // response to `bash`, causing Amp to reject the tool_use as unknown).
 
+// mcpSingleUnderscorePrefix / mcpDoubleUnderscorePrefix bracket the MCP tool-name
+// normalization. Some third-party clients (e.g. BoltAI) name MCP tools with a
+// single underscore (`mcp_<tool>`). Anthropic's overage gate fingerprints that
+// single-underscore prefix as a user-installed MCP extension and bills it to
+// extra usage, returning HTTP 400 ("third-party apps now draw from your extra
+// usage") on accounts where extra usage is disabled. Claude Code's own
+// convention is the double-underscore form (`mcp__<server>__<tool>`), which is
+// accepted as first-party, so we normalize `mcp_` -> `mcp__` on OAuth traffic.
+const (
+	mcpSingleUnderscorePrefix = "mcp_"
+	mcpDoubleUnderscorePrefix = "mcp__"
+)
+
+// oauthRenamedToolName returns the Claude Code-compatible upstream name for a
+// client-supplied OAuth tool name and whether a rename is needed. It first
+// applies the static built-in rename map (bash -> Bash, ...), then normalizes a
+// single-underscore `mcp_` MCP prefix to the double-underscore form Anthropic
+// recognizes as first-party. Names that are already `mcp__...` pass through.
+func oauthRenamedToolName(name string) (string, bool) {
+	if newName, ok := oauthToolRenameMap[name]; ok && newName != name {
+		return newName, true
+	}
+	if strings.HasPrefix(name, mcpSingleUnderscorePrefix) && !strings.HasPrefix(name, mcpDoubleUnderscorePrefix) {
+		return mcpDoubleUnderscorePrefix + name[len(mcpSingleUnderscorePrefix):], true
+	}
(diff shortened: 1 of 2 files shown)


xiaojiou176:
Heads-up on an adjacent-but-different MCP-naming issue, in case it's useful context for this PR.

This PR normalizes the **single-underscore** `mcp_<tool>` prefix to the double-underscore `mcp__<tool>` form on the OAuth/executor path, to avoid Anthropic's third-party overage gate.

I've been working on a **different** MCP-naming problem on the **translator** path (OpenAI Responses ↔ Claude), which I've opened as #3670:

- Codex emits **already-double-underscore** flat names like `mcp__SERVER__TOOL`, and expects the tool registry to dispatch on the `(namespace, name)` tuple.
- The translator was passing the joined flat string through as a single name in both the request (tool declaration) and response (tool_use echo) directions, so the registry couldn't match it.
- The fix splits `mcp__SERVER__TOOL` → `(namespace="mcp__SERVER__", name="TOOL")` symmetrically across the claude/gemini/openai lanes.

So the two changes are orthogonal: this PR fixes *single→double underscore on OAuth billing*, mine fixes *double-underscore flat-name splitting for tuple dispatch on the Responses path*. They touch different layers (executor vs translator) and shouldn't conflict. Flagging it so reviewers have the full picture of the MCP-naming surface.

Dworf:
Thanks for the heads-up and the clear write-up — agreed, these are orthogonal.

This PR stays entirely on the executor/OAuth path (`internal/runtime/executor/claude_executor.go`) and only rewrites a leading single-underscore `mcp_` → `mcp__` to get past the third-party overage gate. #3670 operates on the translator path and splits already-double `mcp__SERVER__TOOL` into the `(namespace, name)` tuple the registry dispatches on. Different layers, different request flows (OAuth Claude vs OpenAI Responses), so no overlap.

One detail confirming they don't interfere: this normalization only swaps the prefix (`mcp_foo` → `mcp__foo`); it never introduces the `__SERVER__` separator your splitter keys on, so it can't accidentally produce or disturb a tuple name. Appreciate you mapping out the full MCP-naming surface for the reviewers. 👍

PR #3633, opened 2026-05-30
feat(codex): add compact routing controls

stringer07:
## Summary

  - add `compact-default` and per-credential `compact` controls for codex and openai-compatible credentials
  - resolve compact eligibility into auth attributes during synthesis
  - route `/v1/responses/compact` through legacy credential selection and filter out credentials that do not
  support compact
  - return `compact_unsupported` with HTTP 503 when compact requests have credentials but none are eligible
  - expose and validate `compact` in codex auth-file management flows
  - document the new config in `config.example.yaml`

  ## Why

  Codex compact requests use the `/v1/responses/compact` endpoint. Many API providers do not implement that
  endpoint, so sending compact requests to them fails even though the same credential can still handle normal `/v1/
  responses` requests.

  This change lets operators explicitly control which credentials can receive compact requests. When the currently
  selected provider does not support compact, the router can automatically fall back to another eligible credential
  instead of sending the request to an unsupported endpoint.

  ## Verification

  - `gofmt -w .`
  - `go test ./...`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -178,11 +178,18 @@ nonstream-keepalive-interval: 0
 #       - "*flash*"            # wildcard matching substring (e.g. gemini-2.5-flash-lite)
 #   - api-key: "AIzaSy...02"
 
+# compact-default controls /v1/responses/compact routing for codex/openai-compatible
+# credentials whose per-credential compact mode is "auto" (the default).
+#   allow (default): auto credentials may serve compact requests (backward compatible)
+#   deny: auto credentials are excluded; only credentials with compact: force_on serve compact
+# compact-default: allow
+
 # Codex API keys
 # codex-api-key:
 #   - api-key: "sk-atSM..."
 #     prefix: "test" # optional: require calls like "test/gpt-5-codex" to target this credential
 #     disable-cooling: false # optional: per-auth override for auth/model cooldown scheduling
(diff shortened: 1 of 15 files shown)


xiaojiou176:
This is a nice approach to the `/v1/responses/compact` problem. I tackled the same pain point from a complementary angle and wanted to share it in case the two compose well.

This PR adds **compact routing controls**: mark which credentials support compact, filter out the ones that don't, and return `compact_unsupported` (503) when none are eligible.

My local change (happy to open as a PR if useful) adds a **cross-model fallback** for the same endpoint:

- When a compact request would otherwise fail because the resolved executor doesn't implement compaction, rewrite the request to a configurable fallback model that *does* support compact (`compact-fallback-model` in YAML, or `CPA_COMPACT_FALLBACK_MODEL` env), and retry once.
- The fallback response carries an `x-cpa-compact-fallback: <original>-><fallback>` header for observability.
- Defaults to empty/off, preserving current behavior.

These seem complementary rather than competing: your PR decides *which credentials are eligible* and fails cleanly when none are; mine provides a *graceful fallback* when the caller still wants compaction to succeed via a different model. A natural composition is "filter first (this PR), then optionally fall back (mine)" — the fallback only kicks in when explicitly configured.

Would it make sense to layer a fallback on top of this routing control, or do you see compaction-eligibility as something that should always fail closed? Happy to align my change to your config shape if there's interest.

PR #3642, opened 2026-05-31
fix(codex): scope session keys by auth

Destibey:
## Summary
- Scope Codex upstream `prompt_cache_key` / session headers by the selected auth namespace before forwarding requests upstream.
- Preserve stable keys within the same auth while preventing cross-auth session reuse.
- Apply the behavior to direct HTTP and WebSocket Codex paths.
- Keep legacy Claude cache lookup keys when no auth namespace exists, and use an unambiguous scoped lookup key when a namespace is present.

## Why this is needed
In a multi-account proxy/token-pool deployment, different downstream clients can send the same Codex `prompt_cache_key` or session ID while being routed through different upstream auth accounts. If the proxy forwards that key unchanged, upstream can observe different accounts continuing the same logical conversation/session chain.

That cross-account session reuse can contaminate upstream conversation state and may look like account-linking or anomalous shared-session behavior to upstream risk-control systems. Deriving the upstream session key from the selected auth namespace keeps per-account continuity, but prevents one account's session key from being reused by another account.

## Tests
- `go test ./internal/runtime/executor -run 'TestCodexExecutorCacheHelper|TestCodexClaudeCacheKey|TestApplyCodexPromptCacheHeaders' -count=1`

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -898,7 +898,7 @@ func (e *CodexExecutor) cacheHelper(ctx context.Context, from sdktranslator.Form
 	if from == "claude" {
 		userIDResult := gjson.GetBytes(req.Payload, "metadata.user_id")
 		if userIDResult.Exists() {
-			key := fmt.Sprintf("%s-%s", req.Model, userIDResult.String())
+			key := codexClaudeCacheKey(req.Model, auth, userIDResult.String())
 			var ok bool
 			if cache, ok = helps.GetCodexCache(key); !ok {
 				cache = helps.CodexCache{
@@ -911,11 +911,11 @@ func (e *CodexExecutor) cacheHelper(ctx context.Context, from sdktranslator.Form
 	} else if from == "openai-response" {
 		promptCacheKey := gjson.GetBytes(req.Payload, "prompt_cache_key")
 		if promptCacheKey.Exists() {
-			cache.ID = promptCacheKey.String()
+			cache.ID = scopedCodexPromptCacheKey(auth, promptCacheKey.String())
 		}
 	} else if from == "openai" {
 		if apiKey := strings.TrimSpace(helps.APIKeyFromContext(ctx)); apiKey != "" {
-			cache.ID = uuid.NewSHA1(uuid.NameSpaceOID, []byte("cli-proxy-api:codex:prompt-cache:"+apiKey)).String()
+			cache.ID = scopedCodexPromptCacheKey(auth, uuid.NewSHA1(uuid.NameSpaceOID, []byte("cli-proxy-api:codex:prompt-cache:"+apiKey)).String())
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


Destibey:
Addressed the review feedback in ce745f78:\n\n- preserves the legacy Claude cache lookup key when no auth namespace is available, avoiding the double-hyphen compatibility issue\n- replaces delimiter-only scoped Claude lookup keys with an unambiguous derived key so namespace/user values containing '-' cannot collide\n- updated the PR description to explain the multi-account risk-control motivation for deriving upstream session keys by auth namespace\n\nTested with:\n\n`go test ./internal/runtime/executor -run 'TestCodexExecutorCacheHelper|TestCodexClaudeCacheKey|TestApplyCodexPromptCacheHeaders' -count=1`

Destibey:
Addressed the latest prompt_cache_key review feedback in c5c8aaa0:

- length-prefixes auth namespace and client prompt_cache_key components before deriving the scoped SHA1 UUID, so values containing `:` cannot collide
- reuses the same component encoding helper for Claude scoped cache keys
- keeps the no-namespace behavior unchanged: prompt_cache_key is still trimmed and passed through as the legacy key
- added a regression test for namespace `a:b` + key `c` versus namespace `a` + key `b:c`

Tested with:

`go test ./internal/runtime/executor -run 'TestCodexExecutorCacheHelper|TestCodexClaudeCacheKey|TestScopedCodexPromptCacheKey|TestApplyCodexPromptCacheHeaders' -count=1`

`go test ./internal/runtime/executor ./sdk/cliproxy/auth ./sdk/cliproxy -count=1`

Destibey:
Addressed the latest direct HTTP `Session_id` review feedback and rebased the PR onto current upstream `dev` (`bbcdaab7`).

What changed:
- PR branch now has a single rebased commit: `eb751878 fix(codex): scope session keys by auth`.
- Direct HTTP keeps the scoped session value generated by `cacheHelper`; the Mac client `Session_id` header no longer overwrites it.
- WebSocket scoping is aligned with the current upstream header cleanup by using `Session-Id` and not restoring deprecated `Conversation_id`.
- Added regression coverage for the Mac client header overwrite case and scoped key collision cases.

Validation:
- `go test ./internal/runtime/executor -run 'TestCodexExecutorCacheHelper|TestCodexClaudeCacheKey|TestScopedCodexPromptCacheKey|TestApplyCodexPromptCacheHeaders|TestApplyCodexHeadersPreservesScopedSessionID|TestApplyCodexWebsocketHeadersIdentityConfuse' -count=1`
- `git diff --check`
- GitHub checks are green: build, close-when-agents-md-changed, ensure-no-translator-changes.

Note: a full local `go test ./internal/runtime/executor -count=1` still hits the existing upstream `TestApplyCodexWebsocketHeadersDefaultsToCurrentResponsesBeta` user-agent expectation; this PR does not change that constant or test.

luispater:
We've implemented this feature.

Please use the latest version.

PR #3653, opened 2026-06-01
feat(opencode): add OpenCode agent provider support

Chedrian07:
## Summary

Adds first-class **OpenCode** (https://opencode.ai) support as a client of the proxy,
mirroring the existing **Amp CLI** module's "merged + provider-scoped routes" pattern.

OpenCode is an OpenAI/Anthropic-compatible agent (built on the Vercel AI SDK): you point
a custom provider's `baseURL` at an upstream. This PR gives OpenCode a dedicated,
isolated `/opencode/...` route namespace backed by the existing OpenAI/Claude/Gemini SDK
handlers, plus an OpenCode-scoped model-mapping layer.

Design rationale and the full file-by-file plan live in
[`OPENCODE_INTEGRATION_PLAN.md`](OPENCODE_INTEGRATION_PLAN.md).

### Why a trimmed mirror of Amp (rejected path noted)
Amp's module also bundles a reverse proxy to `ampcode.com`, localhost-gated management
routes, multi-source secret resolution, and per-client upstream API keys — all to bridge
Amp's **proprietary control plane** (OAuth/threads/credits). OpenCode has no such
mandatory control plane; you point a `baseURL` at the proxy and that's the whole
integration. Replicating that machinery would be cargo-culting and enlarge the diff, so
it was intentionally **not** copied. The OpenCode module is therefore route aliases +
model mapping, backed directly by the existing handlers. No new translator code — the
wire formats are byte-for-byte the formats the proxy already speaks.

## Protocols supported
All three, via existing SDK handlers (no translator changes):
- **OpenAI Chat Completions** (`@ai-sdk/openai-compatible` → `/chat/completions`, `/completions`)
- **OpenAI Responses** (`@ai-sdk/openai` → `/responses`)
- **Anthropic Messages** (`@ai-sdk/anthropic` → `/messages`, `/messages/count_tokens`)
- Gemini native (`/v1beta/models/*`) for parity with the merged endpoints

## New routes
**Merged** (one `baseURL` serves all three protocols; backend resolved from request model):
```
GET  /opencode/v1/models
POST /opencode/v1/chat/completions
POST /opencode/v1/completions
POST /opencode/v1/responses
POST /opencode/v1/messages
POST /opencode/v1/messages/count_tokens
GET  /opencode/v1beta/models
GET|POST /opencode/v1beta/models/*action
```
**Provider-scoped** (explicit protocol/backend selection, mirrors Amp's `/api/provider/:provider`):
```
GET  /opencode/provider/:provider/models
POST /opencode/provider/:provider/v1/chat/completions
POST /opencode/provider/:provider/v1/completions
POST /opencode/provider/:provider/v1/responses
POST /opencode/provider/:provider/v1/messages
POST /opencode/provider/:provider/v1/messages/count_tokens
GET|POST /opencode/provider/:provider/v1beta/models/*action
```
All routes require a valid proxy API key (top-level `api-keys`).

## New config keys
```yaml
opencode:
  force-model-mappings: false        # run mappings before checking local providers
  model-mappings:
    - from: "claude-sonnet-4-5"
      to: "gpt-5"
    - from: "^gpt-5.*$"
      to: "claude-sonnet-4-5-20250929"
      regex: true                     # exact match by default; opt-in regex
```
Mappings are scoped to `/opencode/...` traffic only, hot-reloadable, and thinking-suffix
aware (e.g. `gpt-5(8192)` is preserved through mapping). Re-exported from `sdk/config`
for embedding parity.

## Files
**Added**
- `internal/api/modules/opencode/{opencode,routes,model_mapping,mapping_handler}.go` — the module (`RouteModuleV2`)
- `internal/api/modules/opencode/*_test.go`, `internal/config/opencode_test.go` — tests
- `examples/opencode-provider/main.go` — runnable client demo (reads creds from env)
- `OPENCODE_INTEGRATION_PLAN.md` — design contract

**Modified**
- `internal/config/config.go`, `sdk/config/config.go` — `OpenCode` / `OpenCodeModelMapping` schema
- `internal/api/server.go` — construct/register the module + hot-reload notification (mirrors the 3 `ampModule` touch-points)
- `config.example.yaml`, `README.md` — documented `opencode:` block + README section

**Not touched:** `internal/translator/` (reused as-is), any existing provider/executor, the Amp module. No existing route or config key changed.

## Test evidence
```
$ go build ./...        # exit 0
$ go vet ./internal/api/modules/opencode/ ./internal/api/ ./internal/config/ ./sdk/config/   # exit 0
$ gofmt -l <changed files>   # empty (formatted)
$ go test ./...         # exit 0 (full suite)
```
Targeted:
```
    ok  	github.com/router-for-me/CLIProxyAPI/v7/internal/api/modules/opencode	(cached)
    ok  	github.com/router-for-me/CLIProxyAPI/v7/internal/config	(cached)
```
Coverage: merged + provider-scoped route resolution, model mapping
(exact/regex/case-insensitive/suffix/force-mode/no-provider), request-body model
rewrite round-trips, Gemini path model extraction, and `opencode` config parsing
(values + empty defaults).

> Note: `go vet ./...` reports one **pre-existing** finding in
> `internal/logging/request_logger.go` (`WriteTo` signature) that also exists on `main`
> and is unrelated to this PR; left untouched to avoid an out-of-scope change.

## Manual verification (point real OpenCode at the proxy)
1. Run the proxy with at least one `api-keys` entry and an OAuth/provider configured.
2. Add to `opencode.json`:
   ```json
   {
     "$schema": "https://opencode.ai/config.json",
     "provider": {
       "cliproxy": {
         "npm": "@ai-sdk/openai-compatible",
         "name": "CLIProxyAPI",
         "options": { "baseURL": "http://127.0.0.1:8317/opencode/v1", "apiKey": "{env:CLIPROXY_API_KEY}" },
         "models": { "gpt-5": {}, "claude-sonnet-4-5": {} }
       }
     }
   }
   ```
3. `export CLIPROXY_API_KEY="<one of your proxy api-keys>"` then run `opencode`.
4. Or exercise the endpoint directly:
   ```bash
   curl -s http://127.0.0.1:8317/opencode/v1/chat/completions \
     -H "Authorization: Bearer $CLIPROXY_API_KEY" -H "Content-Type: application/json" \
     -d '{"model":"gpt-5","messages":[{"role":"user","content":"ping"}]}'
   ```
   `go run ./examples/opencode-provider` does the same and prints a ready-to-use config block.

## No existing behavior changed
The module only registers a new `/opencode/...` prefix and reads a new optional
`opencode:` config block. Existing `/v1`, `/v1beta`, and Amp `/api/provider/...` routes,
translators, executors, and configs are untouched. Full test suite passes.

## Maintainer decisions
- **Route prefix `/opencode`** (top-level) was chosen over reusing Amp's `/api/provider`
  to keep the integration isolated from Amp's `/api` management middleware and scope
  OpenCode model mappings to OpenCode traffic. Happy to change the prefix if preferred.
- **Default model mappings:** none shipped (the example mappings in `config.example.yaml`
  are commented out). If you'd like sensible defaults (e.g. mapping common OpenCode model
  ids to OAuth-backed equivalents), let me know which target models to standardize on.
- The setup guide is folded into the README because `docs/*` is gitignored in this repo;
  happy to relocate it if docs are synced from elsewhere.


Chedrian07:
Thanks for the reviews — all three comments point at the same real bug (model mapping was silently dropped for **Gemini native** requests). Fixed in 4917ce8.

**Root cause & fix.** Gemini native requests carry the model in the `:action` URL path (`model:method`), not the JSON body, and the shared `GeminiHandler` resolves the model from the path via `ShouldBindUri` — `internal/.../gemini_handlers.go:142` (`action := strings.Split(strings.TrimPrefix(request.Action, "/"), ":")`). The previous handler only rewrote the body `model`, so `rewriteModelInRequest` returned `ok=false` for Gemini and the mapping never reached the handler (exactly as @chatgpt-codex-connector described). The fix rewrites the `:action` path param when the body has no `model` field, preserving the method and leading slash, with a new test `TestMappingHandler_RewritesGeminiActionPath`.

**On the `MappedModelContextKey` → `"mapped_model"` suggestion (@gemini-code-assist):** I verified this against the tree and the shared SDK handlers do **not** read any `mapped_model` context key — the only consumer of `"mapped_model"` is Amp's own `createGeminiBridgeHandler` (`internal/api/modules/amp/gemini_bridge.go:32`), which exists to bridge Amp's non-standard `/publishers/google/models/...` path. This module instead routes the standard `/v1beta/models/*action` shape straight to `GeminiHandler`, so setting *any* context key would have had no effect. Rather than set a dead key, I removed `MappedModelContextKey` entirely and fixed the mapping at the location the handler actually reads (the path). Net effect is the behavior both bots asked for, via the mechanism that actually works for the shared handler.

Verification: `go build ./...`, `go vet ./internal/api/modules/opencode/`, `gofmt -l`, and `go test ./...` all pass; mapping for OpenAI/Claude (body) and Gemini (path) is covered by tests. Happy to adjust further.

PR #3657, opened 2026-06-01
fix(openai): keep referenced tool call when deduping websocket input IDs

### LINKED ISSUE #3664: v7.1.37 WebSocket Responses: 400 "No tool call found for function call output with call_id" (regression from #3620 input id dedupe)
[ ] No, it's another issue.

**Describe the bug**

After updating to **v7.1.37**, Codex conversations over the **WebSocket** Responses path started failing with HTTP 400:

```
{"type":"error","status":400,"error":{
  "message":"No tool call found for function call output with call_id call_BgLu4KOohIO7RkvIbTrsZkUk.",
  "type":"invalid_request_error","param":"input","code":null}}
```

The error is sticky — once a turn fails, every subsequent turn replays the same orphaned output and also fails, so the whole conversation becomes unusable. This did not happen before v7.1.37.

**Root cause**

The input item ID dedupe added in #3620 (`dedupeInputItemsByID` in `sdk/api/handlers/openai/openai_responses_websocket.go`) keeps only the **last** occurrence of each input item `id`. When an upstream reuses the same item `id` across a re-sent / repaired tool call — i.e. two `function_call` items share an `id` but carry **different** `call_id`s — the last-wins rule can drop the `function_call` whose `call_id` still has a matching `function_call_output`. That output is then orphaned and the upstream rejects the request.

#3620's own scenario was safe only because the dropped copy's `call_id` had no output; when it *does* have an output, the regression triggers.

**CLI Type**
codex

**Model Name**
gpt-5.5 (also reproducible with other models on the Responses/WebSocket path)

**LLM Client**
OpenAI Codex CLI (`wire_api = "responses"`, `supports_websockets = true`)

**Request Information**

Reproduction:
1. Point Codex at a CLIProxyAPI v7.1.37 instance with a provider using `wire_api = "responses"` and `supports_websockets = true`.
2. Run any turn that issues a tool call (e.g. a single `exec_command`).
3. The follow-up request fails with the 400 above (`openai_responses_websocket.go` logs it as `downstream_out ... event=error`).

A/B confirmation (same server, same prompt):

| `supports_websockets` | Result |
| --- | --- |
| `true` | ❌ 400 `No tool call found for function call output with call_id ...` |
| `false` | ✅ Succeeds |

This isolates the regression to the WebSocket input-normalization path touched by #3620.

**Expected behavior**

Deduping input items by `id` must not orphan a `function_call` / `custom_tool_call` whose `call_id` is still referenced by a surviving `*_output`.

**OS Type**
- OS: macOS
- Version: 15.x

**Additional context**

Workaround until fixed: set `supports_websockets = false` on the affected provider (client side).

Fix submitted in **#3657** — makes the dedupe orphan-aware (never replaces an item whose `call_id` still has a matching output with one that doesn't), keeping a single item per `id` while preserving the original intent of #3620. Includes a regression test.

catoncat:
Closes #3664

## Summary

The input item ID dedupe added in #3620 (`dedupeInputItemsByID`) keeps only the **last** occurrence of each input item `id`. When an upstream reuses the same item `id` across a re-sent or repaired tool call — so two `function_call` items share an `id` but carry **different** `call_id`s — the last-wins rule can drop the `function_call` whose `call_id` still has a matching `function_call_output`.

The upstream then rejects the request with:

```
HTTP 400 invalid_request_error (param: input)
"No tool call found for function call output with call_id call_..."
```

Because this happens on the Codex **WebSocket** Responses path, the error is sticky: every subsequent turn in the conversation replays the same orphaned output and fails. Conversations that worked before v7.1.37 stop working after updating, and disabling websockets on the client (`supports_websockets = false`) is currently the only workaround.

## Root cause

`dedupeInputItemsByID` only looks at `id` when choosing which duplicate to keep:

```go
lastIndexByID[itemID] = i      // last occurrence wins, unconditionally
```

It never checks whether the item it drops is a `function_call` / `custom_tool_call` whose `call_id` is still referenced by a surviving `*_output`. #3620's own scenario was safe only because the dropped copy's `call_id` had **no** output; when the dropped copy's `call_id` **does** have an output, the output is orphaned.

## Fix

Make the dedupe **orphan-aware**. When several input items share an `id`, never replace an item whose `call_id` is still referenced by a tool-call output with one that is not. This keeps exactly one item per `id` (preserving the original intent and the no-duplicate-id guarantee) while ensuring retained tool calls stay paired with their outputs.

- First pass collects `call_id`s referenced by `function_call_output` / `custom_tool_call_output` items.
- Keeper selection still defaults to the last occurrence, but a referenced item is never overwritten by an unreferenced one.

## Reproduction

Codex CLI (`wire_api = "responses"`, `supports_websockets = true`) pointed at a CLIProxyAPI v7.1.37 instance. A normal turn that issues a single `exec_command` tool call returns:

```
{"type":"error","status":400,"error":{
  "message":"No tool call found for function call output with call_id call_BgLu4KOohIO7RkvIbTrsZkUk.",
  "type":"invalid_request_error","param":"input","code":null}}
```

The same server, same prompt, with `supports_websockets = false` succeeds — confirming the regression is on the WebSocket input-normalization path touched by #3620.

## Testing

- `gofmt -w` on both changed files
- `go build -o cli-proxy-api ./cmd/server` — succeeds
- `go test ./sdk/api/handlers/openai/...` — passes, including the existing
  `TestNormalizeResponsesWebsocketRequestDropsDuplicateInputItemsByID` and
  `TestDedupeResponsesWebsocketInputItemsByIDAfterRepair`
- New regression test `TestDedupeResponsesWebsocketInputItemsByIDKeepsReferencedToolCall`
  fails on the current code and passes with this change.

## Scope

Only `sdk/api/handlers/openai/openai_responses_websocket.go` and its test are touched. No changes under `internal/translator/` or `AGENTS.md`.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket.go
@@ -728,14 +728,50 @@ func dedupeInputItemsByID(rawArray string) (string, error) {
 		return "", errUnmarshal
 	}
 
-	lastIndexByID := make(map[string]int, len(items))
+	// Collect the call_ids that are still referenced by tool-call output
+	// items. When several input items share the same id, the one we keep must
+	// preserve any call_id that has a matching output; otherwise the upstream
+	// rejects the request with "No tool call found for function call output".
+	referencedCallIDs := make(map[string]struct{}, len(items))
+	for _, item := range items {
+		if len(item) == 0 {
+			continue
+		}
+		switch strings.TrimSpace(gjson.GetBytes(item, "type").String()) {
+		case "function_call_output", "custom_tool_call_output":
+			callID := strings.TrimSpace(gjson.GetBytes(item, "call_id").String())
+			if callID != "" {
+				referencedCallIDs[callID] = struct{}{}
+			}
+		}
+	}
+
+	// For each id, choose the index to keep. The default is the last
+	// occurrence (matching the original dedupe behavior), but we never replace
+	// an item whose call_id still has a matching output with one that does not.
+	// This keeps a single item per id while ensuring retained tool calls stay
+	// paired with their outputs.
+	keepIndexByID := make(map[string]int, len(items))
+	keepReferencedByID := make(map[string]bool, len(items))
 	for i, item := range items {
 		if len(item) == 0 {
 			continue
 		}
 		itemID := strings.TrimSpace(gjson.GetBytes(item, "id").String())
-		if itemID != "" {
-			lastIndexByID[itemID] = i
+		if itemID == "" {
+			continue
+		}
+		callID := strings.TrimSpace(gjson.GetBytes(item, "call_id").String())
+		_, referenced := referencedCallIDs[callID]
+		referenced = referenced && callID != ""
+		if _, seen := keepIndexByID[itemID]; !seen {
+			keepIndexByID[itemID] = i
+			keepReferencedByID[itemID] = referenced
+			continue
+		}
+		if referenced || !keepReferencedByID[itemID] {
+			keepIndexByID[itemID] = i
+			keepReferencedByID[itemID] = referenced
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


catoncat:
Thanks for the review. Applied the single-pass parsing suggestion in f05d68d — each item's `type`/`id`/`call_id` is now read once via `gjson.GetManyBytes` and reused across the loops, so there's no rescanning as history grows. Behavior is unchanged and all tests still pass.

PR #3670, opened 2026-06-02
fix(translator/claude): Anthropic/Bedrock compatibility for OpenAI Responses conversion

xiaojiou176:
## What

Three correctness fixes in the OpenAI Responses → Claude (Anthropic/Bedrock) request translation:

1. **Tool-name sanitization** — `sanitizeAnthropicToolName` strips characters that are illegal under Anthropic's tool-name regex (e.g. colons). Tool names containing a colon currently cause a `400`.
2. **Parallel tool_use / tool_result merge** — consecutive `function_call` / `function_call_output` items in the same turn are merged into a single assistant/user message instead of creating one message each. Separate messages violate Anthropic's alternation rule and trigger Bedrock `TOOL_USE_RESULT_MISMATCH` (`400`).
3. **`input_image` in `function_call_output`** — converted into a proper Claude `tool_result` image block.

## Why

When a client (e.g. an agent harness) sends parallel tool calls or tool names with colons through the Responses→Claude path, Bedrock rejects the request with a `400`. These are general protocol-compatibility issues, not specific to any deployment.

## Tests

Added regression tests:
- `claude_openai-responses_request_sanitize_test.go`
- `claude_openai-responses_response_websearch_test.go`
- extended `claude_openai-responses_request_test.go`

`go build ./...` clean; `go test ./internal/translator/...` passes.

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_request.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_request.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math/big"
+	"regexp"
 	"strings"
 
 	"github.com/google/uuid"
@@ -51,6 +52,13 @@ func ConvertOpenAIResponsesRequestToClaude(modelName string, inputRawJSON []byte
 
 	// Base Claude message payload
 	out := []byte(fmt.Sprintf(`{"model":"","max_tokens":32000,"messages":[],"metadata":{"user_id":"%s"}}`, userID))
+
+	// toolNameMap maps the *Anthropic-safe* name (i.e. what we send upstream)
+	// back to the original Codex tool name. Populated both when we sanitize a
+	// history function_call name and when a namespace tool's qualified name
+	// gets sanitized. Used downstream by tool_choice + by the response-side
+	// translator via setMcpToolNameOnItem to recover the original name.
+	toolNameMap := map[string]string{}
 
 	root := gjson.ParseBytes(rawJSON)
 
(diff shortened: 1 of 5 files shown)


xiaojiou176:
Thanks for the careful review — the string-content edge case is a real bug and is now fixed.

**Fixed (Gemini, HIGH):** both the assistant `tool_use` merge and the user `tool_result` merge now handle the case where the previous message's `content` is a plain string (the single-text legacy path). When `lastContent.Type == gjson.String`, the content is promoted to a `[{"type":"text","text":...}]` array first, then the tool block is appended — so it merges into the existing message instead of producing a consecutive same-role message (which would re-trigger the 400 this PR is meant to prevent).

Added a regression test `TestConvertOpenAIResponsesRequestToClaude_MergeToolUseIntoStringContent` that fails on the old code (two consecutive `assistant` messages) and passes after the fix. Full `go test ./internal/translator/... ./sdk/api/handlers/...` is green.

The two P2 suggestions from Codex (restoring original tool names on the response side for sanitized names, and tool_choice for sanitized names) are valid follow-ups; I'd prefer to keep this PR focused on the alternation/400 fixes and address those separately if maintainers agree.

xiaojiou176:
Follow-up fix for the Codex review on `cd98445d`:

**Fixed (Codex P2, @397 "Attach pending reasoning before merged tool calls"):** the string-content merge I added in the previous revision appended the `tool_use` to the prior assistant message but left `pendingReasoningParts` unflushed — so a later `function_call_output` would `flushPendingReasoning()` a standalone thinking message *between* the `tool_use` and its `tool_result`, breaking the sequence.

The merge path now flushes pending reasoning parts into the *same* assistant message, before the `tool_use`, so reasoning stays attached to the merged tool call. Added regression test `TestConvertOpenAIResponsesRequestToClaude_ReasoningAttachedToMergedToolCall` (asserts strict role alternation + reasoning co-located with the merged tool_use); it fails on the prior revision (`[assistant assistant user]`) and passes now. Full translator/handlers suite green.

Thanks for catching the regression — that interaction wasn't covered before.

PR #3703, opened 2026-06-03
feat(models):Add Antigravity Gemini 3.5 Flash extra-low model to models list

AbdoKnbGit:
Summary
- add Antigravity `gemini-3.5-flash-extra-low` to the embedded static model catalog
- mirror the existing `gemini-3.5-flash-low` limits/thinking metadata for the new backend low tier id
- extend existing static registry lookup coverage for the Antigravity entry without adding a new test file

Tests
- `go test ./internal/registry ./cmd/fetch_antigravity_models`
- `git diff --check`
- `git diff --cached --check`


AbdoKnbGit:
@luispater

PR #3716, opened 2026-06-05
fix(xai): normalize custom tool call input items

### LINKED ISSUE #3704: xai Responses executor sends unsupported custom_tool_call items and orphaned tool_choice, causing intermittent 422/400 on Grok models
### Describe the bug

When using xAI/Grok models through the Responses API path (e.g. `grok-composer-2.5-fast`, `grok-4.3`), requests intermittently fail with one of:

```
HTTP 422
{"error":"Failed to deserialize the JSON body into the target type: data did not match any variant of untagged enum ModelInput"}
```

```
HTTP 400
{"code":"Client specified an invalid argument","error":"Invalid request content: A tool_choice was set on the request but no tools were specified."}
```

It looks random, but it's content-dependent: it triggers based on what the conversation history and tool config contain, not on load or timing.

### Root cause

`internal/runtime/executor/xai_executor.go` translates incoming requests to the `codex` Responses format, then sanitizes them before POSTing to xAI's `/responses`. `normalizeXAITools` / `normalizeXAITool` already know xAI does not support custom tools — they drop `apply_patch` and convert `custom`-type tool **definitions** to `function`. But two related cases are not handled:

1. **`custom_tool_call` / `custom_tool_call_output` items in `input`.** When a prior turn used a custom tool (e.g. `ApplyPatch`/`apply_patch`), the history carries `custom_tool_call` items. xAI's `input` field is an untagged enum (`ModelInput`) with no `custom_tool_call` variant, so a single such item makes the **entire** request fail to deserialize → HTTP 422. The codex format and the real Codex API accept `custom_tool_call` natively, but xAI does not — the same asymmetry that `normalizeXAITool` already handles for tool definitions, just not for input items.

2. **Orphaned `tool_choice`.** When tool normalization drops all tools (e.g. the only tool was a custom `apply_patch`), a leftover `tool_choice` with an empty `tools` array makes xAI return HTTP 400.

`prepareResponsesRequest` currently runs:
```go
body = normalizeXAITools(body)
body = normalizeXAIInputReasoningItems(body)
body = normalizeCodexInstructions(body)
body = sanitizeXAIResponsesBody(body, baseModel)
```
There is no normalization of `custom_tool_call*` input items and no `tool_choice` cleanup.

### Reproduction

Against a running proxy with an xAI auth, on `POST /v1/responses`:

Root cause #1 — fails with 422:
```json
{"model":"grok-composer-2.5-fast","stream":false,
 "input":[{"role":"user","content":[{"type":"input_text","text":"hi"}]},
          {"type":"custom_tool_call","call_id":"c0","name":"ApplyPatch","input":"*** Begin Patch"}]}
```

Root cause #2 — fails with 400:
```json
{"model":"grok-composer-2.5-fast","stream":false,
 "input":[{"role":"user","content":[{"type":"input_text","text":"hi"}]}],
 "tools":[],"tool_choice":"auto"}
```

Verified: `custom_tool_call_output` alone, normal `function_call`/`function_call_output`, and assistant messages are all accepted — only `custom_tool_call` (the call) and the orphaned `tool_choice` are rejected.

### Suggested fix

Mirror the existing `normalizeXAITools` approach for the request body, xAI-only:

- Convert `custom_tool_call` → `function_call`, moving the freeform `input` string into a JSON-**object** `arguments` string (xAI rejects a non-object arguments value with `expected JSON object for tool arguments`); convert `custom_tool_call_output` → `function_call_output`.
- Drop `tool_choice` when no `tools` remain.

I have a tested patch for this (with unit tests + live verification against the xAI backend) and am happy to open a PR if useful.

### Environment
- CLIProxyAPI version: observed on v7.1.43; code path confirmed present on `main` HEAD.
- Models affected: xAI/Grok via Responses API (`grok-composer-2.5-fast`, etc.).

wilsonicdev:
Fixes #3704.

## Summary
- normalize xAI Responses `input` items from `custom_tool_call` to `function_call`
- map custom tool freeform `input` into JSON-object `arguments` strings accepted by xAI
- normalize `custom_tool_call_output` to `function_call_output` while preserving call output fields
- keep the existing xAI-only cleanup for orphaned `tool_choice`/`parallel_tool_calls` covered in the executor flow test
- incorporated Gemini review feedback to avoid repeated `input.Array()` calls and reduce JSON marshal/unmarshal overhead

## Tests
- `/usr/local/go/bin/go test ./internal/runtime/executor -run "TestXAIExecutorNormalizesCustomToolCallInput|TestNormalizeXAIInputCustomToolCalls|TestNormalizeXAIToolChoiceForTools" -count=1`
- `/usr/local/go/bin/go test ./internal/runtime/executor -count=1`
- `/usr/local/go/bin/go test ./... -count=1`
- `/usr/local/go/bin/go build -o test-output ./cmd/server && rm test-output`

Note: this stays in `internal/runtime/executor` and does not touch standalone translator code.

### DIFF
--- a/internal/runtime/executor/xai_executor.go
+++ b/internal/runtime/executor/xai_executor.go
@@ -33,7 +33,11 @@ const (
 	xaiImageHandlerType         = "openai-image"
 	xaiVideoHandlerType         = "openai-video"
 	xaiCustomToolType           = "custom"
+	xaiCustomToolCallType       = "custom_tool_call"
+	xaiCustomToolCallOutputType = "custom_tool_call_output"
 	xaiFunctionToolType         = "function"
+	xaiFunctionCallType         = "function_call"
+	xaiFunctionCallOutputType   = "function_call_output"
 	xaiImageGenerationToolType  = "image_generation"
 	xaiNamespaceToolType        = "namespace"
 	xaiToolSearchType           = "tool_search"
@@ -507,6 +511,7 @@ func (e *XAIExecutor) prepareResponsesRequest(ctx context.Context, req cliproxye
 	body, _ = sjson.DeleteBytes(body, "stream_options")
 	body = normalizeXAITools(body)
 	body = normalizeXAIToolChoiceForTools(body)
+	body = normalizeXAIInputCustomToolCalls(body)
 	body = normalizeXAIInputReasoningItems(body)
 	body = normalizeCodexInstructions(body)
 	body = sanitizeXAIResponsesBody(body, baseModel)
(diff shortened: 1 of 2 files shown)


wilsonicdev:
Applied the Gemini review suggestions in the latest force-push:

- cache `input.Array()` before iterating
- return existing JSON object arguments directly where possible
- wrap freeform strings/primitives with fewer marshal/unmarshal passes

Re-ran:
- `go test ./internal/runtime/executor -run "TestXAIExecutorNormalizesCustomToolCallInput|TestNormalizeXAIInputCustomToolCalls|TestNormalizeXAIToolChoiceForTools" -count=1`
- `go test ./... -count=1`

@codex review

PR #3719, opened 2026-06-05
fix Kimi Claude tool-use reasoning replay

MunMunMiao:
## Summary

- Normalize Kimi Claude-source requests before delegating to the Claude executor.
- Add `reasoning_content` to Claude-format assistant messages that contain `tool_use` blocks, using existing reasoning, thinking, or text content as fallback.
- Apply the same preprocessing to non-streaming, streaming, and count_tokens Kimi Claude-source paths.

## Root cause

Kimi Claude-source traffic was delegated directly to `ClaudeExecutor`, which skipped Kimi's existing tool-message normalizer. When thinking is enabled, Kimi rejects assistant tool-call history that lacks `reasoning_content`, causing 400 responses such as `thinking is enabled but reasoning_content is missing in assistant tool call message`.

## Testing

- `go test ./... -count=1`


AndersJet:
useful! thanks

PR #3733, opened 2026-06-05
fix(management): fallback latest version checks

GGBond6:
## Summary
- keep the existing GitHub releases API path for latest version checks
- fall back to GitHub /releases/latest redirect parsing when the API request fails, such as anonymous API rate limits
- add management handler tests covering the fallback path

## Testing
- go test ./internal/api/handlers/management


luispater:
Our team has optimized this feature.

PR #3737, opened 2026-06-06
fix(codex): strip unsupported image generation tools

### LINKED ISSUE #3725: OpenAI Responses → Codex translator passes through image_generation tools, causing "gpt-image-2 does not exist" error
### Problem

When a client sends a Responses API request through the Codex executor with an `image_generation` tool, the translator passes it through to Copilot unchanged. Copilot returns:

```json
{
  "error": {
    "message": "The model \u0027gpt-image-2\u0027 does not exist.",
    "type": "image_generation_user_error",
    "param": "tools",
    "code": "invalid_value"
  }
}
```

### Root Cause

In `internal/translator/codex/openai/responses/codex_openai-responses_request.go`, `ConvertOpenAIResponsesRequestToCodex()` clones the input JSON and never modifies the `tools` array. When the original request includes a Responses API built-in tool of type `image_generation` (with `model: "gpt-image-2"`), it is forwarded verbatim to the Copilot `/responses` endpoint, which rejects it.

The Responses API supports these built-in tool types: `image_generation`, `web_search_preview`, `code_interpreter`, `file_search`, etc. These are not valid tool types for the Codex endpoint.

### Impact

Users who route GPT-5.x models through the Copilot (Codex) provider get persistent errors even though they are not using image generation — the tool is added by their client SDK automatically for models that support it.

### Suggested Fix

In `ConvertOpenAIResponsesRequestToCodex()`, filter the `tools` array to remove non-function tool types before forwarding. The Codex executor only handles `function` type tools (as the chat-completions translator already does at line 278 of `codex_openai_request.go`).

Specifically, add tool filtering after line 103 and before `return rawJSON`:

```go
// Strip non-function tools (image_generation, web_search_preview, etc.)
// The Codex endpoint only supports function tools.
if tools := gjson.GetBytes(rawJSON, "tools"); tools.Exists() && tools.IsArray() {
    var filtered []map[string]any
    tools.ForEach(func(_, tool gjson.Result) bool {
        if tool.Get("type").String() == "function" {
            filtered = append(filtered, tool.Value().(map[string]any))
        }
        return true
    })
    if len(filtered) > 0 {
        rawJSON, _ = sjson.SetBytes(rawJSON, "tools", filtered)
    } else {
        rawJSON, _ = sjson.DeleteBytes(rawJSON, "tools")
    }
}
```

### Workaround

Affected users can temporarily use Chat Completions format instead of Responses API format, which correctly strips non-function tools during translation.

### Environment

- ProxyPal v0.4.38 (bundles CLIProxyAPI v7.x)
- Copilot (Codex) provider
- GPT-5.5 model (e.g., `gpt-5.5`)

yinkev:
Fixes #3725.

## Summary
- remove unsupported `image_generation` tools before forwarding OpenAI Responses requests to Codex
- drop the `tools` field when no supported tools remain
- preserve function tools and existing Codex built-in tool normalization

## Tests
- `go test ./internal/translator/codex/openai/...`
- `go test ./...`

### DIFF
--- a/internal/translator/codex/openai/responses/codex_openai-responses_request.go
+++ b/internal/translator/codex/openai/responses/codex_openai-responses_request.go
@@ -37,6 +37,7 @@ func ConvertOpenAIResponsesRequestToCodex(modelName string, inputRawJSON []byte,
 
 	// Delete the user field as it is not supported by the Codex upstream.
 	rawJSON, _ = sjson.DeleteBytes(rawJSON, "user")
+	rawJSON = filterUnsupportedTools(rawJSON)
 
 	// Convert role "system" to "developer" in input array to comply with Codex API requirements.
 	rawJSON = convertSystemRoleToDeveloper(rawJSON)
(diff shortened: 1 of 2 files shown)


yinkev:
I missed the repository path guard before opening this. CI correctly reports that changes under `internal/translator/**` are not accepted through pull requests and should be handled by the maintenance team through an issue.

Closing this to respect the repo workflow. The branch contains a tested implementation sketch for #3725 if useful to maintainers.

yinkev:
Closing because the repository path guard explicitly disallows pull requests that modify `internal/translator/**`. See #3725 for the issue-tracking path.

PR #3742, opened 2026-06-06
feat: add kiro support

qing8895:


luispater:
You can use our latest plugin mechanism to implement this feature.

https://github.com/router-for-me/CLIProxyAPI/tree/main/examples/plugin

PR #3747, opened 2026-06-07
fix: preserve OpenAI-compatible alias matching before canonical fallback

ht1072:
## Summary
- normalize executor provider keys consistently when registering and resolving executors
- preserve exact route-model matching before falling back to canonical/selection keys
- add a regression test for mixed-case provider identifiers with suffixed OpenAI-compatible aliases

## Problem
OpenAI-compatible aliases that include a visible suffix such as `(high)` can be listed successfully, but fail during auth selection when the runtime checks support only against canonicalized model keys. If the registry stores the exact alias while selection strips the suffix first, the candidate auth can be filtered out before any upstream request is attempted.

## Root cause
`authSupportsRouteModel()` checked canonicalized route keys before considering the exact route model string, while OpenAI-compatible registries can expose the exact alias string as the supported model id. In mixed provider-id casing scenarios, executor lookup also benefits from using the same normalization helper everywhere.

## Fix
- add `executorProviderKey()` and use it consistently for executor register/lookup/unregister and legacy picker paths
- make `authSupportsRouteModel()` try the raw `routeModel` first, then fall back to canonical/selection keys
- add regression coverage for `Identifier() == "YT"`, auth provider `yt`, and route model `YT-GPT-5.5(high)`

## Validation
```bash
go test ./sdk/cliproxy/auth -run 'TestManager(ExecutorReturnsRegisteredExecutor|MixedLegacyFindsExecutorWithCanonicalProviderKey|RegisterExecutorClosesReplacedExecutionSessions)$'
go build -o /tmp/cli-proxy-api-pr-verify ./cmd/server && rm -f /tmp/cli-proxy-api-pr-verify
```

### DIFF
--- a/sdk/cliproxy/auth/auto_refresh_loop.go
+++ b/sdk/cliproxy/auth/auto_refresh_loop.go
@@ -233,7 +233,7 @@ func (l *authAutoRefreshLoop) handleDueAuth(ctx context.Context, now time.Time, 
 	}
 	next, shouldSchedule := nextRefreshCheckAt(now, auth, l.interval)
 	shouldRefresh := manager.shouldRefresh(auth, now)
-	exec := manager.executors[auth.Provider]
+	exec := manager.executors[executorProviderKey(auth.Provider)]
 	manager.mu.RUnlock()
 
 	if !shouldSchedule {
--- a/sdk/cliproxy/auth/auto_refresh_loop_test.go
+++ b/sdk/cliproxy/auth/auto_refresh_loop_test.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"context"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


ht1072:
Thanks for the automated review feedback. I pushed a follow-up commit that:

- canonicalizes the auto-refresh executor lookup using the shared provider-key helper
- adds a regression test for mixed-case OpenAI-compatible provider keys in the auto-refresh path
- adds defensive nil guards in the legacy picker paths

Validation:
- `go test ./sdk/cliproxy/auth -run 'TestAutoRefreshLoopUsesCanonicalProviderKeyForExecutorLookup|TestManagerMixedLegacyFindsExecutorWithCanonicalProviderKey' -count=1`
- `go build -o /tmp/cli-proxy-api-pr3747 ./cmd/server`

GitHub checks are green on the latest commit.

ht1072:
Closing this PR rather than leaving a stale/conflicted branch open. The fix still looks directionally useful, but the branch is now diverged from `dev` and not mergeable; if still needed, it should be recut cleanly from current `dev` with the same focused regression coverage.

PR #3765, opened 2026-06-08
Fix Claude tool streaming via OpenAI-compatible providers

### LINKED ISSUE #3763: 小米mimo无法执行命令报错Invalid tool parameters
[  ] No, it's another issue.

**Describe the bug**

**CLI Type**
What type of CLI account do you use?  (gemini-cli, gemini, codex, claude code or openai-compatibility)
v7.1.50

**Model Name**
What model are you using? (example: gemini-2.5-pro, claude-sonnet-4-20250514, gpt-5, etc.)

**LLM Client**
What LLM Client are you using? (example: roo-code, cline, claude code, etc.)

**Request Information**
The best way is to paste the cURL command of the HTTP request here.
Alternatively, you can set `request-log: true` in the `config.yaml` file and then upload the detailed log file.

**Expected behavior**

**Screenshots**

**OS Type**
 - OS: [e.g. macOS]
 - Version [e.g. 15.6.0]

**Additional context**

Image

Image

Silence-dream:
Handle joined OpenAI-compatible SSE data frames before stream translation, suppress raw upstream fallback when native stream translators emit no output, and normalize double-encoded tool arguments for Claude tool_use inputs.

Closes #3763

## Summary

Fix Claude tool calls through OpenAI-compatible providers by making streaming response handling more robust.

## Changes

- Split joined OpenAI-compatible SSE `data:` frames before stream translation.
- Prevent native stream translators from falling back to raw upstream chunks when they intentionally emit no output.
- Normalize double-encoded tool call arguments into JSON objects for Claude `tool_use.input`.
- Add regression tests for joined SSE frames, empty native stream output, and double-encoded tool arguments.

## Root Cause

Some OpenAI-compatible providers can return multiple SSE `data:` frames joined on the same line. The previous stream reader passed the joined line to the translator as a single event. When the translator could not emit output for malformed/intermediate chunks, the SDK registry could fall back to the raw upstream chunk, causing OpenAI-formatted SSE data to leak into Claude streams.

Additionally, some providers return `function.arguments` as double-encoded JSON strings. These were not always normalized back into JSON objects, which could make Claude Code reject tool calls with `Invalid tool parameters`.

## Validation

- `go test ./internal/translator/openai/claude`
- `go test ./sdk/translator`

Closes #3763


luispater:
Fixed: 9985976ebd70849cd3e76baae14606c8c14cdd88

PR #3802, opened 2026-06-11
feat: expose DeepSeek cache split usage

cnYui:
## Summary
- Parse DeepSeek/OpenAI-style `prompt_cache_hit_tokens` and `prompt_cache_miss_tokens` into usage detail.
- Add cache hit/miss input token fields to the core usage record and Redis queue payload.
- Cover parser and queue payload behavior with focused tests.

## Test Plan
- go test -run 'TestParseOpenAIUsageDeepSeek|TestParseOpenAIUsage' ./internal/runtime/executor/helps && go test ./internal/redisqueue\n- go build -o test-output ./cmd/server && rm test-output\n- go test ./...\n- git diff --check\n\n## Notes\nThis lets downstream billing systems distinguish cache-hit input, cache-miss input, and output tokens instead of deriving everything from aggregate input/cached totals.

### DIFF
--- a/internal/redisqueue/plugin.go
+++ b/internal/redisqueue/plugin.go
@@ -72,10 +72,14 @@ func (p *usageQueuePlugin) HandleUsage(ctx context.Context, record coreusage.Rec
 		CacheCreationTokens:  record.Detail.CacheCreationTokens,
 		TotalTokens:          record.Detail.TotalTokens,
 	}
+	hasInputCacheStats := tokens.CachedTokens > 0 || tokens.CacheHitInputTokens > 0 || tokens.CacheMissInputTokens > 0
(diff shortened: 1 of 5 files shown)


cnYui:
Fixed in 4f7519e362139e3bb3f3cb37732ca8f8a94f9e0c.

Changes:
- infer OpenAI-style input tokens after cached token details are parsed, so cached-only usage reports input tokens correctly
- defensively infer Redis payload input tokens when cached tokens are present but input tokens are zero
- only infer cache_miss_input_tokens when input cache stats exist, so plain input_tokens are not counted as 100% cache miss

Verified locally:
- go test ./internal/runtime/executor/helps ./internal/redisqueue -count=1
- go test ./... -count=1
- go build ./...
- git diff --check

PR #3820, opened 2026-06-12
fix(management): preserve truncated API keys on config save

### LINKED ISSUE #3819: API keys truncated to xxx...xxxx on config save via management panel
### Description
When saving configuration through the management panel, API keys in config.yaml get truncated. Full keys like `sk-297bd661d39244348933f26adfaa63f5` become `sk-297...63f5` (literal dots 0x2E 0x2E 0x2E). This causes all API calls routed through those credentials to fail with 401.

### Evidence
- Backup `config_20260608_010202.yaml` has full keys
- Backup `config_20260612_122103.yaml` has truncated keys
- Binary update does NOT cause truncation (update script only replaces binary)

### Root Cause
The frontend `maskApiKey()` uses asterisks, not dots. The `...` pattern suggests the Go backend's config save/serialization logic truncates keys during processing.

### To Reproduce
1. Start cliproxyapi with valid config
2. Open management.html config editor
3. Save config (with or without changes)
4. Check config.yaml - keys now have `...`

### Fix Suggestion
Backend: On PUT config, compare incoming key values with current in-memory config. If incoming value matches a truncated pattern (`...` or `****`), preserve original key.

### Environment
cliproxapi v7.1.70, Debian 12

yubeilengzhi:
## Description

When the management panel saves configuration (e.g. PUT /v0/management/claude-api-key), API keys that appear truncated in the UI (containing literal "..." like "sk-xxx...xxxx") can be written back to config.yaml.

This adds a defense-in-depth check: before persisting, compare incoming API key values against the current in-memory config. If an incoming key looks truncated (contains "...") and the in-memory key is full, the original is preserved.

## Testing

- Verified "..." pattern (0x2E 0x2E 0x2E) is literal in corrupted configs
- Go backend normalize/sanitize functions do NOT truncate keys
- Fix is defensive: logs warning, doesn't block save

Fixes #3819

### DIFF
--- a/internal/api/handlers/management/config_lists.go
+++ b/internal/api/handlers/management/config_lists.go
@@ -301,6 +301,18 @@ func (h *Handler) PutClaudeKeys(c *gin.Context) {
 	}
 	h.mu.Lock()
 	defer h.mu.Unlock()
+	// Protect truncated API keys from being saved
+	for i := range arr {
+		if isTruncatedKey(arr[i].APIKey) && i < len(h.cfg.ClaudeKey) {
+			if h.cfg.ClaudeKey[i].APIKey != "" && !isTruncatedKey(h.cfg.ClaudeKey[i].APIKey) {
+				log.WithFields(log.Fields{
+					"handler": "PutClaudeKeys",
+					"index":   i,
+				}).Warn("detected truncated API key, preserving original from in-memory config")
+				arr[i].APIKey = h.cfg.ClaudeKey[i].APIKey
+			}
+		}
+	}
 	h.cfg.ClaudeKey = arr
 	h.cfg.SanitizeClaudeKeys()
 	h.persistLocked(c)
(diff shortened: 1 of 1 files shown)


luispater:
Thanks for the proposed fix. After reviewing the backend save paths in CLIProxyAPI, I did not find any code that rewrites or truncates API keys passed from the frontend. I also checked the Cli-Proxy-API-Management-Center project and did not find code that would transform a full key such as `sk-297bd661d39244348933f26adfaa63f5` into `sk-297...63f5`.

The backend may persist a truncated-looking value if that exact value is submitted, but it does not appear to generate it. Since the linked issue cannot currently be reproduced and may involve a third-party panel or external tooling, this backend preservation workaround is not something we should merge at this time.

If this can be reproduced without third-party tooling, please reopen with complete reproduction steps and the exact versions of both Cli-Proxy-API-Management-Center and CLIProxyAPI.

PR #3843, opened 2026-06-14
fix(management): preserve auth metadata and Codex priority on save

drwpls:
## Summary

When updating an auth file via the management API (status/fields patch), the save path previously serialized only the in-memory metadata payload. Any field that existed **only on disk** — and was not part of the update — was silently dropped. For Codex accounts this notably lost the account `priority` and subscription metadata on a simple enable/disable toggle.

This PR makes the save merge the existing on-disk metadata with the incoming update, so fields the caller did not send are preserved.

## Behavior

- On save, existing metadata is read from the auth file and merged under the incoming metadata (incoming wins).
- The provider `type` is backfilled only when missing/`unknown`, and the `disabled` flag is always written.
- Status/fields patches preserve unspecified metadata (including Codex `priority`) instead of overwriting the record wholesale.

## Changes

- `sdk/auth/filestore.go` — add `mergeAuthMetadataForSave(path, metadata, provider, disabled)`; the token store `Save` now merges with on-disk metadata before writing.
- `internal/api/handlers/management/auth_files.go` — `mergeAuthFileStatusMetadata` / `readAuthMetadataForStatusPatch` preserve existing metadata and Codex priority on a status patch.
- `internal/api/handlers/management/config_lists.go` — keep metadata intact on the related delete path.
- Unit tests: disabled-flag persistence, partial-metadata merge, and a regression test asserting an on-disk `priority` survives a partial-metadata save.

## Testing

- `go build ./...`
- `go test ./internal/api/handlers/management/... ./sdk/auth/...` — all pass
- `gofmt`/`go vet` clean

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -1214,8 +1214,11 @@ func (h *Handler) PatchAuthFileStatus(c *gin.Context) {
 	}
 
 	var req struct {
-		Name     string `json:"name"`
-		Disabled *bool  `json:"disabled"`
+		Name     string         `json:"name"`
+		Disabled *bool          `json:"disabled"`
+		Metadata map[string]any `json:"metadata"`
+		Type     string         `json:"type"`
+		Provider string         `json:"provider"`
 	}
 	if err := c.ShouldBindJSON(&req); err != nil {
 		c.JSON(http.StatusBadRequest, gin.H{"error": "invalid request body"})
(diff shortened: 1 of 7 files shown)


drwpls:
@hkfires bro can you have a look

GGGODLIN:
Hit this from a second path today: a plain `--codex-login` re-auth also drops the on-disk `priority`, not just management-API status/field patches.

Repro, observed on v7.2.75 with two Codex OAuth auths:

**Before** — `codex-<personal>-pro.json` had `"priority": 10`; `codex-<hash>-<work>-team.json` had `"priority": 0`.

Ran `cli-proxy-api --config ~/.cli-proxy-api/config.yaml --codex-login` to re-auth the personal account after its refresh token was invalidated server-side (`refresh_token_invalidated`).

**After** — the rewritten file's keys are:

```
["access_token","account_id","disabled","email","expired","id_token","last_refresh","refresh_token","type"]
```

`priority` is gone. `authPriority()` then defaults it to `0` (`sdk/cliproxy/auth/selector.go`), both auths tie at 0, and selection falls through to the `available[i].ID < available[j].ID` tie-break — so the **work** account won on filename ordering. Confirmed with a probe request: traffic kept going to the team account right after a successful re-auth of the account that was supposed to be preferred. Re-adding `"priority": 10` by hand restored the intended routing, picked up immediately by the auth-dir watcher.

The failure mode is quiet in a way that matters here: the re-login succeeds, subsequent requests succeed, and the account you just re-authed is silently demoted — so you keep spending the fallback account's quota with no error anywhere.

Since the `case auth.Storage != nil:` branch in `Save` writes only in-memory metadata, this looks like the same root cause this PR addresses, just reached through the login path instead of the management API. That branch is still unmerged on master (checked against v7.2.101), so it would be helpful to know whether the merge here is expected to cover the login path as well, or whether login needs its own on-disk metadata carry-over.

Working around it locally for now by snapshotting `email` -> `priority` before login and reapplying after.

PR #3859, opened 2026-06-15
fix: suspend auth on 400 credit balance too low, allowing fill-first to try next key

### LINKED ISSUE #3858: fill-first routing doesn't suspend auth on 400 'credit balance too low' — never tries next API key
## Description

When using fill-first routing with multiple Claude API keys, if the current key returns `HTTP 400 "Your credit balance is too low to access the Anthropic API"`, the conductor does **not** suspend the auth and does **not** try the next available key. It returns the error to the client immediately.

## Root Cause

In `sdk/cliproxy/auth/conductor.go`, the `MarkResult` method has a `switch statusCode` that handles:

| Status | Suspends? |
|--------|-----------|
| 401 | ✅ `unauthorized` |
| 402, 403 | ✅ `payment_required` |
| 404 | ✅ `not_found` |
| 429 | ✅ `quota` |
| 408, 500+ | ❌ cooling only |
| **400** (default) | **❌ does nothing** |

`400` falls through to the `default` case which only clears `NextRetryAfter` without suspending the auth. The next request picks the same key and gets the same error.

Additionally, `isRequestInvalidError` classifies the 400 response containing "invalid_request_error" as a request-shape failure, telling the routing layer **not** to retry with another key — even though switching keys would fix the problem.

## Evidence

From ProxyPal logs (which runs CLIProxyAPI as a sidecar):

```
[13:41:48] Use API key sk-a...wAAA for model claude-opus-4-8
[13:41:49] 400: Your credit balance is too low to access the Anthropic API
```
No `Suspended client` entry follows (unlike 401 which correctly triggers suspension).

## Expected Behavior

1. `400` response with "credit balance too low" should suspend the auth for ~30 minutes
2. `isRequestInvalidError` should **not** classify insufficient-balance errors as request-shape failures
3. The routing layer should then pick the next available key

## Suggested Fix

Already implemented in PR: add `case 400` with message-based detection for credit balance errors in `MarkResult`, and exclude them from `isRequestInvalidError`.

izzzzzi:
## Description

When using `fill-first` routing with multiple Claude API keys, a `400 "Your credit balance is too low"` error was never triggering auth suspension, causing the routing layer to keep selecting the same broken key and returning the error to the client — even if another configured key had sufficient balance.

### Root Cause

In `MarkResult` (`sdk/cliproxy/auth/conductor.go`), the `switch statusCode` handles 401, 402/403, 404, 429, and 5xx — but **400 fell through to `default`**, which only clears `NextRetryAfter` without suspending.

Additionally, `isRequestInvalidError` matched the "invalid_request_error" type in Anthropic's 400 response, classifying it as a request-shape failure — telling the router **not to retry** with another key.

### Changes

| File | Change |
|------|--------|
| `sdk/cliproxy/auth/conductor.go` | +31 lines |

1. **`case 400:`** in `MarkResult` — checks if the error message indicates insufficient balance via `isInsufficientBalanceResultError`. If matched, suspends the auth for 30 minutes with reason `insufficient_balance`, allowing `fill-first` to try the next key.

2. **`isInsufficientBalanceResultError`** — new helper that detects patterns: `credit balance`, `insufficient balance`, `balance is too low`, `insufficient_quota` (case-insensitive).

3. **`isRequestInvalidError`** — excludes insufficient-balance messages from the "invalid request" classification, so the routing layer attempts other keys.

### Testing

- [x] Verified `isInsufficientBalanceResultError` matches Anthropic's exact error: `Your credit balance is too low to access the Anthropic API`
- [x] 400 with other messages (e.g. invalid JSON, bad model name) still falls through to `default` (no suspension)
- [x] `isRequestInvalidError` no longer blocks retry for balance errors

Closes #3858

### Related

- ProxyPal issue: heyhuynhgiabuu/proxypal#227
- ProxyPal PR (workaround): heyhuynhgiabuu/proxypal#228

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -2849,6 +2849,19 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 								shouldSuspendModel = true
 								setModelQuota = true
 							}
+						case 400:
+							if isInsufficientBalanceResultError(result.Error) {
+								if disableCooling {
+									state.NextRetryAfter = time.Time{}
+								} else {
+									next := now.Add(30 * time.Minute)
+									state.NextRetryAfter = next
+									suspendReason = "insufficient_balance"
+									shouldSuspendModel = true
+								}
+							} else {
+								state.NextRetryAfter = time.Time{}
+							}
 						case 408, 500, 502, 503, 504:
 							if disableCooling {
 								state.NextRetryAfter = time.Time{}
@@ -3247,6 +3260,19 @@ func isRequestScopedNotFoundResultError(err *Error) bool {
 	return isRequestScopedNotFoundMessage(err.Message)
 }
 
+// isInsufficientBalanceResultError returns true if the error indicates the
+// API key or account has insufficient credit balance to process the request.
+func isInsufficientBalanceResultError(err *Error) bool {
+	if err == nil {
+		return false
+	}
+	lower := strings.ToLower(strings.TrimSpace(err.Message))
+	return strings.Contains(lower, "credit balance") ||
+		strings.Contains(lower, "insufficient balance") ||
+		strings.Contains(lower, "balance is too low") ||
+		strings.Contains(lower, "insufficient_quota")
+}
+
 // isRequestInvalidError returns true if the error represents a client request
 // error that should not be retried. Specifically, it treats 400 responses with
 // "invalid_request_error", request-scoped 404 item misses caused by `store=false`,
@@ -3267,6 +3293,11 @@ func isRequestInvalidError(err error) bool {
 	switch status {
 	case http.StatusBadRequest:
 		msg := err.Error()
+		if strings.Contains(msg, "credit balance") ||
+			strings.Contains(msg, "insufficient balance") ||
+			strings.Contains(msg, "balance is too low") {
+			return false
+		}
 		return strings.Contains(msg, "invalid_request_error") ||
 			strings.Contains(msg, "INVALID_ARGUMENT") ||
 			strings.Contains(msg, "FAILED_PRECONDITION")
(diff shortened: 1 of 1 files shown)


luispater:
The HTTP 400 status code stands for 'Bad Request'.

A 'credit balance too low' scenario falls under the 401 error code 'Payment Required'.

Generally, a 400 status code indicates a malformed or invalid request payload.

izzzzzi:
> The HTTP 400 status code stands for 'Bad Request'. A 'credit balance too low' scenario falls under the 402 error code 'Payment Required'. Generally, a 400 status code indicates a malformed or invalid request payload.

I understand the HTTP spec concern, but unfortunately Anthropic's API does **not** return 402 for insufficient balance — it returns **400** with the same `invalid_request_error` type as malformed requests:

```
HTTP/1.1 400 Bad Request

  "message":"Your credit balance is too low to access the Anthropic API."}}
```

This is a known, long-standing issue that the community has raised with Anthropic multiple times:

- **anthropics/anthropic-sdk-typescript#618** (Nov 2024, still open): *"Insufficient credit error has the same error code as bad requests"* — the SDK maintainers themselves asked Anthropic to return 402 or a distinct error type, but it was never changed.
- **anthropics/anthropic-sdk-python#1646** (Jun 2026): *"Add machine-readable error code/type for billing/quota errors"* — same request, still unresolved.

Since Anthropic provides no machine-readable signal (no separate status code, no distinct `error.type`, no `error.code`), the only reliable way to distinguish billing exhaustion from a genuine bad request is matching the error message text. This is the approach used by every major proxy and SDK in the ecosystem (litellm, portkey.ai, continue.dev, etc.).

The fix here correctly handles the **actual behavior** of the upstream API. Anthropic returning 402 would be ideal per spec, but that's an upstream change beyond our control.

PR #3875, opened 2026-06-17
fix(openai/gemini): avoid duplicate tool call IDs

### LINKED ISSUE #3874: Gemini-to-OpenAI translator assigns duplicate tool_call_id for multiple function responses
## Summary

When the Gemini-to-OpenAI request translator handles multiple Gemini `functionResponse` parts after multiple `functionCall` parts, it assigns every translated OpenAI `tool` message the last generated `tool_call_id`.

OpenAI-compatible upstreams that validate tool result chains reject the converted request with a duplicate `tool_call_id` error.

## Observed error

This was observed when using Gemini CLI against an `/llm-gateway` Gemini endpoint routed to a DeepSeek OpenAI-compatible upstream:

```text
Duplicate value for 'tool_call_id' of call_<redacted> in message[5]
```

The request path was Gemini wire format -> `ConvertGeminiRequestToOpenAI` -> DeepSeek OpenAI-compatible API.

## Root cause

In `internal/translator/openai/gemini/openai_gemini_request.go`, the translator tracks generated tool call IDs in `toolCallIDs`, but each `functionResponse` currently uses the last generated ID:

```go
toolMsg, _ = sjson.SetBytes(toolMsg, "tool_call_id", toolCallIDs[len(toolCallIDs)-1])
```

For a Gemini message containing parallel/multiple function calls and responses, this produces multiple OpenAI `tool` messages with the same `tool_call_id`.

## Expected behavior

Each `functionResponse` should consume one pending tool call ID so that translated OpenAI `tool` messages keep a valid one-to-one relationship with prior assistant `tool_calls`.

## Proposed fix

Use FIFO consumption for `toolCallIDs`:

```go
var toolCallIDs []string
var toolCallConsumeIdx int

// ... when translating functionResponse:
if toolCallConsumeIdx < len(toolCallIDs) {
    toolMsg, _ = sjson.SetBytes(toolMsg, "tool_call_id", toolCallIDs[toolCallConsumeIdx])
    toolCallConsumeIdx++
} else {
    toolMsg, _ = sjson.SetBytes(toolMsg, "tool_call_id", genToolCallID())
}
```

This keeps existing fallback behavior when there are more responses than calls, while preventing duplicate IDs in normal multi-tool conversations.

## Test coverage to add

- Multiple `functionCall` + multiple `functionResponse` parts produce distinct `tool_call_id` values.
- FIFO consumption works for three calls/responses.
- A `functionResponse` without a prior `functionCall` still receives a generated fallback ID.
- Extra responses beyond the pending call count receive generated fallback IDs without duplicating consumed IDs.

yizishou:
## Summary

- Fix Gemini-to-OpenAI tool result translation so each `functionResponse` consumes the next pending `tool_call_id` instead of reusing the last one.
- Preserve generated fallback IDs when a function response has no pending matching call.
- Add unit tests for multi-call/multi-response, FIFO consumption, missing calls, and extra responses.

Fixes #3874.

## Validation

```text
go test ./internal/translator/openai/gemini/ -v -count=1
go build -o /tmp/cliproxy-test-output ./cmd/server
```

### DIFF
--- a/internal/translator/openai/gemini/openai_gemini_request.go
+++ b/internal/translator/openai/gemini/openai_gemini_request.go
@@ -112,7 +112,8 @@ func ConvertGeminiRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 	out, _ = sjson.SetBytes(out, "stream", stream)
 
 	// Process contents (Gemini messages) -> OpenAI messages
-	var toolCallIDs []string // Track tool call IDs for matching with tool results
+	var toolCallIDs []string   // Track tool call IDs for matching with tool results
+	var toolCallConsumeIdx int // Next index into toolCallIDs to consume for function responses
 
 	// System instruction -> OpenAI system message
 	// Gemini may provide `systemInstruction` or `system_instruction`; support both keys.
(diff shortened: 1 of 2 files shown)


yizishou:
Updated the tests to address the review feedback: they now collect assistant `tool_calls[*].id` and translated `tool` message `tool_call_id`s, then assert FIFO alignment as well as uniqueness. Re-ran validation locally:

```text
go test ./internal/translator/openai/gemini/ -v -count=1
go build -o /tmp/cliproxy-test-output ./cmd/server
```

Both passed.

PR #3895, opened 2026-06-18
fix: rebind session affinity after fallback success

yueziji:
## Summary

Fix session affinity so a request that falls back to another auth after an initial auth fails will stick to the auth that actually succeeds.

Previously, session affinity was bound when the selector first picked an auth. If that auth failed and the same request retried another auth successfully, the sticky binding could still point to the earlier auth until a later invalidation happened.

## What changed

- add a narrow `Bind(...)` hook to `SessionAffinitySelector`
- rebind session affinity on successful execute/count/stream success paths
- preserve fallback session-hash inheritance when rebinding
- add a manager-level regression test covering:
  - first auth fails
  - fallback auth succeeds
  - next request sticks to the successful fallback auth

## Validation

- `go test ./...`
- `go build -o cli-proxy-api ./cmd/server`

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -123,6 +123,10 @@ type Selector interface {
 	Pick(ctx context.Context, provider, model string, opts cliproxyexecutor.Options, auths []*Auth) (*Auth, error)
 }
 
+type sessionAffinityBinder interface {
+	Bind(ctx context.Context, provider, model string, opts cliproxyexecutor.Options, authID string)
+}
+
 type PluginScheduler interface {
 	PickAuth(context.Context, pluginapi.SchedulerPickRequest) (pluginapi.SchedulerPickResponse, bool, error)
 }
@@ -1218,6 +1222,7 @@ func (m *Manager) executeStreamWithModelPool(ctx context.Context, executor Provi
 			close(closedCh)
 			remaining = closedCh
 		}
+		m.bindSessionAffinitySuccess(ctx, provider, routeModel, opts, auth.ID)
 		return m.wrapStreamResult(ctx, auth.Clone(), provider, resultModel, streamResult.Headers, buffered, remaining), nil
 	}
 	if lastErr == nil {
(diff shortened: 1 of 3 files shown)


yueziji:
Addressed the reasonable review points:

- removed `t.Parallel()` from the new session-affinity test because it mutates the shared global registry
- fixed the rebind path to use the original request `opts` instead of post-interceptor `execOpts`, so successful fallback rebinding targets the same session-affinity key that `Pick()` used
- added a regression test covering interceptor body rewrite so this does not regress

I left the extra `extractSessionIDs` parse as-is for now. I agree it is a real optimization opportunity, but it is not a correctness issue after the original-request fix, and threading precomputed session IDs through every execute path felt too large for this PR.

yueziji:
Addressed the new mixed-key issue as well.

`SessionAffinitySelector.Pick()` reads/writes under the selector provider key, which is `"mixed"` on the mixed-provider path, while my success rebind hook was using the selected auth provider (for example `claude`). That meant the rebind could update `claude::...` while subsequent mixed selection still looked up `mixed::...`.

I changed the success rebind to use the same selector provider key that selection used, and added a regression test covering the mixed-provider path.

yueziji:
Addressed the latest mixed-stream affinity feedback.

This follow-up removes the provider-scoped success rebind from the shared stream helper on mixed-provider paths, so mixed `ExecuteStream()` now only persists affinity under the `mixed::<session>::<model>` key.

Also added a regression test covering mixed stream fallback + rebind behavior.

yueziji:
Closing this in favor of the upstream implementation.

`9169ad56` ("fix(auth): make session-affinity updates safe for rebound sessions") solves the same problem with a broader approach: it propagates the session-affinity provider/model namespace through `Options.Metadata` into `Result`, then lets `SessionAffinitySelector.OnResult()` handle binding via `Touch` and release via `CompareAndDelete`, so a stale failure can no longer delete a session that was already rebound to another auth. `00c4377a` additionally shares affinity across model variant suffixes via `canonicalModelKey`.

That covers the mixed-provider success rebind this PR was targeting, and it does so on the result path rather than at individual call sites, so no further work is needed here.

PR #3941, opened 2026-06-20
fix(usage): group openai-compatibility api-key-usage by bare provider name

### LINKED ISSUE #3600: OpenAI-compat entry with matching name overrides built-in executor
Is it a request payload issue?
  [  ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an
  unexpected error.
  [x] No, it's another issue.

  Describe the bug
  When the config contains an OpenAI-compatible provider whose name matches a built-in provider (e.g. "kimi"), a config
  reload can cause the OpenAICompatExecutor to override the dedicated executor (KimiExecutor). After that, scheduled
  token refresh becomes a no-op and the auth token eventually expires.

  The root cause is in ensureExecutorsForAuthWithMode: the OpenAI-compat check runs before the built-in provider switch.
   rebindExecutors() iterates over a map in random order, so if the OpenAI-compat auth named "kimi" is visited after the
   OAuth "kimi" auth, OpenAICompatExecutor("kimi") overwrites KimiExecutor in m.executors["kimi"].

  CLI Type
  Server-side issue (not client-specific)

  Model Name
  N/A

  LLM Client
  N/A

  Request Information
  Log evidence (sensitive info redacted):

  Normal refresh (before config reload):
  [kimi_executor.go:575] kimi executor: refresh called
  [conductor.go:4162] refreshed kimi, <auth-id>, <nil>
  [events.go:82] file system event detected: WRITE <auth-file-path>
  [events.go:122] auth file changed (WRITE): <auth-file-name>, processing incrementally

  After config reload (bug):
  [auto_refresh_loop.go:194] auto-refresh scheduler due auths: 1
  [openai_compat_executor.go:606] openai compat executor: refresh called
  [conductor.go:4162] refreshed kimi, <auth-id>, <nil>
  Note: No executor refresh called log from KimiExecutor, and no file WRITE event follows.

  Expected behavior
  Built-in provider executors should always take precedence over OpenAI-compatible entries with matching names.
  KimiExecutor.Refresh() should be called for provider: kimi auths regardless of whether an OpenAI-compat provider also
  happens to be named "kimi".

  Screenshots
  N/A

  OS Type
  - OS: Linux
  - Version: N/A (server-side)

  Additional context
  Affected providers: any built-in name that collides with an OpenAI-compat compat_name / provider_key: kimi, claude,
  gemini, vertex, gemini-cli, aistudio, antigravity, xai, codex.

  Suggested fix: move the built-in provider switch before the OpenAI-compat check in ensureExecutorsForAuthWithMode
  (sdk/cliproxy/service.go).

### LINKED ISSUE #3940: bug(usage): openai-compatibility providers grouped as openai-compatible-<name> break Management Center recent-requests match (regression from #3600 / 4c78e40d)
## Bug

Recent-requests / per-key usage for `openai-compatibility` providers no longer matches in the Management Center panel after the provider-key unification in `4c78e40d` (v7.2.21+). The management `GET /v0/management/api-key-usage` now groups OpenAI-compatible providers under a namespaced key (`openai-compatible-<name>`), while the panel looks them up by the bare provider `name`. As a result the recent-request status bar and counts for every `openai-compatibility` provider show as empty/idle, even though traffic is flowing.

## Reproduction

`config.yaml`:

```yaml
openai-compatibility:
  - name: VAST
    base-url: https://www.vastnum.com/v1
    api-key-entries:
      - api-key: sk-...
```

1. Send traffic through the provider.
2. `GET /v0/management/api-key-usage` returns:

```json
{
  "openai-compatible-vast": {
    "https://www.vastnum.com/v1|sk-...": { "success": N, "failed": 0, "recent_requests": [...] }
  }
}
```

3. The Management Center "AI Providers" page aggregates OpenAI-compatible providers by `provider.name.toLowerCase()` (`"vast"`), so it looks up key `"vast"`, which is absent. The recent-request status bar and totals stay empty.

## Root cause

`4c78e40d` (Closes #3600) made `synthesizeOpenAICompat` store `auth.Provider = util.OpenAICompatibleProviderKey(providerName)`, i.e. `"openai-compatible-vast"` instead of the bare `"vast"` it stored before.

`GetAPIKeyUsage` groups by `strings.ToLower(strings.TrimSpace(auth.Provider))`:

```go
provider := strings.ToLower(strings.TrimSpace(auth.Provider))
// = "openai-compatible-vast"  for openai-compatibility auths
```

The panel (`Cli-Proxy-API-Management-Center`) matches OpenAI-compatible providers by the bare config `name` lowercased (`"vast"`), with no `openai-compatible-` prefix (confirmed the installed bundle contains no such prefix logic). So the two sides never agree.

Before `4c78e40d`, `auth.Provider` was the bare lowercased `providerName`, which matched the panel. So this is a regression introduced by #3600's fix — the namespacing is needed for executor routing (#3600), but it leaked into the usage/statistics grouping, which does not need it.

## Expected

OpenAI-compatible provider recent-requests and totals display correctly in the Management Center, i.e. the usage grouping uses the same key the panel looks up.

## Suggested fix (backend, minimal)

In `GetAPIKeyUsage`, group OpenAI-compatible auths by their bare `compat_name` (already stored on the auth attributes) instead of the namespaced `auth.Provider`. This keeps #3600's executor namespacing intact (that path is unaffected) while restoring the panel match:

```go
provider := strings.ToLower(strings.TrimSpace(auth.Provider))
if strings.EqualFold(strings.TrimSpace(auth.Provider), "openai-compatibility") ||
   (auth.Attributes != nil && strings.TrimSpace(auth.Attributes["compat_name"]) != "") {
    if name := strings.TrimSpace(auth.Attributes["compat_name"]); name != "" {
        provider = strings.ToLower(name)
    }
}
```

(Alternative: have the panel prepend `openai-compatible-` when looking up OpenAI-compatible providers. Backend fix is preferred since the usage API is the source of truth and `compat_name` is already available.)

## Environment

- CLIProxyAPI v7.2.22 (regression entered at v7.2.21 via `4c78e40d`)
- Management Center: current

Vxider:
## Summary

Fixes #3940. Restores recent-requests / per-key usage display for `openai-compatibility` providers in the Management Center, broken as a side effect of the provider-key unification in `4c78e40d` (which closed #3600).

## Problem

`4c78e40d` made `synthesizeOpenAICompat` store `auth.Provider = util.OpenAICompatibleProviderKey(name)`, i.e. the namespaced key `openai-compatible-<name>` (e.g. `openai-compatible-vast`) instead of the bare lowercased `name`. That namespace is the correct fix for #3600 at the **executor/routing** layer.

However, `GetAPIKeyUsage` groups usage statistics by `strings.ToLower(auth.Provider)`:

```go
provider := strings.ToLower(strings.TrimSpace(auth.Provider))
// = "openai-compatible-vast"  for openai-compatibility auths
```

The Management Center panel addresses OpenAI-compatible providers by their bare config `name` lowercased (e.g. `vast`), with no `openai-compatible-` prefix. Result: the two sides never match, so the recent-request status bar and totals are empty for **every** `openai-compatibility` provider, even though traffic flows normally.

Before `4c78e40d`, `auth.Provider` was the bare lowercased `name`, which matched the panel — so this is a regression introduced when fixing #3600.

## Fix

Group usage by the **bare `compat_name`** (already stored on the auth attributes by `synthesizeOpenAICompat`) for OpenAI-compatible auths, instead of the namespaced `auth.Provider`. `auth.Provider` is left untouched, so #3600's executor namespacing is fully preserved — only the usage/statistics grouping is realigned.

```go
func usageProviderKey(auth *coreauth.Auth) string {
	provider := strings.ToLower(strings.TrimSpace(auth.Provider))
	if provider == "" {
		return "unknown"
	}
	if auth.Attributes != nil {
		if compatName := strings.TrimSpace(auth.Attributes["compat_name"]); compatName != "" {
			return strings.ToLower(compatName)
		}
	}
	return provider
}
```

A unit test (`TestGetAPIKeyUsage_GroupsByProviderAndAPIKey`) is extended to register an openai-compatibility auth whose `auth.Provider` is the namespaced `openai-compatible-vast` with `compat_name = "VAST"`, and assert it is grouped under `vast` (not under `openai-compatible-vast`). Verified the test fails on `main` without the fix and passes with it.

## Why backend, not panel

The usage API is the source of truth and `compat_name` is already available on the auth, so the backend change is minimal and self-contained. The alternative (have the panel prepend `openai-compatible-` when looking up OpenAI-compatible providers) would couple every consumer of the usage API to the internal namespace.

## Verification

- `go vet ./internal/api/handlers/management/` — clean
- `go test ./internal/api/handlers/management/ -run TestGetAPIKeyUsage` — PASS (with fix), FAIL (without fix, on the new assertion)

A pre-existing unrelated failure (`TestWriteOAuthCallbackFileForPendingSessionCreatesMissingAuthDirForCallbackProviders/gemini`) is present on clean `main` as well and is not affected by this change.

## Risk

Low. The change is confined to the read-only usage-aggregation handler; request routing, executor registration, cooldown, and the #3600 collision fix are all keyed off `auth.Provider` / `provider_key` elsewhere and are untouched.

Fixes #3940

### DIFF
--- a/internal/api/handlers/management/api_key_usage.go
+++ b/internal/api/handlers/management/api_key_usage.go
@@ -40,6 +40,30 @@ func mergeRecentRequestBuckets(dst, src []coreauth.RecentRequestBucket) []coreau
 	return dst
 }
 
+
+// usageProviderKey resolves the provider key used to group usage statistics.
+//
+// For OpenAI-compatible providers, auth.Provider carries the namespaced key
+// introduced by provider-key unification (e.g. "openai-compatible-vast"); that
+// namespace is required for executor routing (see #3600). The usage/statistics
+// surface and the Management Center panel, however, address these providers by
+// their bare config name (e.g. "vast"), which is preserved on the auth as
+// compat_name. Returning the bare name here keeps the usage grouping aligned
+// with what the panel looks up, restoring recent-requests/totals display for
+// OpenAI-compatible providers. Other provider kinds keep strings.ToLower(auth.Provider).
+func usageProviderKey(auth *coreauth.Auth) string {
+	provider := strings.ToLower(strings.TrimSpace(auth.Provider))
+	if provider == "" {
+		return "unknown"
+	}
+	if auth.Attributes != nil {
+		if compatName := strings.TrimSpace(auth.Attributes["compat_name"]); compatName != "" {
+			return strings.ToLower(compatName)
+		}
+	}
+	return provider
+}
+
 // GetAPIKeyUsage returns recent request buckets for all in-memory api_key auths,
 // grouped by provider and keyed by "base_url|api_key".
 func (h *Handler) GetAPIKeyUsage(c *gin.Context) {
@@ -78,10 +102,7 @@ func (h *Handler) GetAPIKeyUsage(c *gin.Context) {
 			}
 		}
 		compositeKey := baseURL + "|" + apiKey
-		provider := strings.ToLower(strings.TrimSpace(auth.Provider))
-		if provider == "" {
-			provider = "unknown"
-		}
+		provider := usageProviderKey(auth)
 
 		recent := auth.RecentRequestsSnapshot(now)
 		providerBucket, ok := out[provider]
--- a/internal/api/handlers/management/api_key_usage.go
+++ b/internal/api/handlers/management/api_key_usage.go
@@ -40,7 +40,6 @@ func mergeRecentRequestBuckets(dst, src []coreauth.RecentRequestBucket) []coreau
 	return dst
 }
 
-
 // usageProviderKey resolves the provider key used to group usage statistics.
 //
 // For OpenAI-compatible providers, auth.Provider carries the namespaced key
@@ -51,7 +50,12 @@ func mergeRecentRequestBuckets(dst, src []coreauth.RecentRequestBucket) []coreau
 // compat_name. Returning the bare name here keeps the usage grouping aligned
 // with what the panel looks up, restoring recent-requests/totals display for
 // OpenAI-compatible providers. Other provider kinds keep strings.ToLower(auth.Provider).
+// If compat_name is missing for some reason, the openai-compatible- namespace
+// prefix is stripped as a fallback so the bare name still matches the panel.
 func usageProviderKey(auth *coreauth.Auth) string {
+	if auth == nil {
+		return "unknown"
+	}
 	provider := strings.ToLower(strings.TrimSpace(auth.Provider))
 	if provider == "" {
 		return "unknown"
@@ -61,7 +65,10 @@ func usageProviderKey(auth *coreauth.Auth) string {
 			return strings.ToLower(compatName)
 		}
 	}
-	return provider
+	// Fallback: an OpenAI-compatible auth that reached here without compat_name
+	// still carries the namespaced provider key; strip the namespace so the panel
+	// (which looks up by the bare name) can match its usage.
+	return strings.TrimPrefix(provider, "openai-compatible-")
 }
 
 // GetAPIKeyUsage returns recent request buckets for all in-memory api_key auths,
--- a/internal/api/handlers/management/api_key_usage_test.go
+++ b/internal/api/handlers/management/api_key_usage_test.go
@@ -127,6 +127,52 @@ func TestGetAPIKeyUsage_GroupsByProviderAndAPIKey(t *testing.T) {
 		t.Fatalf("vast totals = %d/%d, want 1/1", vastSuccess, vastFailed)
 	}
 }
+func TestUsageProviderKey(t *testing.T) {
+	tests := []struct {
+		name string
+		auth *coreauth.Auth
+		want string
+	}{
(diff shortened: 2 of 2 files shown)


Vxider:
Both review comments are addressed in `0fe8208e`:

- **gemini-code-assist**: `usageProviderKey` now guards `auth == nil` and strips the `openai-compatible-` prefix as a fallback when `compat_name` is absent, so openai-compat usage is grouped under the bare name in every case.
- **codex review (P2, gofmt)**: ran `gofmt -w`; `gofmt -d` is now clean (removed the extra blank line).

Added `TestUsageProviderKey` table test covering the nil / empty / passthrough / compat_name / fallback branches.

Local verification (the repo's `pr-test-build` CI only compiles, it doesn't run `go test`):
- `go build ./cmd/server` → OK
- `go test ./internal/api/handlers/management/ -run 'TestGetAPIKeyUsage|TestUsageProviderKey'` → 6 cases PASS
- `gofmt -d` → no diff

Issue #3817, opened 2026-06-12
Bug: 接入Mimo时，Third-party providers with empty thinking signatures cause Claude client failure


msuadOf:
## Bug Description

When using a third-party API provider (mimo-v2.5-pro-ultraspeed) through CPA with Anthropic format, Claude Desktop/Code client fails to process responses because thinking blocks have empty signatures.

## Steps to Reproduce

1. Configure CPA to use a third-party API provider with Anthropic format endpoint
2. Use API key "speedtest" 
3. Send a request to the CPA endpoint
4. Observe that Claude client fails to process the response

## Expected Behavior

Claude client should successfully process the response from the third-party provider.

## Actual Behavior

Claude client fails to process the response. The issue is that thinking blocks in the response have empty signature fields.

## Root Cause

The third-party API provider returns thinking blocks with empty `signature` fields:

```json
{
  "type": "thinking",
  "thinking": "The user just said hello...",
  "signature": ""  // Empty signature
}
```

And in streaming mode:
```json
{
  "type": "signature_delta",
  "signature": ""  // Empty signature
}
```

Claude Desktop/Code client validates thinking block signatures and rejects responses with empty signatures.

## Evidence

Testing with curl shows the response is technically valid, but has empty signatures:

```bash
curl -X POST http://<cpa-server>:8317/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: <token-key>" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "model": "mimo-v2.5-pro-ultraspeed",
    "max_tokens": 1024,
    "messages": [{"role": "user", "content": "Hello"}]
  }'
```

Response:
```json
{
  "type": "message",
  "content": [
    {
      "type": "thinking",
      "thinking": "The user just said hello...",
      "signature": ""
    },
    {
      "type": "text",
      "text": "Hello! How can I help you today?"
    }
  ]
}
```

## Suggested Fix

CPA should handle third-party providers that return empty signatures in one of these ways:

1. **Generate a valid signature** - If the provider returns an empty signature, CPA could generate a valid signature before forwarding to the client
2. **Strip thinking blocks** - Remove thinking blocks entirely if they have empty signatures
3. **Add a configuration option** - Allow users to configure how to handle empty signatures (e.g., `strip-empty-signatures: true`)

## Related Issues

- #3052 - Claude Desktop Cowork multi-turn 400 "Invalid signature in thinking block"
[...]

Issue #3601, opened 2026-05-28
kimi 有问题 claude调用新版本有错误


q2467481940:
kimi 有问题 claude调用新版本有错误             Thought for 6s, searched for 1 pattern, listed 1 directory (ctrl+o to expand)
  ⎿  API Error: 400 thinking is enabled but reasoning_content is missing in assistant tool call message at index 7

luispater:
kimi建议使用OAuth的方式登录使用，你需要上传request log，来定位具体问题。

orange-XD:
同样的问题，使用OAuth也是一样，已经回退CPA版本了

MunMunMiao:
我提交了修复 https://github.com/router-for-me/CLIProxyAPI/pull/3719

q2467481940:
API Error: 400 thinking is enabled but reasoning_content is missing in assistant tool call message at 
     index 6

q2467481940:
> kimi建议使用OAuth的方式登录使用，你需要上传request log，来定位具体问题。

 API Error: 400 thinking is enabled but reasoning_content is missing in assistant tool call message at 
     index 6

MunMunMiao:
@luispater 看下能否合并

q2467481940:
> [@luispater](https://github.com/luispater) 看下能否合并
你测试了吗 这个不会异常消耗token吗 我之前本地修了一下 导致没有缓存了

luispater:
> [@luispater](https://github.com/luispater) 看下能否合并

不能合并，你在claude执行器中增加了第三方兼容的代码，这是不允许的。

claude执行器是为了claude而设计工作的。

正确的做法应该是修改kimi执行器。或者使用插件对请求及响应进行规整。

sususu98:
## Proposed fix direction (aligned with maintainer guidance)

This error is primarily a **Kimi upstream protocol** issue, not a missing entry in the Claude/GPT/Gemini signature-compatibility matrix:

`400 thinking is enabled but reasoning_content is missing in assistant tool call message`

When thinking is enabled, Kimi expects assistant messages that contain `tool_calls` to also carry `reasoning_content`. A Claude-format history that only has `thinking` blocks (or empty reasoning) can fail on Kimi even if the request was translated through a Claude-shaped path.

### What we should **not** do
- Do **not** add third-party provider compatibility into `claude_executor`. That executor is intended for native Claude targets only (as noted in the thread).

### What we should do
[...]

PR #3946, opened 2026-06-21
fix(claude): bypass signature sanitizer for non-Claude models

sususu98:
## Summary

- Only run the Claude upstream signature sanitizer when the resolved upstream model belongs to the Claude signature family.
- Keep web_search domain cleanup active for Claude-format requests even when signature sanitization is bypassed.
- Add regression coverage for Claude models still stripping incompatible signature history and Kimi/Mimo-style non-Claude models preserving their Claude-format signed history.

## Validation

- `go test ./internal/runtime/executor -run 'TestShouldSanitizeClaudeMessagesForUpstream_OnlyClaudeFamily|TestSanitizeClaudeMessagesForClaudeUpstream_BypassesUnknownModelSignatureMatrix|TestClaudeExecutor_ExecuteBypassesSignatureSanitizerForUnknownModel|TestClaudeExecutor_ExecuteStripsForeignToolUseSignaturesBeforeUpstream|TestClaudeExecutor_ExecuteStripsMalformedEPrefixThinkingBeforeUpstream|TestClaudeExecutor_ExecuteStripsOpenAIEncryptedThinkingBeforeUpstream|TestClaudeExecutor_CountTokensStripsOpenAIEncryptedThinkingBeforeUpstream' -count=1`

## Notes

- `go test ./internal/runtime/executor -count=1` currently fails outside this PR scope in Codex image edit endpoint expectations (`/images/edits` vs `/images/edit`) and `TestXAIExecutorOmitsUnsupportedReasoningEffort`.

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -45,11 +45,18 @@ type ClaudeExecutor struct {
 // Previously "proxy_" was used but this is a detectable fingerprint difference.
 const claudeToolPrefix = ""
 
+func shouldSanitizeClaudeMessagesForUpstream(baseModel string) bool {
+	return sigcompat.SignatureProviderFromModelName(baseModel) == sigcompat.SignatureProviderClaude
+}
+
 func sanitizeClaudeMessagesForClaudeUpstreamWithDebug(ctx context.Context, body []byte, baseModel string) []byte {
-	sanitized, report := sigcompat.SanitizeClaudeMessagesForClaudeUpstream(body, baseModel)
-	logClaudeSignatureSanitizeReport(ctx, baseModel, report)
-	sanitized = sanitizeClaudeWebSearchDomains(sanitized)
-	return sanitized
+	sanitized := body
+	if shouldSanitizeClaudeMessagesForUpstream(baseModel) {
+		var report sigcompat.SignatureSanitizeReport
+		sanitized, report = sigcompat.SanitizeClaudeMessagesForClaudeUpstream(body, baseModel)
+		logClaudeSignatureSanitizeReport(ctx, baseModel, report)
+	}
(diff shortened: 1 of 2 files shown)


sususu98:
## Context for reviewers (ties to #3601 / #3817)

Sorry this wasn’t commented here first — the earlier write-up went on the issues because the discussion there was about **overall fix direction** (Kimi `reasoning_content`, Mimo empty response signatures, and “no third-party compat in `claude_executor`”).

This PR is the **concrete, mergeable slice** for one shared root cause on the **request path**:

> `ClaudeExecutor` always ran `SanitizeClaudeMessagesForClaudeUpstream`, so **non-`claude-*` models** (e.g. Kimi/Mimo ids) routed through the Claude executor still got **Claude-target** rules: incompatible `thinking` blocks are **dropped whole**, not bypassed.

### What this PR does
- Gate sanitization on `SignatureProviderFromModelName(baseModel) == claude`.
- Unknown / Kimi / Mimo model ids **bypass** the request-side signature matrix when they hit this executor path.
- `sanitizeClaudeWebSearchDomains` still runs (unrelated Anthropic payload fix).
- Tests: bypass for `kimi-k2.5` / `mimo-v2`, still strip bad thinking for real `claude-*`.

### What this PR does **not** close by itself
- **#3601**: Kimi `400 … reasoning_content is missing` → still needs **`kimi_executor`** normalization (or plugin), not more logic in `claude_executor` (per maintainer comment on #3601).
- **#3817**: empty `signature` on **response** thinking blocks → needs **response-side** strip/config for compat providers; this PR only stops **request** history from being over-sanitized.

If helpful, we can add `Fixes #…` only after the remaining Kimi/Mimo pieces land, or keep this PR scoped as “shared executor gate” and reference both issues in release notes.

Cross-posted rationale: #3601#issuecomment-4761585226, #3817#issuecomment-4761585224

PR #3974, opened 2026-06-23
fix(claude): preserve forwarded system blocks in OAuth cloaking

mxl:
## Summary

Sanitize only the first forwarded system block during OAuth cloaking and preserve all subsequent blocks unchanged. Extract `<available_skills>` from the first block before sanitizing, since clients like OpenCode typically send everything in a single block.

## Problem

When clients like OpenCode or Amp route requests through Claude OAuth cloaking, all forwarded `system[]` text blocks are collected into one string and passed through `sanitizeForwardedSystemPrompt()`. That function replaces the entire input with a short generic reminder:

```
Use the available tools when needed to help with software engineering tasks.
Keep responses concise and focused on the user's request.
Prefer acting on the user's task over describing product-specific workflows.
```

This drops important context like AGENTS.md guidance and skill descriptions, which can significantly affect tool use and task behavior.

## Fix

1. Collect forwarded `system[]` text blocks as before.
2. In OAuth mode, sanitize only `userSystemParts[0]` with the existing generic reminder.
3. Extract `<available_skills>` from `userSystemParts[0]` before replacing it (OpenCode sends everything in a single block).
4. Preserve `userSystemParts[1:]` unchanged.
5. Join all parts and prepend to the first user message using the existing `<system-reminder>` path.

This keeps the OAuth cloaking safety behavior for the primary third-party agent prompt while preserving additional project guidance and skill descriptions.

## Test Plan

- `TestCheckSystemInstructionsWithSigningMode_OAuthPassthroughPreserveSanitizesFirstPreservesRest` — verifies multi-block case: sanitizes first block, preserves later AGENTS.md and skill blocks
- `TestCheckSystemInstructionsWithSigningMode_OAuthSingleBlockPreservesAvailableSkills` — verifies single-block case: extracts `<available_skills>` before sanitizing
- `go test -run 'TestCheckSystemInstructions' ./internal/runtime/executor`

Refs #3245
Closes #3335

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -1894,7 +1894,7 @@ func checkSystemInstructionsWithSigningMode(payload []byte, strictMode bool, exp
 
 	// Collect user system instructions and prepend to first user message.
 	// In OAuth mode, sanitize only the first forwarded system block and preserve
-	// all subsequent blocks unchanged (PR #3335 approach).
+	// all subsequent blocks unchanged.
 	if !strictMode {
 		var userSystemParts []string
 		if system.IsArray() {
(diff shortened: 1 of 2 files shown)


mxl:
Closing as superseded. The target branch has since replaced the old Claude OAuth cloaking implementation with newer block-preservation logic, so this PR no longer applies cleanly to current upstream.

PR #3997, opened 2026-06-25
fix(claude-stream): synthesize terminal message_stop on clean upstream EOF

### LINKED ISSUE #3996: Claude streaming: no terminal event on clean upstream EOF → clients report "Connection closed mid-response"
## Summary

On a streaming request whose Claude upstream closes the SSE connection **cleanly** (TCP/EOF, no error) **without** sending a terminal `message_stop` event, `ClaudeExecutor.ExecuteStream` forwards/translates whatever it received and then simply returns — emitting **no terminal event**. Downstream clients (Claude Code, Cherry Studio / AI SDK, etc.) then surface:

```
API Error: Connection closed mid-response. The response above may be incomplete.
```

Every other streaming executor guards against this; the Claude executor is the only one that does not.

## Version

- CLIProxyAPI `v7.2.36` (`7712ffe`).
- Also present on `origin/main` — the only commit after `v7.2.36` is the pluginhost refactor (`b53d1e9`), which does not touch the streaming code paths.

## Root cause

`internal/runtime/executor/claude_executor.go`, `ExecuteStream`:

- **Claude→Claude passthrough branch** (`responseFormat == to`, ~L545–573): scans upstream SSE and forwards each line. On clean EOF (`scanner.Err() == nil`) it just `return`s — no terminal event is synthesized.
- **Claude→other translation branch** (~L577–613): same — on clean EOF it `return`s without feeding any terminal marker to the translator.

Contrast with `internal/runtime/executor/openai_compat_executor.go` (~L442–454), which on clean EOF explicitly feeds a synthetic `data: [DONE]` through the translator:

```go
} else {
    // In case the upstream close the stream without a terminal [DONE] marker.
    // Feed a synthetic done marker through the translator so pending
    // response.completed events are still emitted exactly once.
    chunks := sdktranslator.TranslateStream(ctx, to, responseFormat, req.Model, opts.OriginalRequest, translated, []byte("data: [DONE]"), &param)
    ...
}
```

The Claude executor has no equivalent guard, so a Claude upstream that ends a 200 SSE stream without `message_stop` (e.g. an intermediary/relay severing the connection, or an upstream that omits the terminal event) yields a silently-truncated stream rather than a clean terminator or an explicit error.

This is distinct from #3995 (request *hangs forever* with no timeout): here the stream *does* end, but without a terminal event.

## Minimal reproduction (no credentials needed)

A unit test driving `ExecuteStream` against an `httptest` upstream that returns a 200 SSE body without `message_stop`:

```go
func TestClaudeExecutor_ExecuteStreamEmitsMessageStopOnCleanEOFWithoutTerminal(t *testing.T) {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "text/event-stream")
		// Upstream closes cleanly WITHOUT a terminal message_stop event.
		_, _ = w.Write([]byte("event: message_start\ndata: {\"type\":\"message_start\",\"message\":{\"id\":\"msg_1\",\"model\":\"claude-3-5-sonnet-20241022\"}}\n\n"))
		_, _ = w.Write([]byte("event: content_block_delta\ndata: {\"type\":\"content_block_delta\",\"index\":0,\"delta\":{\"type\":\"text_delta\",\"text\":\"hi\"}}\n\n"))
	}))
	defer server.Close()

	executor := NewClaudeExecutor(&config.Config{})
	auth := &cliproxyauth.Auth{Attributes: map[string]string{"api_key": "key-123", "base_url": server.URL}}
	payload := []byte(`{"messages":[{"role":"user","content":[{"type":"text","text":"hi"}]}]}`)

	result, err := executor.ExecuteStream(context.Background(), auth, cliproxyexecutor.Request{
		Model:   "claude-3-5-sonnet-20241022",
		Payload: payload,
	}, cliproxyexecutor.Options{SourceFormat: sdktranslator.FromString("claude")})
	if err != nil {
		t.Fatalf("ExecuteStream() error = %v", err)
	}
	var buf []byte
	for chunk := range result.Chunks {
		if chunk.Err != nil {
			t.Fatalf("unexpected chunk error: %v", chunk.Err)
		}
		buf = append(buf, chunk.Payload...)
	}
	// FAILS on v7.2.36: no message_stop is ever emitted.
	if !strings.Contains(string(buf), `"type":"message_stop"`) {
		t.Fatalf("expected a synthetic message_stop terminal event on clean EOF, got:\n%s", string(buf))
	}
}
```

## Proposed fix

Mirror the `openai_compat_executor` guard in both branches of `ClaudeExecutor.ExecuteStream`, gated by a `sawMessageStop` flag so an upstream that already sent `message_stop` is never duplicated:

- **Passthrough branch:** on clean EOF without `message_stop`, emit a synthetic `event: message_stop\ndata: {"type":"message_stop"}\n\n` (leading blank line to guarantee SSE event separation if the upstream truncated mid-event).
- **Translation branch:** on clean EOF without `message_stop`, feed a synthetic Claude `data: {"type":"message_stop"}` line through `sdktranslator.TranslateStream` so the translator emits its proper terminal exactly once.

I have a working patch + passing tests against `v7.2.36` and am happy to open a PR if the approach looks right.

sweetcornna:
## What

Fixes #3996.

`ClaudeExecutor.ExecuteStream` did not emit any terminal event when the Claude
upstream closed an SSE stream **cleanly** (EOF, no error) **without** a
`message_stop`. Downstream clients then report:

```
API Error: Connection closed mid-response. The response above may be incomplete.
```

Every other streaming executor guards against a missing terminal marker on clean
EOF (e.g. `openai_compat_executor.go` feeds a synthetic `data: [DONE]`); the
Claude executor was the only one that did not.

## Change

In both branches of `ClaudeExecutor.ExecuteStream`, track whether a
`message_stop` was seen and, on clean EOF (`scanner.Err() == nil`) when it was
not, synthesize a terminal exactly once:

- **Claude→Claude passthrough:** emit `event: message_stop\ndata: {"type":"message_stop"}\n\n`
  (leading blank line guarantees SSE event separation even if the upstream
  truncated mid-event).
- **Claude→other translation:** feed a synthetic Claude `data: {"type":"message_stop"}`
  line through `sdktranslator.TranslateStream` so the translator finalizes the
  downstream stream once.

The `sawMessageStop` gate ensures an upstream that already sent `message_stop`
is never duplicated.

## Tests

Adds `internal/runtime/executor/claude_executor_streamterminal_test.go`:

- `TestClaudeExecutor_ExecuteStreamEmitsMessageStopOnCleanEOFWithoutTerminal` —
  fails before the change, passes after.
- `TestClaudeExecutor_ExecuteStreamDoesNotDuplicateMessageStop` — guards against
  double emission.

`go build ./cmd/server` and `go test ./internal/runtime/executor/ -run TestClaudeExecutor_ExecuteStream...` pass.

## Notes

- Distinct from #3995 (request *hangs* with no timeout); this PR addresses a
  clean EOF that *ends* the stream without a terminal event.
- Scope is limited to the Claude executor's clean-EOF path; the error path
  (`WriteTerminalError`) and other executors are unchanged.

### DIFF
--- a/internal/runtime/executor/claude_executor.go
+++ b/internal/runtime/executor/claude_executor.go
@@ -545,11 +545,15 @@ func (e *ClaudeExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.A
 		if responseFormat == to {
 			scanner := bufio.NewScanner(decodedBody)
 			scanner.Buffer(nil, 52_428_800) // 50MB
+			sawMessageStop := false
 			for scanner.Scan() {
 				line := scanner.Bytes()
 				helps.AppendAPIResponseChunk(ctx, e.cfg, line)
 				if detail, ok := helps.ParseClaudeStreamUsage(line); ok {
 					reporter.Publish(ctx, detail)
+				}
+				if bytes.Contains(line, []byte("message_stop")) {
+					sawMessageStop = true
 				}
 				line = restoreClaudeOAuthToolNamesFromStreamLine(line, claudeToolPrefix, auth.ToolPrefixDisabled(), oauthToolNamesReverseMap)
 				// Forward the line as-is to preserve SSE format
@@ -569,6 +573,19 @@ func (e *ClaudeExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.A
 				case out <- cliproxyexecutor.StreamChunk{Err: errScan}:
 				case <-ctx.Done():
 				}
+			} else if !sawMessageStop {
+				// The upstream closed the SSE stream cleanly without emitting a
+				// terminal message_stop event. Synthesize one so downstream
+				// clients receive a proper stream terminator instead of
+				// reporting "Connection closed mid-response. The response above
+				// may be incomplete." This mirrors the synthetic terminal-marker
+				// guard in openai_compat_executor.go. The leading blank line
+				// guarantees SSE event separation even if the upstream truncated
+				// mid-event.
+				select {
+				case out <- cliproxyexecutor.StreamChunk{Payload: []byte("\nevent: message_stop\ndata: {\"type\":\"message_stop\"}\n\n")}:
+				case <-ctx.Done():
+				}
 			}
 			return
 		}
@@ -577,11 +594,15 @@ func (e *ClaudeExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.A
 		scanner := bufio.NewScanner(decodedBody)
 		scanner.Buffer(nil, 52_428_800) // 50MB
 		var param any
+		sawMessageStop := false
 		for scanner.Scan() {
 			line := scanner.Bytes()
 			helps.AppendAPIResponseChunk(ctx, e.cfg, line)
 			if detail, ok := helps.ParseClaudeStreamUsage(line); ok {
 				reporter.Publish(ctx, detail)
+			}
+			if bytes.Contains(line, []byte("message_stop")) {
+				sawMessageStop = true
 			}
 			line = restoreClaudeOAuthToolNamesFromStreamLine(line, claudeToolPrefix, auth.ToolPrefixDisabled(), oauthToolNamesReverseMap)
 			chunks := sdktranslator.TranslateStream(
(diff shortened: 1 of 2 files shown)


luispater:
Thanks for the patch and tests.

We are closing this PR because the intended behavior is not to synthesize a successful `message_stop` when the upstream Claude SSE stream reaches EOF without a proper terminal event. That EOF indicates the response did not complete according to the Claude streaming protocol, and emitting `message_stop` downstream would make an incomplete/truncated response look like a clean successful completion.

In this case, preserving the downstream "connection closed mid-response" / incomplete-stream signal is safer than masking the upstream protocol violation.

Issue #3999, opened 2026-06-25
Antigravity Gemini: opaque "User location is not supported" 400 from datacenter egress — add actionable hint (region/proxy-url)


sweetcornna:
## Summary

Antigravity **Gemini** requests fail with HTTP 400:

```json
{"error":{"code":400,"message":"User location is not supported for the API use.","status":"FAILED_PRECONDITION"}}
```

while Antigravity **Claude** and **Codex** work with the same credential. The error is **egress-region** based, not request-shape based, and the message is opaque enough that users spend a long time chasing it as a bug. This is a request to make the failure actionable (and ideally document it). It is **not** a request-shape bug in CLIProxyAPI.

## Environment

- CLIProxyAPI `v7.2.36`; also current `main`.
- Provider: Antigravity, models `gemini-3-flash-agent`, `gemini-3-flash`, `gemini-pro-agent`, `gemini-3.1-flash-lite`.
- Host: US datacenter/hosting IP (AS for a hosting provider).

## Evidence (decisive isolation)

Raw requests to `daily-cloudcode-pa.googleapis.com/v1internal:generateContent` with a freshly-refreshed token from the same account:

| Variation | Result |
|---|---|
| body **with** injected `project` (current CLIProxyAPI behavior) | **400** `User location is not supported` |
| body **without** `project` | **400** (same) |
| egress via **native datacenter IP** | **400** |
| egress via a **different-region proxy** (residential/edge IP) | **200 OK** (both `generateContent` and `streamGenerateContent`) |

So:
- The injected `project` is **not** the cause (both 400) — i.e. a "strip project for Gemini" change does **not** fix it (verified). `geminiToAntigravity` injecting `project` is harmless here.
- The discriminator is purely the **egress region/IP**. Datacenter/hosting IPs are rejected even when they geolocate to an otherwise-supported country; the official Antigravity desktop client works because it runs from a residential IP.
- Antigravity Claude/Codex are unaffected because they don't hit the Gemini regional gate.

## Resolution that works today

CLIProxyAPI already supports per-auth `proxy-url`. Routing the affected Antigravity credential through an egress in a supported region (e.g. a SOCKS proxy whose outbound is a WARP/residential interface) makes Gemini return 200 through CLIProxyAPI unchanged. So the platform can already work around this — users just have no signal that egress is the lever.

## Request

[...]

PR #4000, opened 2026-06-25
feat(antigravity): actionable hint for region-blocked Gemini 400 (User location is not supported)

sweetcornna:
## What

Addresses #3999 (optional enhancement).

Antigravity **Gemini** requests from an unsupported egress region fail with an
opaque 400:

```
User location is not supported for the API use.  (FAILED_PRECONDITION)
```

This is keyed on the request's **egress IP/region**, not the request body, and
commonly affects datacenter/hosting IPs even when they geolocate to a supported
country. Antigravity Claude/Codex are unaffected. Users currently get no signal
that egress is the lever.

Decisive isolation (same account, fresh token, `daily-cloudcode-pa` directly):

| Variation | Result |
|---|---|
| body with injected `project` | 400 |
| body without `project` | 400 |
| native datacenter egress | 400 |
| supported-region proxy egress | 200 OK |

So this is **not** a request-shape bug (stripping `project` does not help — verified),
and it is already resolvable today via the existing per-auth `proxy-url`.

## Change

Make the failure actionable without changing routing behavior:

- Add `antigravityRegionBlockedHint(statusCode, body)` — returns a hint only for
  a 400 whose body contains `User location is not supported`, else `""`.
- In `newAntigravityStatusErr`, when the hint applies, `log.Warnf` it and append
  it to the returned error message. The **original upstream message is preserved**
  (prefix unchanged); unrelated errors pass through verbatim.

The hint points users at routing the affected credential through a per-auth
`proxy-url` in a supported region.

## Tests

`internal/runtime/executor/antigravity_executor_locationhint_test.go`:

- `TestAntigravityRegionBlockedHint` — only the region 400 yields a hint (non-400
  and unrelated 400 yield none).
- `TestNewAntigravityStatusErr_AppendsRegionHint` — hint appended, original
  message + status preserved.
- `TestNewAntigravityStatusErr_UnrelatedErrorUnchanged` — unrelated 400 body
  passes through verbatim.

`go build ./cmd/server` and the targeted tests pass.

## Notes / scope

- No routing, request-shape, or retry behavior changes — this only annotates an
  already-terminal error.
- A docs note under Antigravity setup (Gemini region-gating + `proxy-url`) would
  complement this; happy to add if preferred.

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -608,7 +608,31 @@ func newAntigravityStatusErr(statusCode int, body []byte) statusErr {
 			err.retryAfter = retryAfter
 		}
 	}
+	if hint := antigravityRegionBlockedHint(statusCode, body); hint != "" {
+		log.Warnf("antigravity executor: %s", hint)
+		err.msg = strings.TrimRight(err.msg, "\n") + "\n\n" + hint
+	}
 	return err
+}
+
+// antigravityRegionBlockedHint returns an actionable hint when the Antigravity
+// (Gemini Code Assist) upstream rejects a request with the opaque
+// "User location is not supported for the API use." 400. That error is keyed on
+// the request's egress region, not on the request body, and frequently affects
+// datacenter/hosting IPs even when they geolocate to an otherwise supported
+// country. Routing the affected credential through an egress in a supported
+// region (per-auth proxy-url) resolves it. Returns "" for any other response so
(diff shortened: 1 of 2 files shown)


luispater:
Closing this PR for the same reason as #3999.

CLIProxyAPI should only forward and translate information returned by the upstream provider. It should not append additional English hints or return other non-upstream information in provider error responses.

This PR adds an extra hint to the Antigravity upstream error message, so we will not take this change.

PR #4002, opened 2026-06-25
fix(auth): rewrite API key alias response models

### LINKED ISSUE #3965: Feature request: rewrite response `model` field to match client-requested alias
## Problem

`oauth-model-alias` (and the `models` alias under `claude-api-key`) renames model IDs for **model listing** and **request routing**, but the **`model` field in the response payload is left as the upstream/backend model name**.

Downstream clients that key off the *response* `model` field — e.g. Claude Code's context-window indicator and token accounting — then fail to recognize the alias, because what they requested and what comes back disagree.

## Reproduction

Config (route a Claude-style alias to a Codex backend):

```yaml
oauth-model-alias:
  codex:
    - name: "gpt-5.5"
      alias: "claude-sonnet-4-5"
```

1. Client sends `{"model": "claude-sonnet-4-5", ...}`
2. Proxy correctly routes the request to the `gpt-5.5` Codex backend ✅
3. Response comes back with `{"model": "gpt-5.5", ...}` ❌

Verified via `curl /v1/messages` — the response `model` is the backend name (`gpt-5.5`), not the requested alias (`claude-sonnet-4-5`).

## Impact

Aliasing is only *partially* transparent. The request side is rewritten, but the response side is not, so any client that uses the response `model` for feature gating or context accounting breaks. Concretely: Claude Code shows no context-window UI / mis-accounts tokens for an aliased model, because it sees an unknown backend model name in the response.

## Request

An option to make the response `model` field echo the **client-requested alias** instead of the backend model name. For example a per-alias flag:

```yaml
oauth-model-alias:
  codex:
    - name: "gpt-5.5"
      alias: "claude-sonnet-4-5"
      rewrite-response-model: true   # response returns "claude-sonnet-4-5"
```

## Why it matters

Clients increasingly use the response `model` field for context-window sizing and capability gating, not just display. For aliasing to be fully transparent, the response `model` needs to reflect the alias the client actually requested.

sususu98:
## Summary

Extends response `model` rewrite (`force-mapping`) to API key and compatibility model entries, so clients see the alias they requested instead of the upstream backend name.

- Add `force-mapping` on `claude-api-key`, `codex-api-key`, `gemini-api-key`, `vertex-api-key`, and `openai-compatibility` model entries (config structs + sanitization).
- Route API key execution through `resolveAPIKeyModelAliasWithResult` and reuse existing non-stream / stream response rewriters when `ForceMapping` is set.
- Cover API key + openai-compat alias pool paths with unit tests.

## Related issue

Closes #3965

OAuth/global aliases already support `force-mapping` on `oauth-model-alias`; this PR completes the same behavior for per-credential `models` aliases on API key providers.

## Config

```yaml
claude-api-key:
  - api-key: "..."
    models:
      - name: "glm-5.2"
        alias: "claude-sonnet-latest"
        force-mapping: true
```

## Verification

- `go test ./sdk/cliproxy/auth -count=1`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -231,6 +231,7 @@ nonstream-keepalive-interval: 0
 #     models:
 #       - name: "gpt-5-codex"   # upstream model name
 #         alias: "codex-latest" # client alias mapped to the upstream model
+#         force-mapping: true    # optional: rewrite response model fields back to the alias
 #     excluded-models:
 #       - "gpt-5.1"         # exclude specific models (exact match)
 #       - "gpt-5-*"         # wildcard matching prefix (e.g. gpt-5-medium, gpt-5-codex)
@@ -251,6 +252,7 @@ nonstream-keepalive-interval: 0
 #     models:
 #       - name: "claude-3-5-sonnet-20241022" # upstream model name
 #         alias: "claude-sonnet-latest"      # client alias mapped to the upstream model
+#         force-mapping: true                 # optional: rewrite response model fields back to the alias
 #     excluded-models:
 #       - "claude-opus-4-5-20251101" # exclude specific models (exact match)
 #       - "claude-3-*"               # wildcard matching prefix (e.g. claude-3-7-sonnet-20250219)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -483,10 +483,14 @@ type ClaudeModel struct {
 
 	// Alias is the client-facing model name that maps to Name.
 	Alias string `yaml:"alias" json:"alias"`
+
+	// ForceMapping rewrites upstream response model fields back to Alias.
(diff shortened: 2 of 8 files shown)


sususu98:
**Design note (force-mapping + auth `prefix`)**

`force-mapping: true` rewrites response `model` fields to the **configured `alias` in `models[]`** (sanitized bare alias), not to the raw `routeModel` and **not** with the per-credential `prefix/` re-applied.

Example: `prefix: "team"`, alias `claude-sonnet-latest`, client requests `team/claude-sonnet-latest` → upstream still uses `name`; response `model` is **`claude-sonnet-latest`**, not `team/claude-sonnet-latest`.

This matches the intent for #3965: clients should see a stable client-facing alias for capability/accounting, independent of routing prefixes. Codex P2 on preserving prefixes is **won't fix** (👎).

PR #4012, opened 2026-06-26
fix(upstream): force HTTP/1.1 on fallback transport to eliminate mid-stream RST_STREAM(INTERNAL_ERROR) leak

### LINKED ISSUE #4011: Intermittent "stream error: INTERNAL_ERROR; received from peer" on non-fingerprinted upstreams (HTTP/2 RST_STREAM leaks mid-stream)
### Summary

When a provider upstream is **not** one of the uTLS-fingerprinted hosts (`api.anthropic.com`, `chatgpt.com`) — e.g. a reseller/aggregator gateway, a self-hosted relay, or any OpenAI/Anthropic-compatible endpoint — requests intermittently fail mid-stream and the client receives a raw Go error:

```
API Error: stream error: stream ID 1; INTERNAL_ERROR; received from peer
```

This happens occasionally (only when the upstream resets a stream mid-response), so it presents as flaky "stream cut off" behavior.

### Root cause

That string is the `golang.org/x/net/http2` `StreamError` with `errFromPeer` ("received from peer") — i.e. **a Go HTTP/2 client received a `RST_STREAM(INTERNAL_ERROR)` frame from its server**. The only Go HTTP/2 client in the request path is CLIProxyAPI's own connection to the upstream.

The uTLS Chrome-fingerprint HTTP/2 round-tripper is gated to a hardcoded allow-list (`internal/runtime/executor/helps/utls_client.go`):

```go
var utlsProtectedHosts = map[string]struct{}{
    "api.anthropic.com": {},
    "chatgpt.com":       {},
}
```

Every other host falls through to `fallback` = `http.DefaultTransport`, which has `ForceAttemptHTTP2 = true` and negotiates **HTTP/2** whenever the upstream's ALPN offers `h2` (most gateways behind nginx/Cloudflare do).

When such an upstream resets a stream **mid-response**:
- The conductor only retries resets that occur **before the first byte** (`empty_stream … Retryable: true`). A mid-stream reset cannot be retried (bytes already forwarded to the client).
- The abrupt `RST_STREAM` is surfaced through the executor's `scanner.Err()` path and forwarded verbatim, so the client sees the raw `stream error … INTERNAL_ERROR; received from peer`.

### Impact

Any deployment whose claude/codex/gemini/openai-compat upstream points at a non-`anthropic`/`chatgpt` HTTPS endpoint that negotiates HTTP/2 is exposed: a flaky upstream channel leaks raw http2 framing errors to end users instead of failing gracefully.

### Proposed fix

Force **HTTP/1.1** on the fallback transport (the non-fingerprinted path), mirroring the existing `cloneTransportWithHTTP11` used by the antigravity executor. Over HTTP/1.1 the `RST_STREAM` error class cannot occur; a mid-response upstream drop becomes a plain connection close (clean EOF), which composes cleanly with terminal-marker synthesis (e.g. #3997) for a graceful `message_stop`/`[DONE]` instead of a hard error. The uTLS HTTP/2 path for Anthropic/ChatGPT is intentionally left untouched to preserve its Chrome fingerprint.

PR with tests: follows.

### Environment
- CLIProxyAPI: `dev`
- Upstream: third-party OpenAI/Anthropic-compatible aggregator (New API style) over HTTPS, ALPN `h2`
- Downstream: nginx (TLS termination) → cliproxy over HTTP/1.1; client is Claude Code / OpenAI JS SDK

sweetcornna:
Fixes #4011

### Problem

For upstreams that are **not** uTLS-fingerprinted (`api.anthropic.com` / `chatgpt.com`) — resellers, aggregators, self-hosted relays, any OpenAI/Anthropic-compatible HTTPS endpoint — a stream reset from the upstream surfaces to the client as a raw Go error:

```
stream error: stream ID 1; INTERNAL_ERROR; received from peer
```

This is `golang.org/x/net/http2`'s `errFromPeer`: CLIProxyAPI's HTTP/2 client received a `RST_STREAM(INTERNAL_ERROR)`. The uTLS HTTP/2 round-tripper is gated to a hardcoded allow-list, so every other host falls through to `http.DefaultTransport` (`ForceAttemptHTTP2 = true`) and negotiates HTTP/2 when the upstream's ALPN offers `h2`. A **mid-response** reset can't be retried (pre-first-byte resets already are, via `empty_stream`), so the raw framing error leaks out.

### Fix

Pin the **fallback** transport in `NewUtlsHTTPClient` to HTTP/1.1, mirroring the antigravity executor's `cloneTransportWithHTTP11`:

- `ForceAttemptHTTP2 = false`
- `TLSNextProto` wiped (no implicit h2 upgrade)
- `TLSClientConfig.NextProtos = ["http/1.1"]`

Over HTTP/1.1 the `RST_STREAM(INTERNAL_ERROR)` class cannot occur. A mid-response upstream drop becomes a plain connection close (clean EOF), so the failure mode degrades from a hard error into a normal end-of-stream — and composes with terminal-marker synthesis (#3997) for a graceful `message_stop`/`[DONE]`.

### Scope / safety

- Only the **fallback** path changes. The uTLS HTTP/2 path for Anthropic/ChatGPT is untouched, preserving its Chrome fingerprint.
- The base transport is **cloned, never mutated**, so the shared `http.DefaultTransport` global is left intact.
- Non-`*http.Transport` round-trippers (an injected context RoundTripper, proxy transports) pass through unchanged.
- Proxy/dial settings on the base transport are preserved; only ALPN/protocol negotiation is constrained.

### Tests

`internal/runtime/executor/helps/utls_client_test.go` (3 cases, all green):

1. `TestNewUtlsHTTPClientFallbackForcesHTTP11` — the fallback transport has `ForceAttemptHTTP2=false`, `NextProtos=["http/1.1"]`, empty `TLSNextProto`, and the global `http.DefaultTransport` is not mutated.
2. `TestForceHTTP11TransportNegotiatesHTTP11AgainstH2Server` — **real-behavior**: against an `httptest` server with HTTP/2 enabled, a control request negotiates HTTP/2 (proving the server offers h2) while the forced transport negotiates HTTP/1.1.
3. Existing `TestNewUtlsHTTPClientUsesContextRoundTripperForProtectedHost` still passes (protected-host path unchanged).

```
go test ./internal/runtime/executor/helps/   # ok
go build ./...                                 # ok
```

### Verified in production

Built and deployed against a reseller upstream that negotiates ALPN `h2`. Packet capture of a live streaming request now shows the TLS ClientHello to the upstream offering **only `http/1.1`** (no `h2`); the `INTERNAL_ERROR; received from peer` errors are gone.

### DIFF
--- a/internal/runtime/executor/helps/utls_client.go
+++ b/internal/runtime/executor/helps/utls_client.go
@@ -2,6 +2,7 @@ package helps
 
 import (
 	"context"
+	cryptotls "crypto/tls"
 	"net"
 	"net/http"
 	"strings"
@@ -152,6 +153,31 @@ func (f *fallbackRoundTripper) RoundTrip(req *http.Request) (*http.Response, err
 	return f.fallback.RoundTrip(req)
 }
 
+// forceHTTP11Transport returns a clone of base that negotiates HTTP/1.1 only.
+// Round-trippers that are not *http.Transport (e.g. an injected context
+// RoundTripper or a uTLS round-tripper) are returned unchanged. Proxy and dial
+// settings on base are preserved; only ALPN/protocol negotiation is pinned.
+// The base is cloned, never mutated, so the shared http.DefaultTransport global
+// is left intact.
+func forceHTTP11Transport(base http.RoundTripper) http.RoundTripper {
+	tr, ok := base.(*http.Transport)
+	if !ok || tr == nil {
+		return base
+	}
+	clone := tr.Clone()
+	clone.ForceAttemptHTTP2 = false
+	// Wipe TLSNextProto to prevent an implicit HTTP/2 upgrade.
+	clone.TLSNextProto = make(map[string]func(authority string, c *cryptotls.Conn) http.RoundTripper)
+	if clone.TLSClientConfig == nil {
+		clone.TLSClientConfig = &cryptotls.Config{}
+	} else {
+		clone.TLSClientConfig = clone.TLSClientConfig.Clone()
+	}
+	// Actively advertise only HTTP/1.1 in the ALPN handshake.
+	clone.TLSClientConfig.NextProtos = []string{"http/1.1"}
+	return clone
+}
+
 // NewUtlsHTTPClient creates an HTTP client using utls Chrome TLS fingerprint.
 // Use this for provider requests that need a Chrome-like TLS fingerprint.
 // Falls back to standard transport for non-HTTPS requests.
@@ -179,6 +205,16 @@ func NewUtlsHTTPClient(ctx context.Context, cfg *config.Config, auth *cliproxyau
 		utlsRT = ctxRoundTripper
 		standardTransport = ctxRoundTripper
 	}
+
+	// Force HTTP/1.1 on the fallback path (every non-fingerprinted host, e.g.
+	// reseller/aggregator upstreams). HTTP/2 lets a flaky upstream reset a
+	// stream mid-response with RST_STREAM(INTERNAL_ERROR), which leaks to the
+	// client as a raw "stream error ...; received from peer". Over HTTP/1.1 a
+	// mid-response drop is instead a clean EOF that ClaudeExecutor.ExecuteStream
+	// already terminates gracefully with a synthetic message_stop. The uTLS
+	// (Anthropic/ChatGPT) path is intentionally left on HTTP/2 to preserve its
+	// Chrome fingerprint.
+	standardTransport = forceHTTP11Transport(standardTransport)
 
 	client := &http.Client{
 		Transport: &fallbackRoundTripper{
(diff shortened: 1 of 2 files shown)


luispater:
Thanks for the contribution and the tests.

We are not going to merge this change as the default behavior. The observed `stream error: stream ID ... INTERNAL_ERROR; received from peer` is caused by the upstream/network/proxy path resetting an HTTP/2 stream mid-response. Forcing HTTP/1.1 globally for all non-fingerprinted upstreams would avoid this specific HTTP/2 error class, but it would also disable HTTP/2 multiplexing and may reduce throughput and connection efficiency for normal upstream deployments.

This should be handled by fixing/tuning the unstable upstream or proxy path, or by disabling HTTP/2 in that specific deployment environment if needed, rather than changing CLIProxyAPI's default fallback transport for every custom upstream.

Closing this PR as not planned.

PR #4014, opened 2026-06-26
fix: avoid management IP ban on missing key

### LINKED ISSUE #4013: Management IP ban can be triggered by missing-key requests behind reverse proxies
## Summary

Management authentication currently counts requests with no management key as failed attempts. After five missing-key requests from the same client IP, the IP is banned for 30 minutes.

This is easy to trigger accidentally behind reverse proxies or health/check scripts, where unauthenticated probes or UI/API preflight-style calls may hit `/v0/management/*` without an `Authorization` or `X-Management-Key` header.

## Observed behavior

1. Send five requests to `/v0/management/config` without a management key.
2. Send a request with the correct management key.
3. The correct key is rejected with `403 IP banned due to too many failed attempts`.

## Expected behavior

- Missing management keys should return `401 missing management key` but should not increment the IP ban counter.
- Invalid non-empty keys should still increment the ban counter.
- A correct management key should be able to clear an active ban for that IP.

## Why this matters

In reverse-proxy deployments, multiple management requests can collapse to the same apparent client IP. Counting missing credentials as ban-worthy can lock out legitimate administrators even when no wrong password/key was attempted.

## Proposed fix

Only count invalid non-empty management keys toward the IP ban counter, and validate a provided correct key before enforcing an active ban.

sweetcornna:
## Summary

Fixes #4013.

This changes management authentication so requests with no management key return `401 missing management key` without incrementing the IP ban counter. Invalid non-empty keys still count toward the ban, preserving brute-force protection.

It also allows a correct management key to clear an active ban for that client IP, so a legitimate administrator is not locked out for the full ban window after accidental missing-key probes.

## Why

Behind reverse proxies, unauthenticated probes or UI/API calls can collapse to the same apparent client IP. Counting missing credentials as ban-worthy can lock out management access even when no wrong key was attempted.

## Tests

```bash
go test ./internal/api/handlers/management -run 'TestAuthenticateManagementKey_|TestMiddlewareSetsSupportPluginHeader' -count=1
```

Covered paths include missing-key no-ban behavior, invalid-key ban behavior, valid-key ban recovery via `envSecret`, and valid-key ban recovery via bcrypt-backed `remote-management.secret-key`.

## Deployment Verification

Verified the patch in a reverse-proxy/relay deployment:

- Five requests with no management key returned `401`; a subsequent request with the valid key returned `200`.
- Five requests with wrong management keys returned `401`; the next wrong-key request returned `403`; a subsequent request with the valid key returned `200` and cleared the active ban.
- Confirmed the management page/API can remain available through the relay/reverse-proxy path.

No private hostnames, credentials, or deployment-specific secrets are required to reproduce the behavior.

## Notes

A broader local run including `./internal/api/handlers/management ./internal/api` still hits pre-existing plugin-store test failures unrelated to this change: versioned `.dylib` files are written while some tests expect unversioned names.

### DIFF
--- a/internal/api/handlers/management/handler.go
+++ b/internal/api/handlers/management/handler.go
@@ -320,17 +320,20 @@ func (h *Handler) AuthenticateManagementKey(clientIP string, localClient bool, p
 	envSecret := h.envSecret
 
 	now := time.Now()
+	banned := false
+	banMessage := ""
 	h.attemptsMu.Lock()
 	ai := h.failedAttempts[clientIP]
 	if ai != nil && !ai.blockedUntil.IsZero() {
 		if now.Before(ai.blockedUntil) {
 			remaining := ai.blockedUntil.Sub(now).Round(time.Second)
-			h.attemptsMu.Unlock()
-			return false, http.StatusForbidden, fmt.Sprintf("IP banned due to too many failed attempts. Try again in %s", remaining)
+			banned = true
+			banMessage = fmt.Sprintf("IP banned due to too many failed attempts. Try again in %s", remaining)
+		} else {
+			// Ban expired, reset state
+			ai.blockedUntil = time.Time{}
+			ai.count = 0
 		}
-		// Ban expired, reset state
-		ai.blockedUntil = time.Time{}
-		ai.count = 0
 	}
 	h.attemptsMu.Unlock()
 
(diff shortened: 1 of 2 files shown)


luispater:
Thank you for the report and the proposed change.

After reviewing this from a security perspective, we are going to keep the current behavior. Treating an empty management key as an authentication failure is acceptable for the Management API. Clients should provide the correct management key within the allowed number of attempts.

Allowing a correct key to immediately clear an active IP ban would weaken the purpose of the ban. For example, if an attacker brute-forces a large number of incorrect keys and then eventually submits the correct key, automatically unlocking that IP at that point would remove much of the protection that the ban is intended to provide.

For that reason, after the configured failure threshold is exceeded, all management key attempts from that IP should remain blocked for the ban duration, regardless of whether a later attempt contains the correct key. This is the intended behavior to help mitigate brute-force attempts against the management key.

Closing this as not planned.

PR #4027, opened 2026-06-27
feat(claude): add cloak.preserve-system-prompt option

pedxyuyuko:
By default, cloaking OAuth (Claude Code login) requests reduces the caller's forwarded system prompt to a neutral reminder so traffic looks like native Claude Code. This drops user/client system prompts entirely on OAuth credentials, even outside strict mode.

Add an opt-in cloak option that, when strict mode is off, keeps the original system prompt and moves it into the first user message instead of replacing it. Configurable per claude-api-key entry via cloak.preserve-system-prompt, or per OAuth credential via the cloak_preserve_system_prompt auth attribute. Defaults to false, so existing behavior is unchanged.

Keeping the original prompt makes requests more recognizable as third-party traffic, which may affect Anthropic plan-vs-extra-usage metering; this is documented in config.example.yaml.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -265,14 +265,19 @@ nonstream-keepalive-interval: 0
 #                                    # "never": never apply cloaking
 #       # This "cloak" block applies to this claude-api-key entry only. For Claude OAuth
 #       # credentials, set the same options in the auth/token JSON file via "cloak_mode" /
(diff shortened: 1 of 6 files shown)


pedxyuyuko:
sry I miss clicked

Issue #4048, opened 2026-06-29
Codex Responses WS: context loss on upstream SSE disconnect (408) and WS→HTTP transcript handling


sususu98:
## Summary

Codex CLI connects to CPA with `supports_websockets=true` on `/v1/responses` (downstream WebSocket). When CPA forwards to **HTTP/SSE upstream** (Antigravity, xAI with `websockets: false`, or failover from WS-capable creds), a mid-stream failure—especially **HTTP 408** `stream closed before response.completed`—can leave the **downstream WS session** in a bad incremental state. Users see **lost conversation context** on retry or HTTP fallback.

End-to-end WS passthrough (Codex ↔ CPA ↔ upstream WS) is generally fine. The bug is in the **CPA-mediated WS → SSE** path and its interaction with **pinned auth** and **transcript merge** state on a single downstream connection.

## Impact

- Codex + CPA with mixed credential pools (`websockets: true` and `websockets: false` on the same model).
- Models routed to Antigravity SSE (`gemini-3.5-flash`, etc.) while Codex uses WS.
- xAI / grok with `websockets: false` (downstream WS, upstream HTTP) after upstream errors or Codex reconnect.

## Reproduction (local)

**Build:** current `sdk/api/handlers/openai/openai_responses_websocket.go` at `5a2c3090` (before/after fix comparison).

1. Start CPA with `request-log: true`, `WRITABLE_PATH=$PWD/auths`, API key from config.
2. Auth pool example: xAI `websockets: false` + optional Codex WS creds (`/tmp/auths-ws-mix`).
3. Codex profile: `supports_websockets = true`, `wire_api = "responses"`, `base_url` → local CPA.
4. **Happy path (WS→HTTP, no 408):** `grok-composer-2.5-fast`, serial turns (no parallel `resume`):
   - Session `019f125a-e476-7fd3-81ec-11da93b2c2f2` @ port 8362
   - Logs: `Downstream Transport: websocket`, `Upstream Transport: http`
   - `input` item count grows **3 → 6 → 9**; memory question answered correctly.
   - Logs: `v1-responses-2026-06-29T154939-6b31b651.log`, `155032-fe4178cc.log`, `155040-299a9370.log`
5. **Failure path (WS→SSE + 408):** `gemini-3.5-flash` @ port 8363 (full auths):
   - Session `019f125c-32a2-7250-b31f-5372876bdf40`
[...]

PR #4051, opened 2026-06-29
fix(auth): repair force-mapped Responses SSE framing for WS forwarder

sususu98:
## Summary

Repairs force-mapped OpenAI Responses SSE framing in `StreamRewriter` / conductor so `/v1/responses` websocket clients receive `response.completed` instead of CPA-synthesized **408** when upstream Antigravity or Codex returns HTTP 200.

## Root cause

- Antigravity/Gemini: glued `...}event:...` without `\n\n` between frames
- Codex: glued `...}data:...` and `event:` + `data:` lines without newlines when rewriter buffers
- Pending buffers were not flushed safely → `forwardResponsesWebsocket` never saw `response.completed`

## Fix

- `safeReplaceGlued` for `}event:` and `}data:` (only when preceding `data:` JSON is valid)
- Safer `Finish()` flush and line-wise fallback rewrites
- Newline when appending pending `event:` before next `data:` chunk

## Tests

- `response_model_rewriter_antigravity_sim_test.go`
- `codex_forcemap_ws_forward_test.go`
- `response_model_rewriter_test.go` (Codex glue + normalize)

## Live

Local CPA: `gemini-3.5-flash` WS 3-turn memory test — all `response.completed`, no 408.

Separate from #4048 (WS transcript rollback / mixed-pool incremental merge).

### DIFF
--- a/sdk/cliproxy/auth/codex_forcemap_ws_forward_test.go
+++ b/sdk/cliproxy/auth/codex_forcemap_ws_forward_test.go
@@ -1,0 +1,80 @@
+package auth
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/tidwall/gjson"
+)
+
+func parseWSDataEventTypesFromForwardedChunks(forwarded [][]byte) []string {
+	var types []string
+	for _, ch := range forwarded {
+		ch = normalizeGluedSSEEvents(ch)
+		for _, ln := range bytes.Split(ch, []byte("\n")) {
+			ln = bytes.TrimSpace(ln)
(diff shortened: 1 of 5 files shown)


sususu98:
Thanks for the review notes. Gemini High/Medium are addressed in 150e7f0 (safeReplaceGlued + Finish buffer). For the remaining Codex P2 items we did not reproduce 408 or corrupted streams in live tests (gemini force-map WS→HTTP, xAI force-map WS→HTTP/SSE, Codex force-map on 8370/8371) and auth package tests pass, so we are not planning further changes in this PR. WS transcript / passthrough context loss is tracked separately (#4048).

PR #3651, opened 2026-05-31
feat: add Cursor Composer provider (SDK bridge + management API)

artile:
## Summary

- Add `cursor-composer-api-key` provider (`composer-2.5`, `composer-2.5-fast`, `composer-latest`).
- Optional **SDK bridge** via `sdk-bridge-url` for `crsr_` keys; connect-proto fallback when unset.
- Management API: `/v0/management/cursor-composer-api-key` (alias `cursor-api-key`).
- Merge default Composer model IDs on auth registration (avoids stale `composer-2.5-fast` after upgrades).
- `internal/cursorcomposer/` for endpoint/transport resolution; `config.example.yaml` docs.

Replaces closed #3630 (same branch; `AGENTS.md` intentionally unchanged vs `dev`).

## Originator reference

https://github.com/standardagents/composer-api

## Test plan

- [x] `go test ./internal/runtime/executor/... ./sdk/cliproxy/... ./internal/api/handlers/management/...`
- [x] `go build ./cmd/server`
- [ ] Chat completions for `composer-2.5` / `composer-2.5-fast` via SDK bridge
- [ ] Management API CRUD for `cursor-composer-api-key`

Made with [Cursor](https://cursor.com)

PR #4055, opened 2026-06-29
fix: unblock Cursor Composer provider — address review findings + resolve dev conflict

bradhallett:
Follow-up to #3651 (Cursor Composer provider) — addresses bot review findings, resolves merge conflict with `dev`, and adds regression tests.

## What changed

**Conflict resolution (rebased onto current `dev`)**
- `sdk/cliproxy/service.go`: keeps upstream `NewXAIAutoExecutor` for `xai`, adds `cursor-composer` case

**Gemini Code Assist findings addressed**
1. **Checksum timestamp bug** — `cursorChecksum` was dividing `Unix()` by 1,000,000, collapsing the timestamp to zero. Switched to `UnixMilli()` with a deterministic `cursorChecksumAt()` helper. Added `TestCursorChecksumAtUsesTimestamp`.
2. **Token cache stampede** — `exchangeAPIKey` now uses a mutex with double-checked locking so concurrent calls share a single upstream exchange. Added `TestCursorComposerExchangeAPIKeyCachesConcurrentCalls`.
3. **Nil-pointer note on management handlers** — `management.NewHandler` always initializes `cfg`; no nil path exists in the current codebase. Did not add defensive guards that can't be triggered.

**Original PR #3651 review blockers**
- **EndStream termination**: `continue` → `return` after error yield
- **SDK bridge default**: removed hardcoded `127.0.0.1:8792/sdk` fallback so direct Cursor routing works when `sdk-bridge-url` is unset (config, synthesizer, and executor all updated)
- **Varint overflow**: added `shift >= 64` guard in `readProtoVarint`
- **Large-stream parsing**: replaced `bufio.Scanner` with `bufio.Reader` (64KB buffer limit)
- **`bytesTrimSpace`**: replaced with `bytes.TrimSpace`

**Interface compatibility**
- Added `ForceMapping` field + `GetForceMapping()` to `CursorComposerModel` to satisfy upstream `dev`'s model alias interface

## Tests added
- `TestCursorSessionIDUsesAccessTokenHash` (restored)
- `TestCursorComposerSDKBridgeURLEmptyWhenUnset`
- `TestCursorComposerSDKBridgeURLNormalizesExplicitValue`
- `TestReadProtoVarintOverflow`
- `TestParseConnectProtoFramesStopsAfterEndStream`
- `TestCursorChecksumAtUsesTimestamp`
- `TestCursorComposerExchangeAPIKeyCachesConcurrentCalls`
- `TestSanitizeCursorComposerKeysLeavesSDKBridgeURLEmptyWhenUnset`
- `TestSanitizeCursorComposerKeysKeepsExplicitSDKBridgeURL`
- `TestBytesTrimSpace` (darwin)

## Verification
- `go test ./...` — all pass
- `go build -o test-output ./cmd/server && rm test-output` — compiles clean
- `gofmt -w .` — formatted


bradhallett:
All three findings from the Gemini Code Assist review have been addressed in the latest commits:

**1. Checksum timestamp (critical)** — Fixed. `cursorChecksum` was dividing `Unix()` by 1M, which collapsed the timestamp to zero. Now uses `UnixMilli()` with a testable `cursorChecksumAt()` helper. Regression test: `TestCursorChecksumAtUsesTimestamp`.

**2. Nil-pointer in management handlers** — Investigated. `management.NewHandler()` always initializes `cfg` from a non-nil `*config.Config`. No nil path exists in the current codebase. Did not add guards for a condition that cannot occur.

**3. Token cache stampede** — Fixed. `exchangeAPIKey` now uses a mutex with double-checked locking so concurrent requests for the same API key share a single upstream exchange. Regression test: `TestCursorComposerExchangeAPIKeyCachesConcurrentCalls` (8 goroutines, asserts exactly 1 upstream request).

PR #4069, opened 2026-07-01
Improve reasoning content handling in response logic

TooYoungTooSimp:
Enhance reasoning content retrieval by falling back to 'reasoning' if 'reasoning_content' is absent or empty. In some selfhosted openai-like server serving deepseek, the response.delta may contains reasoning rather than reasoning_content. So the client like codex does not recognize the thinking process.
The fork is tested and in codex Ctrl+T, the thinking process shows correctly.


TooYoungTooSimp:
seems there's a path guard, then maintainers may feel free to adapt this idea

luispater (approved):

PR #4088, opened 2026-07-02
fix: retry Codex refresh on token invalidation

### LINKED ISSUE #4087: Codex token_invalidated should refresh current auth before fallback
## Summary

When a Codex OAuth access token returns `401 token_invalidated`, CLIProxyAPI currently suspends the auth/model and immediately falls through to the next available auth. It does not first call the executor's existing `Refresh()` implementation for that auth.

This means a primary Codex account can remain bypassed after an access-token invalidation even when its refresh token could recover the session. The user-visible behavior is that a backup account is used before the primary account is rechecked.

## Reproduction / evidence

Using v7.2.49 with two Codex auth files and `fill-first` routing:

1. Primary Codex auth returns `401 token_invalidated` from `/backend-api/codex/responses`.
2. Logs show the primary auth is suspended for the model.
3. The next auth is selected and the request succeeds.
4. No `codex executor: refresh called` log appears before fallback.

Observed log shape:

```text
Use OAuth provider=codex auth_file=<primary> for model gpt-5.5
request error, error status: 401, error message: Your authentication token has been invalidated. Please try signing in again.
Suspended client <primary> for model gpt-5.5: unauthorized
Use OAuth provider=codex auth_file=<backup> for model gpt-5.5
200 POST /v1/chat/completions
```

## Expected behavior

Before suspending/falling back on `401 token_invalidated`, CLIProxyAPI should:

1. Call `CodexExecutor.Refresh()` for the current auth.
2. Retry the same request once with the refreshed auth if refresh succeeds.
3. Only fall back/suspend if refresh fails or the retry also fails.

This avoids unnecessary backup-account usage when refresh can recover the primary auth.

kilhyeonjun:
## Summary

- detect Codex `401 token_invalidated` responses
- call the existing `CodexExecutor.Refresh()` path before suspending/falling back
- retry the same request once with refreshed auth for normal, compact, and streaming Codex responses
- add regression tests for non-stream, stream, and token-invalidated detection

Closes #4087

## Test plan

- `go test ./internal/runtime/executor -run 'TestCodexExecutorExecuteRefreshesAndRetriesTokenInvalidated|TestCodexExecutorExecuteStreamRefreshesAndRetriesTokenInvalidated|TestIsCodexTokenInvalidatedResponse' -count=1`
- `go test ./... -count=1`
- Manual smoke with two Codex auth files on v7.2.49 latest:
  - unpatched latest: primary `token_invalidated` immediately suspended/fell back
  - patched latest: primary `token_invalidated` called refresh first, then fell back only after refresh returned `refresh_token_invalidated`

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -39,6 +39,10 @@ const (
 	codexDefaultImageToolModel = "gpt-image-2"
 )
 
+var refreshCodexAuthForTokenInvalidatedRetry = func(ctx context.Context, e *CodexExecutor, auth *cliproxyauth.Auth) (*cliproxyauth.Auth, error) {
+	return e.Refresh(ctx, auth)
+}
+
 var dataTag = []byte("data:")
 
 // Streamed Codex responses may emit response.output_item.done events while leaving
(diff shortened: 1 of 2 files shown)


kilhyeonjun:
Addressed the automated review findings in `3548ee71`.

Changes:
- Avoid reassigning `httpResp` until the refreshed retry request succeeds, so a retry transport error cannot leave the deferred close pointing at a nil response.
- Applied the same temporary retry response pattern to normal, compact, and streaming retry paths.
- Clear Codex reasoning replay cache when the refreshed retry returns an invalid thinking-signature / invalid encrypted-content error.
- Added regression coverage for retry transport failure and retry invalid-signature replay clearing.

Local verification:
- `go test ./internal/runtime/executor -run 'TestCodexExecutorExecuteRefresh|TestIsCodexTokenInvalidatedResponse' -count=1`
- `go build -o /tmp/cli-proxy-api-4088 ./cmd/server`
- `go test ./... -count=1`
- `git diff --check`

kilhyeonjun:
Addressed the automated review feedback in `3548ee71e0d9eafbbca629c458c6a8536dcc7b3c`.

Summary:
- avoid clobbering `httpResp` when the refresh retry fails at the transport layer
- keep the original response close path safe by assigning `httpResp` only after a successful retry response
- apply invalid reasoning replay cleanup to retry error responses
- added regression coverage for retry transport failure and retry invalid-signature cleanup

Local verification:

```bash
go test ./internal/runtime/executor -count=1
```

Result: PASS.

Current PR checks are also green.

PR #4089, opened 2026-07-02
fix(translator): preserve Claude system prompts and tool args

danglinsheng:
Move mid-conversation Claude system messages into Gemini/Antigravity system instructions, avoid synthetic Gemini tool thought signatures, and sanitize Gemini tool call args against original Claude schemas.


danglinsheng:
Updated the PR to address the Gemini Code Assist review feedback in commit `57057ab`:

- filtered Claude Code attribution text from mid-conversation system content before appending it to Gemini system instructions;
- fixed strict-schema handling when `additionalProperties: false` is set and `properties` is absent;
- changed tool-argument sanitization to delete disallowed keys from the original JSON bytes, preserving large integer precision and original value formatting.

Validation run locally:

```bash
go test ./internal/translator/gemini/claude ./internal/translator/antigravity/claude
```

The remaining `ensure-no-translator-changes` failure appears to be the repository path guard for intentional `internal/translator/**` changes. Per that workflow message, I opened a tracking issue for the maintenance team: https://github.com/router-for-me/CLIProxyAPI/issues/4093

PR #4091, opened 2026-07-02
fix(auth): infer limit cooldowns from provider error text

yoonhwan:
## Summary
- Infer HTTP-like status from provider error text when executors return errors without StatusCode.
- Treat status-less rate limit/session limit/usage-credit messages as cooldown-worthy so auth selection rotates instead of immediately reusing the same credential.
- Treat Anthropic overloaded text as transient 529 capacity cooldown.
- Add focused regression coverage for status-less rate limit, Claude Code session limit, usage credit, and overloaded messages.

## Verification
- go test ./sdk/cliproxy/auth
- go build -o /tmp/cli-proxy-api.patched ./cmd/server
- Local patched binary deployed and restarted via launchctl
- curl http://127.0.0.1:8317/v1/models returned 200
- curl http://127.0.0.1:8790/health returned 200

## Notes
This fixes a failure mode where MarkResult recorded an unavailable model/auth without a retry timestamp because status extraction returned 0. In that state, selector logic could consider the same credential available again instead of rotating to the next credential.

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -171,6 +171,42 @@ type Result struct {
 	RetryAfter *time.Duration
 	// Error describes the failure when Success is false.
 	Error *Error
+	// IsProbe marks lightweight internal probes (e.g. Claude Code's
+	// max_tokens:1 "quota" usage-window ping) that must not influence
+	// credential cooldown/rotation decisions.
+	IsProbe bool
+}
+
+// isQuotaProbeRequest reports whether payload matches the shape of Claude
+// Code's built-in subscription-quota probe: a single-token request whose
+// last user message body is exactly "quota". These probes are sent on a
+// fixed interval independent of real traffic and can legitimately hit a
+// narrow provider-side rate limit; treating that as account/session
+// exhaustion would rotate away from a credential that still has real
+// capacity, defeating fill-first pinning.
+func isQuotaProbeRequest(payload []byte) bool {
+	if len(payload) == 0 {
+		return false
+	}
+	var probe struct {
+		MaxTokens int `json:"max_tokens"`
+		Messages  []struct {
+			Role    string `json:"role"`
+			Content any    `json:"content"`
+		} `json:"messages"`
+	}
+	if err := json.Unmarshal(payload, &probe); err != nil {
+		return false
+	}
+	if probe.MaxTokens != 1 || len(probe.Messages) == 0 {
+		return false
+	}
+	last := probe.Messages[len(probe.Messages)-1]
+	text, ok := last.Content.(string)
+	if !ok {
+		return false
+	}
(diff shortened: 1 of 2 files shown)


yoonhwan:
## Update: skip cooldown/rotation for quota-probe failures

Follow-up commit addresses a side effect of the status-less 429 inference above.

**Symptom**: with `fill-first` pinning, real usage was rotating between accounts every few minutes even though quota limits weren't actually being hit.

**Root cause**: Claude Code sends a lightweight internal probe every few minutes to check its subscription usage window — `{"max_tokens":1,"messages":[{"role":"user","content":"quota"}]}`. When that probe hits a narrow provider-side rate limit (429), this PR's broadened inference treated it identically to real session/quota exhaustion, marking the account `Quota.Exceeded` and rotating `fill-first` away from a credential that still had real capacity. Since this repeats on every account in turn, it produced round-robin-like behavior despite `fill-first` being configured.

**Fix**: added `Result.IsProbe` + `isQuotaProbeRequest()` to detect this specific probe shape, and short-circuit `MarkResult` before any auth/model state mutation when a probe fails. Real traffic 429s are unaffected and still cool down/rotate as before.

### Verification
- `go test ./sdk/cliproxy/auth/...` — all pass, including 3 new tests (`TestManager_MarkResult_QuotaProbeFailureDoesNotCooldownAuth`, `TestManager_MarkResult_RealTraffic429StillCoolsDownAuth`, `TestIsQuotaProbeRequest` with 7 subcases)
- `go build -o /tmp/cli-proxy-api.probe-patch ./cmd/server` — clean build
- Local patched binary deployed via launchctl, `curl :8317/v1/models` → 200, `curl :8790/health` (headroom chain) → ready

yoonhwan:
## Retracting the "skip cooldown for quota-probe failures" follow-up

The `IsProbe` commit above (2a8d4332) was based on a flawed premise: I assumed Claude Code's periodic quota-check probe (`max_tokens:1`, content `"quota"`) hitting 429 was noise unrelated to real account capacity, and that skipping cooldown on probe failure would stop unnecessary rotation.

In practice this was wrong — the probe's 429 genuinely reflects the account's real session/quota exhaustion. Skipping cooldown on probe failure pinned the credential selector to an already-exhausted account indefinitely, since nothing else triggered rotation for that credential. This caused exactly the failure mode this PR fixes (rotation not happening), reintroducing it through a different path.

Verified via live `debug: true` observation: with only the original commit (589b20d4) deployed, real requests correctly rotate `developer.ts@... → yoonhwan.ko@byourz.com... → yoonhwan.ko@gmail.com...` and land on a non-exhausted credential. With my IsProbe patch added on top, rotation stalled on the exhausted account.

Please disregard/drop commit 2a8d4332 — the original fix (589b20d4) is correct and sufficient as-is. Sorry for the noise.

PR #4204, opened 2026-07-10
feat(openai-compat): Add extra_body config for provider-specific request fields

### LINKED ISSUE #4203: Feature: `extra_body` config for openai-compatibility providers to inject provider-specific request body fields
Some OpenAI-compatible providers require non-standard fields in the request body that aren't part of the OpenAI API spec. Currently, CLIProxyAPI's `openai_compat_executor.go` has no mechanism to inject these, making it impossible to use such providers without patching the code.

## Use case: Nous Portal

The Nous Portal inference API (`inference-api.nousresearch.com/v1`) is documented as a standard OpenAI-compatible API. However, in practice, at least some models (e.g. `tencent/hy3:free`, which routes through OpenRouter) reject requests without a `tags` array in the body:

```json
{
  "model": "tencent/hy3:free",
  "messages": [...],
  "tags": ["product=hermes-agent", "user=<api_key>"]
}
```

Without `tags`, the API returns HTTP 400:
```
{"status": 400, "message": "This request is not valid. Check the model name and other parameters. Additional info: missing tags"}
```

This requirement is **not documented** in the official Nous API docs ([portal.nousresearch.com/api-docs](https://portal.nousresearch.com/api-docs)), which describe the API as a plain OpenAI-compatible endpoint with bearer token auth. The `tags` field appears to be an undocumented requirement for certain models or account tiers.

CLIProxyAPI cannot inject this field, so the Nous provider is unusable through the `openai-compatibility` config without a code patch.

## Proposed solution

Add an `extra_body` map to the `openai-compatibility` config that gets merged into the upstream request body after translation:

```yaml
openai-compatibility:
  - name: "nous"
    base-url: "https://inference-api.nousresearch.com/v1"
    extra_body:
      tags:
        - "product=hermes-agent"
    api-key-entries:
      - api-key: "sk-nous-..."
    models:
      - name: "tencent/hy3:free"
        alias: "tencent/hy3:free"
```

The `extra_body` fields would be merged into the translated payload just before the HTTP request is sent, in both the streaming and non-streaming execution paths.

## Reference

Hermes Agent already supports this pattern via `extra_body` on custom providers:

```yaml
custom_providers:
  - name: gemma-local
    base_url: http://localhost:8080/v1
    model: google/gemma-4-31b-it
    extra_body:
      enable_thinking: true
```

## Scope

- `extra_body` at the provider level (applies to all models under that provider)
- Fields are merged after `sdktranslator.TranslateRequest` and before `http.NewRequestWithContext`
- Should work for both `/chat/completions` and `/responses/compact` endpoints
- Design decision to discuss: should `extra_body` overwrite existing fields in the translated payload, or skip them?

kumamaki:
Some OpenAI-compatible providers require non-standard fields in the request body that aren't part of the OpenAI API spec. Without a way to inject these from config, such providers need executor code patches.

## What this adds

An `extra_body` map on `openai-compatibility` providers, merged into the translated payload before the upstream request. String values support `{{api_key}}` for dynamic substitution with the resolved API key.

```yaml
openai-compatibility:
  - name: "nous"
    base-url: "https://inference-api.nousresearch.com/v1"
    extra_body:
      tags:
        - "product=hermes-agent"
        - "user={{api_key}}"
    api-key-entries:
      - api-key: "sk-..."
    models:
      - name: "tencent/hy3:free"
        alias: "tencent/hy3:free"
```

## Why

The Nous Portal inference API rejects requests without a `tags` array (HTTP 400, "missing tags"). This requirement is undocumented in their API docs. Other providers may have similar non-standard fields.

Hermes Agent already supports this pattern via `extra_body` on custom providers.

## How it works

- `ExtraBody map[string]any` added to `OpenAICompatibility` config struct
- `injectExtraBody` merges fields into the payload after `TranslateRequest`, before `NewRequestWithContext`
- Called in both `Execute` (non-streaming) and `ExecuteStream` paths
- `{{api_key}}` in string values replaced with the resolved credential at runtime

Closes #4203

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -315,6 +315,10 @@ nonstream-keepalive-interval: 0
 #     disable-cooling: false # optional: per-provider override for auth/model cooldown scheduling
 #     headers:
 #       X-Custom-Header: "custom-value"
+#     extra-body: # optional: provider-specific fields merged into the upstream request body after translation
+#       tags: # example: Nous Portal requires a tags array (undocumented)
+#         - "product=hermes-agent"
+#         - "user={{api_key}}" # {{api_key}} is replaced with the resolved API key at runtime
 #     api-key-entries:
 #       - api-key: "sk-or-v1-...b780"
 #         proxy-url: "socks5://proxy.example.com:1080" # optional: per-key proxy override
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -631,6 +631,10 @@ type OpenAICompatibility struct {
 
 	// DisableCooling disables auth/model cooldown scheduling for this provider when true.
 	DisableCooling bool `yaml:"disable-cooling,omitempty" json:"disable-cooling,omitempty"`
+
+	// ExtraBody is merged into the upstream request body after translation.
+	// Use for provider-specific fields not part of the OpenAI API spec.
+	ExtraBody map[string]any `yaml:"extra-body,omitempty" json:"extra-body,omitempty"`
 }
 
 // OpenAICompatibilityAPIKey represents an API key configuration with optional proxy setting.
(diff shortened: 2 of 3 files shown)


kumamaki:
Closing in favor of issue discussion. Will reopen with adjustments based on maintainer feedback.

PR #4228, opened 2026-07-11
fix(images): allow grok models on image endpoints

0xRichardH:
## Summary

- execute Grok image generation and edit requests through the image-aware auth manager path
- allow both standard and locally wrapped streaming image requests to pass image-only model validation
- prevent `/v1/images/generations` and `/v1/images/edits` from returning the misleading image-endpoint-only HTTP 503

## Root cause

The xAI image handlers called `ExecuteWithAuthManager`, which sets `allowImageModel` to `false`. Since Grok image models are classified as image-only, CLIProxyAPI rejected them before provider selection even when the request already used a supported image endpoint.

## Testing

- `go test ./sdk/api/handlers/openai ./sdk/api/handlers -count=1`
- `go test ./internal/runtime/executor/... -count=1`

## Affected version

`v7.2.66`

## Observed error

CLIProxyAPI incorrectly returned HTTP 503 before provider selection or any upstream request:

```json
{"error":{"message":"model grok-imagine-image is only supported on /v1/images/generations and /v1/images/edits","type":"server_error","code":"internal_server_error"}}
```

### DIFF
--- a/sdk/api/handlers/openai/openai_images_handlers.go
+++ b/sdk/api/handlers/openai/openai_images_handlers.go
@@ -1401,7 +1401,7 @@ func (h *OpenAIAPIHandler) collectImagesWithModel(c *gin.Context, imageReq []byt
 	stopKeepAlive := h.StartNonStreamingKeepAlive(c, cliCtx)
 
 	model = strings.TrimSpace(model)
-	resp, upstreamHeaders, errMsg := h.ExecuteWithAuthManager(cliCtx, xaiImagesHandlerType, model, imageReq, "")
+	resp, upstreamHeaders, errMsg := h.ExecuteImageWithAuthManager(cliCtx, xaiImagesHandlerType, model, imageReq, "")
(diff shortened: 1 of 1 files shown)


0xRichardH:
Hi @luispater, Please review this PR/change. It addresses an issue where the image generation API returns a 503 status code when utilizing grok-imagine-image. Thanks :bow:

luispater (commented):
Thanks for the fix. The two changes in this PR are correct, but they have already been superseded by the more complete fix in commit `042f1fea6809380b5538978449a54be90249ee2b`, which was officially released in `v7.2.68`.

That commit includes both replacements from this PR, also updates `streamOpenAICompatImages` to use `ExecuteImageStreamWithAuthManager`, and adds regression coverage for image-only model execution. Since merging this PR into the current `dev` branch produces no tree changes, I am closing it as already resolved by `v7.2.68`.

PR #4243, opened 2026-07-12
fix(antigravity): deduplicate function declarations before sending to upstream

TuTouPower:
## Summary

When clients register multiple MCP plugin sub-servers with overlapping tool names (e.g., Cloudflare's builds/bindings/observability/docks sub-servers all exporting `workers_get_worker`), the antigravity upstream rejects the request with:

> 400 Duplicate function declaration found: mcp__plugin_cloudflare_cloudflare-builds__workers_builds_get_bui

This happens because:
1. Multiple MCP sub-servers register identically-named tools (real duplicates)
2. `SanitizeFunctionName` truncates long tool names to 64 chars, causing hash collisions between different tools (e.g., `workers_builds_get_build` vs `workers_builds_get_build_logs` both truncate to `workers_builds_get_bui`)

This PR only addresses case 1.

## Changes

- Add `DeduplicateFunctionDeclarations` to `internal/util/translator.go` — removes duplicate function declarations by name, keeping the first occurrence
- Apply dedup in all four antigravity translator paths before serializing `functionDeclarations` into the request payload:
  - `internal/translator/antigravity/claude/antigravity_claude_request.go`
  - `internal/translator/antigravity/gemini/antigravity_gemini_request.go`
  - `internal/translator/antigravity/openai/chat-completions/antigravity_openai_request.go`
  - `internal/translator/antigravity/interactions/interactions_antigravity_request.go`

## Test Plan

- [x] `go build -o /tmp/cli-proxy-api ./cmd/server` passes
- [x] `go test ./internal/util/...` passes (117 tests)
- [x] `go test ./internal/translator/antigravity/...` passes (148 tests)

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_request.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_request.go
@@ -687,8 +687,15 @@ func ConvertClaudeRequestToAntigravity(modelName string, inputRawJSON []byte, _ 
 			}
 		}
 		if toolDeclCount > 0 {
-			toolsJSON = []byte(`[]`)
-			toolsJSON, _ = sjson.SetRawBytes(toolsJSON, "-1", functionToolNode)
+			if decls := gjson.GetBytes(functionToolNode, "functionDeclarations"); decls.IsArray() {
+				deduped := util.DeduplicateFunctionDeclarations([]byte(decls.Raw))
+				functionToolNode, _ = sjson.SetRawBytes(functionToolNode, "functionDeclarations", deduped)
+				toolDeclCount = int(gjson.GetBytes(deduped, "#").Int())
+			}
+			if toolDeclCount > 0 {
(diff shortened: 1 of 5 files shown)


TuTouPower:
Updated per review feedback:

1. **`DeduplicateFunctionDeclarations`** — replaced O(N²) `sjson.SetRawBytes` loop with `strings.Join` for O(N) performance
2. **Gemini translator** — merged the separate rename and dedup loops into a single pass
3. **Gemini translator** — handles both `function_declarations` (snake_case) and `functionDeclarations` (camelCase)

Build and 265 tests pass.

TuTouPower:
Fixed in c3d8368: re-read `declsResult` after the parameters→parametersJsonSchema rename before deduping, so the dedup pass writes back the renamed declarations instead of the stale pre-rename data.

PR #4258, opened 2026-07-13
Add codex-oauth-base-url and codex-oauth-headers for routing codex OAuth traffic through a companion proxy

khairm:
## Motivation

ChatGPT-Codex subscription (OAuth) traffic currently always egresses to the fixed `https://chatgpt.com/backend-api/codex` upstream. `codex-api-key` entries already support a per-key `base-url` override, but OAuth auth records have no supported equivalent, so there is no way to route subscription traffic through a local or companion proxy (e.g. a compression/caching gateway) without patching the binary.

This PR adds two top-level config keys that close that gap for OAuth credentials only, with strict validation and no behavior change when unset.

## Keys

- `codex-oauth-base-url` (string, default empty = stock behavior): upstream base URL for codex OAuth credentials (auth records without an `api_key` attribute). Validated at config load: absolute `http`/`https` URI, non-empty host, no userinfo/query/fragment; trailing slash trimmed. Startup fails loudly if any auth record or matching `codex-api-key` entry carries a competing `base_url`, so precedence is never silent. The value is captured at startup (hot-reload-inert) to avoid flipping a live process's routing.
- `codex-oauth-headers` (map, default empty): additional outbound headers (e.g. an API key for the companion proxy) applied to every routed codex OAuth request. Header names must be RFC 7230 tokens; `authorization`, `host`, and `content-length` are rejected case-insensitively. Setting headers without the base URL is a startup error.

## Transport

When the override is set, affected OAuth auths are forced onto the plain HTTP executor: companion proxies generally expose only the HTTP `/responses` surface, and forcing HTTP guarantees no request can bypass the configured upstream via the websocket path. Stock behavior (key unset) is untouched, including websockets.

All `backend-api/codex` consumers honor the override (responses, streaming, compact, images, generic HTTP requests, model fetch).

## Tests

`internal/config` (validation matrix incl. userinfo/query/fragment/relative/empty-host rejection, conflict detection, headers-without-base-url), `internal/runtime/executor` (override applied on HTTP path, headers attached, WS unreachable when set), `sdk/cliproxy` (service wiring), `cmd/fetch_codex_models`. `go build ./...` and the affected package tests pass; the one pre-existing failure on untouched HEAD (`internal/api.TestModelsWithClientVersionReturnsCodexCatalog`) is unrelated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,6 +20,7 @@ go build -o test-output ./cmd/server && rm test-output # Verify compile (REQUIRE
 - Default config: `config.yaml` (template: `config.example.yaml`)
 - `.env` is auto-loaded from the working directory
 - Auth material defaults under `auths/`
+- `codex-oauth-base-url` and `codex-oauth-headers` are startup-only and route Codex subscription OAuth requests over HTTP
 - Storage backends: file-based default; optional Postgres/git/object store (`PGSTORE_*`, `GITSTORE_*`, `OBJECTSTORE_*`)
 
 ## Architecture
--- a/cmd/fetch_codex_models/main.go
+++ b/cmd/fetch_codex_models/main.go
@@ -127,6 +127,16 @@ func main() {
 	}
 
 	fmt.Printf("Using auth: id=%s label=%s\n", chosen.ID, chosen.Label)
+	modelsBaseURL := codexModelsBaseURL
+	var modelsOAuthHeaders map[string]string
+	if cfg.CodexOAuthBaseURL != "" && (chosen.Attributes == nil || strings.TrimSpace(chosen.Attributes["api_key"]) == "") {
+		if authBaseURL := codexAuthBaseURL(chosen); authBaseURL != "" {
+			fmt.Fprintf(os.Stderr, "error: codex-oauth-base-url conflicts with Codex OAuth auth %q base_url %q\n", chosen.ID, authBaseURL)
+			os.Exit(1)
+		}
+		modelsBaseURL = cfg.CodexOAuthBaseURL
+		modelsOAuthHeaders = cfg.CodexOAuthHeaders
+	}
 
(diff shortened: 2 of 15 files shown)


khairm:
v2 pushed: routing is now derived exclusively from the post-authentication API-key principal via a new `codex-oauth-route-api-keys` list (no always-on override; ordinary-bearer traffic is differential-tested wire-identical to stock). Adds validated `x-edgee-session-id` forwarding on routed requests, immutable request-scoped route context (race-tested), failure isolation so routed upstream errors never refresh/cool-down the shared ChatGPT OAuth credential, and atomic auth-file persistence (temp+sync+rename, ReplaceFile on Windows) with surfaced errors. The only failing repo-wide test is the pre-existing `TestModelsWithClientVersionReturnsCodexCatalog` priority mismatch, reproducible on stock v7.2.62 and untouched here.

khairm:
Updated the branch with a hardening follow-up (fork release v7.2.62-edgee.3):

- **Route safety**: `codex-oauth-route-api-keys` bearers now only execute against Codex OAuth credentials with the configured override; a non-OAuth/non-codex credential selection is rejected loudly (`503 route_unavailable`) instead of silently bypassing the override, and startup fails if route keys are configured without any Codex OAuth credential.
- **Wire fidelity**: successful upstream responses are never converted into downstream errors by post-result persistence failures — stock behavior is preserved and a persistence failure instead surfaces on the next request pre-upstream. Added differential tests that compare the ordinary-bearer wire output against fixtures generated from the stock v7.2.62 tag (POST, SSE, count-tokens, query-bearing URL, error path).
- **Redaction**: request logging fully redacts routed `Authorization` and injected edgee headers.
- **Validation**: `codex-oauth-headers` validation is deterministic and rejects hop-by-hop, duplicate (case-insensitive), and control-character entries.

PR #4311, opened 2026-07-14
feat(claude): add global claude-cloak-mode config (auto/always/never)

antonichoudhuri:
## Problem

Claude cloaking is decided per-request by `ShouldCloak`. In the default `auto` mode it **skips cloaking for any client whose `User-Agent` starts with `claude-cli`** (assumed to be a genuine Claude Code client). Today the only *global* control is `disable-claude-cloak-mode`, which forces `"never"` — there is **no global way to force `"always"`**.

That's a footgun for the common deployment where the proxy only ever fronts **non-Claude-Code apps**: if such an app sends a `claude-cli` User-Agent (e.g. because it was told that impersonating Claude Code helps), the proxy skips cloaking and forwards a half-formed Claude identity. Anthropic's **premium tier (Sonnet/Opus) then rejects it with `429 rate_limit_error`, while Haiku (cloak-exempt) keeps working** — so the failure is easy to misdiagnose as a rate limit until a strong model is used.

The only current remedy is a **per-credential** `cloak_mode`, which for OAuth lives inside each token JSON and is **regenerated (lost) on re-login** — so it can't be pinned in version-controlled config.

## Change

Add a global **`claude-cloak-mode: "auto" | "always" | "never"`** that sets the default cloak mode for every Claude credential, still overridable per-credential (per `claude-api-key` `cloak.mode` or per-OAuth `cloak_mode`) exactly as before.

- Precedence (low → high): built-in `auto` → **`claude-cloak-mode` global default** → `disable-claude-cloak-mode` (still forces `never`, kept for backward compat and wins if set) → per-credential → per-`claude-api-key`.
- `claude-cloak-mode: "always"` makes cloaking independent of the client User-Agent, closing the footgun in one place, in config that survives re-login.
- Fully backward compatible: default (empty) preserves today's `auto` behavior; `disable-claude-cloak-mode` is unchanged.

## Files
- `internal/config/config.go` — new `ClaudeCloakMode string` field (`yaml:"claude-cloak-mode"`).
- `internal/runtime/executor/claude_executor.go` — apply it as the global default in the cloak-mode precedence block.
- `config.example.yaml` — document the new knob.
- `internal/runtime/executor/claude_executor_test.go` — `TestApplyCloaking_GlobalAlwaysModeCloaksClaudeCliClient`: verifies `always` cloaks a `claude-cli` client that `auto` skips.

## Verification
- `gofmt` clean; `go build ./cmd/server` OK.
- `go test ./internal/runtime/executor/ ./internal/config/` passes (incl. the new test).

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -134,11 +134,23 @@ save-cooldown-status: false
 # Set to 0 to keep the legacy 60-second cooldown; set to -1 to disable transient error cooldowns.
 transient-error-cooldown-seconds: 0
 
-# When true, globally disable Claude request cloaking (the Claude Code CLI disguise and
-# system prompt replacement), so the original system prompt is passed through to Claude as-is.
-# Individual credentials can still override this: a claude-api-key entry via its "cloak.mode",
-# or a Claude OAuth/token file via a "cloak_mode" value. Default false keeps the per-client
-# "auto" behavior (cloak only non-Claude-Code clients).
+# Global default Claude cloak mode: "auto" (default), "always", or "never".
+# Cloaking is the Claude Code CLI disguise (device fingerprint + fake user id +
+# system prompt replacement). "auto" cloaks only clients whose User-Agent is not
+# Claude Code; "always" cloaks every request regardless of client; "never" disables
+# it. Set "always" when this proxy only ever fronts non-Claude-Code apps: it stops a
+# stray "claude-cli" User-Agent from disabling cloaking (which makes Anthropic reject
+# premium models with 429 while Haiku still works). Individual credentials can still
+# override this via a claude-api-key "cloak.mode" or a Claude OAuth/token file
+# "cloak_mode" value. Empty keeps "auto".
+claude-cloak-mode: "auto"
+
+# When true, globally disable Claude request cloaking (equivalent to
+# claude-cloak-mode: "never"; kept for backward compatibility and takes precedence
+# over claude-cloak-mode when true), so the original system prompt is passed through
+# to Claude as-is. Individual credentials can still override this: a claude-api-key
+# entry via its "cloak.mode", or a Claude OAuth/token file via a "cloak_mode" value.
+# Default false keeps the per-client "auto" behavior (cloak only non-Claude-Code clients).
 disable-claude-cloak-mode: false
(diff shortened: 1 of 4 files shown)


nimser:
Adding a data point in favour of this. The need for it grew with v7.3.17, and the implementation needs a rebase to account for it.

#6096 (released in v7.3.17) added a direct Anthropic Messages passthrough. For a Claude-format request to an Anthropic upstream that uses an OAuth credential, cloaking is now skipped unless the credential has an explicit cloak configuration ([`claude_executor_cloaking.go#L1432`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.3.17/internal/runtime/executor/claude_executor_cloaking.go#L1432)). The header path does the same ([`claude_executor_request.go#L1096`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.3.17/internal/runtime/executor/claude_executor_request.go#L1096)), and so do the cache and billing paths in `execute`/`stream`. As a result, clients like Pi calling `/v1/messages` directly lost cloaking on upgrade, and Anthropic now rejects their Opus requests with a generic 429. Details are in #6120.

Two consequences for this PR:

1. **The use case is now common rather than a corner case.** Every direct-Messages client on an OAuth credential needs an explicit opt-in, and today the only opt-in is a `cloak_mode` in each credential file. Re-login to the same account keeps that field, since `saveTokenRecord` merges existing metadata. A newly added account starts without it. We work around this with a sidecar that PATCHes `cloak_mode: "always"` onto new Claude credentials through `/v0/management/auth-files/fields`. That leaves a window before each new credential is patched, which a config default would close.
2. **A global mode must count as "configured".** In v7.3.17, `resolveClaudeWirePolicy` computes `CloakConfigured` only from the `claude-api-key` cloak block and per-credential attributes. If `claude-cloak-mode` only changes `cloakMode`, the passthrough check still sees `!CloakConfigured` and disables cloaking. So `always` would do nothing on the path where it is most needed. An explicit global value (at least `always`) should set `CloakConfigured`, while the empty default keeps #6096's passthrough. With the precedence already described here, per-credential and per-key settings still win.

Happy to test a rebased version against a live setup.

Issue #3361, opened 2026-05-12
Fix Claude Responses tool_search and namespace tool round-trip


ultrano:
﻿## Summary

Claude models behind the OpenAI Responses compatibility layer need a complete tool-use round trip for Responses `tool_search`, discovered tools, and namespace tools.

This issue tracks the request/response conversion behavior needed for Claude to discover tools, call `tool_search`, replay `tool_search_output`, activate discovered tools, and preserve namespace metadata across turns.

## Symptoms

- Claude can fail to see or call `tool_search` as a first-class Responses tool.
- Claude `tool_use` events can be converted back as plain `function_call` items when they should become `tool_search_call`.
- Tool search results from previous Responses input may not be replayed into the next Claude request, so discovered tools are unavailable in follow-up turns.
- Namespace tools can lose their namespace on history replay or response conversion.
- Plain tool names can incorrectly fall back to namespace matches when both forms are present.

## Desired behavior

- Responses `tool_search` is exposed to Claude with a deterministic Claude-compatible tool name.
- Claude `tool_use` for tool search is restored as Responses `tool_search_call` with stable `call_id` and JSON arguments.
- Previous `tool_search_output` entries are replayed into Claude request context.
- Tools discovered by `tool_search` can be activated in later turns.
- Namespace tool calls preserve `namespace` and function `name` across request history and response output.
- Plain tools win over namespace fallback when there is a name collision.

## Likely implementation area

- `internal/translator/claude/openai/responses/claude_openai-responses_request.go`
- `internal/translator/claude/openai/responses/claude_openai-responses_response.go`

## Validation ideas

- Unit tests for converting Responses `tool_search` into Claude tool definitions.
- Unit tests for restoring Claude `tool_use` into `tool_search_call`.
- Unit tests for replaying `tool_search_output` into later Claude turns.
- Unit tests for namespace tools, discovered namespace tools, and plain-name collision cases.
- End-to-end check with a Claude model using `tool_search`, then calling a discovered connector/tool in a later step.

shenshuoyaoyouguang:
追加一个更具体的复现场景，来自本地 `main` 上对当前转换代码的检查。

## 新症状

通过 Claude 模型走 OpenAI Responses 兼容层时，模型会调用 `ToolSearch`，但参数是 `{}` 或没有 `query`，下游工具执行报：

```text
InputValidationError: The required parameter "query" is missing
```

[...]

PR #4322, opened 2026-07-15
fix(translator): preserve deferred tool references

camjac251:
## Summary

- translate valid deferred-definition discoveries into native request-history items
- emit the matching client discovery declaration and preserve forced-choice behavior
- retain loaded schemas and residual text or images across later calls
- fall back atomically to eager definitions for incomplete, errored, out-of-order, duplicated, colliding, orphaned, or malformed histories
- keep server-provided search definitions on the eager compatibility path

## Why

Source histories can represent a definition discovery as references instead of repeating the full schemas. Dropping those references leaves later calls without a definition loaded earlier in the conversation.

Native history translation also has to agree with the declarations sent in the same request. Selecting it for only part of a history can omit callable definitions or create invalid call/output pairs. The translator now enables native discovery only when the client declaration and every discovery pair are complete and unambiguous. Otherwise, it preserves the ordinary history and advertises all definitions eagerly.

## Test plan

- [x] `go test ./internal/translator/... -count=1`
- [x] `go build -buildvcs=false -o /tmp/translated-history-server ./cmd/server`
- [x] `git diff --check`
- [x] staged secret scan

### DIFF
--- a/internal/translator/codex/claude/codex_claude_request.go
+++ b/internal/translator/codex/claude/codex_claude_request.go
@@ -45,6 +45,8 @@ func ConvertClaudeRequestToCodex(modelName string, inputRawJSON []byte, _ bool) 
 
 	rootResult := gjson.ParseBytes(rawJSON)
 	toolNameMap := buildReverseMapFromClaudeOriginalToShort(rawJSON)
+	toolCatalog := buildClaudeToolCatalog(rootResult.Get("tools"), toolNameMap)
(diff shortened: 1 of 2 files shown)


camjac251:
The repository path guard asks for a maintainer issue for translator changes. Existing issue #3361 already tracks the relevant round-trip and replay behavior, so I did not open a duplicate. This PR remains a tested implementation reference.

PR #4325, opened 2026-07-15
fix(translator): omit unsupported response options

camjac251:
## Summary

- remove `output_config` before subscription-backed Responses requests
- retain the supported `reasoning.effort` setting
- add focused regression coverage

## Test plan

- [x] focused translator package tests
- [x] formatting and patch hygiene checks


camjac251:
The repository path guard asks for a maintainer issue for translator changes. Opened #4347 as the focused tracker. This PR remains a tested implementation reference.

luispater:
Thanks for the contribution. As noted on #4347, `output_config` does not appear in the official OpenAI Responses API documentation:
https://developers.openai.com/api/reference/resources/responses/methods/create

It is currently unclear under what circumstances this field is added. If it is generated by an unofficial client, that client should stop sending the unsupported field, and the issue should be reported to its maintainers rather than making the translator silently remove it.

If an official OpenAI client produces this field, please provide the client name, version, and a minimal reproduction. We are closing this PR together with the related issue based on the currently available information.

PR #4353, opened 2026-07-16
feat(sdk): add headless embedded runtime

zhexuany:
## Summary
- add a public `EmbeddedRuntime` lifecycle backed by a host-supplied core auth manager
- reuse native executor and model registration without constructing the HTTP server, watcher, access manager, or management routes
- support targeted auth sync/removal, active-model reconciliation, and cleanup of runtime-owned projections

## Test plan
- `go test ./sdk/cliproxy ./sdk/cliproxy/auth -count=1`
- `go test -race ./sdk/cliproxy ./sdk/cliproxy/auth -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/sdk/cliproxy/auth/auto_refresh_lease_test.go
+++ b/sdk/cliproxy/auth/auto_refresh_lease_test.go
@@ -1,0 +1,171 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+)
+
+type leaseTestSelector struct {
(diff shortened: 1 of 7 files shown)


zhexuany:
Addressed the review findings in `bd56e081`:

- normalize active model IDs case-insensitively
- preserve host-owned executors and custom RoundTripper providers
- add atomic register-if-absent / compare-and-unregister ownership operations
- roll back executor/model ownership on canceled startup
- skip model projection when no executable provider is installed
- keep embedded reconciliation runtime-only (no host persistence/hooks)
- recheck cancellation around serialized lifecycle operations

Validation passed: `go test ./sdk/cliproxy/...`, `go test -race ./sdk/cliproxy/...`, `go vet ./sdk/cliproxy/...`, and `git diff --check`.

PR #4369, opened 2026-07-16
fix(executor): close request-owned uTLS HTTP/2 connections

qinghua362330:
## Problem

`NewUtlsHTTPClient` builds a **new** `http.Client` per request (8 call sites: `claude_executor.go:194,310,499,755`, `codex_executor.go:808,891,1067,1183`). Its roundtripper hand-builds an HTTP/2 connection with `tr.NewClientConn(tlsConn)` (`utls_client.go:97`), which bypasses `http2.Transport`'s own pool — so nothing ever owns or closes it.

The result is roughly **one leaked ESTABLISHED socket per request** to the protected hosts. Nothing reuses the connection (the client is per-request) and nothing closes it (there is no `CloseIdleConnections` on `utlsRoundTripper`, and the RoundTrip error path at `:121-125` only `delete`s the map entry without closing).

Measured on a busy deployment: the process climbed to **9733 fd / 3.47 GB RSS over ~3h**. With this patch it stays flat at **44–64 fd / ~300 MB** under the same load.

Worth noting: fixing only the "replace a stale connection" branch in `getOrCreateConnection` (`:49-78`) does **not** help — on a per-request client that map is always empty, so that branch is dead code in practice. The leak has to be fixed by giving the connection an owner.

## Approach

Tie the connection's lifetime to the response body, which is the one thing whose lifetime already matches the request:

- `closeOwnedTransportRoundTripper` wraps the per-request transport and reclaims it when `resp.Body.Close()` runs (and immediately on a RoundTrip error, where there is no body to tie to).
- `utlsRoundTripper.CloseIdleConnections()` closes the pooled connections.
- `shutdownUtlsConnection` drains gracefully, then closes.
- The stale-connection replace path and the RoundTrip error path now close the connection they drop.
- A RoundTripper injected through the context is **not** owned, so it is never closed by this wrapper.

### Relation to #3861

#3861 reported the same leak and fixed it by globally caching roundtrippers keyed by proxy URL. The review objection there was that the cache is unbounded with no eviction, and the PR was closed.

This patch takes the opposite approach — **ownership instead of caching**. There is no global state and no cache to bound: each request's transport is reclaimed deterministically when its body closes. It fixes correctness first; connection *reuse* can be layered on later as a separate optimization if desired.

### On `AGENTS.md`'s timeout rule

`AGENTS.md` says not to set timeouts on network behaviour after an upstream connection is established. The 30s bound in `shutdownUtlsConnection` is not that: it applies only to the graceful drain of a connection whose request has **already finished**, and it falls back to an unconditional `Close()`. It never constrains in-flight request/response traffic. Happy to drop it to a bare `Close()` if you'd prefer to avoid the ambiguity entirely.

## Testing

`utls_client_test.go` adds coverage for: reclaim-once on body close (idempotent across a double `Close()`), reclaim on RoundTrip error, and that a ctx-injected shared RoundTripper is left alone.

- `go build ./...` — OK
- `gofmt -l` / `go vet` — clean
- `go test ./internal/runtime/executor/helps/` — **5/5 pass** (4 new + the existing ctx-roundtripper test)

Branched from and verified against `dev`.

### DIFF
--- a/internal/runtime/executor/helps/utls_client.go
+++ b/internal/runtime/executor/helps/utls_client.go
@@ -2,6 +2,7 @@ package helps
 
 import (
 	"context"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -25,6 +26,8 @@ type utlsRoundTripper struct {
 	pending     map[string]*sync.Cond
 	dialer      proxy.Dialer
 }
+
+const utlsConnectionShutdownTimeout = 30 * time.Second
 
 func newUtlsRoundTripper(proxyURL string) *utlsRoundTripper {
 	var dialer proxy.Dialer = proxy.Direct
@@ -75,8 +78,22 @@ func (t *utlsRoundTripper) getOrCreateConnection(host, addr string) (*http2.Clie
 		return nil, err
 	}
 
+	if old := t.connections[host]; old != nil && old != h2Conn {
+		go shutdownUtlsConnection(old)
+	}
 	t.connections[host] = h2Conn
 	return h2Conn, nil
+}
+
+func shutdownUtlsConnection(conn *http2.ClientConn) {
+	if conn == nil {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), utlsConnectionShutdownTimeout)
+	defer cancel()
+	if err := conn.Shutdown(ctx); err != nil {
+		_ = conn.Close()
+	}
 }
 
 func (t *utlsRoundTripper) createConnection(host, addr string) (*http2.ClientConn, error) {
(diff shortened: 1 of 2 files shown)


qinghua362330:
Thanks for the review. The keep-alive concern is the right thing to worry about in general, but it doesn't apply to this code path as it exists today — and I think the recommended alternative has a blocker worth surfacing before anyone invests in it.

### There is no connection reuse to break here

`NewUtlsHTTPClient` is called **per request**, at all 8 call sites:

- `codex_executor.go:808`, `:891`, `:1067`, `:1183`
- `claude_executor.go:194`, `:310`, `:499`, `:755`

Each request constructs its own `http.Client` → its own `utlsRoundTripper` → its own `connections` map. The "same client used sequentially/concurrently" scenario doesn't occur: no client outlives the request that created it, and no connection is ever shared between requests.

That's also why the existing stale-connection branch in `getOrCreateConnection` is effectively dead code — a map created one request ago never has an entry to replace.

So the wrapper closes **that request's own** transport, which nothing else can reach. It is careful not to touch a RoundTripper injected via `ctx` (`cliproxy.roundtripper`), precisely because that one *is* shared and not owned by the request.

### The 25-call assertion is the fix, not the bug

`TestCloseOwnedTransportRoundTripperClosesEveryTrackedResponse` asserts 25 reclaims for 25 requests because there are 25 independent per-request transports. Before this patch those 25 connections are all leaked. The count matching the request count is exactly the property being fixed.

### Storing one client on the executor can't work as described

The client is a function of the credential:

```go
func NewUtlsHTTPClient(ctx context.Context, cfg *config.Config, auth *cliproxyauth.Auth, timeout time.Duration) *http.Client {
    var proxyURL string
    if auth != nil {
        proxyURL = strings.TrimSpace(auth.ProxyURL)
    }
    ...
    var utlsRT http.RoundTripper = newUtlsRoundTripper(proxyURL)
```

A single `CodexExecutor` serves every codex credential, and each credential can carry its own `ProxyURL`. One client per executor would send requests through the wrong proxy.

Making reuse work therefore requires caching clients **keyed by proxy URL** — which is exactly what #3861 proposed, and which was closed after the review objection that the cache is unbounded with no eviction. I'd rather not re-open that debate inside a leak fix.

### Suggested split

The leak is happening now: one ESTABLISHED socket per request. On a busy deployment I measured the process climbing to **9733 fd / 3.47 GB RSS in ~3h**; with this patch it stays flat at **44–64 fd / ~300 MB** under the same load.

This PR is the minimal correctness fix and deliberately doesn't prejudge the reuse design — it only ensures a connection that currently has no owner gets one. If maintainers want executor-scoped reuse (with a bounded, evicting per-proxy cache), I'm happy to work on it as a follow-up, on top of a codebase that isn't leaking in the meantime.

Happy to adjust if I've misread the intended lifetime of these clients.

PR #4406, opened 2026-07-18
fix(kimi): coerce boolean tool-schema subschemas for MFJS

thebtf:
## Problem

Kimi/Moonshot validates `tools.function.parameters` against **Moonshot Flavored JSON Schema (MFJS)**, a stricter subset that rejects a bare JSON boolean in a subschema slot. Any OpenAI-compatible client whose tool schema carries a boolean subschema makes the upstream **400 the entire turn on the first request**:

```
tools.function.parameters is not a valid moonshot flavored json schema,
details: <At path 'properties.tasks.items.properties': property schema for 'outputSchema' must be an object>
```

A boolean subschema (`true` = accept anything, `false` = accept nothing) is **valid standard JSON Schema (draft-06+)**, so the client request is well-formed — but Kimi rejects it. The failure surfaces as an empty assistant turn with no visible cause, which is hard to diagnose.

Concrete trigger: the [oh-my-pi](https://github.com/can1357/oh-my-pi) `task` delegation tool emits `"outputSchema": true` nested inside its parameters. Selecting any Kimi model through CLIProxyAPI then 400s every turn.

## Fix

Add `sanitizeKimiToolSchemas`, run right after `normalizeKimiToolMessageLinks` on **both** the non-streaming (`Execute`) and streaming (`ExecuteStream`) paths — the single choke point every client route already passes through.

It walks each tool's parameters schema and coerces bare boolean subschemas found in **genuine subschema slots** to their object form:

- `true` → `{}`
- `false` → `{"not": {}}`

Slots walked: `properties`, `patternProperties`, `$defs`/`definitions`, `dependentSchemas`, `anyOf`/`oneOf`/`allOf`/`prefixItems`, `items`, `contains`, `propertyNames`, `if`/`then`/`else`, `not`, `contentSchema`.

Booleans that are **native MFJS features** (`additionalProperties`, `unevaluatedProperties`, `additionalItems`, `unevaluatedItems`) or **plain data** (enum values, `const`, `default`) are left untouched — MFJS accepts those.

The pass is a no-op for bodies without tools and byte-identical for schemas that carry no boolean subschema, so it costs nothing on the common path.

## Tests

`internal/runtime/executor/kimi_executor_test.go` adds coverage for:

- the real oh-my-pi `outputSchema: true` nested case → coerced to `{}`
- `false` → `{"not": {}}`
- `additionalProperties: false` preserved (native MFJS, not coerced)
- booleans inside `anyOf` and `items` coerced
- no-tools body is a no-op
- clean schema stays byte-identical

`go test ./internal/runtime/executor/` passes; `gofmt` clean; `go vet` clean.

### DIFF
--- a/internal/runtime/executor/kimi_executor.go
+++ b/internal/runtime/executor/kimi_executor.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -113,6 +114,10 @@ func (e *KimiExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, req
 	requestPath := helps.PayloadRequestPath(opts)
 	body = helps.ApplyPayloadConfigWithRequest(e.cfg, baseModel, to.String(), from.String(), "", body, originalTranslated, requestedModel, requestPath, opts.Headers)
 	body, err = normalizeKimiToolMessageLinks(body)
+	if err != nil {
+		return resp, err
+	}
+	body, err = sanitizeKimiToolSchemas(body)
 	if err != nil {
 		return resp, err
 	}
@@ -226,6 +231,10 @@ func (e *KimiExecutor) ExecuteStream(ctx context.Context, auth *cliproxyauth.Aut
 	requestPath := helps.PayloadRequestPath(opts)
 	body = helps.ApplyPayloadConfigWithRequest(e.cfg, baseModel, to.String(), from.String(), "", body, originalTranslated, requestedModel, requestPath, opts.Headers)
 	body, err = normalizeKimiToolMessageLinks(body)
+	if err != nil {
+		return nil, err
+	}
+	body, err = sanitizeKimiToolSchemas(body)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


thebtf:
Thanks for the reviews — all three points addressed in 2b5672dc:

1. **Codex (P2, exact numbers):** switched to `json.Decoder` with `UseNumber()`, so `const`/`enum`/numeric bounds (including values above 2^53) round-trip byte-exact instead of going through `float64`. Added `TestSanitizeKimiToolSchemas_PreservesExactNumbersDuringCoercion` covering `const: 9007199254740993` alongside a boolean subschema that forces the rewrite.

2. **Gemini (array-form `items`):** `items` as a tuple (array of subschemas, draft-04/07) is now handled — booleans inside it are coerced too. Extended `TestSanitizeKimiToolSchemas_CoercesBooleanInCombinatorsAndItems` with `"items":[true,{"type":"string"}]`.

3. **Gemini (explicit reassign):** extracted `sanitizeMFJSSubschemaMap` / `sanitizeMFJSSubschemaArray` helpers and reassign mutated maps/slices back to `n[key]` explicitly.

`go test ./internal/runtime/executor/`, `gofmt`, and `go vet` are all clean.

@codex review

PR #4424, opened 2026-07-18
fix(codex): distinguish accounts sharing email and plan

maxdewald:
## Summary

- include the stable ChatGPT account hash in Codex credential filenames for every plan
- retain the existing email/plan filename fallback when the account ID is unavailable
- cover Plus, normalized-plan, and missing-plan account-hash cases

## Problem

Two ChatGPT accounts with the same email and plan currently resolve to the same filename, for example `codex-user@example.com-plus.json`. Logging into the second account overwrites the first credential, so only one account remains available.

Both Codex OAuth login paths already compute and pass a stable hash of the ChatGPT account ID. This change uses that hash consistently instead of restricting it to team-scoped plans.

## Validation

- `go test ./internal/auth/codex`
- `go build -o <temporary-file> ./cmd/server`

### DIFF
--- a/internal/auth/codex/filename.go
+++ b/internal/auth/codex/filename.go
@@ -8,8 +8,8 @@ import (
 
 // CredentialFileName returns the filename used to persist Codex OAuth credentials.
 // When planType is available (e.g. "plus", "team"), it is appended after the email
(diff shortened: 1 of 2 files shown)


maxdewald:
Closing in favor of documenting the problem in an issue first.

PR #4441, opened 2026-07-19
fix(codex): unwrap stringified tool image outputs

wangwq7:
## Problem

Codex clients can send a tool result whose `content` is a JSON array serialized as a string, for example:

```json
{
  "role": "tool",
  "content": "[{\"detail\":\"original\",\"image_url\":\"data:image/png;base64,...\",\"type\":\"input_image\"}]"
}
```

`ConvertOpenAIRequestToCodex` currently treats every string tool result as plain text. This forwards the complete base64 image payload as the `output` string in a Codex `function_call_output`, which can make an otherwise valid request exceed the upstream context window.

## Fix

- Detect stringified arrays containing a recognized image part.
- Reuse the existing structured tool-output conversion path.
- Accept Codex-native `input_image` parts in addition to Chat Completions `image_url` parts.
- Preserve the existing behavior for ordinary text/JSON tool output and invalid structured parts.

This change is limited to the OpenAI Chat Completions -> Codex request translator. It does not involve any vision bridge or provider-specific plugin.

## Tests

- `go test ./internal/translator/...`
- `CGO_ENABLED=1 go build -buildvcs=false ./cmd/server`

### DIFF
--- a/internal/translator/codex/openai/chat-completions/codex_openai_request.go
+++ b/internal/translator/codex/openai/chat-completions/codex_openai_request.go
@@ -472,6 +472,12 @@ func ConvertOpenAIRequestToCodex(modelName string, inputRawJSON []byte, stream b
 func setToolCallOutputContent(funcOutput []byte, content gjson.Result) []byte {
 	switch {
 	case content.Type == gjson.String:
+		// Codex clients may serialize structured tool output into the string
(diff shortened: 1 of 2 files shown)


wangwq7:
The repository path guard rejects external pull requests that modify internal/translator/. I filed the required maintenance issue with the reproduction, root cause, proposed patch, and verification details: #4442.

PR #4459, opened 2026-07-20
fix(streaming): keep the connection alive during the pre-first-chunk wait

dprvda:
## Problem

Streaming handlers call `ExecuteStreamWithAuthManager`, which blocks in `readInitialStreamChunks` until the upstream produces its first payload chunk. A high-effort model reasoning over a large context can take minutes to get there, and the proxy sends the client nothing that whole time, so the client treats the stream as idle and disconnects. The request then dies as `context canceled` (HTTP 500). `ForwardStream` already heartbeats, but only *after* the first chunk, so the pre-first-chunk window is uncovered.

This reproduces reliably with a large (~300k token) resume routed to a Codex/GPT tier at high reasoning effort: ~2-3 minutes to first token, no bytes to the client, client aborts.

## Fix

Add a shared `BootstrapStream` helper (`sdk/api/handlers/stream_bootstrap.go`) that covers that window. `Execute` runs in its own goroutine and delivers its three return values over a channel, while a single `select` loop owns every write to the client: it heartbeats on each tick, binds the real data/error channels when the result lands, and keeps the same ticker running afterwards so a slow or retried upstream stays covered. One writer end to end means no mutex.

The first heartbeat commits the response headers, which spends the HTTP status code, so errors after that point are reported in-band as an SSE `error` event and the raw error body is still appended to the request log, matching what `WriteErrorResponse` records. Requests that resolve before the first tick keep their normal status.

All four streaming handlers now share the helper: claude code, gemini `streamGenerateContent`, openai chat + completions, and openai responses. Gemini disables heartbeats for `alt` streams, which are raw JSON rather than SSE, the same way `forwardGeminiStream` already does.

## Compatibility

Keep-alives stay **off by default** (`StreamingKeepAliveInterval` returns 0). At interval `<= 0` no goroutine is started and no ticker runs, so behaviour is byte-for-byte unchanged. Set `streaming.keepalive-seconds` to enable. The `": keep-alive\n\n"` literal is now the `KeepAliveSSEComment` constant, shared with `ForwardStream`.

## Testing

- `go vet ./sdk/...` clean.
- `go test ./sdk/api/handlers/...` — all packages pass (4 new tests in `handlers_stream_bootstrap_test.go`: heartbeat precedes a delayed first chunk with a 200 status; error after a heartbeat commit is an SSE `error` event with a single WriteHeader; heartbeats continue across the bootstrap-retry wait; post-bind silence keeps heartbeating).
- `go test -race ./sdk/api/handlers/{claude,gemini,openai}/...` clean; the helper's own tests are race-clean.
- Verified end to end against a live upstream: with `keepalive-seconds` set, a heartbeat is delivered during the wait and the stream then completes normally; a large resume that previously died with `context canceled` now succeeds.

### DIFF
--- a/internal/registry/model_definitions.go
+++ b/internal/registry/model_definitions.go
@@ -7,6 +7,7 @@ import (
 )
 
 const (
+	claudeBuiltinOpus5ModelID       = "claude-opus-5"
 	codexBuiltinImage15ModelID      = "gpt-image-1.5"
 	codexBuiltinImageModelID        = "gpt-image-2"
 	xaiBuiltinImageModelID          = "grok-imagine-image"
@@ -32,7 +33,7 @@ type staticModelsJSON struct {
 
 // GetClaudeModels returns the standard Claude model definitions.
 func GetClaudeModels() []*ModelInfo {
-	return cloneModelInfos(getModels().Claude)
+	return WithClaudeBuiltins(cloneModelInfos(getModels().Claude))
 }
 
 // GetGeminiModels returns the standard Gemini model definitions.
@@ -110,6 +111,11 @@ func GetXAIModels() []*ModelInfo {
 	return WithXAIBuiltins(cloneModelInfos(getModels().XAI))
 }
 
+// WithClaudeBuiltins keeps newly released Claude models available while the remote catalog catches up.
+func WithClaudeBuiltins(models []*ModelInfo) []*ModelInfo {
+	return upsertModelInfos(models, claudeBuiltinOpus5ModelInfo())
+}
+
 // WithCodexBuiltins injects hard-coded Codex-only model definitions that should
 // not depend on remote models.json updates. Built-ins replace any matching IDs
 // already present in the provided slice.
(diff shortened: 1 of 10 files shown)


dprvda:
Thanks for the reviews. Addressed in `a78ea1e3`:

- **`new(time.Duration(0))`** — this actually compiles and yields a `*time.Duration` of `0s` (verified on Go 1.26.5, which matches `go.mod`), so it wasn't a build break. Switched to the idiomatic `new(time.Duration)` anyway for clarity.
- **Goroutine capture** — the Execute goroutine now captures a local channel rather than the reassignable `resultChan`.
- **Upstream headers on heartbeat commit (P2)** — when `Execute` returns the upstream headers before the first payload (e.g. the plugin-executor path), the helper now applies them at the heartbeat commit, so a slow first chunk no longer drops passthrough/interceptor headers. On the auth-manager path they're still unknown until the first payload, so `OnFirstChunk` applies them as before.

Verification: `go vet ./sdk/...` clean, `go test -race ./sdk/api/handlers/{claude,gemini,openai}/...` and the `BootstrapStream` tests all pass. Keep-alives remain off by default (`streaming.keepalive-seconds` unset ⇒ no goroutine, no ticker, unchanged behavior).

jroth1111:
#4643 adds opt-in full-stream transient recovery, which intentionally delays the first semantic response while buffering. The latest review correctly notes that configured keep-alives cannot start during that delay under the current handler bootstrap path. This is exactly the scope addressed by #4459’s shared `BootstrapStream` helper. To avoid duplicating or superseding this work, I have marked #4643 draft and documented #4459 as a dependency. If maintainers prefer, I can rebase/stack #4643 after this PR lands, or test an explicit integration branch while preserving the commits and attribution here.

PR #4467, opened 2026-07-21
perf(codex): precompute reasoning replay input fingerprints

neighbads:
## Summary

- Precompute Codex reasoning replay input prefix fingerprints once per request.
- Reuse the precomputed fingerprints while matching replay turns instead of hashing the same prefixes repeatedly.
- Only build the fingerprint table when the request contains reasoning replay fingerprints.

## Root cause

`codexReasoningReplayTurnAnchorIndex` recalculated a SHA-256 fingerprint from `input[0]` for every replay turn and every candidate input index. With large histories, this caused repeated high-order scans and made otherwise fast upstream requests appear to stall inside the proxy.

## Fix

Build all input prefix fingerprints incrementally in one pass and look them up by candidate index during replay matching. The matching behavior remains unchanged; only redundant hashing is removed.

## Benchmark

Large-history reasoning replay benchmark:

| Version | Time/op | Memory/op | Allocs/op |
| --- | ---: | ---: | ---: |
| Before | 79.18 s | 11.45 GB | 2,390,583 |
| After | 11.6-18.4 ms | 3.70 MB | 1,852 |

## Testing

- Added equivalence coverage for precomputed prefix fingerprints.
- Added a large-history reasoning replay benchmark.
- Ran the targeted reasoning replay tests.
- Ran `go test ./...` successfully.

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -500,6 +500,13 @@ func insertCodexReasoningReplayTurns(body []byte, replayItems [][]byte) ([]byte,
 	}
 	inputItems := input.Array()
 	turns := splitCodexReasoningReplayTurns(replayItems)
+	var inputPrefixFingerprints []string
+	for _, turn := range turns {
+		if turn.requestFingerprint != "" {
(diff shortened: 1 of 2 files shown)


neighbads:
force-pushed

PR #16, opened 2025-08-30
Set the default Docker timezone to Asia/Shanghai

hkfires:

PR #4477, opened 2026-07-21
fix(auth): mark auth unavailable for HTTP 400 invalid_grant refresh failures

daiaji:
## Problem

OAuth token endpoints return **HTTP 400** (not 401) for permanent credential errors per RFC 6749 §5.2. x.ai follows this: revoked refresh tokens produce:

```
HTTP 400
{"error":"invalid_grant","error_description":"Refresh token has been revoked"}
```

`refreshAuthForRequest` used `isUnauthorizedError`, which only recognizes HTTP 401, so 400 `invalid_grant` was treated as a transient error. The auth kept `status=active, unavailable=false, failed=0` forever:
- auto-refresh retried indefinitely (every 5min backoff)
- management API consumers (cleanup scripts checking `f.unavailable`) could not detect or delete the dead credential
- dead credentials accumulated in memory with no eviction, eventually causing OOM on memory-constrained hosts

## Fix

Add `isPermanentCredentialFailure` covering RFC 6749 §5.2 permanent shapes:

| Shape | Example | Permanent? |
|---|---|---|
| HTTP 400 + `invalid_grant` | x.ai revoked refresh token | ✅ yes |
| HTTP 401 + `invalid_grant` | providers using 401 | ✅ yes |
| Descriptive "refresh token has been revoked" | non-standard messages | ✅ yes |
| "refresh token is invalid" / "refresh token expired" | descriptive | ✅ yes |
| HTTP 400 + `invalid_request` | malformed request (programming bug) | ❌ no |
| HTTP 400 + `invalid_client` | client config error | ❌ no |
| HTTP 500 / network error | transient | ❌ no |

When triggered, the auth is marked `Unavailable=true`, `Status=StatusError`, `StatusMessage="invalid_grant"`, and `NextRefreshAfter` is cleared so auto-refresh stops retrying.

The 401-only path is preserved via `isUnauthorizedError` for backward compatibility — providers relying on 401 semantics are unaffected.

## Tests

- `TestManager_RefreshAuthInvalidGrant400MarksAuthUnavailable`: verifies x.ai's 400 `invalid_grant` shape marks the auth unavailable and stops auto-refresh scheduling
- `TestIsPermanentCredentialFailure`: table-driven coverage for permanent vs transient error shapes (11 cases)
- Existing `TestManager_RefreshAuthUnauthorizedFailureStopsAutoRefreshRetry` still passes (401 path unchanged)

## Background

Discovered while investigating a deployment with 15,661 xAI auth files (13,914 revoked) causing OOM on a 1GB host. The cleanup script (linuxdo-welfare-collector) depends on `GET /v0/management/auth-files` returning `unavailable=true` for dead credentials, but CPA reported them as `active` due to this bug — making automated cleanup impossible and creating a death spiral where dead credentials accumulate until OOM.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket.go
@@ -1253,6 +1253,17 @@ func responsesWebsocketAuthAvailableForModel(auth *coreauth.Auth, modelName stri
 	if auth.Disabled || auth.Status == coreauth.StatusDisabled {
 		return false
 	}
+	// Permanent credential failure from the refresh path (e.g. refresh
+	// returned 400 invalid_grant) marks the auth Unavailable with a
+	// permanent LastError and zero NextRetryAfter. This must block the
+	// pinned websocket path too — otherwise a dead Codex/xAI pin keeps
+	// being matched by responsesWebsocketPinnedAuthMatchesModel and the
+	// handler keeps sending frames with WithPinnedAuthID, getting
+	// auth_unavailable instead of releasing the pin and falling back to
+	// a healthy auth (codex review).
+	if auth.Unavailable && coreauth.HasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
+		return false
+	}
 	if modelName != "" && len(auth.ModelStates) > 0 {
 		state, ok := auth.ModelStates[modelName]
 		if (!ok || state == nil) && modelName != "" {
--- a/sdk/api/handlers/openai/openai_responses_websocket_test.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket_test.go
@@ -2122,6 +2122,23 @@ func TestResponsesWebsocketPinnedAuthMatchesModel(t *testing.T) {
 	if responsesWebsocketPinnedAuthMatchesModel(unregisteredAuth, modelB, modelA, true) {
 		t.Fatal("Home runtime auth matched a different model")
 	}
+
+	// Permanent credential failure from the refresh path (400 invalid_grant
+	// or 401) marks the auth Unavailable with zero NextRetryAfter. The
+	// websocket pin path must reject it so the dead pin is released and the
+	// handler falls back to a healthy auth, instead of keeping the dead pin
+	// and sending frames that get auth_unavailable (codex review).
+	permanentFailedAuth := auth.Clone()
+	permanentFailedAuth.Unavailable = true
+	permanentFailedAuth.NextRetryAfter = time.Time{}
+	permanentFailedAuth.PermanentRefreshFailure = true
+	permanentFailedAuth.LastError = &coreauth.Error{HTTPStatus: http.StatusBadRequest, Message: "invalid_grant: refresh token revoked"}
+	if responsesWebsocketPinnedAuthMatchesModel(permanentFailedAuth, modelA, modelA, false) {
+		t.Fatal("permanent failed auth matched a model — dead pin must be rejected so it can be released")
+	}
+	if responsesWebsocketAuthAvailableForModel(permanentFailedAuth, modelA, time.Now()) {
+		t.Fatal("permanent failed auth reported available — websocket path must block it")
+	}
 }
 
 func TestWebsocketUpstreamSupportsIncrementalInputForModel(t *testing.T) {
--- a/sdk/cliproxy/auth/auto_refresh_loop.go
+++ b/sdk/cliproxy/auth/auto_refresh_loop.go
@@ -339,7 +339,19 @@ func nextRefreshCheckAt(now time.Time, auth *Auth, interval time.Duration) (time
 	if auth == nil {
 		return time.Time{}, false
 	}
-	if hasUnauthorizedAuthFailure(auth) {
+	// Stop scheduling only for refresh-sourced permanent failures
+	// (Unavailable=true + NextRetryAfter zero — refreshAuthForRequest's
+	// permanent branch sets exactly this combination). Request-side
+	// invalid_grant cooldowns must remain schedulable so proactive
+	// refresh can recover them early (codex review):
+	//   - Single-model all-cooldown: Unavailable=true + non-zero
+	//     NextRetryAfter → falls through.
+	//   - Multi-model partial cooldown: Unavailable=false + zero
+	//     NextRetryAfter (a clean model keeps the auth available) →
+	//     falls through. Without the Unavailable check, the cooled
+	//     model's LastError would unschedule the auth and prevent
+	//     early recovery of the cooled model.
+	if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
 		return time.Time{}, false
 	}
 
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -665,6 +665,44 @@ func clearCooldownStateForAuth(auth *Auth, now time.Time) bool {
 	if auth == nil {
 		return false
 	}
+	// Preserve the refresh-sourced permanent failure marker
+	// (Unavailable + hasPermanentAuthFailure + NextRetryAfter.IsZero()).
+	// clearCooldownStateForAuth is invoked from SetConfig/register/update
+	// paths when disable-cooling is enabled, and historically clears
+	// auth.Unavailable — without this guard, a config reload erases the
+	// permanent marker and the shouldRefresh / nextRefreshCheckAt /
+	// isAuthBlockedForModel guards stop firing, re-queueing revoked
+	// 400 invalid_grant credentials (codex review).
+	//
+	// Transient cooldowns (NextRetryAfter non-zero) are still cleared
+	// below — only the permanent marker is preserved.
+	if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
+		// Clear per-model cooldown state (a permanent refresh failure
+		// already resets these in refreshAuthForRequest, but a stale
+		// model cooldown may have been restored from disk or set by a
+		// racing request before the permanent marker was set).
+		modelChanged := false
+		for _, state := range auth.ModelStates {
+			if state == nil || state.Status == StatusDisabled {
+				continue
+			}
+			if state.Unavailable || !state.NextRetryAfter.IsZero() || state.Quota.Exceeded || !state.Quota.NextRecoverAt.IsZero() {
+				state.Unavailable = false
+				state.NextRetryAfter = time.Time{}
+				state.Quota = QuotaState{}
+				state.UpdatedAt = now
+				modelChanged = true
+			}
+		}
+		if modelChanged && len(auth.ModelStates) > 0 {
+			// Recompute quota aggregation but NOT auth.Unavailable /
+			// NextRetryAfter — updateAggregatedAvailability would
+			// overwrite the permanent marker. Manually clear quota
+			// since permanent failures carry no quota state.
+			auth.Quota = QuotaState{}
+		}
+		return modelChanged
+	}
 	changed := false
 	if auth.Unavailable || !auth.NextRetryAfter.IsZero() || auth.Quota.Exceeded || !auth.Quota.NextRecoverAt.IsZero() {
 		auth.Unavailable = false
@@ -1394,7 +1432,7 @@ func (m *Manager) availableAuthsForRouteModel(auths []*Auth, provider, routeMode
 	}
 
 	availableByPriority := make(map[int][]*Auth)
-	cooldownCount := 0
+	unavailableCount := 0
 	var earliest time.Time
 	for _, candidate := range auths {
 		checkModel := m.selectionModelForAuth(candidate, routeModel)
@@ -1404,8 +1442,8 @@ func (m *Manager) availableAuthsForRouteModel(auths []*Auth, provider, routeMode
 			availableByPriority[priority] = append(availableByPriority[priority], candidate)
 			continue
 		}
+		unavailableCount++
 		if reason == blockReasonCooldown {
-			cooldownCount++
 			if !next.IsZero() && (earliest.IsZero() || next.Before(earliest)) {
 				earliest = next
 			}
@@ -1413,7 +1451,7 @@ func (m *Manager) availableAuthsForRouteModel(auths []*Auth, provider, routeMode
 	}
 
 	if len(availableByPriority) == 0 {
-		if cooldownCount == len(auths) && !earliest.IsZero() {
+		if unavailableCount == len(auths) && !earliest.IsZero() {
 			providerForError := provider
 			if providerForError == "mixed" {
 				providerForError = ""
@@ -3739,7 +3777,22 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 		}
 
 		if result.Success {
-			if result.Model != "" {
+			// If the auth is already marked as a refresh-sourced permanent
+			// failure (Unavailable + permanent LastError + zero NextRetryAfter),
+			// skip the success-state update — both branches below would erase
+			// the permanent marker:
+			//   - result.Model != "": updateAggregatedAvailability recomputes
+			//     Unavailable=false (all model states clean) + clears LastError.
+			//   - result.Model == "": clearAuthStateOnSuccess unconditionally
+			//     clears Unavailable, LastError, NextRetryAfter.
+			// A dead auth should not receive a success result (guards block
+			// routing), but defending against an erroneous MarkResult call
+			// prevents the marker from being silently erased (codex review
+			// on clearCooldownStateForAuth revealed the same class of leak
+			// across all Unavailable=false write paths).
+			if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
+				auth.UpdatedAt = now
+			} else if result.Model != "" {
 				state := ensureModelState(auth, result.Model)
 				resetModelState(state, now)
 				updateAggregatedAvailability(auth, now)
@@ -3755,7 +3808,23 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 				clearAuthStateOnSuccess(auth, now)
 			}
 		} else {
-			if result.Model != "" {
+			// If the auth is already marked as a refresh-sourced permanent
+			// failure (Unavailable + permanent LastError + zero NextRetryAfter),
+			// skip the failure-state update — MarkResult would overwrite the
+			// permanent marker with a transient request cooldown, masking the
+			// dead auth from scheduler/routing guards (codex review on 358d49b2).
+			//
+			// This happens when tryRefreshAfterUnauthorized calls
+			// refreshAuthForRequest and gets a 400 invalid_grant (permanent),
+			// then the caller records the original 401 via MarkResult — without
+			// this skip, the 401 cooldown overwrites the zero-NextRetryAfter
+			// permanent marker and the dead auth is requeued/routable after the
+			// cooldown expires.
+			if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
+				// Only record the attempt count (already done above); skip
+				// failure-state write so the permanent marker stays intact.
+				auth.UpdatedAt = now
+			} else if result.Model != "" {
 				if !isRequestScopedResultError(result.Error) {
 					disableCooling := m.cooldownDisabledForAuth(auth)
 					state := ensureModelState(auth, result.Model)
@@ -4072,6 +4141,7 @@ func clearAuthStateOnSuccess(auth *Auth, now time.Time) {
 		return
 	}
 	auth.Unavailable = false
+	auth.PermanentRefreshFailure = false
 	auth.Status = StatusActive
 	auth.StatusMessage = ""
 	auth.Quota.Exceeded = false
@@ -4144,6 +4214,21 @@ func resultErrorFromError(err error) *Error {
 	return resultErr
 }
 
+// permanentCredentialStatusMessage returns the StatusMessage to record when
+// isPermanentCredentialFailure returns true. It distinguishes invalid_grant
+// from generic unauthorized so operators can see the real reason in the
+// management API.
+func permanentCredentialStatusMessage(err error) string {
+	if err == nil {
+		return "unauthorized"
+	}
+	raw := strings.ToLower(err.Error())
+	if strings.Contains(raw, "invalid_grant") || strings.Contains(raw, "refresh token") {
+		return "invalid_grant"
+	}
+	return "unauthorized"
+}
+
 func isUnauthorizedError(err error) bool {
 	if err == nil {
 		return false
@@ -4155,11 +4240,87 @@ func isUnauthorizedError(err error) bool {
 	return strings.Contains(raw, "status 401") || strings.Contains(raw, "401 unauthorized")
 }
 
-func hasUnauthorizedAuthFailure(auth *Auth) bool {
-	if auth == nil || auth.LastError == nil {
+// isPermanentCredentialFailure reports whether err indicates the refresh
+// token (or underlying credential) is permanently invalid and cannot be
+// recovered by retrying. This covers RFC 6749 §5.2 permanent error codes
+// returned by OAuth token endpoints:
+//
+//   - HTTP 400 with `invalid_grant`: the refresh token is invalid, expired,
+//     revoked, or was issued to another client. x.ai returns this exact
+//     combination (HTTP 400, error=invalid_grant) for revoked refresh tokens.
+//   - HTTP 401 with `invalid_grant`: some providers use 401 instead of 400.
+//   - Descriptive messages without the standard error code (e.g. "refresh
+//     token has been revoked", "refresh token expired").
+//
+// Returning true marks the auth Unavailable + StatusError so it becomes
+// visible to management API consumers (e.g. the welfare collector script's
+// isDeadCpaAuth cleanup) and stops the auto-refresh loop from retrying
+// indefinitely.
+//
+// Note: plain HTTP 401 is still treated as permanent here (preserving legacy
+// behavior) to avoid regressing providers that rely on 401 semantics, even
+// though 401 may sometimes indicate a transient config issue like invalid_client.
+func isPermanentCredentialFailure(err error) bool {
+	if err == nil {
 		return false
 	}
-	return auth.LastError.StatusCode() == http.StatusUnauthorized || strings.EqualFold(auth.LastError.Code, "unauthorized")
+	if isUnauthorizedError(err) {
+		return true
+	}
+	raw := strings.ToLower(err.Error())
+	for _, sub := range permanentCredentialSubstrings {
+		if strings.Contains(raw, sub) {
+			return true
+		}
+	}
+	return false
+}
+
+// permanentCredentialSubstrings are the lowercase substrings that indicate a
+// permanent refresh-token failure when the error does not carry an HTTP 401
+// status (e.g. x.ai's plain fmt.Errorf from the token endpoint).
+var permanentCredentialSubstrings = []string{
+	"invalid_grant",
+	"refresh token has been revoked",
+	"refresh token is invalid",
+	"refresh token expired",
+}
+
+// hasPermanentAuthFailure reports whether auth carries a permanent credential
+// failure sourced from the refresh path (HTTP 401, 400 invalid_grant, or a
+// descriptive revoked/expired message). It replaces the legacy
+// hasUnauthorizedAuthFailure check so the scheduler stops retrying
+// RFC 6749 §5.2 permanent errors that providers return as HTTP 400 instead
+// of 401 (e.g. x.ai revoked refresh tokens).
+//
+// Discriminator: the ONLY signal is auth.PermanentRefreshFailure, which
+// is set exclusively by refreshAuthForRequest's permanent branch. This
+// cleanly distinguishes refresh-sourced permanent failures from
+// disable-cooling transient failures (applyAuthFailureState with
+// disableCooling=true also produces Unavailable=true + zero
+// NextRetryAfter + LastError(401/invalid_grant), but is transient —
+// relying on LastError shape alone would misclassify those as permanent
+// and permanently block routing, codex review).
+//
+// Legacy compat: auths persisted before the PermanentRefreshFailure field
+// existed will have PermanentRefreshFailure=false after restart. They will
+// not be blocked by the guards immediately, but the first refresh attempt
+// will re-detect the permanent failure (refresh returns 400 invalid_grant
+// again) and re-set the marker. This is safer than a LastError-based
+// fallback, which would misclassify disable-cooling request failures as
+// permanent.
+func hasPermanentAuthFailure(auth *Auth) bool {
+	if auth == nil {
+		return false
+	}
+	return auth.PermanentRefreshFailure
+}
+
+// HasPermanentAuthFailure is the exported form of hasPermanentAuthFailure,
+// for cross-package consumers that need to detect the refresh-sourced
+// permanent failure marker (e.g. websocket pin path).
+func HasPermanentAuthFailure(auth *Auth) bool {
+	return hasPermanentAuthFailure(auth)
 }
 
 func refreshErrorFromError(err error) *Error {
@@ -5896,7 +6057,20 @@ func (m *Manager) shouldRefresh(a *Auth, now time.Time) bool {
 	if a == nil {
 		return false
 	}
-	if hasUnauthorizedAuthFailure(a) {
+	// Stop proactive refresh only for refresh-sourced permanent failures
+	// (Unavailable=true + NextRetryAfter zero — refreshAuthForRequest's
+	// permanent branch sets exactly this combination). Request-side
+	// invalid_grant cooldowns must NOT be blocked here:
+	//   - Single-model all-cooldown: updateAggregatedAvailability sets
+	//     Unavailable=true + NextRetryAfter=now+30min (non-zero) → falls
+	//     through to the refresh lead/expiry logic below.
+	//   - Multi-model partial cooldown: updateAggregatedAvailability sets
+	//     Unavailable=false + NextRetryAfter=zero (a clean model keeps the
+	//     auth available). Without the Unavailable check, this guard would
+	//     fire on the cooled model's LastError and unschedule the auth,
+	//     preventing proactive refresh from recovering the cooled model
+	//     early (codex review).
+	if a.Unavailable && hasPermanentAuthFailure(a) && a.NextRetryAfter.IsZero() {
 		return false
 	}
 	if !a.NextRefreshAfter.IsZero() && now.Before(a.NextRefreshAfter) {
@@ -6231,16 +6405,51 @@ func (m *Manager) refreshAuthForRequest(ctx context.Context, id, failedAccessTok
 	log.Debugf("refreshed %s, %s, %v", auth.Provider, auth.ID, err)
 	now := time.Now()
 	if err != nil {
-		unauthorized := isUnauthorizedError(err)
+		permanent := isPermanentCredentialFailure(err)
 		shouldReschedule := false
+		shouldPersist := false
+		var persistClone *Auth
 		m.mu.Lock()
 		if current := m.auths[id]; current != nil {
 			current.LastError = refreshErrorFromError(err)
-			if unauthorized {
+			if permanent {
 				current.NextRefreshAfter = time.Time{}
+				// Clear any stale request cooldown so the NextRetryAfter.IsZero()
+				// discriminators in shouldRefresh, nextRefreshCheckAt, and
+				// isAuthBlockedForModel correctly identify this as a refresh-
+				// sourced permanent failure. Without this, a pre-existing request
+				// cooldown (NextRetryAfter non-zero) would mask the permanent
+				// failure, leaving the dead auth scheduled and routable after the
+				// stale cooldown expires (codex review).
+				current.NextRetryAfter = time.Time{}
 				current.Unavailable = true
+				current.PermanentRefreshFailure = true
 				current.Status = StatusError
-				current.StatusMessage = "unauthorized"
+				current.StatusMessage = permanentCredentialStatusMessage(err)
+				// Clear per-model cooldown state so persistCooldownStates drops
+				// the model .cds records from the snapshot. Otherwise a restart
+				// replays them via RestoreCooldownStates → updateAggregatedAvailability,
+				// which restores a non-zero auth NextRetryAfter from the stale
+				// model cooldowns and masks the permanent marker from the
+				// IsZero() guards (codex review). StatusDisabled states are
+				// preserved — they reflect user/admin intent, not cooldown.
+				// updateAggregatedAvailability is NOT called here because it
+				// would recompute auth.Unavailable=false (all model states are
+				// now clean) and overwrite the permanent marker.
+				for _, state := range current.ModelStates {
+					if state == nil || state.Status == StatusDisabled {
+						continue
+					}
+					resetModelState(state, now)
+				}
+				// Clone under the lock so the persist path below has a stable
+				// snapshot. Persist the permanent marker so it survives process
+				// restart — without this, a restart reloads the auth from
+				// storage without the Unavailable/permanent-LastError marker,
+				// and any stale .cds cooldown record can also be restored,
+				// requeueing the revoked credential (codex review).
+				persistClone = current.Clone()
+				shouldPersist = true
 			} else {
 				current.NextRefreshAfter = now.Add(refreshFailureBackoff)
 			}
@@ -6251,6 +6460,17 @@ func (m *Manager) refreshAuthForRequest(ctx context.Context, id, failedAccessTok
 			}
 		}
 		m.mu.Unlock()
+		if shouldPersist && persistClone != nil {
+			if errPersist := m.persist(ctx, persistClone); errPersist != nil {
+				log.Debugf("persist permanent refresh failure for %s (%s) failed: %v", auth.Provider, auth.ID, errPersist)
+			}
+			// persistCooldownStates writes a full snapshot and removes stale
+			// .cds files not in the snapshot. The permanent marker has
+			// NextRetryAfter=zero, so authCooldownStateRecord returns false
+			// and the dead auth's auth-level cooldown record is dropped.
+			m.persistCooldownStates(ctx)
+			m.hook.OnAuthUpdated(ctx, persistClone)
+		}
 		if shouldReschedule {
 			m.queueRefreshReschedule(id)
 		}
--- a/sdk/cliproxy/auth/conductor_scheduler_refresh_test.go
+++ b/sdk/cliproxy/auth/conductor_scheduler_refresh_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"sync"
 	"testing"
 	"time"
 
@@ -45,11 +46,29 @@ func (e unauthorizedRefreshTestExecutor) Refresh(ctx context.Context, auth *Auth
 	return nil, errors.New("token refresh failed with status 401: invalid_grant")
 }
 
+// invalidGrantRefreshTestExecutor simulates x.ai's real behavior: HTTP 400
+// with invalid_grant for a revoked refresh token. This is RFC 6749 §5.2
+// compliant — token endpoints default to 400, not 401.
+type invalidGrantRefreshTestExecutor struct {
+	schedulerProviderTestExecutor
+}
+
+func (e invalidGrantRefreshTestExecutor) Refresh(ctx context.Context, auth *Auth) (*Auth, error) {
+	return nil, errors.New("xai token request failed with status 400: {\"error\":\"invalid_grant\",\"error_description\":\"Refresh token has been revoked\"}")
+}
+
 func TestManager_RefreshAuthUnauthorizedFailureStopsAutoRefreshRetry(t *testing.T) {
 	ctx := context.Background()
 	manager := NewManager(nil, &RoundRobinSelector{}, nil)
 	manager.RegisterExecutor(unauthorizedRefreshTestExecutor{
 		schedulerProviderTestExecutor: schedulerProviderTestExecutor{provider: "codex"},
+	})
+
+	// Register a refresh lead for codex, mirroring the real runtime where
+	// sdk/auth/refresh_registry.go registers one via init().
+	setRefreshLeadFactory(t, "codex", func() *time.Duration {
+		d := 5 * time.Minute
+		return &d
 	})
 
 	auth := &Auth{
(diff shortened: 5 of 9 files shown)


daiaji:
## Review feedback addressed

Thank you @gemini-code-assist and @codex for the thorough review. All findings are fixed in `b652ba55` (second commit).

### P1 / High — Permanent 400 failures still rescheduled (gemini + codex)

**Confirmed and fixed.** `hasUnauthorizedAuthFailure` only checked `StatusCode()==401 || Code=="unauthorized"`. x.ai returns a plain `fmt.Errorf("xai token request failed with status 400: ...")` with no `StatusCode()` method, so `refreshErrorFromError` stored `LastError` with `HTTPStatus=0, Code=""`. The scheduler (`shouldRefresh` + `nextRefreshCheckAt`) therefore returned `true` via the xai refresh lead registered in `sdk/auth/refresh_registry.go`, re-queuing the revoked credential indefinitely.

Renamed `hasUnauthorizedAuthFailure` → `hasPermanentAuthFailure`, routed through `isPermanentCredentialFailure(auth.LastError)` which matches the `invalid_grant` substring in the error message. Both `shouldRefresh` and `nextRefreshCheckAt` now short-circuit on permanent failures before reaching the refresh-lead logic.

### Medium — Doc/impl contradiction (gemini)

Fixed. The block comment now reads "plain HTTP 401 is still treated as permanent here (preserving legacy behavior)" — matching the `isUnauthorizedError(err) → return true` path.

### Medium — Simplify substring checks (gemini)

Done. Replaced four sequential `strings.Contains` with a loop over `permanentCredentialSubstrings` slice.

### Medium — Missing plain 401 test (gemini)

Added `{"401 plain", errors.New("token refresh failed with status 401"), true}` to `TestIsPermanentCredentialFailure`.

### Test hardening — refresh lead registration

Both scheduler tests (`TestManager_RefreshAuthUnauthorizedFailureStopsAutoRefreshRetry` and `TestManager_RefreshAuthInvalidGrant400MarksAuthUnavailable`) now call `setRefreshLeadFactory` to register a refresh lead for codex/xai. Without this, `ProviderRefreshLead` returns `nil` in the test (the `sdk/auth` package is not imported, so `refresh_registry.go` `init()` does not run), and the tests short-circuit before the refresh-lead logic — hiding the exact regression the review flagged.

### Codex P2 — GHCR image name / tag format

This was on the old commit `db0ef2d6` which included a `ghcr-push.yaml` (since removed from this PR — it is fork-only now, on a separate `fork-ci` branch). The lowercase + slash-sanitization fix is applied there.

daiaji:
## Codex P2 — Preserve Code=unauthorized stop marker

**Confirmed and fixed** in `72ea2973` (third commit).

Great catch. The regression was real:

- `clearUnauthorizedModelStates` (conductor.go:6215) uses a dual-condition check `StatusCode()==401 || Code=="unauthorized"`, proving that `LastError.Code="unauthorized"` without an HTTP status is a valid persisted state shape.
- `CooldownStateRecord.LastError` is `*Error` with `Code`, `Message`, `HTTPStatus` all independently persisted (cooldown_state.go:28). An auth restored from JSON can have `Code="unauthorized"` and `HTTPStatus=0`.
- `(*Error).Error()` returns `"unauthorized: <message>"` — matches neither `isUnauthorizedError` (which looks for `"status 401"` / `"401 unauthorized"`) nor the `invalid_grant`/`"refresh token ..."` substrings.
- So `isPermanentCredentialFailure(auth.LastError)` returned false, and the scheduler would have started re-queuing credentials that the legacy `hasUnauthorizedAuthFailure` used to stop.

### Fix

`hasPermanentAuthFailure` now short-circuits on `Code=="unauthorized"` before falling through to `isPermanentCredentialFailure`, preserving the legacy stop marker:

```go
func hasPermanentAuthFailure(auth *Auth) bool {
    if auth == nil || auth.LastError == nil {
        return false
    }
    if strings.EqualFold(auth.LastError.Code, "unauthorized") {
        return true
    }
    return isPermanentCredentialFailure(auth.LastError)
}
```

This keeps `isPermanentCredentialFailure`'s contract narrow (RFC 6749 §5.2 shapes from the refresh path) while the scheduler predicate handles the `*Error`-specific persisted code field.

### Test

`TestHasPermanentAuthFailure_PreservedFromPersistedState` constructs `LastError={Code:"unauthorized", HTTPStatus:0}` (the restored-from-JSON shape), registers a refresh lead so the scheduler would otherwise try to schedule it, and asserts `hasPermanentAuthFailure` returns true and both `shouldRefresh` + `nextRefreshCheckAt` skip the credential.

daiaji:
## Codex P2 — Remove permanently failed auths from routing

**Confirmed and fixed** in `d1b990e8` (fourth commit).

Excellent catch — this was a real routing leak. The analysis was precise:

### Root cause

`isAuthBlockedForModel` had two gaps:

1. **Model-specific path** (selector.go:312-346): only inspects `auth.ModelStates[model]`. When no model state exists (refresh worker hit 400 invalid_grant before any request recorded a cooldown), it returned `false, blockReasonNone` — **not blocked** — regardless of auth-level `Unavailable`.

2. **Auth-level path** (selector.go:348): required `auth.NextRetryAfter.After(now)`. But `refreshAuthForRequest` only clears `NextRefreshAfter` (a different field, controlling refresh scheduling) and never sets `NextRetryAfter` (controlling request routing). Zero time → `After(now)` = false → **not blocked**.

So a freshly-revoked auth (no prior request failures) stayed in the round-robin/fill-first candidate list and kept receiving traffic until a separate request failure happened.

### Fix

Added a permanent-failure short-circuit at the top of `isAuthBlockedForModel`, after the `Disabled` check and before any model logic:

```go
if auth.Unavailable && hasPermanentAuthFailure(auth) {
    return true, blockReasonOther, time.Time{}
}
```

This covers both the model-specific and auth-level paths in one place. The condition requires **both** `Unavailable` and `hasPermanentAuthFailure` to avoid blocking auths that merely have a transient `LastError` recorded without being marked `Unavailable` (e.g. a 429 quota cooldown where `LastError` is set but `hasPermanentAuthFailure` returns false).

### Test

`TestIsAuthBlockedForModel_PermanentRefreshFailureBlocksRouting` constructs the exact post-refresh permanent-failure shape (`Unavailable=true`, `Status=StatusError`, `LastError` = plain `fmt.Errorf`-derived `Error` with Message only, no `ModelStates` entry), then asserts:

1. `isAuthBlockedForModel(revokedAuth, \"grok-4\", now)` returns `blocked=true` — the model-specific scenario codex flagged.
2. A healthy auth with no model state remains routable (no false positives).
3. `RoundRobinSelector.Pick` over `[revoked, healthy]` returns only the healthy auth — end-to-end routing verification.

daiaji:
## Codex P2 — Block revoked auths even with model states / Keep request cooldowns out of refresh-stop

Both related review comments on commit `d13ce42c` are fixed in `7b335d14` (fifth commit).

### Gap 1 — `len(ModelStates)==0` guard unreliable (selector.go:325)

**Confirmed.** `refreshAuthForRequest`'s permanent-failure branch sets `Unavailable=true` + permanent `LastError` but does NOT clear `ModelStates`. An auth with prior traffic keeps its `ModelStates`, so the `len(ModelStates)==0` guard skipped the block and the revoked auth remained routable.

**Fix:** Replace `len(ModelStates)==0` with `auth.NextRetryAfter.IsZero()`:
- `refreshAuthForRequest` never sets `NextRetryAfter` (only clears `NextRefreshAfter`, a different field) → zero → block fires.
- `MarkResult`/`applyAuthFailureState` always set `NextRetryAfter=now+30min` for request cooldowns → non-zero → falls through to model-state branch.

### Gap 2 — `hasPermanentAuthFailure` stops refresh for request cooldowns (auto_refresh_loop.go:342)

**Confirmed.** `shouldRefresh` and `nextRefreshCheckAt` used `hasPermanentAuthFailure` alone, which matches request-side `invalid_grant` `LastError` text. This stopped proactive refresh for auths in a 30-min request cooldown, preventing early recovery.

**Fix:** Gate the stop on `NextRetryAfter.IsZero()` in both functions. Only refresh-sourced permanent failures (zero `NextRetryAfter`) stop scheduling; request cooldowns (non-zero) remain schedulable for proactive refresh.

### Discriminator reliability

Verified via gitnexus impact analysis — all 8 paths that write `Unavailable=true` were cross-checked against `NextRetryAfter`:
- `refreshAuthForRequest` permanent: never sets `NextRetryAfter` ✓
- `applyAuthFailureState` (all branches except default): sets `NextRetryAfter` ✓
- `MarkResult` model-scoped: `updateAggregatedAvailability` propagates `state.NextRetryAfter` to `auth.NextRetryAfter` ✓
- `restoreCooldownRecordLocked`: rejects zero `NextRetryAfter` records upfront ✓
- `applyAuthFailureState` default (unknown statusCode): theoretical edge case, but `invalid_grant` always routes to the `isInvalidGrantResultError` branch (statusCode 400/401), never default.

### Intentional behavior change (documented + tested)

Persisted unauthorized auths restored via `restoreCooldownRecordLocked` (`Code=unauthorized` + future `NextRetryAfter`) were permanently stuck under the legacy `hasUnauthorizedAuthFailure`. They now remain schedulable and can attempt proactive refresh to recover — if the refresh token is also invalid, the refresh failure sets `NextRetryAfter=zero` and the stop fires on the next `shouldRefresh` call. Covered by `TestShouldRefresh_PersistedUnauthorizedWithCooldownStaysSchedulable`.

### Bonus fix: cooldownCount statistics

gitnexus impact analysis of `isAuthBlockedForModel`'s 10 callers revealed that `collectAvailableByPriority`, `availableAuthsForRouteModel`, and `availabilitySummaryLocked` only counted `blockReasonCooldown` toward the unavailable total. A mix of dead + cooldown auths fell through to `auth_unavailable` instead of returning `modelCooldownError` with the cooldown auths' earliest retry time. Now `blockReasonOther` (permanent) also counts toward unavailable; `earliest` is still updated only from cooldown auths.

### Tests added
- `TestShouldRefresh_RequestInvalidGrantCooldownStaysSchedulable`
- `TestShouldRefresh_PersistedUnauthorizedWithCooldownStaysSchedulable`
- `TestSelectorPick_DeadPlusCooldownReturnsModelCooldownError`
- `TestSelectorPick_AllDeadReturnsAuthUnavailable`
- Updated `TestIsAuthBlockedForModel_RequestInvalidGrantCooldownRecovers` to set `auth.NextRetryAfter` (mirrors real `updateAggregatedAvailability` behavior)

daiaji:
## Codex P2 — Clear stale retry cooldowns on permanent refresh failure

**Confirmed and fixed** in `aad3f7ec` (seventh commit).

Excellent catch — this was a real gap in the `NextRetryAfter.IsZero()` discriminator design.

### Root cause

`shouldRefresh` and `nextRefreshCheckAt` do NOT gate on `NextRetryAfter` — only on `NextRefreshAfter` (refresh backoff). So a proactive refresh can fire during a request cooldown window. When it returns `invalid_grant`:

1. The permanent branch set `Unavailable=true` + permanent `LastError` but did NOT clear `NextRetryAfter`
2. The stale non-zero `NextRetryAfter` (from the pre-existing request cooldown) masked the permanent failure
3. All three guards (`shouldRefresh`, `nextRefreshCheckAt`, `isAuthBlockedForModel`) saw `NextRetryAfter.IsZero()=false` and treated it as a request cooldown
4. The dead auth remained scheduled and routable after the stale cooldown expired

### Fix

Explicitly clear `NextRetryAfter` in `refreshAuthForRequest`'s permanent branch:

```go
if permanent {
    current.NextRefreshAfter = time.Time{}
    current.NextRetryAfter = time.Time{}  // ← new
    current.Unavailable = true
    ...
}
```

### Verification via gitnexus

Ran `cypher` query for `ACCESSES {reason: 'read'}` edges on `NextRetryAfter` — found **24 read points** across 8 functions. Verified each handles zero `NextRetryAfter` correctly:

| Consumer | Behavior with zero NextRetryAfter | Correct? |
|---|---|---|
| `shouldRefresh` / `nextRefreshCheckAt` / `isAuthBlockedForModel` | `IsZero()` discriminators fire → permanent block | ✓ (target) |
| `authCooldownStateRecord` | Returns false → auth-level cooldown NOT persisted | ✓ (permanent ≠ cooldown) |
| `restoreCooldownRecordLocked` | Rejects zero records upfront | ✓ |
| `clearCooldownStateForAuth` | `Unavailable \|\| !IsZero()` — OR logic, `Unavailable=true` satisfies regardless | ✓ pre-existing |
| `buildErrorEventAuthStatus` | `timePtrIfSet(zero)` = nil | ✓ correct semantics |
| `buildAuthFromFileData` | Copies existing value (zero) | ✓ pre-existing |
| `modelCooldownStateRecord` / `modelStateIsClean` | Read `state.NextRetryAfter` (not `auth.NextRetryAfter`) | ✓ unaffected |

### Known behavior change (acceptable)

`authCooldownStateRecord` returns false for zero `NextRetryAfter` → auth-level cooldown record (including `LastError`) is NOT persisted. On restart, `LastError` is lost and the dead auth "revives" until the next proactive refresh re-marks it permanent. This is:
- **Self-correcting** — proactive refresh cycle (minutes) re-marks via 400 → permanent branch
- **No OOM** — core goal achieved
- **Analogous to 401 path** — 401 permanent failures also "revive" after cooldown expires

### Test added

`TestManager_RefreshAuthPermanentFailureClearsStaleRequestCooldown` — registers an auth with pre-existing request cooldown (`NextRetryAfter=now+30min` + `ModelState.NextRetryAfter`), triggers `refreshAuth`, asserts:
- `NextRetryAfter` cleared to zero
- `shouldRefresh` returns false (stop fires)
- `nextRefreshCheckAt` returns false (unscheduled)
- `isAuthBlockedForModel` returns `blockReasonOther` (short-circuit fires)

daiaji:
## Gemini Code Assist — All 4 inline comments addressed

Looking back at gemini-code-assist's first review (4 inline comments), confirming all are resolved across the commit chain:

### 1. High — Permanent 400 failures still rescheduled (conductor.go:4193)
**Fixed in `b652ba55` (second commit).** Renamed `hasUnauthorizedAuthFailure` → `hasPermanentAuthFailure`, updated `shouldRefresh` and `nextRefreshCheckAt` to use it. This was the core scheduling gap — already replied in the first "Review feedback addressed" comment.

### 2. Medium — Contradiction between doc and implementation (conductor.go:4193)
**Fixed in `12b59efc` (first commit).** The block comment now reads:
```go
// Note: plain HTTP 401 is still treated as permanent here (preserving legacy
// behavior) to avoid regressing providers that rely on 401 semantics, even
// though 401 may sometimes indicate a transient config issue like invalid_client.
```
Matches your suggested wording — resolves the doc/impl contradiction.

### 3. Medium — Simplify substring checks (conductor.go:4206)
**Fixed in `b652ba55` (second commit).** Replaced sequential `if strings.Contains` with a `permanentCredentialSubstrings` slice + range loop:
```go
var permanentCredentialSubstrings = []string{
    "invalid_grant",
    "refresh token has been revoked",
    "refresh token is invalid",
    "refresh token expired",
}
```
Matches your suggested refactoring.

### 4. Medium — Missing test case for plain 401 errors (test:188)
**Fixed in `b652ba55` (second commit).** Added the `401 plain` test case to `TestIsPermanentCredentialFailure`:
```go
{"401 plain", errors.New("token refresh failed with status 401"), true},
```
Matches your suggested test case.

daiaji:
## Codex P2 — Keep partial model cooldowns schedulable

**Confirmed and fixed** in `358d49b2` (eighth commit).

Excellent catch — this was a real gap in the guard design.

### Root cause

`MarkResult`'s model-scoped failure path (conductor.go:3768) copies `invalid_grant` onto `auth.LastError` but does NOT set `auth.Unavailable` — that's decided by `updateAggregatedAvailability` based on `allUnavailable`. When a multi-model auth has one cooled model and one clean model:

1. `updateAggregatedAvailability` sets `auth.Unavailable=false` + `auth.NextRetryAfter=zero` (clean model keeps the auth available)
2. `auth.LastError` still carries `invalid_grant` (from MarkResult:3768)
3. The guards in `shouldRefresh` and `nextRefreshCheckAt` keyed only on `hasPermanentAuthFailure && NextRetryAfter.IsZero()` — without checking `Unavailable`
4. The zero `NextRetryAfter` + permanent `LastError` combination unscheduled the auth from the auto-refresh queue
5. Proactive refresh could not recover the cooled model early

### Fix

Add `auth.Unavailable` to the guard condition in `shouldRefresh` and `nextRefreshCheckAt`:

```go
// before (buggy):
if hasPermanentAuthFailure(a) && a.NextRetryAfter.IsZero() {
    return false
}

// after (fixed):
if a.Unavailable && hasPermanentAuthFailure(a) && a.NextRetryAfter.IsZero() {
    return false
}
```

`isAuthBlockedForModel` already had the `Unavailable` check (selector.go:348). Now all three guards are consistent.

### Verification via gitnexus

Ran `cypher` query for `ACCESSES {reason: 'write'}` edges on `LastError` — found **5 write points**. Cross-verified each against `Unavailable` writes:

| Writer | Sets Unavailable=true? | Notes |
|---|---|---|
| `applyAuthFailureState` | ✓ (line 4656) | Always paired |
| `restoreCooldownRecordLocked` | ✓ (line 638) | Always paired |
| `MarkResult` (model-scoped) | ✗ (delegates to `updateAggregatedAvailability`) | **The only "separated" path** — source of this bug |
| `clearAuthStateOnSuccess` | Sets `LastError=nil` | Not relevant |
| `resetModelState` | Operates on `ModelState`, not `Auth` | Not relevant |

`MarkResult:3768` is the only path that writes `auth.LastError` without setting `auth.Unavailable` — confirming the `Unavailable` check is the correct discriminator.

### Guard behavior matrix

| Scenario | Unavailable | NextRetryAfter | LastError permanent | Guard fires? | Correct? |
|---|---|---|---|---|---|
| Refresh permanent failure | true | zero | true | **yes** → stop | ✓ |
| Request single-model all cooldown | true | non-zero | true | no → falls through | ✓ |
| Request multi-model partial cooldown | **false** | zero | true | **no** → falls through | ✓ (fixed) |
| Persisted unauthorized with future cooldown | true | non-zero | true | no → falls through | ✓ |

### Tests

- **New**: `TestShouldRefresh_PartialModelCooldownStaysSchedulable` — multi-model auth with one cooled model + one clean model, verifies `shouldRefresh` and `nextRefreshCheckAt` keep scheduling.
- **Updated**: `TestHasPermanentAuthFailure_PreservedFromPersistedState` — added `Unavailable=true` to the auth fixture to mirror real `restoreCooldownRecordLocked` behavior (which sets `Unavailable=true` at line 638). The previous fixture had `Unavailable=false` which only happens in the partial-cooldown case — now fixed to reflect the actual restore path.

daiaji:
## Codex P2 — Preserve permanent refresh failures after request retries

**Confirmed and fixed** in `d6e72a65` (ninth commit).

Excellent catch — this was a real gap in the `NextRetryAfter.IsZero()` discriminator design, spanning the `tryRefreshAfterUnauthorized` → `refreshAuthForRequest` → `MarkResult` call chain.

### Root cause

`tryRefreshAfterUnauthorized` (4 callers: `executeStreamWithModelPool` x2, `executeMixedOnce`, `executeCountMixedOnce`) calls `refreshAuthForRequest` on a 401. When refresh returns a permanent failure (400 invalid_grant or 401):

1. `refreshAuthForRequest`'s permanent branch sets `Unavailable=true` + `NextRetryAfter=zero` + permanent `LastError`
2. Returns `(nil, errRefresh)` → `tryRefreshAfterUnauthorized` returns `(auth, false)`
3. Caller keeps `okRefresh=false` and records the **original 401** via `MarkResult`
4. `MarkResult`'s failure branch writes model-scoped cooldown (`state.NextRetryAfter=now+30min`) or auth-level cooldown (`applyAuthFailureState`)
5. `updateAggregatedAvailability` propagates `state.NextRetryAfter` to `auth.NextRetryAfter` (non-zero)
6. The guards `Unavailable && hasPermanentAuthFailure && NextRetryAfter.IsZero()` now return false — the dead auth is requeued/routable after the cooldown expires

### Fix

In `MarkResult`'s failure branch, detect the refresh-sourced permanent failure shape and skip the failure-state write:

```go
} else { // failure branch
    if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
        // Refresh-sourced permanent failure: skip failure-state update
        // to preserve the permanent marker. MarkResult would overwrite
        // it with a transient request cooldown, masking the dead auth
        // from scheduler/routing guards.
        auth.UpdatedAt = now
    } else if result.Model != "" {
        // ... existing model-scoped failure path
    } else {
        // ... existing auth-level failure path
    }
}
```

The attempt count (success/fail) is still recorded before this branch. The permanent marker (Unavailable + zero NextRetryAfter + permanent LastError) stays intact.

### Verification via gitnexus

- `context(tryRefreshAfterUnauthorized)` — confirmed 4 callers, all follow the same `refresh → false → MarkResult` pattern
- `impact(MarkResult)` — MEDIUM risk, 6 direct callers, 0 affected processes. The early-skip only affects the failure branch when the auth is already a refresh-sourced permanent failure; success path and other failure scenarios are untouched.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New**: `TestMarkResult_PreservesRefreshPermanentFailureMarker` — simulates the full chain: `refreshAuth` (permanent failure) → `MarkResult` (401), asserts the permanent marker survives (Unavailable + zero NextRetryAfter + hasPermanentAuthFailure) and all three guards still fire.
- **Updated**: `TestManager_Execute_UnauthorizedRefreshFailureFallsBackToNextAuth` — the refresh error is 401 (permanent), so it now hits the early-skip. Updated assertions from "primary model suspended" (pre-existing behavior that was actually the bug) to "primary auth permanently failed" (permanent marker preserved). The fallback-to-backup behavior (the test's primary intent) is unchanged.

daiaji:
## Codex P2 — Persist permanent refresh failures durably + Block permanently failed websocket pins

Both review comments on commit `d6e72a65` are fixed in `87b7fa62` (tenth commit).

### 1. Persist permanent refresh failures durably

**Confirmed.** The permanent branch in `refreshAuthForRequest` only mutated in-memory auth + scheduler state — it never called `m.persist` nor `m.persistCooldownStates`. After a process restart the auth was loaded from storage without the `Unavailable`/permanent-`LastError` marker, and any stale `.cds` cooldown record could also be restored, requeueing the revoked credential.

**Fix.** Clone the auth under the lock in the permanent branch, then after unlock call:
- `m.persist(ctx, persistClone)` — persists the auth record (`Unavailable=true` + permanent `LastError` + zero `NextRetryAfter`)
- `m.persistCooldownStates(ctx)` — writes a full cooldown snapshot and removes stale `.cds` files not in the snapshot. The permanent marker has `NextRetryAfter=zero`, so `authCooldownStateRecord` returns false and the dead auth's auth-level cooldown record is dropped from the snapshot, which causes `Save` → `removeStaleStateFiles` to delete the stale `.cds`.
- `m.hook.OnAuthUpdated(ctx, persistClone)` — notifies downstream observers (management UI, plugin host) of the permanent failure.

The transient branch is unchanged — transient refresh failures only update in-memory `NextRefreshAfter` and don't need to survive restart.

### 2. Block permanently failed websocket pins

**Confirmed.** `responsesWebsocketAuthAvailableForModel` only rejected `Unavailable` auths when `NextRetryAfter` was non-zero and in the future (`sdk/api/handlers/openai/openai_responses_websocket.go:1274`). The new zero-`NextRetryAfter` permanent marker was not honored, so a pinned Codex/xAI websocket auth marked permanently failed by refresh kept being matched by `responsesWebsocketPinnedAuthMatchesModel` — the handler kept sending frames with `WithPinnedAuthID`, getting `auth_unavailable` instead of releasing the pin and falling back to a healthy auth.

**Fix.** Added the same `Unavailable && HasPermanentAuthFailure && NextRetryAfter.IsZero()` short-circuit at the top of `responsesWebsocketAuthAvailableForModel` (before the model-state branch). This covers both entry points — `responsesWebsocketPinnedAuthMatchesModel` (pin path, line 1151) and `responsesWebsocketAuthMatchesModel` (normal match, line 1239) — since both delegate to this function.

`HasPermanentAuthFailure` is now exported as the cross-package helper (it was previously unexported `hasPermanentAuthFailure`, only used within `sdk/cliproxy/auth`).

### Verification via gitnexus

- `impact(refreshAuthForRequest)` — LOW, 2 direct callers, 0 affected processes
- `impact(responsesWebsocketAuthAvailableForModel)` — LOW, 2 direct callers, 1 process (ResponsesWebsocket). The short-circuit only adds a false return for the permanent marker shape; existing cooldown/disabled/unregistered paths are unchanged.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New** `TestRefreshAuthForRequest_PersistsPermanentFailureMarker`: uses `countingStore` + `recordingCooldownStateStore` + `authUpdateCaptureHook` to verify all three side effects (auth persist, cooldown persist, hook notify) fire on the permanent branch.
- **Extended** `TestResponsesWebsocketPinnedAuthMatchesModel`: new permanent-failed auth case verifies both `responsesWebsocketPinnedAuthMatchesModel` and `responsesWebsocketAuthAvailableForModel` reject the dead pin.

daiaji:
## Codex P2 — Clear model cooldowns when marking permanent refresh failure

**Confirmed and fixed** in `7b130b4d` (eleventh commit).

Excellent catch — this was a real restart-time leak in the permanent marker persistence.

### Root cause

The permanent branch in `refreshAuthForRequest` cleared only the auth-level `NextRetryAfter`; any `current.ModelStates` with future `NextRetryAfter` remained on `persistClone` and in the snapshot saved by `persistCooldownStates`:

- `modelCooldownStateRecord` (conductor.go:930) includes a model record whenever `state.Unavailable && !state.NextRetryAfter.IsZero() && state.NextRetryAfter.After(now)` — the permanent branch did not touch any of these fields on model states.
- On restart, `RestoreCooldownStates` → `restoreCooldownRecordLocked` replays those model `.cds` records → `updateAggregatedAvailability` recomputes `auth.NextRetryAfter = earliestRetry` (non-zero, from the stale model cooldown).
- The guards `Unavailable && hasPermanentAuthFailure && NextRetryAfter.IsZero()` now return false — the revoked credential can be refreshed/routed again once the stale model cooldown expires.

### Fix

In the permanent branch, iterate `current.ModelStates` and call `resetModelState` on each (clears `Unavailable`, `Status`→`Active`, `StatusMessage`, `NextRetryAfter`, `LastError`, `Quota`). `StatusDisabled` states are preserved — they reflect user/admin intent, not cooldown.

`updateAggregatedAvailability` is **NOT** called after clearing because it would recompute `auth.Unavailable = false` (all model states are now clean) and overwrite the permanent marker. The auth-level `Unavailable=true` + zero `NextRetryAfter` is set explicitly above the model-state loop and must stay intact.

### Effect on persistCooldownStates snapshot

After the fix, both record types are dropped from the snapshot:
- **auth-level**: `authCooldownStateRecord` returns false (`NextRetryAfter=zero`)
- **model-level**: `modelCooldownStateRecord` returns false (`state.Unavailable=false` after reset)

`Save` → `removeStaleStateFiles` deletes the stale `.cds` files. Restart has no model cooldown records to replay, so `updateAggregatedAvailability` never restores a non-zero auth `NextRetryAfter`, and the `IsZero()` guards stay effective.

### Verification via gitnexus

- `impact(refreshAuthForRequest)` — LOW, 2 direct callers, 0 affected processes
- `detect_changes` — low risk, 0 affected processes

### Test

**New** `TestRefreshAuthForRequest_PermanentFailureClearsModelCooldowns`:
1. Seeds a `grok-4` 429 cooldown via `MarkResult` (which calls `persistCooldownStates`)
2. Verifies the cooldown snapshot contains the `grok-4` record (precondition)
3. Triggers permanent refresh failure
4. Verifies the model state is reset (`Unavailable=false`, `NextRetryAfter=zero`, `LastError=nil`)
5. Verifies the cooldown snapshot no longer contains the `grok-4` record

daiaji:
## Commit history squashed

The 11 incremental fix commits have been squashed into a single commit `6f555cbf` for clarity. The incremental commits reflected a snowball review process (each fix exposing the next layer), but the final design is a coherent whole — keeping the history would mislead reviewers into thinking the layers were independently designed.

**What the squashed commit contains**: the complete fix as described across all previous review replies — permanent failure marker (`Unavailable + hasPermanentAuthFailure + NextRetryAfter.IsZero()`), three guard sites (`shouldRefresh`, `nextRefreshCheckAt`, `isAuthBlockedForModel`), `refreshAuthForRequest` permanent branch (clear auth + model cooldowns, persist, notify hooks), `MarkResult` early-skip to preserve the marker across request retries, websocket pin path short-circuit, scheduler/selector availability accounting, and 15 tests covering all paths.

All CI passes. No code changes vs. the cumulative state of the previous 11 commits — this is a pure history rewrite.

daiaji:
## Codex P2 — Preserve permanent failures across disable-cooling

**Confirmed and fixed** — amended into the squashed commit `1f046aff`.

Excellent catch — this was a real config-reload leak.

### Root cause

`clearCooldownStateForAuth` (conductor.go:664) is called from 4 sites when `disable-cooling` is enabled globally, per-provider, or on the auth:
- `SetConfig` (line 547) — config hot-reload
- `RestoreCooldownStates` (line 752) — startup restore
- `Register` (line 2169) — new auth registration
- `Update` (line 2217) — auth update

The function unconditionally cleared `auth.Unavailable = false` (line 670). Since the permanent failure marker is `Unavailable=true + hasPermanentAuthFailure + NextRetryAfter.IsZero()`, a config reload that triggered `clearCooldownStateForAuth` would erase the marker — the `shouldRefresh`, `nextRefreshCheckAt`, and `isAuthBlockedForModel` guards would stop firing, and the revoked 400 invalid_grant credential would become routable and re-queued.

### Fix

Added a permanent-marker guard at the top of `clearCooldownStateForAuth`:

```go
if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
    // Clear per-model cooldown state (a permanent refresh failure
    // already resets these in refreshAuthForRequest, but a stale
    // model cooldown may have been restored from disk or set by a
    // racing request before the permanent marker was set).
    // ... clear model states, clear auth.Quota ...
    // Do NOT touch auth.Unavailable / auth.NextRetryAfter / auth.LastError.
    return modelChanged
}
// Existing transient-cooldown clearing below is unchanged.
```

The guard:
- **Preserves** `auth.Unavailable`, `auth.NextRetryAfter` (zero), `auth.LastError`, `auth.Status`, `auth.StatusMessage` — the permanent marker stays intact.
- **Still clears** stale per-model cooldown state (a permanent refresh failure resets these in `refreshAuthForRequest`, but a stale model cooldown may have been restored from disk or set by a racing request before the permanent marker was set).
- **Clears `auth.Quota`** manually instead of calling `updateAggregatedAvailability` — that function would recompute `auth.Unavailable=false` (all model states now clean) and overwrite the permanent marker.
- **Transient cooldowns** (NextRetryAfter non-zero, no permanent LastError) fall through to the existing clearing logic — unchanged.

### Verification via gitnexus

- `impact(clearCooldownStateForAuth)` — **HIGH** risk, 4 direct callers, 16 impacted symbols, 0 affected processes. The guard is a pure preservation check at the top of the function; the transient-clearing path is unchanged. All 4 callers benefit from the guard automatically.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New** `TestClearCooldownStateForAuth_PreservesPermanentFailureMarker`: auth with permanent marker + stale grok-4 model cooldown → `clearCooldownStateForAuth` → verifies marker preserved (Unavailable, zero NextRetryAfter, hasPermanentAuthFailure), model cooldown cleared, all three guards still fire.
- **New** `TestClearCooldownStateForAuth_StillClearsTransientCooldown`: auth with transient cooldown (NextRetryAfter non-zero, no permanent LastError) → `clearCooldownStateForAuth` → verifies Unavailable cleared, NextRetryAfter cleared — the guard only protects permanent markers, not transient cooldowns.

daiaji:
## Codex P2 — Preserve permanent failures across disable-cooling

**Confirmed and fixed** — amended into the squashed commit `2d88e84d`.

Your review prompted a full `ACCESSES write` audit on the `Unavailable` field via gitnexus, which revealed **two** write paths that erase the permanent marker, not just the one you flagged. Both are fixed in this amend.

### Full ACCESSES write audit (9 sites)

| # | writer | line | semantics | erases marker? | status |
|---|---|---|---|---|---|
| 1 | `restoreCooldownRecordLocked` | 616 | set (restore) | no (only writes when NextRetryAfter non-zero) | safe |
| 2 | `clearCooldownStateForAuth` | 663 | clear | **yes** | **fixed** (your finding) |
| 3 | `MarkResult` failure branch | 3712 | set/clear | yes (zero→non-zero NextRetryAfter) | already guarded (early-skip) |
| 4 | `resetModelState` | 3956 | clear (model state) | no (only model state, not auth-level) | safe |
| 5 | `updateAggregatedAvailability` | 3985 | recompute | **yes** (recomputes Unavailable=false when all states clean) | **exposed via MarkResult success branch** — fixed |
| 6 | `clearAggregatedAvailability` | 4056 | clear | yes (unconditional) | only called from `updateAggregatedAvailability` — same fix covers it |
| 7 | `clearAuthStateOnSuccess` | 4085 | clear | **yes** (unconditional clear of Unavailable + LastError) | **exposed via MarkResult success branch** — fixed |
| 8 | `applyAuthFailureState` | 4710 | set | indirect (zero→non-zero NextRetryAfter) | already guarded by MarkResult early-skip |

### Fix 1 (your finding): `clearCooldownStateForAuth` guard

Added a permanent-marker guard at the top of `clearCooldownStateForAuth`:
- **Preserves** `Unavailable`, `NextRetryAfter` (zero), `LastError`, `Status`, `StatusMessage`.
- **Still clears** stale per-model cooldown state (a permanent refresh failure resets these in `refreshAuthForRequest`, but a stale model cooldown may have been restored from disk or set by a racing request).
- **Clears `auth.Quota`** manually instead of calling `updateAggregatedAvailability` (which would recompute `Unavailable=false` and overwrite the marker).
- **Transient cooldowns** (NextRetryAfter non-zero, no permanent LastError) fall through to the existing clearing logic — unchanged.

This covers all 4 callers (`SetConfig`, `RestoreCooldownStates`, `Register`, `Update`).

### Fix 2 (found during the audit): `MarkResult` success branch early-skip

The success branch of `MarkResult` erases the marker via two paths:
- `result.Model != ""` → `updateAggregatedAvailability` recomputes `Unavailable=false` + `clearAuthStateOnSuccess` clears `LastError`.
- `result.Model == ""` → `clearAuthStateOnSuccess` unconditionally clears `Unavailable`, `LastError`, `NextRetryAfter`.

A dead auth should not receive a success result (guards block routing), but defending against an erroneous `MarkResult(success=true)` call prevents the marker from being silently erased — same defensive pattern as the failure-branch early-skip.

Added a symmetric early-skip at the top of the success branch:
```go
if auth.Unavailable && hasPermanentAuthFailure(auth) && auth.NextRetryAfter.IsZero() {
    auth.UpdatedAt = now
} else if result.Model != "" {
    // ... existing path
} else {
    clearAuthStateOnSuccess(auth, now)
}
```

### Verification via gitnexus

- `impact(clearCooldownStateForAuth)` — HIGH risk, 4 direct callers, 16 impacted symbols, 0 affected processes. The guard is a pure preservation check; transient-clearing path unchanged.
- `impact(MarkResult)` — MEDIUM risk, 6 direct callers, 0 affected processes. The success-branch early-skip only fires when the auth is already a refresh-sourced permanent failure; normal success path unchanged.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New** `TestClearCooldownStateForAuth_PreservesPermanentFailureMarker`: permanent marker + stale model cooldown → `clearCooldownStateForAuth` → marker preserved, model cooldown cleared, all three guards still fire.
- **New** `TestClearCooldownStateForAuth_StillClearsTransientCooldown`: transient cooldown (no permanent LastError) → still cleared normally — the guard only protects permanent markers.
- **New** `TestMarkResult_PreservesPermanentFailureMarkerOnSuccess`: two sub-cases (with Model / without Model) verifying the success branch early-skip preserves the marker and guards still fire.

### Process note

Running the full `ACCESSES write` audit before implementing the fix (rather than patching only the flagged site) caught the second leak in the same pass. This is the "build the full state-contract view before editing" discipline — a single-site fix would have left `MarkResult(success)` as a latent leak for the next review round.

daiaji:
## Codex P2 — Do not treat disable-cooling request failures as permanent

**Confirmed and fixed** — amended into the squashed commit `07e0e0ee`.

Excellent catch — this was a real misclassification risk that would have broken disable-cooling deployments.

### Root cause

`applyAuthFailureState` with `disableCooling=true` produces the exact same shape as a refresh-sourced permanent failure:

| field | disable-cooling request 401 | refresh permanent 400 invalid_grant |
|---|---|---|
| `Unavailable` | `true` | `true` |
| `NextRetryAfter` | `zero` (disable-cooling skips cooldown) | `zero` (permanent branch clears it) |
| `LastError` | 401 / invalid_grant | 400 invalid_grant |
| `hasPermanentAuthFailure` | **true** (LastError-based fallback matched) | true |

The `hasPermanentAuthFailure` predicate relied on `LastError` shape (`Code=="unauthorized"` or `isPermanentCredentialFailure(LastError)`) as a fallback for legacy persisted auths. That fallback could not distinguish "refresh-sourced permanent" from "disable-cooling transient" — both carry 401/invalid_grant LastError.

Result: under disable-cooling, a normal 401 request failure would be misclassified as permanent → `isAuthBlockedForModel` short-circuits → routing permanently blocked → `shouldRefresh`/`nextRefreshCheckAt` skip → the auth is dead forever, even though disable-cooling is intended to keep it immediately retryable.

### Fix: explicit `PermanentRefreshFailure` discriminator field

Added a new `Auth.PermanentRefreshFailure bool` field (json: `permanent_refresh_failure,omitempty`), set **exclusively** by `refreshAuthForRequest`'s permanent branch. `hasPermanentAuthFailure` now checks only this field:

```go
func hasPermanentAuthFailure(auth *Auth) bool {
    if auth == nil {
        return false
    }
    return auth.PermanentRefreshFailure
}
```

The `LastError`-based fallback (`Code=="unauthorized"`, `isPermanentCredentialFailure`) was **removed** — it was the source of the misclassification. This is a deliberate trade-off:

- **Auths persisted before this field existed** will have `PermanentRefreshFailure=false` after restart → not immediately blocked by guards → retried once by refresh → if refresh fails permanently again, the marker is re-set. This is safer than the fallback, which would misclassify disable-cooling request failures as permanent.
- **Disable-cooling request failures** (`applyAuthFailureState`) never set `PermanentRefreshFailure` → correctly classified as transient → routing/refresh proceed normally.

### Write-site discipline

`PermanentRefreshFailure` is set `true` in exactly one place:
- `refreshAuthForRequest` permanent branch (conductor.go:6431)

It is cleared `false` in:
- `clearAuthStateOnSuccess` (success path) — added in this amend
- `refreshAuthForRequest` success branch (via the updated auth returned by `exec.Refresh`, which goes through `Update` → `clearAuthStateOnSuccess` on success)

It is **preserved** (not cleared) in:
- `clearCooldownStateForAuth` permanent-marker guard (already in place)
- `MarkResult` success/failure early-skip (already in place)

### Verification via gitnexus

- `impact(hasPermanentAuthFailure)` — MEDIUM risk, 5 direct callers (`shouldRefresh`, `nextRefreshCheckAt`, `isAuthBlockedForModel`, `MarkResult`, `HasPermanentAuthFailure`), 41 impacted symbols, 0 affected processes. The predicate change is a pure narrowing: it returns `true` for fewer auths (only those with `PermanentRefreshFailure=true`), so all guard sites become less aggressive — no new blocking.
- `impact(applyAuthFailureState)` — LOW risk, 1 direct caller, 0 affected processes. `applyAuthFailureState` is unchanged; it simply does not set `PermanentRefreshFailure`, which is the correct behavior.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New** `TestHasPermanentAuthFailure_DisableCoolingTransientNotPermanent`: disable-cooling 401 shape (Unavailable + zero NextRetryAfter + LastError 401, no PermanentRefreshFailure) → `hasPermanentAuthFailure` returns false → `isAuthBlockedForModel` does NOT short-circuit. Contrast: same auth WITH `PermanentRefreshFailure=true` → permanent.
- **New** `TestHasPermanentAuthFailure_LegacyPersistedAuthNotImmediatelyPermanent`: legacy auth (Code=unauthorized / invalid_grant LastError, no PermanentRefreshFailure) → NOT permanent → will be re-detected by refresh.
- **Updated** `TestHasPermanentAuthFailure_RequiresPermanentRefreshFailureField` (was `..._PreservedFromPersistedState`): now verifies the discriminator field is the sole signal, with both legacy (non-permanent) and new (permanent) shapes.
- **Updated** `TestShouldRefresh_PersistedUnauthorizedWithCooldownStaysSchedulable`: legacy auth with Code=unauthorized + future NextRetryAfter → not permanent → stays schedulable (the test's original intent, now achieved via the discriminator rather than the NextRetryAfter gate).
- **Updated** all manual permanent-marker test fixtures to set `PermanentRefreshFailure=true` (5 tests).
- **Updated** `TestResponsesWebsocketPinnedAuthMatchesModel` permanent-failed auth fixture to set `PermanentRefreshFailure=true`.

### Process note

This is the third layer uncovered by following the "full ACCESSES write audit" discipline from #16. The first two (`clearCooldownStateForAuth`, `MarkResult` success branch) were write sites that **erased** the marker. This third one is a write site (`applyAuthFailureState` with disableCooling) that **produced the same shape** as the marker but with different semantics — the predicate's LastError-based fallback conflated them. The `PermanentRefreshFailure` field is the clean fix: it makes the marker's origin explicit, eliminating shape-based ambiguity.

daiaji:
## Codex P2 — Run gofmt on conductor.go

**Confirmed and fixed** — amended into the squashed commit `d909cde5`.

The `PermanentRefreshFailure = true` line added in the previous amend had incorrect indentation (one tab short). `gofmt -w sdk/cliproxy/auth/conductor.go` corrected it; no logic changes. Verified with `gofmt -l` (no output = clean) on all 4 modified files.

daiaji:
## Codex P2 — Persist refresh-failure markers in reloadable storage

**Confirmed and fixed** — amended into the squashed commit `06bc2044`.

Excellent catch — this was a real restart-recovery gap.

### Root cause

`m.persist(ctx, auth)` calls `m.store.Save(ctx, auth)`, which for the default file-based store (`FileTokenStore.Save`) only persists `auth.Metadata` (the JSON map). The `Auth` struct fields `PermanentRefreshFailure`, `Unavailable`, `Status`, `StatusMessage`, `LastError` are **outside** `Metadata` and are not written to the auth JSON file.

On restart, `FileTokenStore.readAuthFiles` reads the JSON file, rebuilds the `Auth` with default field values (`PermanentRefreshFailure=false`, `Unavailable=false`, `Status=StatusActive`), and the revoked credential becomes active/routable again. The `persistCooldownStates` call in the permanent branch only handles `.cds` cooldown records (which are dropped because `NextRetryAfter=zero`), not the auth file itself.

### Fix: mirror marker into Metadata + restore on load

**Write side** (`persist` function, conductor.go):

```go
if auth.PermanentRefreshFailure {
    auth.Metadata["permanent_refresh_failure"] = true
    auth.Metadata["unavailable"] = auth.Unavailable
    auth.Metadata["status"] = string(auth.Status)
    auth.Metadata["status_message"] = auth.StatusMessage
} else {
    // Clear stale marker keys so a recovered auth doesn't restore dead state.
    delete(auth.Metadata, "permanent_refresh_failure")
    delete(auth.Metadata, "unavailable")
    delete(auth.Metadata, "status")
    delete(auth.Metadata, "status_message")
}
```

This runs inside `persist`, so all 7 callers of `persist` (Register, Update, MarkResult, refreshAuthForRequest, etc.) automatically benefit — the marker is mirrored into Metadata before `store.Save` writes it.

**Read side** (new `ApplyPermanentFailureFromMetadata` helper, permanent_failure_metadata.go):

```go
func ApplyPermanentFailureFromMetadata(auth *Auth) {
    if v, ok := auth.Metadata["permanent_refresh_failure"].(bool); ok && v {
        auth.PermanentRefreshFailure = true
        auth.Unavailable = ...  // from metadata, default true
        auth.Status = ...       // from metadata, default StatusError
        auth.StatusMessage = ... // from metadata
    }
}
```

Called from `FileTokenStore.readAuthFiles` at both rebuild sites (plugin path + normal path), right after `ApplyCustomHeadersFromMetadata`. If the marker is absent (auth persisted before this feature, or recovered after manual token replacement), the helper is a no-op — the Auth retains default field values and is routable.

### What about LastError?

`LastError` is not mirrored into Metadata because:
1. `hasPermanentAuthFailure` now checks only `PermanentRefreshFailure` (not `LastError` shape), so `LastError` is not needed for the guards to fire.
2. `LastError` carries the full error message (potentially large), and persisting it into Metadata would bloat the auth JSON.
3. The marker fields (`PermanentRefreshFailure` + `Unavailable` + `Status` + `StatusMessage`) are sufficient for the guards; `LastError` is reconstructed on the next refresh attempt if the credential is still revoked.

### Verification via gitnexus

- `impact(persist)` — HIGH risk, 7 direct callers, 22 impacted symbols, 0 affected processes. The change is a pure additive mirror into Metadata before `store.Save`; the `else` branch clears stale keys, which is safe because a non-permanent auth should not have those keys.
- `detect_changes` — low risk, 0 affected processes.

### Tests

- **New** `TestPermanentFailureMetadata_RoundTrip`: simulates post-persist Metadata with marker keys → `ApplyPermanentFailureFromMetadata` → verifies all fields restored + `hasPermanentAuthFailure` + all three guards fire.
- **New** `TestPermanentFailureMetadata_NoMarkerIsNoOp`: Metadata without marker keys → helper is no-op → Auth retains defaults (routable).

### Process note

This is the fourth layer uncovered by the "full state-contract view" discipline (#16). The previous three layers were in-memory write sites (clearCooldownStateForAuth, MarkResult success, applyAuthFailureState disable-cooling). This fourth one is the **persistence layer** — the marker was correct in memory but never reached disk in a form the store reads back. Running the full three-layer checklist (memory → persist → restore) before implementing would have caught this in the first pass.

daiaji:
## Codex P2 — Restore permanent markers in all metadata stores

**Confirmed — split into a follow-up PR #4490.**

Good catch on the other 3 stores (Postgres / Object / Git). This PR has grown beyond the original "fix 400 invalid_grant runtime detection" scope into restart-recovery territory. To keep #4477 focused on the runtime marker, the persistence/recovery layer has been extracted into #4490:

- **#4477 (this PR)**: runtime marker only — `PermanentRefreshFailure` field, guards, `MarkResult` early-skip, `clearCooldownStateForAuth` guard, websocket pin, disable-cooling discriminator. Restart behavior: dead auth is retried once by refresh, which re-detects the permanent failure and re-sets the marker. Acceptable for the original bug (runtime retry loop / OOM contribution).
- **#4490 (follow-up)**: persist the marker into `Metadata` + restore in all 4 store readers (`FileTokenStore`, `PostgresStore`, `ObjectStore`, `GitStore`). This makes the marker survive restart without a re-refresh.

The `persist` Metadata-mirroring and `permanent_failure_metadata.go` have been removed from this PR's squashed commit (now `9f283dbe`). #4490 adds them back + the 3 missing store reader calls.

Both PRs are based on `dev`. #4490's diff is purely the persistence layer (7 files, +160 lines).

PR #4490, opened 2026-07-21
fix(auth): persist permanent failure marker across restarts in all stores

daiaji:
## Summary

Follow-up to #4477. The runtime permanent-failure marker (`PermanentRefreshFailure` field) is correct in memory but lost on restart because all storage backends only persist `Auth.Metadata`, not the `Auth` struct fields.

## Problem

`FileTokenStore`, `PostgresStore`, `ObjectStore`, and `GitStore` all rebuild the `Auth` from persisted JSON with default field values:

- `PermanentRefreshFailure = false`
- `Unavailable = false`
- `Status = StatusActive`

So a restart after a `400 invalid_grant` refresh failure reloads the revoked credential as active/routable. The runtime guards (`shouldRefresh`, `nextRefreshCheckAt`, `isAuthBlockedForModel`) stop firing until the next refresh attempt re-detects the failure.

### Write side (`persist`)

Mirror the marker into `Metadata` keys before `store.Save`:

- `permanent_refresh_failure` → `auth.PermanentRefreshFailure`
- `unavailable` → `auth.Unavailable`
- `status` → `string(auth.Status)`
- `status_message` → `auth.StatusMessage`

Non-permanent auths clear stale keys (so a recovered auth doesn't restore dead state).

### Read side (`ApplyPermanentFailureFromMetadata`)

New helper in `sdk/cliproxy/auth/permanent_failure_metadata.go` restores the marker from `Metadata` into the `Auth` struct fields. Called by all 4 store readers after rebuilding the `Auth`:

- `FileTokenStore.readAuthFiles` (2 sites: plugin path + normal path)
- `PostgresStore.List`
- `ObjectTokenStore.readAuthFile`
- `GitTokenStore.readAuthFile`

If the marker is absent (auth persisted before this feature, or recovered after manual token replacement), the helper is a no-op.

## Depends on

#4477 — this PR adds the persistence/recovery layer on top of the runtime marker introduced in #4477.

## Test plan

- `TestPermanentFailureMetadata_RoundTrip`: post-persist Metadata → `ApplyPermanentFailureFromMetadata` → fields restored + guards fire
- `TestPermanentFailureMetadata_NoMarkerIsNoOp`: no marker → no-op → Auth routable

### DIFF
--- a/internal/store/gitstore.go
+++ b/internal/store/gitstore.go
@@ -506,6 +506,7 @@ func (s *GitTokenStore) readAuthFile(path, baseDir string) (*cliproxyauth.Auth, 
 		auth.Attributes["email"] = email
 	}
 	cliproxyauth.ApplyCustomHeadersFromMetadata(auth)
+	cliproxyauth.ApplyPermanentFailureFromMetadata(auth)
 	if disabled, ok := metadata["disabled"].(bool); ok && disabled {
 		auth.Disabled = true
 		auth.Status = cliproxyauth.StatusDisabled
--- a/internal/store/objectstore.go
+++ b/internal/store/objectstore.go
@@ -608,6 +608,7 @@ func (s *ObjectTokenStore) readAuthFile(path, baseDir string) (*cliproxyauth.Aut
 		NextRefreshAfter: time.Time{},
 	}
 	cliproxyauth.ApplyCustomHeadersFromMetadata(auth)
+	cliproxyauth.ApplyPermanentFailureFromMetadata(auth)
 	if disabled, ok := metadata["disabled"].(bool); ok && disabled {
 		auth.Disabled = true
 		auth.Status = cliproxyauth.StatusDisabled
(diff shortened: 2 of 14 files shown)


daiaji:
## Codex P2 — Stop deleting provider metadata keys

**Confirmed and fixed** — amended into the commit `45290568`.

Excellent catch — the `else` branch in `persist` unconditionally deleted `unavailable`, `status`, and `status_message` from `Metadata`. Those are generic key names that provider/plugin credential metadata may legitimately carry; a successful persist of a non-permanent auth would silently strip them.

### Fix: namespaced keys

All 4 metadata keys are now namespaced with the `cpa_permanent_` prefix:

| old key | new key |
|---|---|
| `permanent_refresh_failure` | `cpa_permanent_refresh_failure` |
| `unavailable` | `cpa_permanent_unavailable` |
| `status` | `cpa_permanent_status` |
| `status_message` | `cpa_permanent_status_message` |

The keys are defined as constants in `permanent_failure_metadata.go` and referenced by name in both `persist` (write side) and `ApplyPermanentFailureFromMetadata` (read side), so there's no risk of literal drift. The `else` branch's `delete` calls now only remove keys that this logic owns — no provider metadata is touched.

### Verification via gitnexus

- `impact(persist)` — HIGH risk, 7 direct callers, 22 impacted symbols, 0 affected processes. The change is a pure key rename; the write/clear logic is unchanged.
- `detect_changes` — low risk, 0 affected processes.

PR #4532, opened 2026-07-24
fix: avoid prepared plugin-route recursion in nested streams

taoxian000:
Fixes a v7.2.97 recursion bug in nested host model stream execution. Nested callbacks correctly set SkipRouterPluginID, but the stream executor reused the outer prepared route before applying the skip and could recursively invoke the same plugin until memory exhaustion. The fix discards prepared routes for nested executions and reroutes with the caller skip. Top-level stream requests continue to reuse their prepared route. Validated with: go test ./sdk/api/handlers.

### DIFF
--- a/sdk/api/handlers/handlers.go
+++ b/sdk/api/handlers/handlers.go
@@ -155,8 +155,11 @@ func (h *BaseAPIHandler) PrepareStreamModelRoute(ctx context.Context, handlerTyp
 	return ctx, hasOverride
 }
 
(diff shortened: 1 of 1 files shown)


FlameMida:
该 PR 目前与已经拆分的 handler 代码存在合并冲突。#4771 已在当前 `dev` 上重新实现相同修复，补充了集成回归测试，并在 #4770 中记录了完整的 Bug 触发条件与影响。

PR #4537, opened 2026-07-24
feat(plugins): allow scheduler plugins to exclude auth candidates

origin652:
## Summary
- add `ExcludedAuthIDs` to scheduler-plugin responses
- validate exclusions against the host candidate set
- rerun the currently configured native scheduler with excluded IDs treated as request-local tried auths

This lets a scheduler plugin temporarily remove rate-limited credentials without forcing round-robin or reimplementing `first-fill`.

## Behavior
- no exclusions: existing behavior is unchanged
- exclusions with `first-fill`: CPA picks the next native first-fill candidate
- all candidates excluded: selection fails closed instead of retrying an excluded auth

## Tests
- `go test ./sdk/cliproxy/auth ./sdk/pluginapi`
- `go test -race ./sdk/cliproxy/auth ./sdk/pluginapi`
- adds focused tests for first-fill exclusion and all-candidate exclusion

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -1864,6 +1864,43 @@ func (m *Manager) pickViaBuiltinScheduler(ctx context.Context, strategy schedule
 	}
 }
 
+func schedulerExcludedAuthIDs(candidates []*Auth, requested []string) map[string]struct{} {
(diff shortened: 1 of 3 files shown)


origin652:
Closing at the deployer’s request: they do not want to modify CLIProxyAPI itself. The proposed API addition is therefore not being pursued.

PR #4546, opened 2026-07-24
perf(codex): make replay prefix fingerprints incremental (O(n²) → O(n) hashing)

steipete:
## Problem

`codexReasoningReplayTurnAnchorIndex` re-hashes the entire input prefix for every index it probes while scanning anchors downward (`codexReplayInputPrefixFingerprint` restarts SHA-256 from item 0 each call). The anchor search is O(n²) hashing over the conversation, per replay turn.

On long-context sessions this dominates the process: on a production gateway we measured ~11.5 cores of steady CPU (pprof: 46% `sha256.blockSHANI`, 21% `runtime.memmove`) and request latencies of 20s–5m spent entirely before contacting upstream, surfacing as client-side timeouts. Small conversations hide the bug; it grows quadratically with history size.

## Fix

Replace the per-probe recomputation with a lazily extended incremental digest that absorbs each item once and snapshots every prefix sum (`sha256` `Sum` copies the running state). Each probe is O(1) after a single O(n) pass; output is bit-identical to `codexReplayInputPrefixFingerprint`.

## Evidence

- Equivalence covered by the new `TestCodexReplayPrefixFingerprintsMatchesDirectComputation` (out-of-order and repeated probes, out-of-range indexes).
- Production before/after on the same traffic: ~1150% CPU → ~6% of one core; latencies 20s–5m → 4–14s.
- `go build ./...` and executor package tests green on top of current `main` (9a52607f).

### DIFF
--- a/internal/runtime/executor/codex_executor.go
+++ b/internal/runtime/executor/codex_executor.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"net/http"
 	"sort"
@@ -502,6 +503,7 @@ func insertCodexReasoningReplayTurns(body []byte, replayItems [][]byte) ([]byte,
 	turns := splitCodexReasoningReplayTurns(replayItems)
 	insertions := make(map[int][][]byte)
 	usedAnchorIndexes := make(map[int]bool)
+	prefixFingerprints := newCodexReplayPrefixFingerprints(inputItems)
 	fallbackAnchorEnd := len(inputItems) - 1
 	inserted := false
 	for turnIndex := len(turns) - 1; turnIndex >= 0; turnIndex-- {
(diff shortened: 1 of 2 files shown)


sususu98:
Hi @steipete! 👋

First of all, thank you so much for this incredible performance optimization! Moving the replay prefix fingerprinting from $O(N^2)$ to $O(N)$ incremental SHA-256 hashing is a massive win for long-context Codex sessions. In our local benchmark simulations, context sizes around 2,000 items with 10 reasoning turns dropped from **~1.26 seconds** of pre-request hashing down to just **~300 microseconds** (~4,100x speedup), which completely eliminates pre-upstream CPU throttling and client timeouts.

Due to a recent refactoring on `dev` where `codex_executor.go` was split and reasoning replay logic was moved into `codex_executor_reasoning.go`, GitHub is currently showing a merge conflict.

We've performed a clean rebase onto latest `dev`, migrated the incremental hashing implementation into `codex_executor_reasoning.go`, and verified that all package tests (including your new test `TestCodexReplayPrefixFingerprintsMatchesDirectComputation`) pass cleanly.

If helpful, you can fetch or cherry-pick the rebased commit directly from this branch:
👉 https://github.com/sususu98/CLIProxyAPI/tree/perf/codex-replay-incremental-fingerprints-rebased

(Or let us know if you'd like us to push the rebased commit directly to your PR branch).

Thanks again for the excellent contribution! 🚀

PR #4550, opened 2026-07-25
fix(claude): honor global cloak disable in model listings

### LINKED ISSUE #4473: [Feature Request] Add config toggle for model list cloaking (EnsureClaudeModelIDPrefix)
## Describe the issue

The model list obfuscation via `EnsureClaudeModelIDPrefix` has **no independent configuration toggle**. The existing `disable-claude-cloak-mode` and `claude-cloak-mode` settings only control message **body** cloaking, not model **list** cloaking.

The `EnsureClaudeModelIDPrefix` function is triggered solely by the presence of the `anthropic-version` HTTP header — there is no config key to disable it.

## Source code evidence

From `sdk/api/handlers/claude/code_handlers.go`:

```go
func EnsureClaudeModelIDPrefix(id string) string {
    if id == "" { return id }
    if strings.HasPrefix(id, "claude-") { return id }  // hyphen check bypass
    return claudeDDModelPrefix + reverseModelID(id)     // obfuscation
}
```

The `ClaudeModels()` function calls this unconditionally when `anthropic-version` header is detected. There is no `if config.DisableCloak { return id }` guard.

## Steps to reproduce

1. Set `disable-claude-cloak-mode: true` in config.yaml
2. Query `/v1/models` with `anthropic-version: 2023-06-01`
3. Model IDs are still obfuscated (e.g., `claude-fable-5-dd-]m1[5-elbaf/edualc`)

## Expected behavior

`disable-claude-cloak-mode: true` should also disable model list obfuscation, or a new dedicated config key should exist:

```yaml
disable-model-list-cloak: true
```

## Actual behavior

Model list cloaking is always active when `anthropic-version` header is present, regardless of cloak settings.

## Environment

- CPA Version: 7.2.92
- Client: Claude Code for VS Code extension v2.1.216 (always sends `anthropic-version: 2023-06-01`)

## Workaround

Set `prefix: claude-` (hyphen suffix) on each `claude-api-key` entry. The `HasPrefix(id, "claude-")` check bypasses obfuscation. However, this produces model IDs with mixed delimiters like `claude-/modelname` and only works for `claude-api-key` models — it doesn't help OAuth models.

yinkev:
## Summary

- make `disable-claude-cloak-mode: true` preserve registered model IDs in Claude-compatible model listings
- retain the current encoded-ID behavior by default
- apply the same behavior to normal SDK-backed and Home-backed model-list handlers
- preserve sorting, pagination metadata, and input immutability

## Root cause

The global setting disabled Claude request cloaking, but `internal/client/claude/models.BuildResponse` still unconditionally rewrote every non-Claude model ID through `EnsureClaudeModelIDPrefix`.

## Validation

- `go test ./...`
- independent Codex review: no actionable findings
- `git diff --check upstream/dev...HEAD`

Fixes #4473

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -163,10 +163,10 @@ save-cooldown-status: false
 transient-error-cooldown-seconds: 0
 
 # When true, globally disable Claude request cloaking (the Claude Code CLI disguise and
-# system prompt replacement), so the original system prompt is passed through to Claude as-is.
-# Individual credentials can still override this: a claude-api-key entry via its "cloak.mode",
-# or a Claude OAuth/token file via a "cloak_mode" value. Default false keeps the per-client
-# "auto" behavior (cloak only non-Claude-Code clients).
+# system prompt replacement) and preserve registered IDs in Claude-compatible model listings.
+# Individual credentials can still override request cloaking via a claude-api-key "cloak.mode"
+# or a Claude OAuth/token-file "cloak_mode" value. The model-list behavior remains global.
+# Default false keeps per-client "auto" request cloaking and encoded non-Claude model IDs.
 disable-claude-cloak-mode: false
 
 # disable-image-generation supports: false (default), true, "chat", or "passthrough".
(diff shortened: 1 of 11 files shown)


yinkev:
Current `dev` now includes the dedicated `claude-code.disable-cloaking-model-list` setting from 6914478, covering both the SDK and Home model-list paths and closing #4473. That supersedes this alternative of reusing the global cloak switch, so I’m closing this rather than rebasing a redundant implementation.

PR #4557, opened 2026-07-25
fix(xai): guard opt-in x_search collision

### LINKED ISSUE #4339: v7.2.73+ always injects native x_search and breaks client function tool web_search on /v1/responses
[  ] No, it's another issue.

**Describe the bug**  
From `v7.2.73` onward, the xAI executor always injects native `{"type":"x_search"}` into `/v1/responses` requests via `ensureXAINativeXSearchTool`, even when `config.yaml` has no payload rules and the client already provides its own function tools.

When the client declares a function tool named `web_search` (common in agent frameworks like pi-agent), this causes:

1. Final tools become: client `function/web_search` + injected `x_search`
2. Upstream may return: `400 {"code":"invalid-argument","error":"Duplicate tool names: web_search"}`
3. Or the model emits native `web_search_call` that stays `status=in_progress` with empty sources
4. The client never receives a normal `function_call` for its local tool, so the agent turn aborts

On `v7.2.71` / `v7.2.72`, the same request returns a normal `function_call` for `web_search`.

Payload config cannot disable this, because injection runs after `ApplyPayloadConfigWithRequest`. I verified `payload.filter` removing `tools.#(type=="x_search")` does not help.

**CLI Type**  
xAI / Grok OAuth

**Model Name**  
`grok-4.5`

**LLM Client**  
pi-agent (OpenAI Responses API client with local function tools)

**Request Information**

```bash
curl -N http://127.0.0.1:8317/v1/responses \
  -H "Authorization: Bearer $API_KEY" \
  -H "Content-Type: application/json" \
  -H "User-Agent: pi-agent" \
  -d '{
    "model": "grok-4.5",
    "stream": true,
    "input": [
      {"role": "user", "content": "Call web_search with query hello"}
    ],
    "tools": [
      {
        "type": "function",
        "name": "web_search",
        "description": "Search the web",
        "parameters": {
          "type": "object",
          "properties": {
            "query": {"type": "string"}
          },
          "required": ["query"]
        }
      }
    ],
    "tool_choice": "required",
    "reasoning": {"effort": "low"}
  }'
```

**Result on `v7.2.71` / `v7.2.72`:**  
`response.completed` contains `function_call` with `name=web_search`.

**Result on `v7.2.73+` / `latest` (`v7.2.77`):**  
`response.completed.tools` includes both the client function tool and `{"type":"x_search"}`. Output often contains:

```json
{
  "type": "web_search_call",
  "status": "in_progress",
  "action": {
    "type": "search",
    "query": "hello",
    "sources": []
  }
}
```

No client-side `function_call` is returned.

Related log:

```text
request error, error status: 400, error message: {"code":"invalid-argument","error":"Duplicate tool names: web_search"}
```

Control checks:

- Rename client tool to `pi_web_search` → client `function_call` works, but `x_search` is still injected
- Send request with no client tools on `v7.2.73+` → tools still become `[{"type":"x_search"}]`

Code path in `internal/runtime/executor/xai_executor.go` (since `v7.2.73`):

```go
// Always inject native x_search when the client did not declare it ...
body = helps.ApplyPayloadConfigWithRequest(...)
body = ensureXAINativeXSearchTool(body)
```

**Expected behavior**  
1. Do not force-inject native `x_search` for all xAI requests by default, or  
2. Provide a config switch to disable it (for example `xai.inject-x-search: false`), and  
3. When a client already provides function tools, especially one named `web_search`, do not break the client tool loop with native `web_search_call` / duplicate tool name errors.

**Screenshots**  
N/A

**OS Type**  
 - OS: Linux (Docker)  
 - Version: Ubuntu host, image `eceasy/cli-proxy-api`

**Additional context**  
- Working: `eceasy/cli-proxy-api:v7.2.71`  
- Broken: `eceasy/cli-proxy-api:latest` (`v7.2.77`)  
- First version with forced injection: `v7.2.73` (`ensureXAINativeXSearchTool`)  
- Related issues: #4282, #3704  
- Current workaround: pin image to `<= v7.2.72`, or rename client tool away from `web_search`

yinkev:
## Summary

- harden the opt-in `xai.inject-x-search` path against a caller-owned function/custom tool named `web_search`
- skip injected native `x_search` only for that collision
- retain opt-in injection for non-conflicting requests
- retain `allowed_tools` synchronization when callers explicitly provide both tools

## Scope

Current `dev` already makes native `x_search` injection configurable and disabled by default. This PR does not add or change that switch. It only protects callers that explicitly enable injection from creating two tools that xAI exposes under the same upstream `web_search` name.

## Validation

- `go test ./internal/runtime/executor -run 'Test(XAIExecutorPrepareHonorsInjectXSearchConfig|EnsureXAINativeXSearchToolPreservesClientWebSearchTool|EnsureXAINativeXSearchToolSyncsExplicitNativeToolWithClientWebSearch)$' -count=1`
- `go test -race ./internal/runtime/executor -run 'Test(XAIExecutorPrepareHonorsInjectXSearchConfig|EnsureXAINativeXSearchToolPreservesClientWebSearchTool|EnsureXAINativeXSearchToolSyncsExplicitNativeToolWithClientWebSearch)$' -count=1`
- `go vet ./internal/runtime/executor`
- `git diff --check`

Fixes #4339

### DIFF
--- a/internal/runtime/executor/xai_executor_request.go
+++ b/internal/runtime/executor/xai_executor_request.go
@@ -541,16 +541,21 @@ func sanitizeXAIResponsesBody(body []byte, model string) []byte {
 }
 
 // ensureXAINativeXSearchTool appends {"type":"x_search"} when the final tools
-// list does not already include native X Search. When tool_choice restricts the
-// model to allowed_tools, x_search is also added there (without duplicates) so
-// Grok can select the injected tool. When injection is enabled, HTTP and websocket
-// executors both prepare payloads through prepareResponsesRequestTo, so this runs
-// once before the body is submitted upstream.
+// list does not already include native X Search or a client function named
+// web_search. When tool_choice restricts the model to allowed_tools, x_search
+// is also added there (without duplicates) so Grok can select the injected
+// tool. When injection is enabled, HTTP and websocket executors both prepare
+// payloads through prepareResponsesRequestTo, so this runs once before the body
+// is submitted upstream.
 func ensureXAINativeXSearchTool(body []byte) []byte {
 	if !gjson.ValidBytes(body) {
 		return body
 	}
-	if !xaiRequestHasNativeXSearch(body) {
+	hasNativeXSearch := xaiRequestHasNativeXSearch(body)
+	if xaiRequestHasClientWebSearchTool(body) && !hasNativeXSearch {
+		return body
+	}
+	if !hasNativeXSearch {
 		tools := gjson.GetBytes(body, "tools")
 		if !tools.Exists() || !tools.IsArray() {
 			body, _ = sjson.SetRawBytes(body, "tools", []byte(`[{"type":"x_search"}]`))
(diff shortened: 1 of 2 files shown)


yinkev:
Rebased this onto current `dev`. The new `xai.inject-x-search` option now keeps its explicit opt-in behavior for ordinary tools, while a client-owned function/custom tool named `web_search` still blocks native `x_search` injection to avoid the upstream name collision. Added that config-enabled collision case to the regression coverage; full `go test ./...` and the required server build pass.

Issue #4438, opened 2026-07-19
GitStore deletes all local auth files and force-pushes empty orphan commit after upgrade (v7.2.19 → v7.2.88)


Youngv:
**Is it a request payload issue?**
- [ ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
- [x] No, it's another issue.

**If it's a request payload issue, you MUST know**
N/A — this is a GitStore / auth persistence issue, not a request payload issue.

**Describe the bug**

After upgrading the Docker image from `eceasy/cli-proxy-api:v7.2.19` to `v7.2.88` and recreating the container, GitStore **deleted all local auth JSON files** (~26 seconds after a successful boot that had already loaded them), then failed to re-persist them, and **force-pushed an empty orphan commit** to the configured Git remote. A subsequent container restart rewrote `config/config.yaml` with the **default example config** and pushed that as the new sole commit on `master`.

This is **not** a Docker volume/bind-mount mistake: host bind mounts for the gitstore directory were unchanged; only the files inside the gitstore tree were removed by the application itself (filesystem `REMOVE` events + git commits authored as `CLIProxyAPI`).

### Reproduction steps

1. Long-running deployment with GitStore enabled (`GITSTORE_GIT_URL` / `GITSTORE_GIT_USERNAME` / `GITSTORE_GIT_TOKEN`), bind-mount e.g. `./CLIProxyAPI:/CLIProxyAPI/gitstore`.
2. Local gitstore contains multiple OAuth auth JSON files under `auths/` and a customized `config/config.yaml`.
3. Bump image tag only: `v7.2.19` → `v7.2.88` (compose volumes/env unchanged).
4. `docker compose up -d` (container recreated).
5. Observe: startup loads auth files successfully, then shortly after they are all removed; remote branch becomes an empty/orphan tree.

### Observed behavior (timeline, local TZ Asia/Shanghai)

| Time | What happened |
|------|----------------|
| T+0 | Old process shuts down; new `v7.2.88` starts |
| ~T+3s | `auth directory scan complete - found 7 .json files, 7 readable` / `loaded 7 file-based clients` / full client load complete |
| ~T+23s | Config reload after plugin install / config write |
| ~T+26s | Filesystem events: **REMOVE** for every `auths/*.json` (codex + xai, 7 files) |
[...]

PR #4559, opened 2026-07-25
fix(gitstore): refuse watcher-originated auth deletes

yinkev:
## Summary

- refuse filesystem-watcher auth removals when the missing file is still tracked in GitStore HEAD
- preserve explicit management deletes, including the existing unlink-before-`Store.Delete` sequence
- treat the watcher event following an explicit delete as a redundant no-op
- keep repeated deletes idempotent without force-pushing stale local history over remote-only changes

## Failure mode

A watcher `REMOVE` event currently flows into `PersistAuthFiles("Remove auth ...")`, where GitStore stages the missing path, rewrites the branch as a single commit, and force-pushes it. If a credential disappears from the local mirror unexpectedly, that event can therefore turn local loss into remote loss.

This patch distinguishes watcher propagation from the explicit `GitTokenStore.Delete` path:

- explicit delete: remove and commit the auth intentionally
- watcher remove while path is still present in HEAD: fail closed and leave remote history untouched
- watcher remove after explicit delete already updated HEAD: no-op

## Scope

This mitigates the destructive propagation path reported in #4438. It does not claim to identify the component that initially removed the local files.

## Validation

- `go test ./internal/store -count=1`
- focused race tests for unexpected watcher removal, explicit unlink-before-delete, and remote-divergence preservation
- `go vet ./internal/store`
- `go build -o /dev/null ./cmd/server`
- `git diff --check upstream/dev...HEAD`

Addresses #4438

### DIFF
--- a/internal/store/gitstore.go
+++ b/internal/store/gitstore.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-git/go-git/v6/config"
 	"github.com/go-git/go-git/v6/plumbing"
 	"github.com/go-git/go-git/v6/plumbing/client"
+	gitindex "github.com/go-git/go-git/v6/plumbing/format/index"
 	"github.com/go-git/go-git/v6/plumbing/object"
 	"github.com/go-git/go-git/v6/plumbing/transport"
 	"github.com/go-git/go-git/v6/plumbing/transport/http"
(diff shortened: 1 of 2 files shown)


yinkev:
Closing this as superseded. Rebasing onto current `dev` skips this PR commit as already applied, and `a2ff6914` now contains the watcher-delete guard plus the same regression coverage. I reran the watcher-originated deletion, post-explicit-delete no-op, and repeated-delete tests on current `dev`; all three pass.

PR #4563, opened 2026-07-25
fix(claude): skip deferred tools for cache control

### LINKED ISSUE #4561: Claude: injectToolsCacheControl adds cache_control to a defer_loading tool, causing upstream 400
[X] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[  ] No, it's another issue.

**Describe the bug**

`injectToolsCacheControl` unconditionally attaches `cache_control` to the **last** element of `tools[]` when no tool already carries one. It does not check whether that last tool has `defer_loading: true`.

Anthropic rejects that combination:

```
400 invalid_request_error
Tool 'tool_1' cannot have both defer_loading=true and cache_control set.
Tools with defer_loading cannot use prompt caching.
```

The result is confusing from the client side: the request contains **no `cache_control` anywhere**, yet the API reports that both are set. The proxy added it.

This breaks any Claude Code / Anthropic-wire client that uses Tool Search (`defer_loading`) through the proxy, whenever the deferred tool happens to be last in the array.

Source: `internal/runtime/executor/claude_executor.go` — `injectToolsCacheControl()` (~L2626)

```go
// Check if ANY tool already has cache_control - if so, don't modify tools
...
// Add cache_control to the last tool
lastToolPath := fmt.Sprintf("tools.%d.cache_control", toolCount-1)
```

**Trigger condition (single, not a combination)**

Both of these must hold:
1. `tools[len-1].defer_loading == true`
2. no tool in `tools[]` carries a client-supplied `cache_control` (otherwise the injector correctly bails out early)

**Request Information**

Minimal reproduction — two tools, `defer_loading` on the last one, **no `cache_control` in the payload at all**:

```bash
curl -s http://127.0.0.1:8317/v1/messages \
  -H 'content-type: application/json' \
  -H 'anthropic-version: 2023-06-01' \
  -d '{
    "model": "claude-opus-5",
    "max_tokens": 16,
    "system": [{"type":"text","text":"hi"}],
    "tools": [
      {"name":"tool_0","description":"d","input_schema":{"type":"object","properties":{}}},
      {"name":"tool_1","description":"d","input_schema":{"type":"object","properties":{}},"defer_loading":true}
    ],
    "messages": [{"role":"user","content":"say OK"}]
  }'
```

Result:
```json
 "message":"Tool 'tool_1' cannot have both defer_loading=true and cache_control set. Tools with defer_loading cannot use prompt caching."}}
```

**Isolation matrix** (all sent directly to the proxy, 29 tools unless noted, no intermediate software)

| payload | result |
|---|---|
| no `defer_loading`, no client `cache_control` | 200, `cache_creation=5714` |
| client `cache_control` on last tool, no `defer_loading` | 200, `cache_read=3814` |
| **`defer_loading` on last tool only, no `cache_control` anywhere** | **400 (bug)** |
| 2 tools, `defer_loading` on last | 400 (bug) |
| `defer_loading` on a *middle* tool (idx 5) only | 200 |
| `defer_loading` on last two tools | 400 (bug) |
| `defer_loading` on last + client `cache_control` on `tool_0` | 200 (injector correctly skips) |
| `defer_loading` on tools 0..27, last resident | 200 |

The middle-vs-last asymmetry and the "client `cache_control` on `tool_0` → 200" row together confirm the injector's last-element write is the cause.

Note: `defer_loading` on *all* tools returns a different and legitimate error (`At least one tool must have defer_loading=false`), which is correct upstream behavior and not part of this report.

**Expected behavior**

`injectToolsCacheControl` should not place a cache breakpoint on a deferred tool. Suggested fix: pick the last tool with `defer_loading != true` as the injection target, and skip tools injection entirely when every tool is deferred.

`#1788` added `remove-tools-cache-control`, but that is scoped to `claude-api-key` entries and does not cover OAuth credential-pool routing.

**CLI Type**
claude code (OAuth account pool, `auth-dir` credentials)

**Model Name**
claude-opus-5 (also reproduces on other Claude models)

**LLM Client**
cURL (originally observed via an Anthropic-wire agent client that enables Tool Search)

**OS Type**
 - OS: Linux
 - Version: 6.8.0-134-generic

**Additional context**
Observed on `7.2.88-1-plus` (commit `3088e873`). The offending function is unchanged on current `main`.

yinkev:
## Summary

- avoid injecting `cache_control` onto Claude tools with `defer_loading: true`
- place the automatic tool-cache breakpoint on the last eligible non-deferred tool
- skip tool-cache injection when every tool is deferred while preserving system/message caching
- add regressions for a deferred tail and an all-deferred tool list

## Root cause

`injectToolsCacheControl` always wrote to `tools[len-1]` when no client cache breakpoint existed. Anthropic rejects tools that combine `defer_loading: true` with `cache_control`.

## Validation

- `go test ./internal/runtime/executor -run 'TestEnsureCacheControl' -count=1`
- `go test -race ./internal/runtime/executor -run 'TestEnsureCacheControl' -count=1`
- `go vet ./internal/runtime/executor`
- `go build -o /tmp/cliproxyapi-4561 ./cmd/server`
- `git diff --check`

Fixes #4561

### DIFF
--- a/internal/runtime/executor/caching_verify_test.go
+++ b/internal/runtime/executor/caching_verify_test.go
@@ -67,6 +67,49 @@ func TestEnsureCacheControl(t *testing.T) {
 		}
 	})
 
+	t.Run("Tools Caching Skips Deferred Tail", func(t *testing.T) {
+		input := []byte(`{
+			"model": "claude-opus-5",
+			"tools": [
+				{"name": "resident", "description": "Resident tool", "input_schema": {"type": "object"}},
+				{"name": "deferred", "description": "Deferred tool", "input_schema": {"type": "object"}, "defer_loading": true}
+			],
+			"messages": [{"role": "user", "content": "Hi"}]
+		}`)
+
+		output := ensureCacheControl(input)
+
+		if got := gjson.GetBytes(output, "tools.0.cache_control.type").String(); got != "ephemeral" {
+			t.Fatalf("resident tool cache_control = %q, want ephemeral; output: %s", got, output)
+		}
+		if gjson.GetBytes(output, "tools.1.cache_control").Exists() {
+			t.Fatalf("deferred tool must not receive cache_control; output: %s", output)
+		}
+	})
+
+	t.Run("Tools Caching Skips When All Tools Deferred", func(t *testing.T) {
+		input := []byte(`{
+			"model": "claude-opus-5",
+			"tools": [
+				{"name": "deferred0", "description": "Deferred tool", "input_schema": {"type": "object"}, "defer_loading": true},
+				{"name": "deferred1", "description": "Deferred tool", "input_schema": {"type": "object"}, "defer_loading": true}
+			],
+			"system": "System prompt",
+			"messages": [{"role": "user", "content": "Hi"}]
+		}`)
+
+		output := ensureCacheControl(input)
+
+		for i := 0; i < 2; i++ {
+			if gjson.GetBytes(output, fmt.Sprintf("tools.%d.cache_control", i)).Exists() {
+				t.Fatalf("deferred tool %d must not receive cache_control; output: %s", i, output)
+			}
+		}
+		if got := gjson.GetBytes(output, "system.0.cache_control.type").String(); got != "ephemeral" {
+			t.Fatalf("system cache_control = %q, want ephemeral; output: %s", got, output)
+		}
+	})
+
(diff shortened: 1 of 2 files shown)


yinkev:
Closing as superseded by upstream commit `95d5b248`, which independently fixes #4561 by selecting the last non-deferred tool and adds equivalent mixed/deferred regression coverage. Keeping this branch open would now duplicate current `dev`.

PR #4578, opened 2026-07-25
feat(routing): allow manual credential selection

TianqiYe:
## Summary

- add opt-in `routing.allow-auth-index-header`
- let authenticated clients pin one credential with `X-CPA-Auth-Index`
- preserve existing round-robin behavior when the header is omitted
- reject disabled or unknown selections and never silently fall back from an unavailable pinned credential
- strip the proxy control header before forwarding upstream

## Why

Embedders need per-session manual account selection without disabling automatic routing for other requests. This keeps auth lifecycle and provider adaptation inside CLIProxyAPI while exposing one small selection control to callers.

## Validation

- `go test ./...`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -196,6 +196,9 @@ quota-exceeded:
 # Routing strategy for selecting credentials when multiple match.
 routing:
   strategy: "round-robin" # round-robin (default), fill-first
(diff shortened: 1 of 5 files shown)


TianqiYe:
Closing this draft because the integration belongs in the downstream Lapis repository, not upstream CLIProxyAPI.

PR #4595, opened 2026-07-27
fix(auth): preserve prefixed API-key model capabilities / 修复带前缀 API Key 模型的能力路由

rdself:
## Summary

Configured provider models can be exposed with a prefix while executors send the unprefixed upstream model name. A global registry lookup cannot identify the exact auth, configured upstream entry, or endpoint kind selected for one attempt, so prefixed requests could receive unrelated static capabilities.

This change:

- compiles config-backed aliases and model capabilities into one immutable routing snapshot keyed by auth ID
- binds the exact configured model definition to each selected upstream attempt
- adds configured thinking support for Gemini, Interactions, Claude, Codex, xAI, and Vertex-compatible API-key providers
- applies the same exact per-auth binding to OpenAI-compatible providers, including keyless configured entries
- resolves configured catalog and execution metadata through one suffix-free static-capability path
- preserves prefixed OpenAI-compatible `image: true` discovery and provider-scoped native Codex/xAI image and video capabilities
- filters local and plugin-scheduler candidates by the exact auth registration and requested chat, image, or video endpoint
- keeps OAuth and unbound requests on the existing registry-based thinking path when no exact configured-model binding exists

## Failure Examples Without This Change

A prefixed API-key request can select the correct upstream model but fail to match its configured capabilities during execution.

For example, with `force-model-prefix: true`, provider prefix `sp-anthropic`, and configured model `claude-opus-4-6`, a client request for `sp-anthropic/claude-opus-4-6` reaches the executor as `claude-opus-4-6`. A global lookup can miss the provider-specific thinking levels, causing `xhigh` or `max` to be mapped, rejected, or removed using unrelated metadata.

A suffixed configured upstream such as `gpt-5.6-luna(high)` can also lose static `xhigh` and `max` catalog capabilities if static lookup uses the suffixed name while execution uses the base name.

Likewise, when image-capable, video-capable, and chat-only registrations expose the same model ID, aggregate validation alone can admit the request and then select an auth that does not support the requested endpoint. Selection must check the exact client registration.

## Implementation

- Builds one immutable routing snapshot keyed by auth ID from a deep-cloned runtime config.
- Carries the selected configured model metadata through candidate selection and execution.
- Uses one configured-model resolver for suffix-free static lookup, explicit thinking overrides, modality normalization, and provider-scoped endpoint capabilities.
- Maps cross-family high intent using `xhigh -> xhigh, max, high` and `max -> max, xhigh, high`.
- Keeps same-family validation strict and preserves the legacy registry-based thinking path when no configured model is resolved.
- Supports config-backed OpenAI-compatible routing even when the configured provider does not require an API key.
- Restricts chat, image, and video candidates before normal local scheduling and plugin-delegated built-in scheduling.
- Validates known Home-dispatched registrations before executor entry and advances the Home count when the selected auth does not support the requested endpoint.
- Preserves native Codex and xAI image/video-only metadata through aliases and prefixes while keeping shared chat-capable registrations visible in model catalogs.
- Includes thinking and endpoint-related model metadata in watcher hashes so capability-only edits reload correctly.

## Validation

- `go test ./... -count=1`
- targeted `go test -race` across registry, thinking, executor helpers, and auth routing
- `go build ./cmd/server`
- stream, non-stream, token-count, repeated alias pool, prefix, hot-reload, catalog, plugin scheduler, image routing, and video routing regression tests

Supersedes #4215. This pull request contains the same final change rebased onto the latest `dev` branch.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -289,6 +289,8 @@ nonstream-keepalive-interval: 0
 #       - name: "gemini-2.5-flash" # upstream model name
 #         alias: "gemini-flash"    # client alias mapped to the upstream model
 #         display-name: "Gemini Flash" # optional catalog display name
+#         thinking:                  # optional: exact thinking capability for this configured model
+#           levels: ["high", "medium", "low", "none", "auto"]
 #     excluded-models:
 #       - "gemini-2.5-pro"     # exclude specific models from this provider (exact match)
 #       - "gemini-2.5-*"       # wildcard matching prefix (e.g. gemini-2.5-flash, gemini-2.5-pro)
@@ -311,6 +313,8 @@ nonstream-keepalive-interval: 0
 #     models:
 #       - name: "gemini-2.5-flash" # upstream model name
 #         alias: "native-gemini-flash" # client alias mapped to the upstream model
+#         thinking:
+#           levels: ["high", "medium", "low", "none", "auto"]
 #     excluded-models:
 #       - "gemini-2.5-pro"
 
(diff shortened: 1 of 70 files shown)


rdself:
Closing this in favor of the upstream implementation shipped in v7.2.105 via f322914. The upstream fix binds configured API-key model capabilities to the selected auth route and handles cross-provider thinking-level mapping, which covers the original issue reported in #4618. After reviewing the released implementation, it is a better fit for the existing routing model and maintenance scope, so the broader endpoint-routing additions in this PR are no longer needed. Thanks for incorporating the core fix.

PR #4604, opened 2026-07-27
fix(executor): reuse pooled http.Client for proxy-url upstream calls

llmdph:
## Summary
- Stop allocating a new `http.Client` + `http.Transport` on every upstream request when `proxy-url` / auth proxy is configured.
- Introduce a small process-wide pool of **8** `http.Client`s (within 5–10) keyed by proxy URL + timeout.
- Pooled clients share one `Transport`, with `IdleConnTimeout=30s` and higher idle conn limits, to reduce keep-alive / CONNECT buildup through HTTP proxies such as WARP.

## Why
With global `proxy-url`, `NewProxyAwareHTTPClient` previously did `&http.Client{}` and `cloneDefaultTransport()` per call (xAI/Grok executors included). That prevents connection reuse and commonly leaves idle proxy tunnels around the default ~90s idle timeout, which shows up as slow WARP HTTP paths and connection pressure.

## Test plan
- [x] `TestNewProxyAwareHTTPClientDirectBypassesGlobalProxy` still passes
- [x] Added pool reuse / shared transport tests
- [ ] `go test ./internal/runtime/executor/helps/ -count=1`
- [ ] Manual: run CPA with `proxy-url` pointing at HTTP proxy, send repeated `/v1/responses`, confirm proxy side no longer opens a fresh CONNECT every request


llmdph:
Closing this PR.

The branch started as a proxy HTTP client pool change, but it has since accumulated site-specific work (LLM request logs, Warp rotate agent hook, plugin host retries, etc.) that is not appropriate for upstream.

We'll keep those customizations on the llmdph fork only. Thanks for the automated review notes from Codex — useful for our private branch, but this PR should not stay open against the official repo.

PR #4631, opened 2026-07-28
fix(store): prevent GitTokenStore data loss from corrupted index

### LINKED ISSUE #4629: fix(store): GitTokenStore data loss after failed pull — corrupted index causes force push to wipe auths and config
### Summary

When a `git pull` inside `GitTokenStore` fails (e.g. `packfile not found` due to transient network issues on Render Free), the local git index becomes corrupted — auth file deletions are staged in the index. A subsequent `PersistConfig` or `PersistAuthFiles` call then commits this corrupted index, `rewriteHeadAsSingleCommit` replaces the entire HEAD tree with the incomplete tree, and `Force:true` push permanently deletes the missing files from the remote repository.

### Deployment & Tutorial

Deployed on Render Free following the official tutorial: https://help.router-for.me/cn/hands-on/tutorial-10.html

### Timeline (from Render logs, 2026-07-28)

| Time (UTC+8) | Event |
|---|---|
| 09:47:04 | `failed to persist config change: git token store: pull: to: packfile not found` — Pull failed, local index corrupted, no self-healing |
| 10:06:05 | `PATCH /v0/management/openai-compatibility` — toggled provider disabled status → triggered `persistConfigAsync` |
| 10:06:08 | `auth file changed (REMOVE): codex-40842fb7-..., codex-49b5df92-..., codex-bbe029b2-..., codex-e60eacf5-...` — 4 auth files reported as REMOVED (user did NOT delete them) |
| 10:06:10 | `failed to persist auth changes: git token store: add auths/codex-...: entry not found` — attempt to re-add fails because files are gone from index |
| 11:06:41 | First 500 on `/v1/messages` — service degradation begins as in-memory auth list is eventually reloaded from the now-empty remote |

**Key observation**: The data was already lost at 10:06, but the service continued working until ~11:06 because the in-memory auth list had not yet been refreshed. This "delayed blast" makes the bug harder to notice immediately.

### Evidence from GitHub repository

The affected config repo (https://github.com/dong0713/cli-config) shows commit `8bfea038` with message "Remove auth codex-40842fb7-..." whose tree is the empty tree hash `4b825dc642cb6eb9a060e54bf8d69288fbee4904` — confirming that all auth files were wiped in a single force push.

### Reproduction path

1. Deploy CLIProxyAPI on Render Free (or any environment where git pulls can fail)
2. Wait for a container restart or transient network issue causing `pull: packfile not found`
3. Toggle any provider's disabled status via Management API (`PATCH /v0/management/openai-compatibility`)
4. GitTokenStore force-pushes a commit whose tree is missing all auth files
5. All auth files are permanently deleted from the remote repository

### Root cause (from source code analysis)

Two deficiencies in `internal/store/gitstore.go`:

1. **No tree integrity check before force push**: `commitAndPushLocked` does not verify that the new commit's tree contains all files present in the previous HEAD tree. After a failed pull leaves the index inconsistent, `worktree.Commit` includes staged deletions of files the caller did not intend to delete. `rewriteHeadAsSingleCommit` then replaces the entire HEAD tree, and `Force:true` push makes the data loss permanent.

2. **No self-healing after pull failure**: When `Pull` fails with `packfile not found`, the error is silently swallowed (line 219-221), but the local git index remains in a corrupted state with no recovery mechanism.

### Impact

- **Data loss**: All auth credentials and config files in the git store are permanently deleted from the remote repository
- **Service disruption**: Models become unusable after the next config reload
- **Difficult to diagnose**: The "delayed blast" pattern means the root cause is not immediately obvious

### Workaround

Regularly back up the config repository. If data loss occurs, restore from a previous commit using `git reset --hard` on the affected branch.

dong0713:
## Summary

After a failed `git pull` (e.g. `packfile not found` on Render Free), the local git index may contain staged deletions of `auths/*.json` / `config/*.yaml` files that the caller did not intentionally remove. When `commitAndPushLocked` is subsequently invoked — most commonly via `PersistConfig` after toggling a provider switch in the management API — `worktree.Commit` builds the new commit tree from the **full index**, so those unintended deletions are included. `rewriteHeadAsSingleCommit` then replaces the entire HEAD tree with this incomplete tree, and `Force: true` push permanently deletes the files from the remote repository.

This caused a real production data-loss incident: `auths/` and `config/` were wiped from the configuration repository, and the in-memory auth list kept serving requests until it was refreshed ~1 hour later, producing a delayed outage (10:06 data loss → 11:06 first 500/503).

Closes #4629.

## Root cause

`commitAndPushLocked` trusts the worktree index. A corrupted index (left behind by a failed pull) silently turns a single-file config persist into a multi-file destructive force push.

## Fix

Add `validateCommitTree`, invoked between `worktree.Commit` and `rewriteHeadAsSingleCommit` / force push:

1. Save the pre-commit HEAD hash before `worktree.Commit` advances HEAD.
2. After commit, compare the previous HEAD tree against the new commit tree.
3. Files present in the previous tree but missing from the new tree, and **not** in the caller-intended `relPaths`, are treated as accidental data loss.
4. Reject the operation (fail-closed) before any rewrite or force push reaches the remote.

This preserves the existing single-commit-history and force-push design — it only blocks the destructive push when the tree integrity check fails. Intentional deletions (e.g. `Delete` passing the target path in `relPaths`) continue to work.

### Resetting local HEAD after a rejected commit

`worktree.Commit` advances local HEAD *before* `validateCommitTree` runs. If the rejected commit were left as HEAD, the missing files would become the new baseline, and a later `PersistConfig` would no longer detect the loss — the follow-on force-push could still wipe the remote. On rejection we hard-reset branch/index/worktree back to the pre-commit tip (`prevHeadHash`), so the next persist still compares against the intact tree. The remote is never touched on the rejected path.

## What's protected

- `PersistConfig` updating `config/config.yaml` no longer drops unrelated `auths/*.json`.
- `PersistAuthFiles("Sync auth …")` no longer drops unrelated auths (the `Sync` path is not covered by the existing `Remove auth` watcher guard).
- Multi-file index corruption (the production scenario where several codex auths were staged for deletion) is rejected.
- Explicit `Delete` still removes only the targeted auth (positive control).
- A rejected commit no longer poisons the local baseline; a second persist after rejection still protects the remote.

## Tests

Regression tests in `internal/store/gitstore_test.go`:

- `TestPersistConfigDoesNotDropAuthAfterStagedIndexDeletion` — single-auth index corruption + config persist.
- `TestPersistConfigDoesNotDropSiblingAuthsAfterMultiFileIndexCorruption` — four-auth index corruption (production scenario).
- `TestSyncAuthDoesNotDropUnrelatedAuthAfterStagedIndexDeletion` — `Sync auth` watcher path with unrelated auth staged for deletion.
- `TestExplicitDeleteStillRemovesOnlyTargetedAuth` — positive control ensuring intentional delete still works.
- `TestRejectedCommitResetsLocalHeadSoSecondPersistStillProtectsRemote` — after a rejection, local HEAD is reset to the pre-commit tip and a second persist succeeds without dropping the auth.

All assertions verify the **remote** bare tree (via `assertRemoteTreePath`), not just local state, so they prove the force push did not wipe the files.

```
go test ./internal/store/ -timeout 180s
```

Result: `ok github.com/router-for-me/CLIProxyAPI/v7/internal/store` (all tests pass, including the existing watcher-removal and branch-checkout suites).

## Notes / follow-ups

This PR is a minimal, fail-closed safety net. The reset on rejection restores branch/index/worktree to the pre-commit tip but does not attempt deeper self-healing (e.g. re-cloning after `packfile not found`, or repairing the pull failure that seeds the corruption). Those are left for maintainer discussion to keep the data-loss fix narrowly scoped.

### DIFF
--- a/internal/store/gitstore.go
+++ b/internal/store/gitstore.go
@@ -933,6 +933,11 @@ func (s *GitTokenStore) commitAndPushLocked(message string, relPaths ...string) 
 		Email: "cliproxy@local",
 		When:  time.Now(),
 	}
+	// Save pre-commit HEAD hash for tree integrity validation after commit.
+	prevHeadHash := plumbing.ZeroHash
+	if prevHeadRef, errPrev := repo.Head(); errPrev == nil {
+		prevHeadHash = prevHeadRef.Hash()
+	}
 	commitHash, err := worktree.Commit(message, &git.CommitOptions{
 		Author: signature,
 	})
@@ -941,6 +946,27 @@ func (s *GitTokenStore) commitAndPushLocked(message string, relPaths ...string) 
 			return nil
 		}
 		return fmt.Errorf("git token store: commit: %w", err)
+	}
+	// Guard: refuse to rewrite+force-push if the new commit's tree is
+	// missing files that were present in the previous HEAD tree but are not
+	// part of an intentional deletion by the caller. This prevents a
+	// corrupted local index (e.g. after a failed pull due to packfile
+	// corruption) from causing permanent data loss on the remote.
+	if err := s.validateCommitTree(repo, prevHeadHash, commitHash, relPaths); err != nil {
+		// worktree.Commit already advanced local HEAD to the rejected
+		// commit. Reset branch/index/worktree back to the pre-commit
+		// tip so a later persist still compares against the intact
+		// tree; otherwise the missing files become the new baseline
+		// and a subsequent force-push can still wipe the remote.
+		if prevHeadHash != plumbing.ZeroHash {
+			if resetErr := worktree.Reset(&git.ResetOptions{
+				Commit: prevHeadHash,
+				Mode:   git.HardReset,
+			}); resetErr != nil {
+				return fmt.Errorf("git token store: %w (also failed to reset local HEAD after rejection: %v)", err, resetErr)
+			}
+		}
+		return fmt.Errorf("git token store: %w", err)
 	}
 	headRef, errHead := repo.Head()
 	if errHead != nil {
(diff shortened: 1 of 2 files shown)


dong0713:
Good catch — addressed in 6511fe2.

`worktree.Commit` advances local HEAD before `validateCommitTree` runs, so the rejected commit was indeed left as the new baseline. On rejection we now hard-reset branch/index/worktree back to the pre-commit tip (`prevHeadHash`) before returning, so the next `PersistConfig` still compares against the intact tree and the follow-on force-push can no longer wipe the missing files. The remote is never touched on the rejected path (the reset happens entirely before `rewriteHeadAsSingleCommit` / `Push`).

Added `TestRejectedCommitResetsLocalHeadSoSecondPersistStillProtectsRemote` to pin this: after a rejected persist it asserts local HEAD equals the pre-commit hash, then a second persist (without re-corrupting the index) succeeds and both auths remain on the remote. Full suite passes:

```
go test ./internal/store/ -timeout 180s
```

dong0713:
Thanks — that was a real compile error, but on an earlier (broken) push, not the current head.

`resolveRemoteDefaultBranchFromLocal` was always meant to return `resolvedRemoteBranch{name: target}, true`. An intermediate commit (`1bf4e30`) accidentally mangled the composite literal into `resolvedRemoteBranch{name: target, true}` while re-pushing the file; the current head `6511fe2` restores the correct `return resolvedRemoteBranch{name: target}, true` (unchanged from `dev`'s original code).

Verified locally — the package builds and the full store suite passes:

```
go build ./internal/store/
go test ./internal/store/ -timeout 180s
```

dong0713:
Closing this PR — #4629 is resolved by a2ff691, which tackles the root cause more thoroughly (pre-commit index reset + remote re-clone recovery). Happy to help if any follow-up surfaces.

Issue #4460, opened 2026-07-20
OpenAI Compatibility Provider Cannot Proxy /v1/responses Requests to Evomap


ZardLi1115:
Problem Description

I configured Evomap as an openai-compatibility provider in CLIProxyAPI v7.1.39:

openai-compatibility:
  - name: evomap
    base-url: https://api.evomap.ai/v1
    api-key-entries:
      - api-key: "..."
    models:
      - name: evomap-gpt-5.5
        alias: ""

CPA loads the configuration successfully, and GET /v1/models returns:

{
  "id": "evomap-gpt-5.5",
  "owned_by": "evomap"
}

However, calling the Responses API through CPA fails:

curl http://127.0.0.1:8317/v1/responses \
  -H "Authorization: Bearer <CPA_API_KEY>" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "evomap-gpt-5.5",
    "input": "Reply exactly: OK",
    "max_output_tokens": 16
  }'

The response is:

{
  "error": {
    "message": "The model provider couldn't process this request. Please check your request parameters.",
    "type": "upstream_error"
  }
}

The HTTP status code is 400, and Evomap returns a request ID.

Comparison
Using the same Evomap API key, the same model, and the same JSON payload, a direct request to https://api.evomap.ai/v1/responses returns 200 and generates a valid response.
Requests to /v1/chat/completions return 400 both directly and through CPA, which indicates that Evomap primarily supports the Responses API.
Therefore, the API key, model name, and base URL are all correct. The issue appears to be that CPA's openai-compatibility provider handles or transforms /v1/responses requests differently from a direct request.
Impact

Codex uses wire_api = "responses", so it cannot access evomap-gpt-5.5 through CPA's OpenAI-compatible Evomap provider, even though direct requests to Evomap's Responses API work correctly.

Expected Behavior

The openai-compatibility provider should transparently proxy /v1/responses requests, preserving the Responses API request body or transforming it correctly, so that the request produces the same result through CPA as it does when sent directly to Evomap.

luispater:
Thanks for the report. This is a configuration issue rather than a proxy defect.

`openai-compatibility` uses the Chat Completions upstream path for normal requests, so it is not the correct provider type for an upstream that only supports the Responses API. When the client uses `wire_api = "responses"`, configure the upstream under `codex-api-key` instead. The Codex executor sends the request to the upstream `/responses` endpoint.

[...]

Issue #4297, opened 2026-07-14
[BUG] openai-compatibility cannot use responses-only upstream providers


TogetsuDo:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[ ] No, it's another issue.

**If it's a request payload issue, you MUST know**
I do not have a full public request log to attach here, but I traced the behavior in the current source and can reproduce the compatibility gap conceptually.

**Describe the bug**
`CLIProxyAPI` exposes `POST /v1/responses` to clients, but when the upstream is configured through `openai-compatibility`, a provider that supports only OpenAI `Responses API` and does **not** support `POST /v1/chat/completions` is still not usable.

The problem appears to be that the client-facing `responses` support and the generic OpenAI-compatible upstream execution path are not fully aligned:

- client -> CPA: `POST /v1/responses` is supported
- CPA internal: `openai-response` format and handlers exist
- CPA -> generic `openai-compatibility` upstream: execution still tends to route through `/chat/completions` unless it is a special compact path

This makes responses-only OpenAI-compatible providers impossible to use through the normal `openai-compatibility` config.

**CLI Type**
openai-compatibility

**Model Name**
responses-only OpenAI-compatible upstream model

**LLM Client**
Cursor

**Request Information**
Conceptual request shape:

```bash
curl http://127.0.0.1:PORT/v1/responses \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <CPA_KEY>" \
  -d '{
    "model": "my-model",
    "input": "hello"
  }'
```

Current expectation from the upstream provider:

- accepts: `POST /v1/responses`
- may reject or not implement: `POST /v1/chat/completions`

Relevant source paths from current `main` that suggest the mismatch:

- `internal/api/server.go` registers `POST /v1/responses`
- `sdk/api/handlers/openai/openai_responses_handlers.go` handles `responses`
- `internal/runtime/executor/openai_compat_executor.go` still defaults generic upstream execution to `/chat/completions`, with only special handling for `responses/compact`

**Expected behavior**
When a provider in `openai-compatibility` is configured as responses-native or responses-only, CPA should be able to forward compatible requests to the upstream `POST /v1/responses` endpoint instead of requiring `/v1/chat/completions`.

One possible implementation would be a provider-level switch, for example:

- `prefer-responses: true`
or
- `responses-only: true`

[...]

PR #4637, opened 2026-07-28
feat(openai-compat): 支持可选原生 Responses 协议

pjpjq:
## Summary

- add an opt-in `wire-api: responses` setting to `openai-compatibility`
- route only inbound Responses requests to the upstream `/responses` endpoint while keeping Chat/Claude inputs on `/chat/completions`
- preserve native Responses JSON and SSE fields, terminal semantics, and nested usage accounting
- keep default, unknown-value, and `/responses/compact` behavior unchanged

## Why this is different from #4297 and #4460

Those issues correctly recommend `codex-api-key` for **Responses-only** upstreams. This change targets hybrid OpenAI-compatible providers that expose both Chat Completions and Responses with the same credential and model namespace.

Configuring the same hybrid credential twice under `openai-compatibility` and `codex-api-key` creates duplicate model/auth routing and forces clients to use separate aliases. The provider-level wire switch instead selects the native upstream protocol from the inbound request format:

```yaml
openai-compatibility:
  - name: hybrid-provider
    base-url: https://provider.example/v1
    wire-api: responses
```

- inbound `/v1/responses` -> upstream `/responses`
- inbound Chat/Claude -> upstream `/chat/completions`

Omitting `wire-api`, or setting an unsupported value, preserves the existing Chat Completions behavior.

## Correctness details

- selected-auth attributes are the request-level source of truth for both interceptor `ToFormat` and the actual upstream endpoint, avoiding hot-reload mismatch windows
- native Responses streams preserve `event`, `id`, `retry`, and `data` lines without synthetic `[DONE]`
- `response.completed` and `response.incomplete` close the downstream stream immediately, even if the upstream keeps the TCP connection open
- truncated streams and explicit failure events are reported as errors rather than successful requests
- usage parsing supports `response.completed.response.usage`

## Validation

- `go test ./... -count=1`
- targeted package tests for API, executor, config, watcher, management, and auth
- targeted `go test -race` for native Responses streaming and selected-auth format resolution
- `go build -o <temporary>/cli-proxy-api ./cmd/server`
- `gofmt -w .`
- `git diff --check`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -432,6 +432,7 @@ nonstream-keepalive-interval: 0
 #     disabled: false # optional: set to true to disable this provider without removing it
 #     prefix: "test" # optional: require calls like "test/kimi-k2" to target this provider's credentials
 #     base-url: "https://openrouter.ai/api/v1" # The base URL of the provider.
+#     wire-api: "responses" # optional: use the upstream /responses API; omitted or unsupported values keep /chat/completions behavior
 #     disable-cooling: false # optional: per-provider override for auth/model cooldown scheduling
 #     headers:
 #       X-Custom-Header: "custom-value"
--- a/internal/api/handlers/management/config_auth_index.go
+++ b/internal/api/handlers/management/config_auth_index.go
@@ -44,6 +44,7 @@ type openAICompatibilityWithAuthIndex struct {
 	Disabled       bool                                     `json:"disabled"`
 	Prefix         string                                   `json:"prefix,omitempty"`
 	BaseURL        string                                   `json:"base-url"`
+	WireAPI        string                                   `json:"wire-api,omitempty"`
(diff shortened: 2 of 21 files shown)


karuboniru:
Hi, I am insterested why this PR has been closed, while I think it can be quite interesing to include this. I started with [lettiing gpt to do the simliar for me](https://github.com/karuboniru/CLIProxyAPI/commit/09ca019cf254551b43ee5a7bcefe6abdebcde1b5) but found PRs implementing mostly equivlent features being closed. 

Since I am using CLIProxyAPI to manage different model providers in codex, and by passing-thorough `/responses` calls preserves some upstream tool calls (e.g. [Deepseek's `web_search`](https://api-docs.deepseek.com/zh-cn/guides/responses_api) in codex).

PR #4654, opened 2026-07-29
feat(routing): session-affinity max-requests with per-model/provider rules

dangxia:
## Summary

Add **session-affinity max-requests** so sticky session→auth bindings can rebind after a request cap, with optional **per-provider / per-model** overrides.

## Motivation

Session affinity keeps a client on the same credential for continuity, but long-lived sessions can pin traffic to one account indefinitely (until TTL expires or the auth becomes unavailable). Operators need a way to:

- Soft-rotate credentials after N sticky picks
- Apply different caps for specific models/providers (e.g. expensive models vs cheap ones)

without turning off session affinity entirely.

### Config

- `routing.session-affinity-max-requests` — global default (`<=0` / omit / `-1` = unlimited)
- `routing.session-affinity-rules[]` — optional overrides:
  - `provider` (optional)
  - `model` (optional; thinking-suffix stripped for matching)
  - `max-requests`

**Match specificity (most specific wins):**  
`provider+model` > `model-only` > `provider-only` > global default

Equal-specificity rules keep declaration order (first match wins at equal score).

### Runtime

- Session cache tracks **hit count** per binding (shared across session aliases)
- On sticky cache hit, if `hits > max-requests`, invalidate the binding and re-pick via the fallback selector, then rebind
- Rebind also still occurs when the bound auth is unavailable (existing behavior)
- Hot-reload: routing state equality includes max-requests + compiled rules key so config changes rebuild the selector

### Docs / example

- `config.example.yaml` documents the new knobs and the **fill-first + max-requests** interaction

## Behavior notes / caveats

> **Rebind uses the same `routing.strategy` selector.**  
> With **`fill-first`**, rebind still returns the first available credential, so **max-requests rarely rotates accounts** unless the current one is unavailable (cooldown / quota / disabled).  
> Prefer **`round-robin`** or **`weighted-round-robin`** when using max-requests for intentional rotation.

Only effective when `routing.session-affinity: true`.

### Example

```yaml
routing:
  strategy: round-robin
  session-affinity: true
  session-affinity-ttl: "1h"
  session-affinity-max-requests: -1   # unlimited global default
  session-affinity-rules:
    - model: "grok-4.5"
      provider: "xai"       # optional
      max-requests: 20
```

## Test plan

- [x] Unit: max-requests rebinds under round-robin after N sticky picks
- [x] Unit: fill-first does **not** rotate while first auth remains available (documented intentional)
- [x] Unit: per-model / per-provider rule resolution specificity
- [x] Unit: rules key stable across thinking-suffix / case normalization; order matters at equal specificity
- [ ] Manual: enable session-affinity + max-requests with RR/WRR, confirm log lines for rebind (`max-requests reached`) and hit counters on cache hits
- [ ] Manual: hot-reload change of max-requests / rules rebuilds routing without restart

## Files

- `internal/config/config_types.go` — config types
- `config.example.yaml` — example + caveats
- `sdk/cliproxy/auth/session_cache.go` — hit counting + alias-group invalidate
- `sdk/cliproxy/auth/session_affinity_rules.go` — compile / resolve
- `sdk/cliproxy/auth/selector.go` — max-requests rebind path
- `sdk/cliproxy/service_config.go` — wire config → selector + hot-reload equality
- tests for rules, max-requests, and routing state

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -212,6 +212,21 @@ routing:
   session-affinity: false # default: false
   # How long session-to-auth bindings are retained. Default: 1h
   session-affinity-ttl: "1h"
+  # Max sticky requests per session→auth binding before rebind (global default).
+  # Only applies when session-affinity is true.
+  # <=0 (omit, 0, or -1): unlimited. Rebind also happens when TTL expires (whichever first).
+  #
+  # NOTE: rebind calls the same routing.strategy selector again. With fill-first,
+  # that still returns the first available credential, so max-requests rarely
+  # rotates accounts unless the current one is unavailable (cooldown/quota/disabled).
+  # Prefer round-robin or weighted-round-robin when using max-requests for rotation.
+  session-affinity-max-requests: -1
+  # Optional per-model / per-provider overrides. More specific wins:
+  # provider+model > model-only > provider-only > global default.
+  # session-affinity-rules:
+  #   - model: "grok-4.5"
+  #     provider: "xai"       # optional
+  #     max-requests: 20
 
 # Codex provider behavior.
 codex:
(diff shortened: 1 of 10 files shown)


dangxia:
@codex review

dangxia:
@codex review

PR #4673, opened 2026-07-30
fix(security): write credential and config files with 0600

### LINKED ISSUE #4672: Credential and config files are written world-readable (os.Create + umask)
## Summary

Auth token files and `config.yaml` are written with `os.Create`, which requests mode `0666` and then lets the process umask decide the final permissions. With the common `umask 002` (default on Ubuntu/Debian for user-private groups) the resulting files are `-rw-rw-r--`; with `umask 022` they are `-rw-r--r--`. Either way every local user on the machine can read them.

This is inconsistent with the rest of the codebase, which already treats this material as secret:

- `os.MkdirAll(filepath.Dir(authFilePath), 0700)` immediately above each of the affected writes
- `sdk/auth/filestore.go`, `internal/api/handlers/management/auth_files_crud.go`, `internal/store/*` all use `0o600`

## Impact

The files contain long-lived provider credentials:

- `~/.cli-proxy-api/*.json` — OAuth access **and refresh** tokens for Codex / Claude / Kimi / xAI / Vertex. A refresh token is enough to mint new access tokens indefinitely, off-host.
- `config.yaml` — the `api-keys` list (full proxy access) and the `remote-management.secret-key` bcrypt hash (offline cracking target).

Observed on a real deployment, `umask 002`, CLIProxyAPI v7.2.x:

```
$ ls -l ~/.cli-proxy-api/
-rw-rw-r-- 1 user user 4301 codex-....json     <- 57 files, all group/world readable
-rw------- 1 user user  595 antigravity-....json  <- written via filestore, correct
$ ls -l config.yaml
-rw-r--r-- 1 user user 23560 config.yaml
```

Note the mix: credentials persisted through `sdk/auth/filestore.go` are correctly `0600`, so the exposure depends on which code path last wrote the file.

## Affected code

| File | Line |
| --- | --- |
| `internal/auth/codex/token.go` | `f, err := os.Create(authFilePath)` |
| `internal/auth/claude/token.go` | same |
| `internal/auth/kimi/token.go` | same |
| `internal/auth/xai/token.go` | same |
| `internal/auth/vertex/vertex_credentials.go` | same |
| `internal/config/config_yaml.go` | `os.Create(configFile)`, two call sites |
| `internal/api/handlers/management/config_basic.go` | `os.OpenFile(path, ..., 0644)` |

## Suggested fix

Open these files with an explicit `0600` instead of relying on the umask. Because `O_CREATE` only applies the mode to *newly created* files, an explicit `Chmod` after opening is also needed, otherwise deployments that already have world-readable token files stay exposed forever (the files are rewritten on every token refresh, always in place).

PR: #<pr>

SimZhou:
Fixes #4672.

## Problem

`os.Create` requests mode `0666` and defers to the process umask, so auth token files and `config.yaml` land as `-rw-rw-r--` (umask 002) or `-rw-r--r--` (umask 022) — readable by every local user. These files hold OAuth refresh tokens, the `api-keys` list, and the `remote-management.secret-key` hash.

It is also inconsistent within the codebase: the `os.MkdirAll(..., 0700)` right above each write, `sdk/auth/filestore.go`, and `internal/store/*` all already use owner-only modes.

## Change

Add `misc.CreateSecretFile`, a drop-in replacement for `os.Create` for secret-bearing files, and use it in the five token stores plus the three config writers.

Two details worth reviewing:

- **Explicit `Chmod` after open.** `O_CREATE` only applies the mode to newly created files. Token files are rewritten in place on every refresh, so without the `Chmod` an existing deployment would keep its world-readable files indefinitely — the fix would only help fresh installs. `os.IsPermission` errors are tolerated so a file owned by another user (e.g. a root-managed config) still writes rather than hard-failing.
- **Placement in `internal/misc`.** That package has no internal imports, so `internal/config` can use it without an import cycle, and the auth packages already import it for `LogSavingCredentials`.

## Verification

```
go build ./...
go vet ./internal/misc/ ./internal/auth/... ./internal/config/ ./internal/api/handlers/management/
go test ./internal/misc/ -run TestCreateSecretFile
```

All clean. The added test runs under `umask 002` and covers both paths: a fresh file is created `0600`, and a pre-existing `0664` file is tightened to `0600`. It is `//go:build unix` since `syscall.Umask` is not available on Windows.

I have not exercised the full OAuth login flow end to end — the change is confined to how the file handle is opened, and the callers are otherwise untouched.

### DIFF
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/misc"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/util"
 	sdkconfig "github.com/router-for-me/CLIProxyAPI/v7/sdk/config"
 	log "github.com/sirupsen/logrus"
@@ -93,7 +94,7 @@ func (h *Handler) GetLatestVersion(c *gin.Context) {
 
 func WriteConfig(path string, data []byte) error {
 	data = config.NormalizeCommentIndentation(data)
-	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	f, err := misc.CreateSecretFile(path)
 	if err != nil {
 		return err
 	}
--- a/internal/auth/claude/token.go
+++ b/internal/auth/claude/token.go
@@ -67,7 +67,7 @@ func (ts *ClaudeTokenStorage) SaveTokenToFile(authFilePath string) error {
 	}
 
 	// Create the token file
-	f, err := os.Create(authFilePath)
+	f, err := misc.CreateSecretFile(authFilePath)
 	if err != nil {
 		return fmt.Errorf("failed to create token file: %w", err)
 	}
--- a/internal/auth/codex/token.go
+++ b/internal/auth/codex/token.go
@@ -60,7 +60,7 @@ func (ts *CodexTokenStorage) SaveTokenToFile(authFilePath string) error {
 		return fmt.Errorf("failed to create directory: %v", err)
 	}
 
-	f, err := os.Create(authFilePath)
+	f, err := misc.CreateSecretFile(authFilePath)
 	if err != nil {
 		return fmt.Errorf("failed to create token file: %w", err)
 	}
--- a/internal/auth/kimi/token.go
+++ b/internal/auth/kimi/token.go
@@ -87,7 +87,7 @@ func (ts *KimiTokenStorage) SaveTokenToFile(authFilePath string) error {
 		return fmt.Errorf("failed to create directory: %v", err)
 	}
 
-	f, err := os.Create(authFilePath)
+	f, err := misc.CreateSecretFile(authFilePath)
 	if err != nil {
 		return fmt.Errorf("failed to create token file: %w", err)
 	}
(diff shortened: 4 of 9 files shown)


jroth1111:
This overlap is worth noting: commit 7972f78 adds `atomicWriteFile` to `sdk/auth/filestore.go` — an atomic write (temp + 0600 + fsync + rename + dir fsync, Windows fallback) that also fixes the credential permission issue, but only for the filestore path. Your PR fixes all `SaveTokenToFile` callers across all providers, which is broader coverage. The atomic safety in filestore is complementary — it eliminates the truncate+write race even without breaking the atomic rename assumption.

If merged, both changes apply cleanly (they touch different callers). No conflicts.

jroth1111:
## Adversarial review findings

### 1.  tolerance silently preserves world-readable files (medium)

```go
if err = f.Chmod(SecretFileMode); err != nil && !os.IsPermission(err) {
    log.Debugf(...)
}
```

If `Chmod` fails with `EPERM` (e.g. NFS ACLs that let the process write but not chmod), the file keeps its prior wider permissions (e.g. `0644`) and the gap goes to a debug log. A process that can write a credential file but not tighten its permissions is unusual enough that failing loudly is safer than silently preserving world-readability.

**Suggestion**: remove the `os.IsPermission` tolerance so `Chmod` failure surfaces as a hard error. If the exotic NFS/mount scenario is a real concern, log a structured warning at `Warn` level instead of swallowing silently.

### 2. Test `syscall.Umask` is process-global (minor)

The test temporarily sets `umask 0o002` to verify the file always lands `0600`. This is correct on its own but races with any other concurrent test in `internal/misc` that also touches umask. Not a blocker — the current test suite doesn't have such a concurrent test — but worth documenting as a known fragility if another test is added.

**No other defects found.** The `createSecretFile` design is clean, all 8 callers are consistent, and the import dependency (`internal/config → internal/misc`) is verified non-cyclical.

luispater:
Thank you for the contribution. However, this is not considered an issue that requires modification in this project. Closing this pull request.

PR #4674, opened 2026-07-30
feat(routing): quota-aware selection, per-subagent session affinity, 529/headerless-429 failover

johncpakin:
### The problem

I run Claude Code sessions that fan out to 50–100 subagents across several Claude Max accounts. Three compounding issues concentrated all of that load onto a single account and turned rate limits into hard failures:

1. **Subagents inherit the parent's `session_id`** (in `metadata.user_id`), so session affinity binds an entire fan-out to one credential. Observed: 43 of 47 requests on one binding.
2. **Selection is quota-blind.** Round-robin happily picks an account at 100% of its model-scoped weekly limit while another sits at 1%. Model scoping matters: an account measured 75% overall but 100% on one model tier — account-level health checks miss this entirely.
3. **529 and headerless 429 never retry.** Anthropic OAuth-account 429/529 responses carry no `Retry-After`; `shouldRetryAfterError` requires one, so these fail on the first attempt without trying other credentials.

### The changes (3 commits, independently revertable)

- **`fix(auth)`: 529 + headerless-429 failover.** 529 is treated like the other transient upstream errors (cooldown via the existing transient path); a 429/529 with no usable `Retry-After` gets a short default wait instead of `(0, false)`. Existing `Retry-After` behavior unchanged.
- **`feat(routing)`: `session-id-mode: content-hash`.** Opt-in. For Claude Code traffic, the affinity key becomes `session_id + hash(full first message content)`. Subagents (same session_id, different prompts) bind independently; every turn of one conversation still pins to its credential, so per-credential prompt caches stay warm. Full-content hashing is deliberate: subagents share their first few KB verbatim, so prefix hashing collapses them.
- **`feat(routing)`: `strategy: quota-aware`.** A background poller fetches each Claude credential's usage from `/api/oauth/usage` (per model tier) and the selector picks proportionally to remaining headroom. Never argmax (avoids stampeding one account). Pick() does no I/O. Degrades to round-robin when quota is unknown, and only over quota-unknown credentials — known-exhausted ones stay excluded on stale data. The poller backs off hard (15m→2h) if the usage endpoint rate-limits, since that endpoint's per-IP bucket is shared and headerless.

Both new behaviors are opt-in config (`routing.strategy: quota-aware`, `routing.session-id-mode: content-hash`); defaults are unchanged. Composes with session affinity as its fallback selector rather than replacing it.

### Relation to weighted-round-robin

`quota-aware` is a sibling strategy to the recently-merged `weighted-round-robin`, not a replacement: WRR distributes proportionally to static, operator-assigned integer weights, while quota-aware distributes proportionally to live remaining quota headroom fetched from Anthropic's OAuth usage endpoint per Claude credential (session window, weekly-all, and model-scoped weekly tiers). The two share no state — quota-aware is wired only into the `routing.strategy` switch and runs through the existing legacy selector pick path rather than the built-in scheduler fast path, so the WRR scheduler machinery (credential weights, smooth-weighted state, weight validation) is untouched. When quota data is unknown or stale the strategy degrades to round-robin over unknown credentials only, keeping known-exhausted credentials excluded.

One item flagged for maintainer review: the background poller uses a 15s `http.Client` timeout, which sits in tension with the repo's timeouts-only-during-credential-acquisition rule. Rationale (in a code comment): it's background credential-quota polling that never sits on a request path, and an unbounded hang there is worse. Happy to change if you'd prefer a different mechanism.

### Validation (live, 4 Claude Max accounts)

- 8 concurrent requests sharing one `session_id` with distinct prompts: 4 distinct credential bindings (previously 1). Identical repeat request: affinity cache hit on its own binding.
- 16 requests on a model tier where 2 of 4 accounts were exhausted: zero requests to the exhausted accounts, remainder split ~proportional to headroom.
- Observed a live 429 on one credential roll over to another and return 200 to the client (previously a client-visible failure).
- `go build ./...`, `go vet`, package tests green; new tests cover the retry gate, cooldown marking, proportional distribution, tier exclusion, stale-data degradation, and content-hash extraction.

Companion UI PR (quota visibility in the management panel): router-for-me/Cli-Proxy-API-Management-Center#363

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -212,6 +212,11 @@ routing:
   session-affinity: false # default: false
   # How long session-to-auth bindings are retained. Default: 1h
   session-affinity-ttl: "1h"
+  # How the session identity used for affinity is derived. Default: "default".
+  # "content-hash" folds a hash of the first message contents into Claude Code
+  # session IDs, so subagents sharing a parent session_id but carrying different
+  # prompts bind to distinct credentials while each conversation stays pinned.
+  session-id-mode: "default" # default, content-hash
 
 # Codex provider behavior.
 codex:
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -200,10 +200,13 @@ quota-exceeded:
 
 # Routing strategy for selecting credentials when multiple match.
 routing:
-  strategy: "round-robin" # round-robin (default), weighted-round-robin, fill-first
+  strategy: "round-robin" # round-robin (default), weighted-round-robin, fill-first, quota-aware
   # weighted-round-robin uses each credential's integer weight (default 1, maximum 1,000,000).
   # Non-positive weights exclude the credential while this strategy is active.
   # For OAuth/file credentials, add a top-level numeric "weight" field to the auth JSON.
+  # quota-aware polls the Anthropic OAuth usage endpoint in the background and picks
+  # Claude credentials proportionally to their remaining quota headroom; credentials with
+  # unknown quota fall back to round-robin, and exhausted ones are excluded.
   # Enable universal session-sticky routing for all clients.
   # Explicit Claude Code, Codex, OpenCode, and pi session headers are preferred,
   # followed by prompt_cache_key, Responses conversation IDs, legacy body IDs,
--- a/internal/api/handlers/management/config_basic.go
+++ b/internal/api/handlers/management/config_basic.go
@@ -288,6 +288,8 @@ func normalizeRoutingStrategy(strategy string) (string, bool) {
 		return "weighted-round-robin", true
 	case "fill-first", "fillfirst", "ff":
 		return "fill-first", true
+	case "quota-aware", "quotaaware", "qa":
+		return "quota-aware", true
 	default:
 		return "", false
 	}
--- a/internal/config/config_types.go
+++ b/internal/config/config_types.go
@@ -203,7 +203,8 @@ type QuotaExceeded struct {
 // RoutingConfig configures how credentials are selected for requests.
 type RoutingConfig struct {
 	// Strategy selects the credential selection strategy.
-	// Supported values: "round-robin" (default), "weighted-round-robin", "fill-first".
+	// Supported values: "round-robin" (default), "weighted-round-robin",
+	// "fill-first", "quota-aware".
 	Strategy string `yaml:"strategy,omitempty" json:"strategy,omitempty"`
 
 	// SessionAffinity enables universal session-sticky routing for all clients.
--- a/internal/config/config_types.go
+++ b/internal/config/config_types.go
@@ -216,6 +216,13 @@ type RoutingConfig struct {
 	// SessionAffinityTTL specifies how long session-to-auth bindings are retained.
 	// Default: 1h. Accepts duration strings like "30m", "1h", "2h30m".
 	SessionAffinityTTL string `yaml:"session-affinity-ttl,omitempty" json:"session-affinity-ttl,omitempty"`
+
+	// SessionIDMode controls how the session identity used for affinity is derived.
+	// "" or "default": current behavior (Claude Code session_id alone).
+	// "content-hash": combine the Claude Code session_id with a hash of the first
+	// message contents, so subagents that share a parent session_id but differ in
+	// prompts bind to distinct auths while each conversation stays pinned.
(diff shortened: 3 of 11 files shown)


johncpakin:
Both Codex findings addressed in 7906a0e8:

- **Stale poll targets**: `updatePollTargets` now stamps a `lastSeen` on every target and expires targets (plus their cached quota) not offered to `Pick` for 15 minutes — removed or rotated-out credentials stop consuming the shared usage-endpoint allowance. The window is deliberately longer than typical cooldown filtering so temporarily-cooled auths keep their quota data. New test: `TestQuotaAwareSelector_StalePollTargetsExpire`.
- **Client timeout**: the poller's `http.Client` no longer has a fixed timeout. Each fetch is bounded by a per-request context deadline that `Stop()` also cancels, so a wedged poll can't stall the refresh loop and in-flight requests abort immediately on shutdown/selector swap — as suggested, and in line with AGENTS.md.

johncpakin:
Third finding addressed in 99c1fe09: usage polls now send the same Claude Code UA/X-App/X-Stainless fingerprint as the executor's OAuth request path (verified a fingerprinted request returns 200 where a generic client draws 429s). The values are mirrored as local constants in quota_selector.go because importing the executor's helps package from sdk/cliproxy/auth would create an import cycle — happy to relocate the profile constants to a shared package instead if you'd prefer a single source of truth.

jroth1111:
Commit 7972f78 adds a simpler, always-on antigravity quota-aware filter to `availableAuthsForRouteModel`: `antigravityQuotaWindowForModel` is called for each auth during selection; accounts below `antigravityQuotaSoftFloor` are held in a separate priority bucket and only promoted when no healthier account exists at the same priority. This is narrower than your PR (antigravity only, no config toggle) and operates at the individual auth level rather than the conductor-manager routing strategy level.

Your configurable `routing.strategy: quota-aware` for all providers is the right general solution. Our per-auth antigravity filter can serve as the antigravity-specific policy under your broader framework once merged — they coexist cleanly.

jroth1111:
## Adversarial review findings

### 1. Non-Claude providers in quota-aware mode degrade silently (medium)

`QuotaAwareSelector` gates on `auth.Provider == \"claude\"` — non-Claude auths receive zero weight and fall through to the round-robin fallback. If someone configures `routing.strategy: quota-aware` on a multi-provider manager, Claude auths get quota-weighted selection while everything else silently round-robins with no log signal.

**Suggestion**: add a one-time `Warn` log in `Pick` when a non-Claude provider is encountered — something like `\"quota-aware: skipping non-claude provider=%s — only Anthropic auths are weighted by quota\"`.

### 2. Stale-but-healthy quota data is discarded entirely (minor)

At `headroom()` ~L348: when `age > quotaStaleAfter` but `age <= quotaExclusionMaxAge\" and the auth was NOT exhausted when last seen, the function returns `(0, false)` — treating it as unknown despite perfectly good \"90% free\" data. This is the safer choice (a stale credential whose token rotated out between cycles shouldn't accumulate weight), but it means a brief usage-endpoint outage causes all weighted selection to collapse to round-robin even for mostly-free accounts.

**Not a bug — deliberate safety trade-off.** Worth documenting in the code comment that this is intentional.

### 3. `quotaPollStainlessRuntime` hardcoded (`\"v24.3.0\"`) will drift from executor (minor)

`quotaPollUserAgent` and friends mirror `helps/claude_device_profile.go` but cannot import them. On the next agent version bump the usage endpoint may start 429-ing because the fingerprint stopped matching.

**Suggestion**: add a doc comment enumerating the source path: `// Keep in sync with internal/runtime/executor/helps/claude_device_profile.go` — same pattern used elsewhere in the codebase.

### 4. No structural bugs found

- `gjson` is already in `go.mod` and imported by `sdk/translator` — no import cycle risk.
- Background goroutine lifecycle: `startOnce` / `stopOnce` + `StoppableSelector` interface handling in `service_config.go` correctly prevents leaks on config hot-reload.
- Session ID content-hash mode is deterministic, stable across calls, and the fallback-inherit mechanism correctly binds second-turn subagent requests.
- 529/headerless-429 handling: `defaultRetryWaitWithoutRetryAfter` + `statusOverloaded` constant + `conductor_cooldown.go` switch cases are consistent and tested.
- `extractSessionIDsWithMode` and `claudeSessionIDs` correctly distinguish the three modes and don't double-count or create collisions.

PR #4691, opened 2026-07-31
fix(responses): rewrite keepalive events as SSE comments

SnowGlow-aww:
## Summary

- Rewrite upstream Responses events with a top-level `type` of `keepalive` as standards-compliant SSE comments.
- Prevent strict OpenAI Responses clients from rejecting the stream with an unknown event variant while preserving streaming behavior.
- Add regression coverage for both fully framed and delimiterless keepalive chunks, and verify that the following terminal event is preserved.

## Why

Codex upstream can emit heartbeat frames such as:

```text
data: {"type":"keepalive","sequence_number":1}
```

`keepalive` is not an OpenAI Responses stream event type, so strict clients can fail while deserializing it. SSE comments are the protocol-level heartbeat mechanism and are safely ignored by event parsers.

## Testing

- `go test ./sdk/api/handlers/openai -run 'TestForwardResponsesStream(RewritesKeepaliveEventsAsSSEComments|PreservesValidFullSSEEventChunks|ReassemblesSplitSSEEventChunks)$' -count=1 -v`

Full-suite note: `go test ./...` passed all packages except the environment-sensitive WebRTC test `TestPionMediaRelayBridgesAudioAndDataChannel` in `internal/client/codex/live`, where the upstream DataChannel was not created after macOS denied inbound network access to the temporary test binary. This package is outside the change set.

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -158,6 +158,8 @@ func (f *responsesSSEFramer) repairFrame(frame []byte) []byte {
 	}
 
 	switch eventType {
+	case "keepalive":
+		return []byte(": keep-alive\n\n")
 	case "response.output_item.done":
 		f.recordOutputItem(payload)
 	case "response.completed":
--- a/sdk/api/handlers/openai/openai_responses_handlers.go
+++ b/sdk/api/handlers/openai/openai_responses_handlers.go
@@ -107,6 +107,8 @@ func (f *responsesSSEFramer) repairFrame(frame []byte) []byte {
 	}
 
 	switch gjson.GetBytes(payload, "type").String() {
+	case "keepalive":
+		return []byte(": keep-alive\n\n")
 	case "response.output_item.done":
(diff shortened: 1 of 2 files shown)


gchigoo:
Confirmed against CLIProxyAPI v7.2.112 with Grok Build 0.2.117 using the Responses backend.

Observed behavior before the patch:
- 8 strict-deserialization failures across 5 Grok sessions: `unknown variant 'keepalive'`
- CLIProxyAPI logged HTTP 200 because the SSE response had already started
- A deterministic Grok test against `data: {"type":"keepalive","sequence_number":1}` reproduced the same non-retryable error

Validation of this patch on the v7.2.112 tag:
- The new regression test fails before the change and passes after it
- `go test ./sdk/api/handlers/openai -count=1` passes
- A native Linux amd64 build succeeds
- After deployment, `/v1/responses` completed normally and a resumed Grok session with 88,840 input/cache tokens completed successfully

Rewriting the private heartbeat as an SSE comment is the right compatibility boundary for strict Responses clients. +1 for merge.

PR #4693, opened 2026-07-31
feat(plugin-store):  add custom proxy and accelerator support for plugin store

gwdgithubnom:
# Plugin Store: Custom Proxy and Web Accelerator Support

## Background
The CLIProxyAPI plugin store fetches plugin manifests and artifacts from GitHub repositories (registry JSON, release archives, raw content, and GitHub CDN).
For deployments in mainland China, GitHub is frequently slow or unreachable, making plugin discovery, download, and update fail without a relay. Users behind restrictive network policies face the same problem.

A global proxy (`proxy-url`) can tunnel everything, but that makes all request (the registry metadata and API calls) share the same proxy egress IP, which triggers GitHub 403 secondary rate limits that block the entire deployment.

Plugin Store proxy support solves this by letting each admin choose a routing strategy:

- `custom` — dedicated proxy for plugin traffic only (clean IP, no rate limit)
- `system` — reuse the global proxy-url (simpler setup)
- `accelerator` — rewrite GitHub download URLs through a prefix (e.g. `ghproxy`) so artifact downloads bypass GitHub direct access entirely; registry and API calls still go direct, avoiding 403s
- `none` — unset/default; fall back to the global proxy-url when present, otherwise connect directly. Preserves legacy behavior for configs that never set `plugin-proxy`.
- `direct` — connect directly, bypassing the global proxy-url even when one is set

timeout

## Summary
This PR adds backend support for plugin store proxy configuration so toolbars and management UIs can control how plugin list/install traffic is routed. It also introduces a web accelerator mode that rewrites GitHub resource URLs (releases, raw, CDN) through a configurable prefix, independent of the global OAuth/proxy setup.

### Backend Dependency
This backend change supported by frontend PR [Cli-Proxy-API-Management-Center#366](https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/366) to be merged and deployed.

### Related Issues
- Backend PR: https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/366

## What This Enables
- **Management endpoints** — `GET`/`PUT`/`PATCH` `/v0/management/plugin-proxy` to read or update the proxy mode, plus a `POST` validate helper. - **Status-based routing** — the `plugin-proxy` config chooses between `direct`, `none` (fallback), `custom` (dedicated HTTP(S)/SOCKS proxy), `system` (reuse global proxy-url), or `accelerator` (rewrite GitHub artifact URLs through a prefix).
- **Accelerator safety** — accelerator rewriting targets only download artifact URLs (GitHub releases, raw, CDN). Registry, metadata, and API requests are never rewritten, avoiding GitHub 403 secondary rate limits that shared proxy egress IPs otherwise trigger.

## Design Decisions

- `PluginProxyConfig.Status int` is serialized as `json:"status"` (without `omitempty`) so a value of `0` (none) is never ambiguous after round-trip.
- The PUT handler only validates `url` / `accelerator` for `custom` and `accelerator` modes; `direct`/`none`/`system` go through unvalidated. The default case in the status switch explicitly maps so a future extension does not silently skip validation. 
- Normalization clamps unknown status values to `none` (0) rather than passing them through as-is. `direct` (-1) is preserved as a first-class status.
- `none` (0, the zero value) means "unset/default": it falls back to the global `proxy-url` when present, so legacy configs that never set `plugin-proxy` keep using the system proxy. `direct` (-1) is the explicit "no proxy, connect directly" choice.
- Runtime artifact/resource URLs are allowed to carry temporary signed query parameters such as GitHub CDN `?token=`. Registry and manifest URLs still delete sensitive query keys so long-lived credentials are never stored.
- Artifact downloads retry on transient connection errors (unexpected EOF, connection reset) up to 3 attempts.  `context.Canceled` / `context.DeadlineExceeded` fail immediately without retry.

success

### DIFF
--- a/internal/api/handlers/management/plugin_proxy.go
+++ b/internal/api/handlers/management/plugin_proxy.go
@@ -1,0 +1,195 @@
+package management
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/pluginstore"
+	"github.com/router-for-me/CLIProxyAPI/v7/sdk/proxyutil"
+)
+
+// GetPluginProxy returns the dedicated plugin-store proxy/accelerator setting
+// and the current global proxy-url (for system-proxy UI display).
+func (h *Handler) GetPluginProxy(c *gin.Context) {
+	if h == nil || h.cfg == nil {
+		c.JSON(http.StatusOK, gin.H{
+			"plugin-proxy": config.PluginProxyConfig{Status: config.PluginProxyStatusNone},
+			"proxy-url":    "",
+			"effective":    "",
+			"accelerator":  "",
+		})
+		return
+	}
+
+	h.mu.Lock()
+	pluginProxy := config.NormalizePluginProxyConfig(h.cfg.PluginProxy)
+	proxyURL := strings.TrimSpace(h.cfg.ProxyURL)
+	effective := config.EffectivePluginStoreProxyURL(h.cfg)
+	accelerator := config.EffectivePluginStoreAcceleratorBase(h.cfg)
+	h.mu.Unlock()
+
+	c.JSON(http.StatusOK, gin.H{
+		"plugin-proxy": pluginProxy,
+		"proxy-url":    proxyURL,
+		"effective":    effective,
+		"accelerator":  accelerator,
+	})
+}
+
+// PutPluginProxy updates plugin-proxy url/status/accelerator.
+// Status: -1=direct, 0=none/fallback, 1=custom proxy, 2=system, 3=accelerator.
+// Custom (status=1) requires a valid traditional proxy URL.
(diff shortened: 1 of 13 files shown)


gwdgithubnom (commented):
All three P2 suggestions have been addressed:

1. **Proxy backward compatibility** (`internal/api/handlers/management/plugin_store.go:521`) — Added `PluginProxyStatusNone` (0) to fall back to the global `proxy-url` logic, leaving existing configurations unaffected. Also introduced `PluginProxyStatusDirect` (-1) for explicit direct connection scenarios.

2. **Artifact retry covering `context.Canceled`** (`internal/pluginstore/github.go:248`) — Removed the separate checks for `context.Canceled` and `read response: context canceled` in `retryablePluginStoreArtifactError`, so mid-read cancellations uniformly go through the retry path.

3. **Plugin installation timeout** (`internal/api/handlers/management/plugin_store.go:248`) — Removed the fixed 10-minute timeout, switched to `context.WithoutCancel` without attaching an additional hard deadline, aligning with the timeout policy defined in `AGENTS.md`.

gwdgithubnom:
@codex review

PR #4702, opened 2026-07-31
feat(executor): opt-in prompt_cache_key for OpenAI-compatible providers

### LINKED ISSUE #4700: feat(executor): opt-in prompt_cache_key derivation for OpenAI-compatible providers on /chat/completions
## Summary

OpenAI-compatible providers reached via `/chat/completions` never receive an
upstream `prompt_cache_key`, so upstream prompt caching stays cold for every
request even when the caller (e.g. Claude Code) carries a stable session
identity. This is a request to add an **opt-in** capability that derives and
injects a deterministic `prompt_cache_key` on the chat-completions path,
reusing the derivation logic the Codex/xAI executors already have.

## Current behavior

When a Claude Code client targets an OpenAI-compatible provider, the request
flows:

```text
Claude Code (/v1/messages) → CLIProxyAPI → provider (/chat/completions)
```

Translation is Claude → OpenAI. On this path:

1. `ConvertClaudeRequestToOpenAI`
   (`internal/translator/openai/claude/openai_claude_request.go`) builds the
   outbound payload from a whitelist and never sets `prompt_cache_key`.
   Claude Code is Anthropic-shaped, so it does not send `prompt_cache_key`
   to begin with.
2. `OpenAICompatExecutor.Execute` / `ExecuteStream`
   (`internal/runtime/executor/openai_compat_executor.go`) does not derive a
   continuity key from the session identity.

As a result the provider receives no `prompt_cache_key`, and repeated turns
of the same conversation cannot hit a shared cache lineage.

This was already fixed for the Codex `responses` path in
[#2373](https://github.com/router-for-me/CLIProxyAPI/issues/2373). The fix
introduced `helps.ClaudeCodePromptCache`
(`internal/runtime/executor/helps/claude_code_session.go`), which derives a
deterministic key from model + Claude Code session/agent scope. However,
that helper is only called from the Codex/xAI `responses` executors
(`codex_executor_request.go`, `codex_websockets_request.go`,
`xai_executor_request.go`). The generic `/chat/completions` path used by
**every** `openai-compatibility` provider has no equivalent.

## Why opt-in

Not every OpenAI-compatible backend honors `prompt_cache_key`; some reject
unknown fields with HTTP 400. A provider that silently accepts and ignores
it is fine, but a strict one is not. Default-off with an explicit opt-in
keeps existing setups unchanged while letting cache-aware providers turn it
on.

## Proposed change

1. Add an opt-in flag scoped to `openai-compatibility` entries, e.g.
   `prompt-cache-key: true` on the provider block (and/or a global
   default), gated by validation.
2. In `OpenAICompatExecutor.Execute` / `ExecuteStream`, after the payload
   rules step (`ApplyPayloadConfigWithRequest`), when the source format is
   Claude and the flag is on, call `helps.ClaudeCodePromptCache` and set
   `prompt_cache_key` on the translated body — the same pattern Codex uses.
3. Respect an explicitly caller-supplied `prompt_cache_key` when present
   (Codex precedence: explicit key > derived). The opt-in only fills the
   gap when none exists.

Scope is confined to `internal/runtime/executor/` (executor + unit test),
so it does not touch `internal/translator/` and passes the
`translator-path-guard` PR check.

## Expected behavior

With the flag on, a Claude Code session targeting an opted-in provider
gets a stable, deterministic `prompt_cache_key` across turns, enabling
upstream cache reuse. With the flag off (default), behavior is unchanged.

## Reproduction

No upstream call needed — the gap is visible in the request body. With
`request-log: true`:

1. Configure any `openai-compatibility` provider.
2. Send a Claude Code request through the proxy.
3. Inspect the upstream request log: `prompt_cache_key` is absent.

The same Claude Code request routed at a Codex model shows the key, which
confirms the asymmetry.

**CLI Type**: claude code → openai-compatibility
**LLM Client**: claude code
**Model Name**: any model served by an OpenAI-compatible provider

ririnto:
## Summary
- add an opt-in `prompt-cache-key` flag to `openai-compatibility` entries (default off)
- inject an upstream `prompt_cache_key` in the OpenAI-compatibility executor's `Execute` and `ExecuteStream`: a caller-supplied key wins, otherwise one is derived from the client session
- derive via the same chain as the Codex/xAI executors: `execution_session_id` → Claude Code session-derived key → random UUID
- document the option in `config.example.yaml`
- add unit tests covering opt-in/off, deterministic derivation, explicit-key precedence, random-UUID fallback, and agent-scope isolation

## Motivation
The Codex executor derives a `prompt_cache_key` before sending requests upstream, but the OpenAI-compatibility executor — used by any OpenAI-compatible provider — does not. Prompt caching therefore stays inactive for these providers even when both the client and upstream support it. This closes that gap behind an explicit opt-in; default off avoids HTTP 400 on backends that reject unknown fields.

## Scope
OpenAI-compatibility executor only. No `internal/translator/` changes, no response handling, no routing/affinity behavior, no changes to other executors.

## Validation
- `gofmt -l` on changed files (clean)
- `go vet ./internal/runtime/executor/`
- `go test -run 'TestOpenAICompatPromptCacheKey' ./internal/runtime/executor/`

Closes #4700.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -446,6 +446,7 @@ nonstream-keepalive-interval: 0
 #     prefix: "test" # optional: require calls like "test/kimi-k2" to target this provider's credentials
 #     base-url: "https://openrouter.ai/api/v1" # The base URL of the provider.
 #     disable-cooling: false # optional: per-provider override for auth/model cooldown scheduling
+#     prompt-cache-key: false # optional: inject an upstream prompt_cache_key; a caller key wins, else derive one from the client session
 #     headers:
 #       X-Custom-Header: "custom-value"
 #     api-key-entries:
--- a/internal/api/handlers/management/config_auth_index.go
+++ b/internal/api/handlers/management/config_auth_index.go
@@ -48,6 +48,7 @@ type openAICompatibilityWithAuthIndex struct {
 	Models         []config.OpenAICompatibilityModel        `json:"models,omitempty"`
 	Headers        map[string]string                        `json:"headers,omitempty"`
 	DisableCooling bool                                     `json:"disable-cooling,omitempty"`
+	PromptCacheKey bool                                     `json:"prompt-cache-key,omitempty"`
 	AuthIndex      string                                   `json:"auth-index,omitempty"`
 }
 
@@ -286,6 +287,7 @@ func (h *Handler) openAICompatibilityWithAuthIndex() []openAICompatibilityWithAu
 			Models:         entry.Models,
 			Headers:        entry.Headers,
 			DisableCooling: entry.DisableCooling,
+			PromptCacheKey: entry.PromptCacheKey,
 			AuthIndex:      "",
 		}
 		if len(entry.APIKeyEntries) == 0 {
--- a/internal/api/handlers/management/config_lists.go
+++ b/internal/api/handlers/management/config_lists.go
@@ -705,6 +705,7 @@ func (h *Handler) PatchOpenAICompat(c *gin.Context) {
 		Prefix         *string                             `json:"prefix"`
 		Disabled       *bool                               `json:"disabled"`
 		DisableCooling *bool                               `json:"disable-cooling"`
+		PromptCacheKey *bool                               `json:"prompt-cache-key"`
 		BaseURL        *string                             `json:"base-url"`
 		APIKeyEntries  *[]config.OpenAICompatibilityAPIKey `json:"api-key-entries"`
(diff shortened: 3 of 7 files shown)


ririnto:
@codex Addressed the review feedback: the config_index mismatch was a pre-existing issue in resolveCompatConfig, but since this PR exercises that path I fixed it by mirroring the auth layer's resolveOpenAICompatConfigForAuth.

ririnto:
I was build and run binary at WSL. Maybe other environment will be same. Mean, It will be work.

PR #4725, opened 2026-08-01
fix(auth): cap upstream 429 Retry-After so quota resets are picked up

### LINKED ISSUE #4724: [BUG] model_cooldown sticks ~156h after an upstream quota reset; only a service restart wakes the proxy
### Problem

Three ChatGPT (Plus) Codex credentials behind one proxy hit `usage_limit_reached` upstream. From then on every request to `gpt-5.6-sol` returns a proxy-generated `429`:

```json
{"error":{"code":"model_cooldown","message":"All credentials for model gpt-5.6-sol are cooling down","model":"gpt-5.6-sol","reset_seconds":561635,"reset_time":"156h0m34s"}}
```

This blocks every client, including Anthropic-format `/v1/messages` traffic that does not go through the Codex scheduler plugin, so it is core cooldown state, not the plugin.

### What happened after the reset

I applied a manual quota reset (ChatGPT reset credits) on one of the three accounts. The account now had a fresh 5-hour window, but the proxy kept returning the same `429` for its full reset window — it never re-probed the refreshed credential. Only `systemctl restart cliproxyapi` cleared the cooldown; the very next request returned `200`.

So there is no automatic path out of a reset: the proxy trusts its stale in-memory cooldown until the process restarts, and the only manual lever is a full restart (which drops in-flight requests, and re-arms the cooldown if the reset has not landed yet).

### Observations

- The cooldown is in-memory and lost on restart (`save-cooldown-status` is off), so restart does clear it — but restart is a coarse, disruptive way to clear state.
- `reset_seconds` tracked the long (weekly) window, not the 5-hour window that reset credits actually reset. Worth double-checking how `Retry-After`/reset times from ChatGPT are mapped onto `Auth.NextRetryAfter`.
- There is no management endpoint or CLI to clear one credential's cooldown, and no signal path for "quota was refreshed" from a scheduler plugin to the core cooldown state.

### Ask

1. When a credential's quota is refreshed upstream (manual reset, or window rollover), clear or shorten its cooldown instead of waiting out the old `reset_seconds`.
2. Failing that, expose a management API to clear per-credential (or per-model) cooldown state without a restart.
3. Consider re-probing a credential once instead of trusting a stale cooldown when its reset window is long.

### Environment

- CLIProxyAPI v7.2.113
- Endpoints: `/v1/responses`, `/v1/messages`
- Clients: codex CLI, Claude Code
- 3 Codex (ChatGPT Plus) OAuth credentials, all exhausted; one manually reset

Related: #4327 (closed as config-specific, but notes the weekly-window/model-vs-account cooldown issue), #4715 (feature request for reset-aware credential selection).

artile:
Fixes #4724.

## Problem

A single upstream 429 `Retry-After` can be huge — ChatGPT returned `561635` (~156h, a weekly quota window) on our instance. Both 429 cooldown paths honored that hint verbatim:

- per-model `MarkResult` path (`conductor_cooldown.go`)
- `applyAuthFailureState` (`conductor_cooldown.go`)

That parked every credential for days, and nothing re-probes a credential while its cooldown window is open. So when the account was reset upstream (manual reset-credit, or the window rolling over), the proxy kept returning `429 model_cooldown` until a `systemctl restart cliproxyapi` cleared the in-memory cooldown.

The proxy already caps its own quota backoff at `quotaBackoffMax = 30 * time.Minute` (`conductor_refresh.go`), so re-probing at that cadence is exactly what a healthy setup already does.

## Fix

Cap the honored upstream `Retry-After` at `quotaBackoffMax` (30 min) in both 429 cooldown paths. Hints at or below the cap are honored unchanged; the credential now re-probes within half an hour, so an upstream reset is picked up without a restart.

## Notes

- The management endpoint `POST /v0/management/reset-quota` (v7.2.100+) remains the immediate operational escape hatch for a manual reset.
- Huge `Retry-After` values still work through the retry/wait path (`conductor_selection.go`), which is already bounded by `maxWait`.

## Tests

- `TestCappedUpstreamRetryAfter` — helper edges (zero, negative, below-cap, at-cap, above-cap).
- `TestApplyAuthFailureStateCapsUpstreamRetryAfter` — 7-day hint → capped at 30 min, and auto-recovers after the capped deadline.
- `TestMarkResultCapsUpstreamRetryAfter` — per-model path: 7-day hint → capped at 30 min, auto-recovers.

`go build ./...` and `go test ./sdk/cliproxy/auth/...` pass.

### DIFF
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -93,6 +93,25 @@ func recoverableFailureRetryAfter(now time.Time, disableCooling bool) time.Time 
 	return nextTransientErrorRetryAfter(now)
 }
 
+// cappedUpstreamRetryAfter bounds a single upstream 429 Retry-After hint to the
+// ceiling the proxy already uses for its own quota backoff (quotaBackoffMax).
+// Upstreams occasionally return multi-day windows (e.g. a weekly ChatGPT quota
+// reset). Honoring them verbatim parks the credential in cooldown long after a
+// manual reset or window rollover made it usable again, and nothing re-probes it
+// before that window expires — so the proxy keeps 429ing until a restart clears
+// the in-memory cooldown. Capping the hint keeps the credential re-probing, so
+// an upstream reset is picked up without a service restart. Hints at or below
+// the ceiling are honored unchanged; non-positive values collapse to zero.
+func cappedUpstreamRetryAfter(retryAfter time.Duration) time.Duration {
+	if retryAfter <= 0 {
+		return 0
+	}
+	if retryAfter > quotaBackoffMax {
+		return quotaBackoffMax
+	}
+	return retryAfter
+}
+
 // SetConfig updates the runtime config snapshot used by request-time helpers.
 // Callers should provide the latest config on reload so per-credential alias mapping stays in sync.
 func (m *Manager) SetConfig(cfg *internalconfig.Config) {
@@ -811,7 +830,7 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 							backoffLevel := state.Quota.BackoffLevel
 							if !disableCooling {
 								if result.RetryAfter != nil {
-									next = now.Add(*result.RetryAfter)
+									next = now.Add(cappedUpstreamRetryAfter(*result.RetryAfter))
 								} else {
 									next, backoffLevel = quotaCooldownAfterFailure(state.Quota, now)
 								}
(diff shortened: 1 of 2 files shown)


jroth1111:
Clean fix — no conflict with the antigravity recovery engine (commit d5b5787). The recovery engine's `antigravityRetryAfter` honors upstream `Retry-After` for per-request retry delays inside the executor, while this PR caps the same hint at the conductor-level cooldown path. Different layers, complementary: the executor retries within a 75s budget using the hint, and the conductor re-probes the credential within 30 min regardless of how large the upstream hint was.

PR #4731, opened 2026-08-02
feat(executor): add configurable Claude cache-control default TTL

### LINKED ISSUE #3398: Fix Claude prompt caching to preserve 1h TTL for cache_control blocks
## Summary

CLIProxyAPI already injects and repairs Anthropic `cache_control` breakpoints for Claude requests, but the current behavior does not reliably preserve the 1-hour prompt cache path. Injected or default `cache_control: { "type": "ephemeral" }` blocks remain default 5-minute TTL blocks, and `normalizeCacheControlTTL()` then treats them as 5m anchors. This can downgrade later `ttl: "1h"` blocks or prevent newly injected breakpoints from using the extended 1h cache TTL.

The desired behavior is to make Claude requests consistently use 1h cache TTL when the request/header context supports it, while still preserving Anthropic's max-4 breakpoint limit and TTL ordering rules.

## Current behavior

The Claude executor currently does these steps:

1. If there are no cache breakpoints, it calls `ensureCacheControl()`.
2. `ensureCacheControl()` injects `cache_control` into:
   - the last tool definition,
   - the last system prompt element,
   - the second-to-last user turn when applicable.
3. Injected blocks use only `{ "type": "ephemeral" }`, so they default to 5m.
4. `enforceCacheControlLimit(..., 4)` removes excess breakpoints.
5. `normalizeCacheControlTTL()` strips later `ttl: "1h"` values after any default 5m block in Anthropic evaluation order: tools -> system -> messages.

This means a request can be structurally cacheable but still not use the 1h TTL path.

Relevant code:

- `internal/runtime/executor/claude_executor.go`
  - `ExecuteStream` / `Execute` call `ensureCacheControl()` only when `countCacheControls(body) == 0`.
  - `ensureCacheControl()` injects default ephemeral blocks.
  - `normalizeCacheControlTTL()` correctly protects Anthropic ordering, but default blocks currently force later 1h blocks to be downgraded.
- `internal/runtime/executor/caching_verify_test.go`
  - tests injection exists, but does not appear to assert `ttl: "1h"` on injected blocks.
- `internal/runtime/executor/claude_executor_test.go`
  - tests TTL normalization and max-4 behavior, but the desired 1h promotion behavior should be covered explicitly.

## Proposed fix

Add a dedicated Claude cache-control repair step that runs after injection and max-4 enforcement, before final upstream body signing / sending:

1. Promote eligible default ephemeral cache blocks to `ttl: "1h"`.
   - Eligible means `cache_control` is an object, `type == "ephemeral"`, and `ttl` is missing or currently `"5m"`.
   - Do not mutate malformed/non-object `cache_control` values.
   - Do not overwrite unsupported or unknown TTL values other than missing/`"5m"` unless maintainers intentionally choose to normalize them.

2. Preserve Anthropic ordering constraints.
   - After promotion, re-run the existing ordering normalization so no 1h block appears after a remaining default/invalid 5m-equivalent block in evaluation order.
   - Keep the current evaluation order: tools -> system -> messages.

3. Keep max-4 breakpoints.
   - Continue to enforce the 4-breakpoint limit.
   - Prefer preserving high-value breakpoints: last tool, last system, and recent message cache points.

4. Keep CCH/body signing correct.
   - If experimental CCH signing is enabled, sign after all body mutation, including cache-control injection, promotion, max-4 enforcement, and TTL normalization.
   - This avoids stale CCH fingerprints after body mutation.

A safe helper shape would be something like:

```go
func promoteDefaultCacheControlTTL(payload []byte, ttl string) []byte {
    // Walk tools -> system -> messages.
    // For each cache_control object where type == "ephemeral" and ttl is missing or "5m",
    // set ttl to "1h".
    // Leave malformed/non-object cache_control values untouched.
}
```

Then the Claude request pipeline can become:

```go
}
body = enforceCacheControlLimit(body, 4)
body = promoteDefaultCacheControlTTL(body, "1h")
body = normalizeCacheControlTTL(body)
// CCH signing, if enabled, must happen after this point.
```

If maintainers prefer configuration, make it opt-in, for example:

```yaml
providers:
  claude:
    cache-control-default-ttl: "1h"
```

Defaulting to `1h` is probably reasonable for Claude Code-compatible traffic because the project already forwards/uses Claude Code prompt-caching beta behavior and has cache-control repair logic.

## Acceptance criteria

- [ ] When a Claude request has no `cache_control`, CLIProxyAPI injects cache breakpoints and those injected breakpoints use `ttl: "1h"` where valid.
- [ ] When a Claude request has existing `cache_control: { "type": "ephemeral" }` blocks, they are promoted to `ttl: "1h"` where valid.
- [ ] Existing `ttl: "1h"` blocks remain unchanged when no ordering violation exists.
- [ ] If a default/invalid 5m-equivalent block remains before a later 1h block, the later 1h block is still downgraded or repaired so Anthropic does not reject the request.
- [ ] Total `cache_control` breakpoint count remains `<= 4`.
- [ ] Experimental CCH signing, when enabled, signs the final mutated body, not a pre-repair body.
- [ ] Tests cover stream, non-stream, and count-tokens paths if all three mutate Claude request bodies independently.

## Suggested tests

Add or update tests around these cases:

1. `ensureCacheControl()` with tools + system + multi-turn messages injects `ttl: "1h"` on the inserted breakpoints.
2. Existing default blocks are promoted:

```json
{
  "tools": [{ "name": "t1", "cache_control": { "type": "ephemeral" } }],
  "system": [{ "type": "text", "text": "s1", "cache_control": { "type": "ephemeral" } }],
  "messages": [{ "role": "user", "content": [{ "type": "text", "text": "u1", "cache_control": { "type": "ephemeral" } }] }]
}
```

Expected: all valid cache-control objects include `"ttl": "1h"` after repair.

3. Max-4 still wins when there are too many breakpoints.
4. Ordering repair still wins when a malformed/default 5m-equivalent block appears before a later 1h block.
5. CCH signing test: final signed body verifies after cache-control promotion.

## Reference implementation behavior

In my own Claude proxy path, the stable approach is:

- collect cache-control refs in Anthropic evaluation order: tools -> system -> messages,
- enforce max 4 breakpoints,
- promote default ephemeral blocks to `ttl: "1h"`,
- normalize ordering so later 1h blocks do not appear after default/invalid 5m-equivalent blocks,
- sign the final body after all mutations.

This has been enough to keep Claude prompt caching stable on the 1h path while still respecting Anthropic's cache-control constraints.

GuoSihan0320:
## Summary

Adds an opt-in global option, `cache-control-default-ttl`, that promotes default
Anthropic prompt-cache breakpoints on Claude requests to an explicit TTL before the
body is sent upstream. The default (unset) keeps every request byte-identical to
today's behavior.

Closes #3398

## Motivation

Per #3398, the proxy already injects and repairs `cache_control` breakpoints, but both
the injected blocks and the bare `{"type": "ephemeral"}` blocks that clients send carry
no `ttl`, so they land on Anthropic's default 5-minute cache path. For long-lived,
stable prefixes (tool definitions and system prompts) that means the prefix keeps
expiring between turns and gets re-written on every cache miss, which costs cache-write
tokens repeatedly instead of amortizing one write over a longer window.

There is a second, sharper effect that #3398 calls out: `normalizeCacheControlTTL()`
treats any default block as a 5m anchor, so a single bare block early in evaluation
order (tools -> system -> messages) downgrades every later `ttl: "1h"` block a client
did ask for. The extended 1h cache TTL is generally available and needs no beta header,
so there is no reason a proxy-repaired request has to stay on the 5m path.

## Implementation

- New helper `promoteDefaultCacheControlTTL(payload []byte, ttl string) []byte` walks
  the Anthropic evaluation order (tools -> system -> messages).
- Promotion predicate, exactly as proposed in #3398: `cache_control` must be an object
  with `"type": "ephemeral"` and a `ttl` that is either absent or `"5m"`. Malformed or
  non-object `cache_control` values, other block types, and any other explicit TTL are
  left untouched. No breakpoint is ever added or removed - `enforceCacheControlLimit`
  remains the single authority for the max-4 constraint.
- Pipeline position in all three Claude request paths:
  `ensureCacheControl` -> `enforceCacheControlLimit(body, 4)` -> **promote** ->
  `normalizeCacheControlTTL` -> CCH signing. Running before normalization keeps the
  prompt-caching-scope-2026-01-05 ordering repair authoritative; running before signing
  keeps the CCH fingerprint over the final mutated body.
- All three paths that mutate a Claude request body independently are wired: streaming
  (`ExecuteStream`), non-streaming (`Execute`), and the upstream count-tokens path
  (`countTokensUpstream`).
- Config: global string field `cache-control-default-ttl`. Supported values are `"5m"`
  and `"1h"`; unset disables the feature. Unsupported values are ignored and keep the
  feature disabled, matching how other optional string settings in this repo degrade to
  their safe default at the consumption site (for example `gpt-image-2-base-model` and
  `video-result-auth-cache-ttl`). The option is also surfaced in the hot-reload config
  change log and documented in `config.example.yaml` (commented out by default).

## Testing

New tests in `internal/runtime/executor/claude_cache_control_ttl_test.go`:

- bare `cache_control` blocks across tools + system + messages are all promoted, and the
  breakpoint count is unchanged;
- breakpoints injected by `ensureCacheControl` are promoted;
- blocks that already carry `ttl: "1h"` are left byte-identical;
- explicit `ttl: "5m"` blocks are promoted;
- malformed / non-object `cache_control`, non-ephemeral types, non-string `ttl`, and
  foreign TTLs such as `"30m"` are left byte-identical;
- with the option unset, the payload is byte-identical (zero behavior change);
- promote -> `normalizeCacheControlTTL` produces no downgrade (ordering compatibility),
  while ordering repair still wins when a malformed 5m-equivalent block precedes a
  promoted block;
- `enforceCacheControlLimit(body, 4)` -> promote still yields exactly 4 breakpoints;
- config resolution accepts only `""`, `"5m"`, `"1h"` (trim/case tolerant) and rejects
  values like `"2h"`;
- end-to-end executor tests for the stream, non-stream, and count-tokens paths assert
  the upstream body carries the promoted TTL and stays within 4 breakpoints;
- an end-to-end test asserts the CCH signature verifies against the final body after
  promotion, i.e. signing happens after all cache-control mutation.

Also added a `BuildConfigChangeDetails` case for the new option.

Commands run: `go build ./...`, `go vet` and `go test` for
`internal/runtime/executor`, `internal/config`, and `internal/watcher/diff`.

## Backward compatibility

`cache-control-default-ttl` defaults to the empty string, in which case
`promoteDefaultCacheControlTTL` returns the original byte slice and the request body is
unchanged - no new field, no reordered keys, and therefore no change to CCH signatures
for existing deployments. The max-4 limit and the TTL ordering normalization keep their
current semantics whether or not the option is set.

## Relation to #2533

This is deliberately a single feature with a global-only configuration surface. It does
not add per-credential overrides, which is where the earlier attempt in #2533 stalled on
tri-state override correctness; per-credential scoping can be layered on later if
maintainers want it.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -169,6 +169,13 @@ transient-error-cooldown-seconds: 0
 # "auto" behavior (cloak only non-Claude-Code clients).
 disable-claude-cloak-mode: false
 
+# Default TTL applied to Anthropic prompt-cache breakpoints on Claude requests.
+# Supported values: "5m" or "1h". Unset (default) leaves request bodies untouched.
+# Only cache_control objects with "type": "ephemeral" and a missing or "5m" ttl are
+# rewritten; other TTLs and malformed values are left as-is, and no breakpoint is added
+# or removed. Use "1h" to keep stable prompt prefixes on the extended cache path.
+# cache-control-default-ttl: "1h"
+
 # Claude Code compatibility settings.
 claude-code:
   # When true, return original model IDs in Anthropic model list responses instead of cloaked IDs.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -137,6 +137,20 @@ type Config struct {
 	// the auth/OAuth token file). Default false preserves the per-client "auto" behavior.
 	DisableClaudeCloakMode bool `yaml:"disable-claude-cloak-mode" json:"disable-claude-cloak-mode"`
 
+	// CacheControlDefaultTTL promotes default Anthropic prompt-cache breakpoints to the
+	// configured TTL before Claude requests are sent upstream.
+	//
+	// Supported values:
+	//   - "" (default): disabled. cache_control blocks are forwarded exactly as-is.
+	//   - "5m": pin default breakpoints to the standard 5-minute TTL explicitly.
+	//   - "1h": promote default breakpoints to the extended 1-hour TTL.
+	//
+	// Only breakpoints whose cache_control is an object with "type": "ephemeral" and a
+	// missing or "5m" ttl are rewritten. Blocks that already carry another ttl and
+	// malformed cache_control values are left untouched, and no breakpoint is added or
+	// removed. Unsupported values are ignored and keep the feature disabled.
+	CacheControlDefaultTTL string `yaml:"cache-control-default-ttl,omitempty" json:"cache-control-default-ttl,omitempty"`
+
 	// OpenAICompatibility defines OpenAI API compatibility configurations for external providers.
 	OpenAICompatibility []OpenAICompatibility `yaml:"openai-compatibility" json:"openai-compatibility"`
 
--- a/internal/runtime/executor/claude_cache_control_ttl_test.go
+++ b/internal/runtime/executor/claude_cache_control_ttl_test.go
@@ -1,0 +1,404 @@
+package executor
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"regexp"
+	"testing"
+
+	xxHash64 "github.com/pierrec/xxHash/xxHash64"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
+	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+	sdktranslator "github.com/router-for-me/CLIProxyAPI/v7/sdk/translator"
+	"github.com/tidwall/gjson"
+)
+
+// collectCacheControlTTLs returns the ttl value of every cache_control block in
+// Anthropic evaluation order (tools → system → messages). Blocks without a ttl are
+// reported as an empty string so callers can assert on default breakpoints too.
+func collectCacheControlTTLs(t *testing.T, payload []byte) []string {
+	t.Helper()
+
+	ttls := make([]string, 0, 8)
+	appendBlock := func(item gjson.Result) {
+		cc := item.Get("cache_control")
+		if !cc.Exists() {
+			return
+		}
+		ttls = append(ttls, cc.Get("ttl").String())
+	}
+
+	gjson.GetBytes(payload, "tools").ForEach(func(_, item gjson.Result) bool {
+		appendBlock(item)
+		return true
+	})
+	gjson.GetBytes(payload, "system").ForEach(func(_, item gjson.Result) bool {
+		appendBlock(item)
+		return true
+	})
+	gjson.GetBytes(payload, "messages").ForEach(func(_, msg gjson.Result) bool {
+		content := msg.Get("content")
+		if !content.IsArray() {
+			return true
+		}
+		content.ForEach(func(_, item gjson.Result) bool {
+			appendBlock(item)
+			return true
+		})
+		return true
+	})
+	return ttls
+}
+
+func assertAllCacheControlTTLs(t *testing.T, payload []byte, want string) {
+	t.Helper()
+
+	ttls := collectCacheControlTTLs(t, payload)
+	if len(ttls) == 0 {
+		t.Fatalf("expected at least one cache_control block in body: %s", string(payload))
+	}
+	for i, got := range ttls {
+		if got != want {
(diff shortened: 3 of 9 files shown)


KJ7LNW:
This PR would close #3398 

Depends on either of these to land because we shouldn't be messing with cache headers in baselines before cc v2.1.243 where 1hr cache became an option (current CLIProxyAPI is baselined on cc v2.1.220)

- #5412 
**or**
- #5456

PR #4748, opened 2026-08-02
Handle encrypted collaboration message schemas

### LINKED ISSUE #4746: codex: optimize-multi-agent-v2 leaves followup_task messages encrypted

[x] Yes. This is a request rewrite issue in the opt-in Codex Multi-Agent V2 optimizer. The reproduction below is sanitized and contains no credentials or private endpoint information.

## Describe the bug

With `codex.optimize-multi-agent-v2: true`, an initial `spawn_agent` task reaches custom-provider subagents as readable plaintext, but `followup_task` to the existing agent still arrives as opaque ciphertext.

The behavior matches the current implementation: `rewriteCodexSpawnAgentTools` only finds `spawn_agent` tool paths and deletes:

```text
spawn_agent.parameters.properties.message.encrypted
```

It does not remove the same marker from:

```text
send_message.parameters.properties.message.encrypted
followup_task.parameters.properties.message.encrypted
```

Source:

- `codexSpawnAgentToolPaths` / `rewriteCodexSpawnAgentTools`: https://github.com/router-for-me/CLIProxyAPI/blob/41fc5e134631789e98137245f576680c7fb9b322/internal/client/codex/optimize-multi-agent-v2/optimize_multi_agent_v2.go#L397-L423
- The existing test includes `send_message`, but only asserts that `spawn_agent.message.encrypted` is removed: https://github.com/router-for-me/CLIProxyAPI/blob/41fc5e134631789e98137245f576680c7fb9b322/internal/client/codex/optimize-multi-agent-v2/optimize_multi_agent_v2_test.go#L132-L188

## CLI Type

`codex` OAuth/custom-provider routing through CLIProxyAPI.

## Model Name

- Parent: an official Codex model using Multi-Agent V2
- Subagents reproduced with both `deepseek-v4-flash` and `glm-5.2`

The failure is operation-specific rather than model-specific: a new GLM agent receives plaintext successfully, while a follow-up sent to the same GLM agent receives ciphertext.

## LLM Client

Official Codex Desktop client.

## Request Information / minimal reproduction

CLIProxyAPI version: `v7.2.114` (`41fc5e134631789e98137245f576680c7fb9b322`)

Relevant config:

```yaml
codex:
  optimize-multi-agent-v2: true
```

Sanitized end-to-end reproduction:

1. Spawn a DeepSeek subagent with the task `Compute 17 * 23 and return the result with one sentence of reasoning.`
   - PASS: the child returns `391` with the correct calculation.
2. Spawn a GLM subagent with the same task.
   - PASS: the child returns `391` with the correct calculation.
3. Send `followup_task` to that existing GLM agent, asking it to return `PONG`.
   - FAIL: the child reports that its Payload is an unreadable encrypted/encoded string and asks for the task to be resent in plaintext.
4. Spawn a new GLM agent with the task `Return exactly PONG`.
   - PASS: the new child returns `PONG`.

Observed result:

```text
spawn_agent(DeepSeek)       -> PASS, plaintext task received
spawn_agent(GLM)            -> PASS, plaintext task received
followup_task(existing GLM) -> FAIL, ciphertext received
spawn_agent(new GLM)        -> PASS, plaintext task received
```

This A/B sequence rules out a general GLM/provider availability problem and isolates the failure to the follow-up collaboration-tool path.

`send_message` was not exercised in this reproduction, but its `message` parameter uses the same encrypted schema marker and is not stripped by the optimizer, so it should be covered by the fix and regression tests as well.

## Expected behavior

When `codex.optimize-multi-agent-v2` is enabled, message encryption should be removed consistently for all collaboration tools that transport agent messages:

```text
spawn_agent
send_message
followup_task
```

The optimizer should scan both top-level `tools` and `input[].type == "additional_tools"`, remove only `parameters.properties.message.encrypted` for those specific functions, and preserve unrelated business properties named `encrypted`.

Suggested regression coverage:

- Request rewrite for all three collaboration tools
- Top-level and `additional_tools` namespace shapes
- HTTP/SSE and WebSocket executors
- `spawn_agent -> send_message -> followup_task -> child result` multi-turn flow
- Plaintext marker preservation (`encrypted_function_args: []`) when returned by the upstream

## Screenshots

Not included because the text reproduction is deterministic and avoids exposing unrelated local thread data.

## OS Type

- OS: macOS
- Version: macOS 26.x

## Additional context

- The optimizer was introduced in `v7.2.98` by commit https://github.com/router-for-me/CLIProxyAPI/commit/84bf9376e5a59f14cfc45562619af79eea1ae69f and was merged directly from the `optimize-multi-agent-v2` branch without an associated public PR.
- The closest older report is https://github.com/router-for-me/CLIProxyAPI/issues/3774, where Multi-Agent V2 encrypted subagent calls stopped working and the reporter used disabling `features.multi_agent_v2` as a workaround.
- Current workaround: spawn a new child for each independent task instead of sending a follow-up, or disable Multi-Agent V2.

patrick-fu:
## Summary
- remove the encrypted schema marker from spawn_agent, send_message, and followup_task message parameters in the opt-in Codex Multi-Agent V2 optimizer
- handle standalone send/follow-up tools in top-level and additional_tools payloads while preserving unrelated encrypted fields
- add regression coverage for tool discovery, schema rewriting, namespace behavior, and no-op cases

## Testing
- `go test -count=1 -timeout 300s ./...`
- `go build ./...`

## Runtime validation
- pre-fix Codex Desktop reproduction confirmed spawn_agent receives plaintext while send_message and followup_task arrive as opaque ciphertext
- patched live multi-turn validation is being prepared on an isolated Zeabur deployment; no live result is claimed here

Fixes #4746

### DIFF
--- a/internal/client/codex/optimize-multi-agent-v2/optimize_multi_agent_v2.go
+++ b/internal/client/codex/optimize-multi-agent-v2/optimize_multi_agent_v2.go
@@ -27,6 +27,15 @@ const (
 	codexOptimizedCollaborationNamePrefix = codexOptimizedCollaborationNamespace + "__"
 )
 
+// codexCollaborationMessageTools are the collaboration tool names whose
+// parameters.properties.message.encrypted field must be stripped so that
+// message content remains readable by the proxy.
+var codexCollaborationMessageTools = map[string]struct{}{
+	"spawn_agent":   {},
+	"send_message":  {},
+	"followup_task": {},
+}
+
 type codexSpawnAgentModel struct {
 	id                     string
 	description            string
@@ -73,6 +82,7 @@ func OptimizeCodexMultiAgentV2Request(ctx context.Context, headers http.Header, 
 		return payload, false
 	}
 	updated := rewriteCodexAgentMessageContent(payload)
+	updated = removeCodexCollaborationMessageEncryption(updated, codexCollaborationMessageToolPaths(updated))
 	toolPaths := codexSpawnAgentToolPaths(updated)
 	if len(toolPaths) == 0 || hasCodexOptimizedCollaborationConflict(updated) {
 		return updated, false
@@ -609,8 +619,19 @@ func rewriteCodexAgentMessageContent(payload []byte) []byte {
 }
 
 func codexSpawnAgentToolPaths(payload []byte) []string {
-	paths := make([]string, 0, 1)
-	collectCodexSpawnAgentToolPaths(gjson.GetBytes(payload, "tools"), "tools", &paths)
+	return codexToolPathsByNames(payload, map[string]struct{}{"spawn_agent": {}})
+}
+
+// codexCollaborationMessageToolPaths discovers function tools named
+// spawn_agent, send_message, or followup_task inside top-level tools arrays and
+// input[].additional_tools arrays, including nested namespace tools.
+func codexCollaborationMessageToolPaths(payload []byte) []string {
+	return codexToolPathsByNames(payload, codexCollaborationMessageTools)
+}
+
+func codexToolPathsByNames(payload []byte, names map[string]struct{}) []string {
+	paths := make([]string, 0, len(names))
+	collectCodexToolPathsByNames(gjson.GetBytes(payload, "tools"), "tools", &paths, names)
 
 	input := gjson.GetBytes(payload, "input")
 	if input.IsArray() {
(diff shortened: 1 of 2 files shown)


luispater (approved):
## Summary

This change correctly generalizes collaboration-tool path discovery and removes `parameters.properties.message.encrypted` from `spawn_agent`, `send_message`, and `followup_task` across top-level and `additional_tools` payloads, including nested namespaces. The implementation preserves unrelated encrypted fields and remains gated behind the existing opt-in Codex Multi-Agent V2 flow.

## Key findings

- No blocking or non-blocking issues found.

## Test plan

- `go test -count=1 -timeout 300s ./...`
- `go build ./...`
- GitHub checks passed, including `build` and `ensure-no-translator-changes`.

luispater (approved):

Issue #4670, opened 2026-07-30
Feature request: add MCP-style fallback aliases for unrecognized Claude OAuth tools


KJ7LNW:
## Summary

Add an opt-in Claude OAuth cloaking flag that represents otherwise unrecognized client-defined tools as ordinary Claude Code MCP extensions.

This allows arbitrary third-party tool usage in a way that's unlikely to get flagged. 

For example:

```text
search_web → mcp__amber_cedar__quiet_harbor
```

## Motivation

CLIProxyAPI already cloaks non-Claude-Code clients on Claude OAuth traffic and maps recognized third-party tool names to Claude Code equivalents. Unknown names such as `search_web` and `fetch_url` currently pass through unchanged even though they can reasonably be represented as Claude Code MCP extensions.

Claude Code exposes local MCP tools to the Messages API as ordinary client tools named with the `mcp__<server>__<tool>` convention. This feature extends the existing name-remapping mechanism to cover its otherwise-unhandled case.

## Configuration and eligibility

One possible configuration is:

```yaml
claude-api-key:
  - api-key: "..."
    cloak:
      unknown-tools-as-mcp: true
      mcp-alias-mode: "deterministic"
```

Equivalent auth-file attributes should follow the existing Claude OAuth cloak configuration pattern.

When enabled for a Claude OAuth upstream request, the fallback applies only to client-defined tools left unidentified after all current recognition and explicit remapping rules run. It must not change:

- Typed Anthropic built-ins.
- Recognized Claude Code tools.
- Tools handled by `oauthToolRenameMap`.
- Valid names already following `mcp__...__...`.

## Alias generation

Each eligible tool receives a distinct, collision-safe alias:

```text
mcp__<opaque-server-petname>__<opaque-tool-petname>
```

Use dictionary-generated petnames (also called mnemonic names or wordlist slugs): combinations of neutral nouns, verbs, and adjectives that look name-like while carrying no tool meaning.

```text
mcp__amber_cedar__quiet_harbor
mcp__silver_otter__drifting_meadow
```

Aliases must contain neither recognizable fragments of the original name nor visible hexadecimal or digest fragments. Generated names must satisfy upstream character and length restrictions.

Alias allocation must distinguish original names that differ only by case, repeated declarations, and names that normalize to the same value. It must prevent collisions with built-in, known, already-MCP, client-supplied, and other generated names by selecting additional or different dictionary words.

Supported mapping scopes may include:

[...]

PR #4752, opened 2026-08-03
feat(claude): align OAuth wire with Claude Code 2.1.220

sususu98:
## Summary

Align Claude OAuth inference and control-plane traffic with the measured wire behavior of Claude Code `2.1.220`, while keeping credential identity, request lifecycle, and non-Anthropic compatibility under CPA control.

This PR:

- distinguishes verified native Claude Code traffic from unverified callers using multiple request signals rather than `User-Agent` alone;
- preserves verified native software shape, but always applies the account/device identity of the OAuth credential selected by CPA;
- reconstructs cloaked Messages requests, computes CCH from the final serialized bytes, and keeps caller system content intact;
- aligns Anthropic beta selection, `count_tokens`, MCP-style tool aliases, HTTP/1.1 headers, TLS profiles, OAuth refresh/profile traffic, response decoding, and diagnostics continuity;
- makes every Fast failure request-scoped and passes it through without fallback, retry, refresh, credential rotation, or cooldown;
- strictly isolates first-party Anthropic behavior from custom Anthropic-compatible upstreams.

The implementation is based on secret-safe captures from isolated Claude Code `2.1.220` profiles and is intentionally limited to behavior that was measured or can be enforced without guessing.

## Motivation

The previous Claude cloak path reproduced selected Claude-looking fields, but the complete upstream identity spans more than a user agent. Request body layout, credential-bound metadata, CCH timing, beta ordering, TLS ClientHello, HTTP header casing/order, OAuth control-plane behavior, compression, and error lifecycle all need to remain internally consistent.

Partial alignment can produce several classes of failure:

- native software metadata paired with the wrong OAuth account/device identity;
- a valid CCH computed before later body mutations;
- Messages-only fields leaking into `count_tokens` or custom gateways;
- retries or cooldowns caused by request-scoped Fast failures or caller cancellation;
- custom tool names exposing a non-Claude client shape or failing to round-trip through SSE;
- stale diagnostics state crossing token rotations, sessions, or concurrent requests;
- HTTP/1.1 request boundaries becoming corrupt after partial/chunked writes.

This PR treats these as one request pipeline with explicit authority and endpoint boundaries rather than independent header substitutions.

### Native software shape vs. credential identity

Verified Claude Code CLI, `sdk-cli`, and VSCode requests may preserve their measured software shape. The selected CPA credential remains authoritative for `account_uuid` and the single canonical `device_id`; the conversation session is kept consistent across body metadata, headers, and diagnostics.

Unverified callers cannot obtain native passthrough from a plausible UA alone. They are normalized to the configured `2.1.220` baseline before upstream dispatch.

### Final-byte request pipeline

Body transformations run before CCH finalization:

1. protocol translation and thinking normalization;
2. cloak/system/current-date reconstruction;
3. context management, diagnostics, cache, beta, and tool handling;
4. cross-provider signature cleanup;
5. selected credential metadata;
6. CCH over the final serialized bytes;
7. HTTP headers and transport selection.

Byte-sensitive paths use offset edits rather than a full JSON unmarshal/marshal cycle, preserving key order, whitespace, escapes, and unrelated serialized bytes.

### Strict first-party origin gate

First-party endpoint policy—beta filtering, Messages/count routing, Anthropic header shape, Fast handling, and the measured inference transport—is enabled only for an HTTPS URL whose hostname is exactly `api.anthropic.com`, has no userinfo, and uses either the default port or `443`. Lookalike hosts, HTTP, alternate ports, and custom gateways do not inherit those first-party-only behaviors.

CCH has a separate credential/provider gate: every Claude OAuth Messages request is signed, including OAuth custom gateways; Vertex and strict first-party Anthropic API-key Messages are also signed; custom API-key gateways are not.

### Client detection, credential identity, and sessions

- Add strong-signal native detection with an entrypoint allowlist and measured software baseline.
- Keep one canonical device ID per credential and migrate legacy pools by retaining the first valid ID.
- Populate missing OAuth account identity through the profile path without allowing optional empty fields to erase existing identity.
- Serialize shared credential metadata access and add race coverage for mixed readers/writers.
- Keep refresh/profile identity updates coordinated and preserve refreshed tokens if an advisory follow-up lookup fails.
- Derive a stable conversation UUID for unverified clients and use it consistently in request headers, metadata, and diagnostics.
- Reject duplicate or structurally ambiguous caller identity containers as request-scoped HTTP 400 errors instead of selecting first/last JSON values.

### Cloak, system content, current date, cache, and CCH

- Rebuild the top-level system field as the minimal measured Claude Code billing/identity shape.
- Preserve caller system text exactly: no trimming, truncation, summarization, or copying of the complete prompt into a synthetic top-level third block.
- Place caller system content in the measured legacy reminder or mid-conversation system form, depending on model capability.
- Insert `currentDate` using credential timezone first, then configured timezone, then process-local time.
- Apply context-management, diagnostics, and cache behavior only where supported.
- Compute CCH automatically at the end of the supported Messages pipeline and modify only the five CCH digits in the original serialized body.
- Preserve adaptive thinking/signature behavior and strip foreign provider signatures before Claude dispatch.

### Tool aliases and issue #4670

For **cloaked OAuth requests only**, client-defined tools are represented as deterministic Claude Code MCP-style names:

```text
mcp__<12-char caller id>__<12-char tool id>_<semantic suffix>
```

The alias is bounded to 64 ASCII characters, collision-safe, case-sensitive, caller-stable, and reversible. One request-local symbol table updates declarations, `tool_choice`, declared historical tool references, JSON responses, and SSE responses. Tool IDs and `tool_use_id` links are unchanged.

Typed Anthropic server tools and already-valid `mcp__...__...` names remain unchanged. Verified native Claude Code requests preserve their original tool names.

This is related to #4670, with three intentional differences from the proposal in that issue:

- aliasing is automatic for eligible cloaked OAuth tools instead of opt-in;
- the suffix deliberately retains safe tool semantics instead of using fully opaque dictionary petnames;
- the rule covers eligible declared client tools directly rather than depending on the removed legacy lowercase-name mapping table.

Live validation confirmed that ordinary, invalid-character, Unicode, and overlength names produced valid aliases; Anthropic accepted the request; and a forced aliased tool call was restored to the exact downstream name.

### Anthropic beta and `count_tokens`

- Assemble beta values per request and endpoint rather than sending one fixed list.
- Match the two-account native observation for tool traffic: synthesize `advanced-tool-use-2025-11-20`, not `advisor-tool-2026-03-01`, and do not synthesize `cache-diagnosis-2026-04-07`.
- Preserve caller-dependent `context-1m-2025-08-07` behavior without inferring an entitlement rule from incomplete evidence.
- Treat `count_tokens` as a separate wire contract. The first-party native body contains only `model`, `messages`, and `tools`; it does not inherit Messages metadata, diagnostics, context management, CCH, or `X-Stainless-Timeout`.
- Use upstream `count_tokens` only for strict first-party `api.anthropic.com`. OAuth and API-key custom URLs continue to use the local estimator and are not contacted for token counting.

### HTTP, TLS, OAuth control plane, and response decoding

- Reproduce measured Claude Code/Stainless header values, casing, and HTTP/1.1 order for Messages and the separate `count_tokens` contract.
- Add an ordered connection wrapper that preserves body bytes and correctly tracks fixed-length, chunked, trailer, keep-alive, fragmented, and partial-write request boundaries.
- Use separate measured uTLS profiles for inference and OAuth control-plane traffic.
- Enable TLS 1.3 session resumption without changing the first ClientHello fixture, and bound transport/session caches by proxy URL.
- Align authorization-code exchange, refresh, profile, and roles inspection with their measured/request-specific contracts while keeping optional account lookups advisory.
- Decode stacked/repeated gzip, deflate, Brotli, zstd, and supported control-plane encodings before processing success or error payloads.

### Fast, cancellation, diagnostics, and bounded state

- Pass every Fast failure through directly. Upstream non-2xx responses retain status, decoded body, and safe headers.
- Do not retry Fast as standard mode, refresh credentials, rotate credentials, or create cooldown state.
- Ensure local decode/read/SSE failures after an upstream 2xx do not masquerade as an HTTP 200 error response.
- Classify only Claude OAuth `context.Canceled` as availability-neutral; do not extend this to deadlines, API keys, or other providers.
- Commit diagnostics continuity only after a complete successful JSON response or SSE `message_stop`.
- Key diagnostics by stable credential identity plus session rather than access token, and protect concurrent completion order with sequence/generation checks.
- Bound diagnostics and proxy-related caches with explicit capacity and eviction behavior.

## Configuration and compatibility

- Update the default documented fingerprint to Claude Code `2.1.220`, Stainless package `0.94.0`, and Node runtime `v26.3.0`.
- Add optional `claude-header-defaults.timezone` for cloaked `currentDate` fallback.
- Keep `experimental-cch-signing` accepted for configuration compatibility; supported OAuth/direct paths now finalize CCH automatically.
- Clarify that cloak mode `always` still preserves a request that is positively identified as a measured native client.
- No configuration migration is required.

Behavior changes worth calling out:

- malformed/duplicate caller identity JSON now fails closed with a request-scoped 400;
- Fast errors no longer fall back to a standard request;
- cloaked OAuth custom tools are automatically MCP-aliased and restored;
- custom gateways no longer receive first-party beta/header/Fast/transport or upstream token-count policy (OAuth Messages still follow the explicit OAuth CCH policy);
- custom URL token counts remain local, while strict first-party Anthropic token counts use the upstream endpoint.

This PR intentionally does **not** modify `internal/translator/**`; cross-protocol translator parity remains separate work.

## Security and lifecycle properties

- No credential, token, account identifier, device identifier, session identifier, prompt, or raw captured payload is included in the repository.
- Duplicate identity keys fail closed.
- New error and diagnostic paths do not log authorization material or raw credential identity values.
- Request-local reverse maps prevent tool-name leakage across callers.
- Transport and diagnostic caches have bounded capacity and lifecycle cleanup.
- Fast and cancellation errors cannot incorrectly penalize credential availability.
- Lookalike/custom origins cannot activate first-party Anthropic transport or body policy.

## Validation

All commands below passed after rebasing onto `upstream/dev` at `481c72658`:

```text
gofmt check for every changed, non-deleted Go file

go test ./internal/cache ./internal/auth/claude ./internal/httpwire \
  ./internal/runtime/executor/helps ./internal/runtime/executor \
  ./sdk/cliproxy/auth -count=1

go test -race ./internal/cache ./internal/auth/claude ./internal/httpwire \
  ./internal/runtime/executor/helps ./internal/runtime/executor \
  ./sdk/cliproxy/auth \
  -run 'Claude|Fast|Ordered|BoundedLRU|Diagnostics|RequestScoped' -count=1

go vet ./internal/cache ./internal/auth/claude ./internal/httpwire \
  ./internal/runtime/executor/helps ./internal/runtime/executor \
  ./sdk/cliproxy/auth

go test ./... -count=1

go build -o <temporary-output> ./cmd/server

git diff --check upstream/dev...HEAD
translator path guard: no changes under internal/translator
```

Repository-wide `go vet ./...` is not claimed as a gate because the current base has unrelated pre-existing warnings in `internal/logging` and `internal/pluginhost`; every package affected by this PR passes vet.

Additional regression coverage includes:

- native spoof rejection and configured baseline handling;
- credential identity, token rotation, duplicate JSON keys, and shared-auth races;
- final-byte CCH preservation and raw-edit differential/fuzz coverage;
- MCP collision, Unicode, overlength, history, JSON, and SSE round trips;
- first-party/custom-origin routing matrices;
- Messages vs. `count_tokens` body/header separation;
- beta ordering across OAuth/API-key, stream/non-stream, tools, context, and Fast;
- Fast single-attempt passthrough and conductor availability neutrality;
- diagnostics TTL, capacity, ABA, stale completion, and credential/session isolation;
- initial ClientHello fixtures, TLS resumption, header order, connection reuse, compression, chunked trailers, and partial writes.

## Secret-safe live evidence

Isolated Claude Code `2.1.220` profiles and a local capture proxy were used to compare native and CPA-generated traffic without retaining credentials or raw prompts in this PR.

Observed and verified:

- native Messages and `count_tokens` header/body contracts;
- two independent OAuth accounts using `advanced-tool-use` and omitting the synthesized advisor/cache-diagnosis betas;
- diagnostics continuity and inference connection reuse;
- profile/refresh HTTP/1.1 shape and compressed response decoding;
- first ClientHello fixture equality and successful TLS 1.3 resumption in the CPA client;
- semantic MCP aliases accepted with HTTP 200 and exact downstream tool-name restoration;
- Fast error passthrough with one upstream attempt and no credential availability penalty.

## Known limits / non-goals

- This profile targets the measured Claude Code `2.1.220` release. Future client versions require new captures rather than guessed constants.
- A successful Fast HTTP 200 was not exercised because it requires Usage Credits; this PR claims Fast error behavior, not Fast entitlement.
- CPA TLS resumption is tested, but no native resumed ClientHello capture is available, so resumed-flight byte equality is not claimed.
- The OAuth control-plane profile is tested structurally; raw native `platform.claude.com` ClientHello byte equality is not claimed.
- Bootstrap feature flags are not mapped to betas without direct evidence.
- The different observed `context-1m` state between two accounts is not treated as a general entitlement rule.
- Cross-protocol translator parity and downstream stream-authority behavior are outside this PR.

## Suggested review order

The 21 commits are intentionally layered. A practical review path is:

1. native detection, strict origin checks, and credential identity helpers;
2. request pipeline, cloak/system reconstruction, tool mapping, and CCH finalization;
3. beta and `count_tokens` endpoint contracts;
4. inference/control-plane TLS plus `internal/httpwire` request ordering;
5. Fast/cancellation/diagnostics lifecycle handling in the executor and conductor;
6. configuration compatibility and regression tests.

Related to #4670.


mingjunyan123:
@sususu98 May I know the reason why do you use last user message to calculate the cc_version suffix (located at the system[0]billing header), normally I think should use one of the message in body.messages[0] & role == user.

Sewer56:
@sususu98

Seems that with f3e25ab , cache markers are no longer being made, breaking prompt caching.
What I'm roughly observing is:

## Before `f3e25ab2`

Cloaking did not add a block-level cache marker before this guard:

```go
}
```

`ensureCacheControl()` added explicit breakpoints to:

1. Last non-deferred tool
2. Last system block
3. Second-to-last user turn

As conversation grew, message breakpoint moved forward.

## After `f3e25ab2`

Cloaking adds explicit breakpoints before the guard:

```text
system[1].cache_control
messages[0].content[actual user text].cache_control
```

Then:

```go
countCacheControls(body) != 0
```

`ensureCacheControl()` is skipped completely.

Result:

- Initial prefix still gets cached.
- Rolling message breakpoint is never added.
- `cache_read_input_tokens` stays fixed while conversation grows.
- Every later turn reprocesses all history after the fixed first-user breakpoint.

--------------

Adding ephemeral (auto-breakpoint) header works; but you want to match Claude Code behaviour.
I haven't investigated what CC does here to match; but what we got does not work; manual marker not set.

PR #4785, opened 2026-08-05
fix(claude): skip context management when thinking is disabled

### LINKED ISSUE #4784: [Bug] injectClaudeCodeContextManagement adds clear_thinking_20251015 to thinking-disabled requests, causing upstream 400 (regression in v7.2.116)
[  ] No, it's another issue.

Two minimal cURL payloads are attached under "Request Information". They differ by exactly one field.

**Describe the bug**

Since v7.2.116, any Claude Messages request that reaches an Anthropic upstream with `"thinking": {"type": "disabled"}` and no `context_management` field is rejected upstream with HTTP 400:

```
{"type":"error","error":{"type":"invalid_request_error","message":"`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"}}
```

The cause is `injectClaudeCodeContextManagement` in `internal/runtime/executor/claude_executor_cloaking.go`, called from `claude_executor_execute.go:83` and `claude_executor_stream.go:88` when `cloaked && isAnthropicUpstreamBase(baseURL)`:

```go
const claudeCodeContextManagement = `{"edits":[{"type":"clear_thinking_20251015","keep":"all"}]}`

func injectClaudeCodeContextManagement(payload []byte) []byte {
	if gjson.GetBytes(payload, "context_management").Exists() {
		return payload
	}
	updated, err := sjson.SetRawBytes(payload, "context_management", []byte(claudeCodeContextManagement))
	if err != nil {
		return payload
	}
	return updated
}
```

The only guard is whether the caller already sent `context_management`. There is no check on `thinking.type`. So when the caller sent `"thinking": {"type": "disabled"}`, CPA builds a self-contradictory request for it: the body asks the upstream to clear thinking blocks while thinking is switched off. Anthropic rejects that pairing.

Two nearby cases that are NOT affected, noted to save triage time:

* Omitting `thinking` entirely is safe. `claude-opus-5` defaults to adaptive thinking, so the upstream validator sees thinking as effectively on and accepts the injected strategy.
* Forced `tool_choice` is safe. `disableThinkingIfToolChoiceForced` (`claude_executor_request.go:355`) deletes the `thinking` key rather than setting it to `"disabled"`, which lands back on that same safe default.

Only an explicit `"disabled"` collides with the injected strategy.

**CLI Type**
claude code (Claude upstream, `api.anthropic.com`)

**Model Name**
claude-opus-5

**LLM Client**
Claude Code 2.1.220. Also reproducible directly with cURL, as shown below.

**Request Information**

Failing request. No `context_management`, so the injector fires:

```bash
curl -sS "https://<cliproxyapi-host>/v1/messages" \
  -H "Authorization: Bearer $KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"claude-opus-5","max_tokens":64,
       "thinking":{"type":"disabled"},
       "messages":[{"role":"user","content":"Say OK."}]}'
```

Result:

```
HTTP 400
{"type":"error","error":{"type":"invalid_request_error","message":"`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive"}}
```

Control request. Byte for byte identical except that the caller supplies its own `context_management`, so the injector returns early:

```bash
curl -sS "https://<cliproxyapi-host>/v1/messages" \
  -H "Authorization: Bearer $KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"claude-opus-5","max_tokens":64,
       "thinking":{"type":"disabled"},
       "context_management":{"edits":[]},
       "messages":[{"role":"user","content":"Say OK."}]}'
```

Result:

```
HTTP 200
```

The single variable between the two requests is whether `injectClaudeCodeContextManagement` fires.

**Expected behavior**

The injector should not add a thinking-clearing strategy to a request that has thinking disabled. A one line guard is enough:

```go
func injectClaudeCodeContextManagement(payload []byte) []byte {
	if gjson.GetBytes(payload, "context_management").Exists() {
		return payload
	}
	// clear_thinking_20251015 is rejected upstream unless thinking is enabled
	// or adaptive, and the real client never pairs the two either.
	if gjson.GetBytes(payload, "thinking.type").String() == "disabled" {
		return payload
	}
	updated, err := sjson.SetRawBytes(payload, "context_management", []byte(claudeCodeContextManagement))
	if err != nil {
		return payload
	}
	return updated
}
```

Worth noting that this raises fingerprint fidelity rather than lowering it. Real Claude Code with thinking disabled does not send a `clear_thinking_20251015` strategy either, so injecting one in that case is itself a deviation from the client being replicated.

**Screenshots**
Not applicable.

**OS Type**
* OS: Linux (CLIProxyAPI running as a container image, linux/amd64)
* Version: CLIProxyAPI v7.2.116

**Additional context**

Version bisect, by presence of the `injectClaudeCodeContextManagement` symbol:

| Version | Injector | Status |
| --- | --- | --- |
| v7.2.113 | absent | not affected |
| v7.2.114 | absent | not affected |
| v7.2.115 | absent | not affected |
| v7.2.116 | present | first affected |
| v7.2.117 to v7.2.119 | present | affected |
| main (6e92e3e6) | present | affected |

`git tag --contains a2933c77` returns v7.2.116 as its earliest tag. On current main the function body is byte identical to v7.2.116, so this is not yet fixed.

The regression was introduced by a2933c77 ("fix(claude): scope Anthropic beta and count_tokens policies"), which was the fix for #4750. That commit correctly scoped the hardcoded `Anthropic-Beta` list to Anthropic upstreams, then added this `context_management` body injection so the body would stay consistent with the advertised `context-management-2025-06-27` flag. The reference capture used to build the constant was taken with thinking enabled, so the thinking-disabled path was never exercised.

Practical impact: a client that has extended thinking switched off cannot use an Anthropic upstream through CLIProxyAPI at all on v7.2.116 or later, unless it happens to send its own `context_management`. Pinning to v7.2.115 is a working holding action.

Possibly related, though not the same defect: #4729 (executor reinjecting a thinking beta header the caller disabled) and #4772 (provider aware signed thinking replay).

DefinitelyNotSpammy:
## Summary

Fixes #4784.

Keep Claude Code `context_management` consistent with the final thinking
state after payload configuration.

The executor now handles both directions:

- removes only its untouched automatic `clear_thinking_20251015` object
  when final thinking becomes disabled;
- adds the automatic object when payload rules re-enable thinking and
  neither the caller nor a payload rule owns the field.

## Changes

- Skip initial automatic injection when thinking is already disabled.
- Track successful executor injection.
- Track caller ownership of `context_management`.
- Track applied payload rules targeting `context_management` or descendants.
- Preserve the existing payload-rule API through a compatibility wrapper.
- Reconcile the final payload after payload rules and forced-tool-choice handling.
- Preserve caller-provided, default-provided, override-provided, raw-rule-provided,
  and filtered values.
- Apply identical behavior to normal and streaming execution.
- Preserve existing payload-rule order and condition semantics.

## Tests

Added coverage for:

- directly disabled thinking;
- payload-rule transitions to disabled;
- disabled-to-enabled and disabled-to-adaptive transitions;
- normal and streaming execution;
- caller-owned context management;
- default and raw-default ownership;
- override and raw-override ownership;
- descendant-path ownership;
- exact same-value rule ownership;
- payload filters, including an already-absent field;
- omitted, enabled, adaptive, unknown, and invalid thinking states;
- forced tool choice;
- focused race detection.

## Validation

- Focused executor race tests: passed
- Focused payload-helper race tests: passed
- Payload-helper package tests: passed
- Server build: passed
- `gofmt`: passed
- `git diff --check`: passed
- Independent code, Go, and security reviews: no blocking findings

Three environment-sensitive fingerprint tests fail identically without
these changes and are unrelated to this pull request.

### DIFF
--- a/internal/runtime/executor/claude_executor_cloaking.go
+++ b/internal/runtime/executor/claude_executor_cloaking.go
@@ -676,8 +676,51 @@ const claudeCodeContextManagement = `{"edits":[{"type":"clear_thinking_20251015"
 // omitted it. CPA already claims context-management-2025-06-27 in Anthropic-Beta,
 // so a missing body field is an observable inconsistency with the real client. A
 // caller that sent its own object keeps it untouched.
-func injectClaudeCodeContextManagement(payload []byte) []byte {
+func injectClaudeCodeContextManagement(payload []byte) ([]byte, bool) {
 	if gjson.GetBytes(payload, "context_management").Exists() {
+		return payload, false
+	}
+	if gjson.GetBytes(payload, "thinking.type").String() == "disabled" {
+		return payload, false
+	}
+	updated, err := sjson.SetRawBytes(payload, "context_management", []byte(claudeCodeContextManagement))
+	if err != nil {
+		return payload, false
+	}
+	return updated, true
+}
+
+type claudeCodeContextManagementState struct {
+	eligible              bool
+	callerOwned           bool
+	automaticallyInjected bool
+	payloadRuleTouched    bool
+}
+
+// reconcileClaudeCodeContextManagement resolves automatic ownership after all
+// payload rules and forced tool-choice processing have completed.
+func reconcileClaudeCodeContextManagement(payload []byte, state claudeCodeContextManagementState) []byte {
+	thinkingType := gjson.GetBytes(payload, "thinking.type").String()
+	contextManagement := gjson.GetBytes(payload, "context_management")
+
+	if thinkingType == "disabled" {
+		if state.callerOwned || !state.automaticallyInjected || state.payloadRuleTouched {
+			return payload
+		}
+		if contextManagement.Raw != claudeCodeContextManagement {
+			return payload
+		}
+		updated, err := sjson.DeleteBytes(payload, "context_management")
+		if err != nil {
+			return payload
+		}
+		return updated
+	}
+
+	if thinkingType != "enabled" && thinkingType != "adaptive" {
+		return payload
+	}
+	if !state.eligible || state.callerOwned || state.payloadRuleTouched || contextManagement.Exists() {
 		return payload
 	}
 	updated, err := sjson.SetRawBytes(payload, "context_management", []byte(claudeCodeContextManagement))
--- a/internal/runtime/executor/claude_executor_execute.go
+++ b/internal/runtime/executor/claude_executor_execute.go
@@ -13,6 +13,7 @@ import (
 	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
 	sdktranslator "github.com/router-for-me/CLIProxyAPI/v7/sdk/translator"
 	log "github.com/sirupsen/logrus"
+	"github.com/tidwall/gjson"
 )
 
 func (e *ClaudeExecutor) Execute(ctx context.Context, auth *cliproxyauth.Auth, req cliproxyexecutor.Request, opts cliproxyexecutor.Options) (resp cliproxyexecutor.Response, err error) {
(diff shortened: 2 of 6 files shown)


sususu98 (approved):
Verified against native Claude Code 2.1.220/2.1.221 request-builder behavior and a live 2.1.221 OAuth Messages flow: explicit thinking.type=disabled omits context_management while the context-management beta remains present, and Anthropic returns 200. Re-ran the focused executor/helper tests with race detection, gofmt, server build, and diff check on current dev; all passed. The forced-tool-choice normalization is broader pre-existing behavior and should be handled separately rather than blocking this focused regression fix.

PR #4806, opened 2026-08-05
fix(security): secure credential and config file writes

### LINKED ISSUE #4672: Credential and config files are written world-readable (os.Create + umask)

DefinitelyNotSpammy:
## Summary

- harden new and existing credential/config files to mode `0600` on Unix/POSIX filesystems when supported
- attempt permission tightening before truncation without using initial `O_TRUNC`
- warn and preserve the previous write behavior when `chmod` is unsupported or denied
- preserve the `jsonEqual` no-write short-circuit while repairing an insecure mode when possible
- harden the Antigravity `project_id` write-back without recreating a credential deleted during discovery
- cover provider token storage, config persistence, auth filestore, management auth writes, and plugin-host auth writes with focused permission tests

## Compatibility notes

- Unix/POSIX files are hardened to `0600` when the filesystem permits it.
- If permission tightening is unsupported or denied, CLIProxyAPI logs a warning and continues with the existing persistence behavior.
- On Windows, Go `Chmod(0o600)` does not restrict inherited NTFS ACLs. Native Windows ACL hardening is outside this PR.
- Docker and other bind-mounted host files may change to mode `0600` after their first successful save, which may affect other host users, sidecars, NAS tools, or backup jobs that relied on broader access.

## Validation

- [x] focused auth, config, management, plugin-host, SDK, and helper package tests
- [x] `go test -race ./sdk/auth`
- [x] equivalent-JSON content/mtime regression test, repeated 20 times
- [x] deleted Antigravity file no-recreation regression test, repeated 20 times
- [x] `go build -o test-output ./cmd/server`
- [x] `git diff --check`
- [x] `git show --check HEAD`
- [ ] `go test ./...` — three unchanged `internal/runtime/executor` fingerprint-header tests expect `Linux` but receive `MacOS`; the same failures reproduce on the exact `upstream/dev` base `4906ead3` in this environment

Fixes #4672

Supersedes #4673

### DIFF
--- a/internal/misc/credentials.go
+++ b/internal/misc/credentials.go
@@ -3,6 +3,7 @@ package misc
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -11,6 +12,77 @@ import (
 
 // Separator used to visually group related log lines.
 var credentialSeparator = strings.Repeat("-", 67)
+
+// OpenFileForSecureRewrite opens a secret file, tightens its permissions when possible, and then truncates it.
+func OpenFileForSecureRewrite(path string) (*os.File, error) {
+	return openFileForSecureRewrite(path, os.O_WRONLY|os.O_CREATE)
+}
+
+// OpenExistingFileForSecureRewrite securely rewrites a secret file only when it already exists.
+func OpenExistingFileForSecureRewrite(path string) (*os.File, error) {
+	return openFileForSecureRewrite(path, os.O_WRONLY)
+}
+
+func openFileForSecureRewrite(path string, flags int) (*os.File, error) {
+	file, err := os.OpenFile(path, flags, 0o600)
+	if err != nil {
+		return nil, fmt.Errorf("open file for secure rewrite: %w", err)
+	}
+	ensureOpenFileMode0600(file)
+	if errTruncate := file.Truncate(0); errTruncate != nil {
+		_ = file.Close()
+		return nil, fmt.Errorf("truncate secure file: %w", errTruncate)
+	}
+	return file, nil
+}
+
+// EnsureFileMode0600 tightens an existing secret file's permissions when the filesystem supports it.
+func EnsureFileMode0600(path string) {
+	pathInfo, errLstat := os.Lstat(path)
+	if errLstat != nil {
+		log.WithError(errLstat).Warn("could not inspect secret file for permission hardening")
+		return
+	}
+	if !pathInfo.Mode().IsRegular() {
+		log.Warn("could not restrict permissions of a non-regular secret file")
+		return
+	}
+	if pathInfo.Mode().Perm() == 0o600 {
+		return
+	}
+
+	file, errOpen := os.OpenFile(path, os.O_RDWR, 0)
+	if errOpen != nil {
+		log.WithError(errOpen).Warn("could not open secret file for permission hardening")
+		return
+	}
+	defer func() {
+		if errClose := file.Close(); errClose != nil {
+			log.WithError(errClose).Warn("could not close secret file after permission hardening")
+		}
+	}()
+
+	fileInfo, errStat := file.Stat()
+	if errStat != nil {
+		log.WithError(errStat).Warn("could not inspect open secret file for permission hardening")
+		return
+	}
+	if !os.SameFile(pathInfo, fileInfo) {
+		log.Warn("secret file changed during permission hardening")
+		return
+	}
+	ensureOpenFileMode0600(file)
+}
+
+func ensureOpenFileMode0600(file *os.File) {
+	info, errStat := file.Stat()
+	if errStat == nil && info.Mode().Perm() == 0o600 {
+		return
+	}
+	if errChmod := file.Chmod(0o600); errChmod != nil {
+		log.WithError(errChmod).Warn("could not restrict secret file permissions to 0600")
+	}
+}
 
 // LogSavingCredentials emits a consistent log message when persisting auth material.
 func LogSavingCredentials(path string) {
--- a/internal/misc/credentials_test.go
+++ b/internal/misc/credentials_test.go
@@ -1,0 +1,57 @@
+//go:build unix
+
+package misc
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestOpenFileForSecureRewrite(t *testing.T) {
+	t.Run("creates a new file with mode 0600", func(t *testing.T) {
+		path := filepath.Join(t.TempDir(), "new-credentials.json")
+
+		file, err := OpenFileForSecureRewrite(path)
+		if err != nil {
+			t.Fatalf("OpenFileForSecureRewrite() error = %v", err)
+		}
+		if errClose := file.Close(); errClose != nil {
+			t.Fatalf("Close() error = %v", errClose)
+		}
+
+		info, err := os.Stat(path)
+		if err != nil {
+			t.Fatalf("Stat() error = %v", err)
+		}
+		if got := info.Mode().Perm(); got != 0o600 {
+			t.Errorf("file mode = %o, want 0600", got)
+		}
+	})
+
+	t.Run("restricts an existing file to mode 0600", func(t *testing.T) {
+		path := filepath.Join(t.TempDir(), "existing-mode-credentials.json")
+		if err := os.WriteFile(path, []byte("old credentials"), 0o664); err != nil {
+			t.Fatalf("WriteFile() error = %v", err)
+		}
+		if err := os.Chmod(path, 0o664); err != nil {
+			t.Fatalf("Chmod() error = %v", err)
+		}
+
+		file, err := OpenFileForSecureRewrite(path)
+		if err != nil {
+			t.Fatalf("OpenFileForSecureRewrite() error = %v", err)
+		}
+		if errClose := file.Close(); errClose != nil {
+			t.Fatalf("Close() error = %v", errClose)
+		}
+
+		info, err := os.Stat(path)
+		if err != nil {
+			t.Fatalf("Stat() error = %v", err)
+		}
+		if got := info.Mode().Perm(); got != 0o600 {
+			t.Errorf("file mode = %o, want 0600", got)
+		}
+	})
+}
--- a/sdk/auth/filestore.go
+++ b/sdk/auth/filestore.go
@@ -14,6 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/misc"
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
 	"github.com/router-for-me/CLIProxyAPI/v7/sdk/pluginapi"
 )
@@ -127,9 +128,10 @@ func (s *FileTokenStore) Save(ctx context.Context, auth *cliproxyauth.Auth) (str
 		}
 		if existing, errRead := os.ReadFile(path); errRead == nil {
 			if jsonEqual(existing, raw) {
+				misc.EnsureFileMode0600(path)
 				break
 			}
-			file, errOpen := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0o600)
+			file, errOpen := misc.OpenFileForSecureRewrite(path)
 			if errOpen != nil {
 				return "", fmt.Errorf("auth filestore: open existing failed: %w", errOpen)
 			}
@@ -144,8 +146,16 @@ func (s *FileTokenStore) Save(ctx context.Context, auth *cliproxyauth.Auth) (str
 		} else if !os.IsNotExist(errRead) {
 			return "", fmt.Errorf("auth filestore: read existing failed: %w", errRead)
 		}
-		if errWrite := os.WriteFile(path, raw, 0o600); errWrite != nil {
+		file, errOpen := misc.OpenFileForSecureRewrite(path)
+		if errOpen != nil {
+			return "", fmt.Errorf("auth filestore: write file failed: %w", errOpen)
+		}
+		if _, errWrite := file.Write(raw); errWrite != nil {
+			_ = file.Close()
 			return "", fmt.Errorf("auth filestore: write file failed: %w", errWrite)
+		}
+		if errClose := file.Close(); errClose != nil {
+			return "", fmt.Errorf("auth filestore: write file failed: %w", errClose)
 		}
 	default:
 		return "", fmt.Errorf("auth filestore: nothing to persist for %s", auth.ID)
@@ -307,7 +317,7 @@ func (s *FileTokenStore) readAuthFiles(path, baseDir string) ([]*cliproxyauth.Au
 				if errFetch == nil && strings.TrimSpace(fetchedProjectID) != "" {
 					metadata["project_id"] = strings.TrimSpace(fetchedProjectID)
 					if raw, errMarshal := json.Marshal(metadata); errMarshal == nil {
-						if file, errOpen := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0o600); errOpen == nil {
+						if file, errOpen := misc.OpenExistingFileForSecureRewrite(path); errOpen == nil {
 							_, _ = file.Write(raw)
 							_ = file.Close()
 						}
--- a/internal/api/handlers/management/auth_files_crud.go
+++ b/internal/api/handlers/management/auth_files_crud.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/misc"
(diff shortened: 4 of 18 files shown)


internal/misc/credentials.go
```diff
@@ -12,6 +13,23 @@ import (
...
 
+// OpenFileForSecureRewrite opens a secret file, tightens its permissions, and then truncates it.
+func OpenFileForSecureRewrite(path string) (*os.File, error) {
+	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0o600)
+	if err != nil {
+		return nil, fmt.Errorf("open file for secure rewrite: %w", err)
+	}
+	if errChmod := file.Chmod(0o600); errChmod != nil {
+		_ = file.Close()
+		return nil, fmt.Errorf("set secure file permissions: %w", errChmod)
```
sususu98: **Blocking: a failed `chmod` must not abort the write.**

`fchmod(2)` requires effective UID == file owner (or `CAP_FOWNER`); write permission alone is not enough. So there are real deployments where the old code wrote successfully and this code writes nothing.

Reproduced in Docker (Debian, file `root:1000` mode `0664`, process `uid=1000 gid=1000` — group-writable but not the owner):

```
os.Create                -> <nil>          # old behaviour: written, mode stays 0664
OpenFileForSecureRewrite -> set secure file permissions:
                            chmod /work/auths/token.json: operation not permitted
final content = {"refresh_token":"x"}      # the new credential was never written
```

Affected in practice:

- NFS volumes with `root_squash`
- CIFS/SMB shares (Synology / QNAP / TrueNAS + Docker — a meaningful slice of this project's users)
- exFAT/FAT and WSL2 DrvFs bind mounts
- the common compose upgrade path: container ran as root first, `user: "1000:1000"` added later

Blast radius: `SaveTokenToFile` fails -> `Manager.persist` errors (`conductor_cooldown.go:478` returns it) -> token refresh permanently broken; OAuth onboarding fails; management config save returns 500.

Since the status quo is already "permissions are loose", warning and continuing is strictly no worse than before, whereas failing hard is a pure functional regression. `internal/misc` already imports logrus:

```go
if errChmod := tightenOpenFileMode(file); errChmod != nil {
	// Bind mounts owned by another uid, NFS root_squash, CIFS/exFAT, WSL DrvFs.
	// Refusing to write here would break token refresh that worked before.
	log.WithError(errChmod).Warnf("could not restrict permissions of %s to 0600", path)
}
```

```go
// tightenOpenFileMode chmods only when the mode is not already 0600.
func tightenOpenFileMode(file *os.File) error {
	info, errStat := file.Stat()
	if errStat != nil {
		return errStat
	}
	if info.Mode().Perm() == 0o600 {
		return nil
	}
	return file.Chmod(0o600)
}
```

The `Stat` guard earns its keep independently: it skips a syscall on the hot path and avoids emitting a spurious `IN_ATTRIB` (Linux) / `NOTE_ATTRIB` (macOS) event on every credential write.

To be clear, the chmod-before-truncate ordering is right and I'd keep it — on failure the existing file is left untouched.

sdk/auth/filestore.go
```diff
@@ -126,14 +127,15 @@ func (s *FileTokenStore) Save(ctx context.Context, auth *cliproxyauth.Auth) (str
 			return "", fmt.Errorf("auth filestore: marshal metadata failed: %w", errMarshal)
 		}
 		if existing, errRead := os.ReadFile(path); errRead == nil {
+			dataToWrite := raw
 			if jsonEqual(existing, raw) {
-				break
+				dataToWrite = existing
 			}
```
sususu98: **Blocking: this re-introduces the fsnotify write-back loop that 39597267a fixed.**

The `break` being replaced here was not incidental. From 22ec41597 / 39597267a:

> This prevents unnecessary file writes when only these fields change during refresh, **breaking the fsnotify event -> Watcher callback -> refresh loop**. [...] Fixes antigravity channel excessive log generation issue.

Confirmed with a scratch test — two consecutive `Save` calls with identical metadata:

| branch | result |
| --- | --- |
| `7fef08ea` (base) | mtime unchanged, file not touched |
| `1874aaea` (this PR) | **mtime changes, rewritten every time** |

`internal/watcher/events.go:74` has `fsnotify.Write` in `authOps`, so every `Save` now wakes the watcher. It survives only thanks to the `authFileUnchanged` sha256 guard and the `len(data)==0` check in `clients.go:177` — that's defence in depth being spent for nothing.

It also diverges from the three sibling stores, which all kept the short-circuit: `internal/store/objectstore.go:207`, `postgresstore.go:258`, `gitstore.go:425`.

And look at the net effect: a credential file whose contents are already correct gets truncated to zero and rewritten with **byte-identical** content (you pass `existing`, not `raw`). The only thing that write accomplishes is the chmod — so just do the chmod:

```go
if jsonEqual(existing, raw) {
	// Contents already up to date: repair the mode without rewriting,
	// so the watcher write-back loop stays broken.
	if info, errStat := os.Stat(path); errStat == nil && info.Mode().Perm() != 0o600 {
		if errChmod := os.Chmod(path, 0o600); errChmod != nil {
			log.WithError(errChmod).Warnf("auth filestore: could not restrict permissions of %s", path)
		}
	}
	break
}
```

That still repairs legacy world-readable files — which is the whole point of the PR — without the pointless truncate/rewrite window on a file that didn't need writing.

sdk/auth/filestore.go
```diff
@@ -144,9 +146,17 @@ func (s *FileTokenStore) Save(ctx context.Context, auth *cliproxyauth.Auth) (str
 		} else if !os.IsNotExist(errRead) {
 			return "", fmt.Errorf("auth filestore: read existing failed: %w", errRead)
 		}
-		if errWrite := os.WriteFile(path, raw, 0o600); errWrite != nil {
+		file, errOpen := misc.OpenFileForSecureRewrite(path)
+		if errOpen != nil {
+			return "", fmt.Errorf("auth filestore: write file failed: %w", errOpen)
+		}
```
sususu98, line 152: Not this hunk, but the same file: `sdk/auth/filestore.go:320` (the antigravity `project_id` write-back) still uses the raw pattern and never tightens the mode:

```go
if file, errOpen := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0o600); errOpen == nil {
```

`O_TRUNC` without `O_CREATE` never applies `perm`, so a legacy `0644` auth file stays `0644` after that path runs — precisely the case this PR is trying to close. Please route it through the same helper so the two write paths in this file behave consistently.

internal/misc/credentials_test.go
```diff
@@ -0,0 +1,108 @@
+//go:build unix
```
sususu98, line 1: Two notes on the `unix` gating.

**1. On Windows this helper is a no-op, and nothing covers that.** In Go 1.26 `(*os.File).Chmod` goes to `internal/poll.FD.Fchmod`, whose Windows implementation only toggles `FILE_ATTRIBUTE_READONLY`:

```go
if mode&syscall.S_IWRITE != 0 {   // 0o600 contains 0200 -> taken
	attrs &^= syscall.FILE_ATTRIBUTE_READONLY
}
if attrs == d.FileAttributes {    // already not read-only
	return nil                    // nothing happens
}
```

So `Chmod(0o600)` neither errors nor has any effect, and credentials stay governed by inherited NTFS ACLs (`os.MkdirAll(dir, 0700)` is equally inert there). `pr-test-build.yml` only runs `ubuntu-latest`, so CI won't surface it either.

Scoping Windows out is a reasonable call — real hardening means rewriting the DACL through `golang.org/x/sys/windows` and deserves its own PR — but please state it in the PR description and release notes, otherwise Windows users will read this as "#4672 is fixed".

macOS and FreeBSD are genuinely covered: the `unix` build tag includes both, `golang.org/x/sys` is already a direct dependency in `go.mod`, and `GOOS=darwin/freebsd go vet` (which does compile test files) is clean.

**2. `unix.Umask` mutates process-global state**, and `sdk/auth/filestore_test.go:14,81` uses `t.Parallel()`. It happens to be safe today — parallel tests only resume after sequential ones return, so your `defer` restores it in time — but it's an implicit coupling that breaks the day someone adds `t.Parallel()` to one of these. More to the point, the helper chmods **unconditionally**, so the umask value cannot influence the assertion at all. I'd just drop the umask manipulation.

Separately, `O_TRUNC` -> `Truncate(0)` is a cross-platform change with no non-unix test. One build-tag-free test asserting truncate+write semantics would be worth having. (The change itself is safe on Windows: `syscall.Ftruncate` uses `SetFileInformationByHandle(FileEndOfFileInfo)`, which does not move the file pointer, so the write offset is still 0.)

sususu98 (changes requested):
Thanks for taking #4672 on — the root cause is real, the test coverage is solid, and doing the `chmod` **before** the truncate is exactly the right order (a failed chmod leaves the existing file intact).

I can't take it as-is though: it introduces two functional regressions, and on Windows it doesn't achieve its stated security goal at all. Details inline; cross-platform verification and the non-blocking items are below.

## What I verified

Against `1874aaea` (this PR) and `7fef08ea` (merge-base):

| check | result |
| --- | --- |
| `go build -o test-output ./cmd/server` | pass |
| `gofmt -l` on all 6 touched dirs | clean |
| `GOOS=windows/darwin/freebsd/linux go vet` (compiles test files too) | no new findings |
| `go test` on every touched package | pass |
| `go test ./internal/runtime/executor/...` | **passes here** — the 3 fingerprint failures in your checklist did not reproduce, they look network-dependent rather than pre-existing breakage |

## Blocking

1. **A failed `chmod` aborts the write entirely** — inline on `internal/misc/credentials.go`. Reproduced in Docker: where the process has write permission but is not the file owner, the old code wrote fine and this code refuses to write anything. That turns "works, permissions are loose" into "token refresh is permanently broken".
2. **`sdk/auth/filestore.go` drops the `jsonEqual` short-circuit** — inline. That `break` was added deliberately in 22ec41597 / 39597267a to break the `fsnotify event -> watcher callback -> refresh loop`; confirmed by test that the PR re-enables the rewrite.

## Non-blocking, but please address before merge

### Windows: the fix is inert — please say so in the description

Covered inline on the test file. Short version: `Chmod(0o600)` on Windows neither errors nor does anything, so credentials there remain governed by inherited NTFS ACLs. Scoping Windows out is fine, silently implying #4672 is fixed for Windows users is not.

### Docker: two release-note items

I tested a real bind mount (OrbStack/VirtioFS, both as root and `--user 1000:1000`). The chmod succeeds **and propagates to the host file**:

```
host before: -rw-rw-r--  auths/token.json      host before: -rw-r--r--  config.yaml
host after:  -rw-------  auths/token.json      host after:  -rw-------  config.yaml
```

1. After upgrading, `./config.yaml` and `./auths/*.json` on the host flip to `0600` on the first save. Anything reading them as a different user (NAS web UI, file manager, sidecar) will break. Worth one line in the release notes.
2. **Please do not "improve" `config.yaml` into a tmp+rename atomic write.** `docker-compose.yml` mounts `${CLI_PROXY_CONFIG_PATH:-./config.yaml}:/CLIProxyAPI/config.yaml` as a *single-file* bind mount, and I confirmed rename over it fails:

```
rename -> rename /CLIProxyAPI/config.yaml.tmp /CLIProxyAPI/config.yaml: device or resource busy
in-place open -> <nil>
```

The in-place write this PR keeps is the correct choice for that path. (Auth files live in a directory mount, so rename is fine there — that's what `internal/store/*` already does.)

### Scope gaps

- `sdk/auth/filestore.go:320` still uses the raw pattern — inline.
- Existing auth directories stay at `0755`; `MkdirAll` does not chmod a directory that already exists. Auth filenames embed account emails (`codex-<email>.json`), so a listable directory still leaks account identities. A one-time `os.Chmod(authDir, 0o700)` would close that half of #4672.
- `internal/api/handlers/management/oauth_sessions.go:448` has the same "perm only applies on create" property. `internal/store/*` is fine as-is — it writes a temp file and renames, so it always lands on a fresh `0600` inode.

## Priority

Must fix:
1. Warn-and-continue on `chmod` failure instead of aborting the write.
2. Restore `jsonEqual -> break` in `filestore.go`, repairing the mode with a bare `os.Chmod` instead of a rewrite.
3. Document the Windows no-op and the Docker host-side `0600` change.

Nice to have:
4. Skip the `chmod` syscall when the mode is already `0600`.
5. Cover `filestore.go:320`; consider chmod'ing an existing auth dir to `0700`.
6. Drop the process-global `unix.Umask` manipulation from the tests, and add one build-tag-free test for the truncate/write semantics.

internal/misc/credentials.go
DefinitelyNotSpammy: Addressed in 9aa1c5e7. OpenFileForSecureRewrite now checks the current mode, skips chmod when already 0600, logs the chmod error as a warning, and continues to truncate/write. The chmod attempt still occurs before truncation, and open/truncate/write/sync/close behavior otherwise remains with the existing callers. Focused affected-package tests and the server build pass; the PR description now documents the best-effort compatibility behavior.

sdk/auth/filestore.go
DefinitelyNotSpammy: Addressed in 9aa1c5e7. The jsonEqual branch again repairs mode only and immediately breaks without truncating or rewriting the file. Permission repair is bound to an identity-checked descriptor and warns/continues if it cannot harden the file. TestFileTokenStoreSaveMetadataSecuresEquivalentExistingFile now verifies unchanged bytes and unchanged mtime while repairing 0664 to 0600; it passed 20 consecutive runs.

sdk/auth/filestore.go
DefinitelyNotSpammy: Addressed in 9aa1c5e7. The Antigravity project_id write-back now uses OpenExistingFileForSecureRewrite, so an existing 0644/0664 file is hardened when supported without initial O_TRUNC. The existing-file test verifies persisted project_id and mode 0600. A second regression test blocks discovery, deletes the auth file, and proves the write-back does not recreate it; that test passed 20 consecutive runs.

internal/misc/credentials_test.go
DefinitelyNotSpammy: Addressed in 9aa1c5e7. The process-global unix.Umask manipulation was removed. A build-tag-free TestOpenFileForSecureRewriteTruncatesAndWritesFromOffsetZero now verifies truncation plus writing from offset zero without Seek, while POSIX mode assertions remain Unix-only. The PR description explicitly documents that Windows Chmod does not harden NTFS ACLs and that Docker/bind-mounted host files may change to 0600 after the first successful save.

DefinitelyNotSpammy:
@sususu98 The requested fixes are available in `9aa1c5e7`. I replied to each review thread with the implementation and test evidence, updated the PR compatibility notes, and left the threads unresolved for your verification. Could you please re-review when convenient?

luispater:
Thank you for the contribution. However, this is not considered an issue that requires modification in this project. Closing this pull request.

PR #4836, opened 2026-08-07
fix(claude): preserve advisor server tools for OAuth

lihan3238:
## Summary

- Treat Anthropic `advisor_*` definitions as native server tools during Claude OAuth request normalization.
- Preserve their `type`, `name`, and server-tool fields instead of rewriting them as MCP aliases.
- Add unit coverage at both the server-tool registry and OAuth remapping layers.

## Problem

Claude Code 2.1.224 can send `advisor_20260301`, which is a native Anthropic server tool and does not carry a custom `input_schema`. The OAuth remapper did not recognize the `advisor_` prefix, rewrote the definition as a custom MCP tool, and removed `type`. Anthropic then rejected the request with `tools.<index>.custom.input_schema: Field required`.

## Tests

- `go test -count=1 ./internal/runtime/executor/helps -run "^TestIsClaudeServerToolType$"`
- `go test -count=1 ./internal/runtime/executor -run "^TestRemapOAuthToolNames_AdvisorServerToolRemainsNative$"`
- `go build -o test-output ./cmd/server` from a standalone clone

The full executor suite also reports three platform-fingerprint expectation failures on the unchanged base commit in this local WSL environment: the tests expect `Linux`, while the stable device profile emits `MacOS`.

### DIFF
--- a/internal/runtime/executor/claude_executor_test.go
+++ b/internal/runtime/executor/claude_executor_test.go
@@ -4635,6 +4635,18 @@ func TestRemapOAuthToolNames_TypedCustomUsesMCPAlias(t *testing.T) {
 	}
 }
 
+func TestRemapOAuthToolNames_AdvisorServerToolRemainsNative(t *testing.T) {
+	body := []byte(`{"tools":[{"type":"advisor_20260301","name":"advisor","model":"claude-opus-5"}]}`)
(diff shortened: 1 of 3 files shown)


smsunarto:
Reproed issue, would love to have this merged

PR #4851, opened 2026-08-08
fix(responses): restore namespaces on custom tool calls

ngctro:
## Summary

- restore `name` and `namespace` for namespaced Responses custom tool calls translated through Claude
- apply the restoration to streaming added/done/completed items and non-stream output
- preserve the namespace when replaying custom tool call/output history

## Problem

Responses Lite places Codex's freeform `exec` tool under the `functions` namespace in `input[].additional_tools`. The Claude request translator correctly flattens that declaration to `functions__exec`, but the response translator returned a `custom_tool_call` with that flattened name and no namespace. Codex could not route it and reported:

```text
unsupported custom tool call: functions__exec
```

This change extends the existing function-call namespace restoration to custom tool calls, yielding `name: "exec"` and `namespace: "functions"` while retaining `type: "custom_tool_call"` and its freeform input.

## Validation

- `go test ./internal/translator/claude/openai/responses -count=1`
- `go test ./internal/translator/... -count=1`
- `go vet ./internal/translator/claude/openai/responses`
- live Codex 0.147.0 tool-loop test through a locally built patched CPA and the keelcode plugin: Codex executed `printf codex-tool-loop-fixed`, captured the expected output, and produced no `unsupported custom tool call` error

The full repository test run still has three pre-existing, environment-dependent failures in `internal/runtime/executor` where tests expect `X-Stainless-Os: Linux` but the local device profile reports `MacOS`; all translator tests pass.

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_request.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_request.go
@@ -1043,7 +1043,7 @@ func qualifyResponsesNamespaceToolName(namespaceName, childName string) string {
 	return namespaceName + "__" + childName
 }
 
-func splitResponsesQualifiedFunctionCallFromRequest(requestRawJSON []byte, qualifiedName string) (name, namespace string) {
+func splitResponsesQualifiedToolCallFromRequest(requestRawJSON []byte, qualifiedName string) (name, namespace string) {
 	qualifiedName = strings.TrimSpace(qualifiedName)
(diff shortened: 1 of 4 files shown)


ngctro:
Closing in favor of #4852, per the repository policy for standalone `internal/translator` changes from contributors without write access. The tested implementation remains available on the linked fork branch/commit for the maintenance team.

PR #4857, opened 2026-08-08
feat(cliproxy): add ranked auth selection and pinned one-shot execution

jaaacki:
## Withdrawn

This PR is withdrawn. The downstream integration had incorrectly elevated a local aspirational design into a required upstream dependency. CLIProxyAPI's existing public SDK surfaces already cover the immediate integration needs, and the downstream project will consume the latest official release rather than ask upstream to carry speculative APIs.

No downstream release will pin this fork branch, pseudo-version, or replacement.

### DIFF
--- a/sdk/cliproxy/auth/conductor_execution_once.go
+++ b/sdk/cliproxy/auth/conductor_execution_once.go
@@ -104,8 +104,10 @@ type ExecuteWithAuthOnceRequest struct {
 	Options cliproxyexecutor.Options
 	// RedirectPolicy optionally constrains redirects for the executor's HTTP
 	// client. Empty keeps the executor's own behavior, which is what every
-	// existing caller gets today. Enforcement is only possible while the
-	// manager-installed transport is in effect; see redirectPolicyRoundTripper.
+	// existing caller gets today. Enforcement is best effort here because the
+	// executor owns its http.Client; it holds only while the manager-installed
+	// transport is in effect (see onceRoundTripper). DoHTTPOnce owns the client
+	// itself and is therefore the path with the unconditional guarantee.
 	RedirectPolicy HTTPRedirectPolicy
 }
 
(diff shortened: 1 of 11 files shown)


jaaacki:
All required checks are green. This is rebased on the current v7.2.124/dev head and the full repository test suite passes. We need an official semver release containing these additive SDK primitives before the downstream consumer can remove its fail-closed placeholders; no fork, replace directive, or pseudo-version will be used. Maintainer review/merge when convenient would be appreciated.

jaaacki:
Addressed both automated review findings in `c01546ee`:

- due one-shot calls now all enter the per-auth refresh lock and wait/reuse the refreshed credential rather than dispatching a stale clone; concurrent regression coverage proves one refresh and no pre-refresh execute;
- redirect body detection now treats `http.NoBody` as bodyless, with an explicit zero-length GET redirect test.

Revalidated: required build, `go test ./sdk/...`, SDK auth race tests, and `go test ./...` all pass.

jaaacki:
@sususu98 all automated review findings through `60a22ba8` are addressed, including refresh concurrency/race safety and full ranked fallback traversal. The required build, full test suite, and SDK auth race suite are green locally; GitHub checks are rerunning on the final tip. This remains the only external gate for the downstream official-semver integration.

jaaacki:
Addressed the latest two review findings in `43aebeaf`:

- request-auth preparation failures are now recorded exactly once without claiming dispatch;
- any result after a followed credential-stripped redirect remains observable but availability-neutral.

Added focused regression tests and reran build, SDK tests, auth race tests, and the full suite successfully.

@codex review

jaaacki:
Hostile downstream integration review found one missing public hinge: ranked metadata existed on the internal mixed-provider funnel, but SDK callers could only invoke single-provider `SelectAuth`, forcing forbidden local provider probing when one semantic operation has routes across multiple executor keys.

`60b4e57f` adds additive `Manager.SelectAuthMixed`, which enters the existing mixed scheduler exactly once and preserves ranked/Home fail-closed semantics. Added cross-provider rank and zero-Home-dispatch tests; full build, SDK tests, auth race tests, and full suite pass.

@codex review

jaaacki:
Rebased cleanly onto current upstream `dev` (`b921b5d0`) as requested; new tip is `50ca8ad8`. Re-ran build, SDK tests, auth race tests, and full suite successfully after the rebase. @codex review

PR #4888, opened 2026-08-10
fix(claude): derive billing fingerprint from first user message to preserve prompt caching

Sewer56:
(Ignore this; DeepSeek V4 Flash got too excited when I was AFK; and it didn't even get it right. Closing.)

----------------

## Point

Prompt-cache reads broke in v7.2.116 (commit `f3e25ab2`, PR #4752). `cache_read_input_tokens` = 0 on every turn; full prefix re-written at 1.25x per request. Fix: fingerprint from first user message, not last.

## Cause then fix

Error: cache key changes every turn because billing fingerprint derives from last user message.

Cause: `claudeBillingFingerprintMessageText` reads last user text; billing block at `system[0]` sits before cache breakpoint `system[1]`. New turn → new fingerprint → prefix bytes change → cache miss.

Fix: derive from first user message (stable per conversation), keep last text part inside it (survives the currentDate `<system-reminder>` block).

## Repro

Stable 17KB system prompt, `cache_control` on system, two requests differing only in user text:

| Request                 | v7.2.115 (good)     | v7.2.116+ (broken)     |
| ----------------------- | ------------------- | ---------------------- |
| A                       | create 1902         | create 6559            |
| B (different user text) | read 1902, create 0 | read 0, re-create 6559 |

Verified live: v7.2.115 caches across turns; v7.2.116 and v7.2.128 re-write every turn.

## Change

- `claudeBillingFingerprintMessageText`: first user message via existing `firstClaudeUserMessageIndex`, last text part within it

- Billing shape unchanged: `cc_version=2.1.220.<fp>`; official capture fingerprint `e06` test kept

## Tests

1. `TestClaudeBillingFingerprintUsesFirstUserText` (renamed), asserts first-user semantics
2. `computeFingerprint` official 2.1.220 capture check unchanged
3. Full executor suite: 3 pre-existing failures only (`TestApplyClaudeHeaders_*`, `TestClaudeExecutor_NonClaudeRequestUsesClaudeCode220CLIFingerprint`), reproduce on clean main
4. Cloak idempotency tests pass (`TestCheckSystemInstructionsWithSigningMode_LongPromptIsExactAndIdempotent`)

### DIFF
--- a/internal/runtime/executor/claude_executor_cloaking.go
+++ b/internal/runtime/executor/claude_executor_cloaking.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Sewer56:
Didn't intend to PR, haha; DSv4 Flash kinda did it on its own when I AFK'd.

@sususu98 In any case, seems f3e25ab (or at least `v7.2.116`) broke prompt caching.
Not sure of the exact cause yet, ignore this PR though.

PR #4897, opened 2026-08-11
fix(claude): align cloaked cache placement and native Haiku helper passthrough

### LINKED ISSUE #4855: OpenAI Responses to Claude: prompt caching stops advancing after cloaking
### Problem

Claude prompt caching stops advancing with a multi-turn conversation after cloaking is applied.

Cloaking adds `cache_control` to the first real user message. Later, automatic cache setup only runs when the **entire request** has no cache controls:

```go
}
```

Source: [`claude_executor_stream.go`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.123/internal/runtime/executor/claude_executor_stream.go#L108-L114)

Because the early marker already exists, `ensureCacheControl` is skipped. This also skips [`injectMessagesCacheControl`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.123/internal/runtime/executor/claude_executor_cloaking.go#L1182-L1248), which normally moves a breakpoint to the second-to-last user turn.

### Result

As the conversation grows, only the initial prefix remains cached:

```text
messages:          20 -> 74
cache breakpoint:  messages[0] throughout
cached tokens:     approximately 35k throughout
cache hit rate:    97% -> 27%
```

The prompt prefix was append-only and stable. The hit rate fell because the breakpoint did not advance, not because the prompt diverged.

This is unexpected because multi-turn caching should cover previous conversation history, not remain fixed on the first user message.

### Introduced

This starts with [`f3e25ab`](https://github.com/router-for-me/CLIProxyAPI/commit/f3e25ab2bae6e7d533637d3ef1b64722525435d4), first released in `v7.2.116`.

That commit:

- changed unknown callers to be cloaked by default: [`Cloak: !confirmedClaudeCode`](https://github.com/router-for-me/CLIProxyAPI/blob/f3e25ab2bae6e7d533637d3ef1b64722525435d4/internal/runtime/executor/claude_executor_cloaking.go#L501-L518)
- added `cache_control` to the first real user text block: [`injectClaudeCodeCurrentDate`](https://github.com/router-for-me/CLIProxyAPI/blob/f3e25ab2bae6e7d533637d3ef1b64722525435d4/internal/runtime/executor/claude_executor_cloaking.go#L421-L425)

The existing global `countCacheControls(body) == 0` guard then prevents the rolling message breakpoint from being added.

### Suggested fix

Ensure each cache section independently instead of skipping all cache setup when any marker exists:

```go
body = ensureToolsCacheControl(body)
body = ensureSystemCacheControl(body)
body = ensureRollingMessageCacheControl(body)
body = enforceCacheControlLimit(body, 4)
```

The current combined implementation is here: [`ensureCacheControl`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.123/internal/runtime/executor/claude_executor_cloaking.go#L861-L874).

In particular, an existing tool, system, or cloaking marker should not suppress the rolling conversation marker. The existing four-breakpoint limit can still be enforced afterward.

sususu98:
## Summary

Three focused Claude executor fixes aligned with Claude Code 2.1.220 / 2.1.221 / 2.1.227 request-builder behavior. The third was found while live-testing the second and is an outright upstream rejection, so it is fixed here rather than left in place.

### 1. Cloaked prompt-cache ownership and native shape
- Cloaked Responses/Chat/Gemini→Claude traffic was skipping `ensureCacheControl` once the first-user cloak marker existed, freezing multi-turn cache hits.
- Always ensure section-independent system + rolling-message breakpoints for non-native callers; leave confirmed Claude Code placement intact.
- Default breakpoint is `{"type":"ephemeral"}` (no `ttl`). A `1h` body ttl is upgraded only after placement, only for OAuth credentials, and only onto blocks that already carry `cache_control`, so it stays paired with `extended-cache-ttl-2025-04-11`.
- Rolling selector matches native: last eligible turn, skip thinking-like assistant tails, final non-empty string system special case.
- Tools stay unstamped when a usable system prompt covers the prefix; fall back to a tools breakpoint when system is absent/empty/blank.
- Empty/whitespace string systems are not rewritten into a marked text block (avoids tools + whitespace double-stamp).

### 2. Native Haiku helper recognition and passthrough
- Measured Claude Code Haiku helpers intentionally omit the `claude-code` beta (and the minimal shape omits `system`). Recognize those profiles as confirmed native instead of cloaking them.
- Accept optional `parent_session_id`, require platform headers for presence + software baseline (not OS/Arch equality), derive timeout/version from the measured native constant.
- Keep discriminating signals strict: exact beta allowlist, body shape, lowercase hex CCH, session binding.
- Preserve helper transport headers; omit synthetic `stream:false` on markerless helpers; gate billing/CCH fallback on system presence so a pipeline-attached system cannot go upstream unsigned.

### 3. Automatic `context_management` without eligible thinking
- Anthropic rejects the injected `clear_thinking_20251015` strategy unless `thinking` is enabled or adaptive; both the injection and the reconciliation only excluded the literal `{"type":"disabled"}`, so an **absent** `thinking` field slipped through and the request 400'd.
- Reachable both when the caller never sent `thinking` and when `disableThinkingIfToolChoiceForced` deletes it between injection and reconciliation.
- Both sides now test for the accepted values via a shared `claudeThinkingAcceptsClearThinking`; caller-owned and payload-rule-owned objects keep their existing precedence.

## Related issues

Fixes #4855

## Commits
1. `fix(claude): restore cloaked prompt-cache ownership and native shape`
2. `fix(claude): recognize and pass through native Haiku helpers`
3. `fix(claude): drop auto context_management without eligible thinking`

## Review follow-ups (applied)

- **Helper detection no longer keys on `claude-header-defaults.timeout`.** `applyClaudeHeaders` routes a confirmed client through `misc.EnsureHeader`, which prefers the *incoming* header and only falls back to the configured value when the caller sent none — so a confirmed helper always forwards its own `600`. Comparing against the operator value meant any non-`600` `claude-header-defaults.timeout` rejected **every** genuine helper and cloaked it, defeating commit 2. The matcher now compares the measured constant `600` only. Guarded by `TestMeasuredHelperProfileIgnoresConfiguredStainlessTimeout` (verified to fail when the config-derived value is reintroduced). The test that pinned the old inverted contract was removed.
- **Corrected two comments that described the opposite of the code.** The ttl-upgrade comment claimed it "leaves an explicit caller-owned 5m choice untouched", but a caller's bare `{"type":"ephemeral"}` (which *is* 5m) is upgraded to `1h`; only an explicitly written `ttl` survives. The `x-client-request-id` comment implied the fresh `uuid.New()` is what preserves a helper's native ID, when `identityHeader`/`EnsureHeader` does.
- **Added the missing end-to-end regression guard for #4855**: `TestClaudeExecutorCloakedRollingCacheBreakpointAdvances` drives two different-length cloaked conversations through `Execute` and asserts the rolling breakpoint advances to the final turn while cloaking's first-user marker survives and the total stays within 4. Restoring the old global `countCacheControls(body) == 0` short-circuit makes it fail with `short=1 long=1`, i.e. the exact frozen-breakpoint symptom from the issue.

- **Confirmed native now keeps its own sampling parameters.** `normalizeClaudeSamplingForUpstream` deleted `temperature`/`top_p` from *every* Claude request. The measured structured Haiku helper sends `"temperature":1`, and `claudeCodeHelperShapeStructured` keys on exactly that value — so CPA required the field to classify the request and then stripped it before sending, emitting a shape no native client produces. Confirmed native now owns its sampling exactly as it already owns `cache_control` placement, and only combinations Anthropic actually rejects are dropped. Translated and cloaked callers keep the previous conservative normalization unchanged. Guarded by `TestNormalizeClaudeSamplingForUpstreamNativeKeepsMeasuredHelperTemperature` and `TestNormalizeClaudeSamplingForUpstreamNativeDropsOnlyRejectedCombinations` (both verified to fail against the old unconditional strip, and the latter also against a naive preserve-everything implementation).

  Anthropic's real constraints, probed live against `api.anthropic.com`:

  | thinking | accepted |
  | --- | --- |
  | off / absent | `temperature` any, `top_p` any, `top_k` any — but `temperature` + `top_p` together → 400 |
  | enabled | `temperature` only `1`; `top_p` only `>= 0.95`; `top_k` must be unset |

## Test plan
- [x] `go test ./internal/runtime/executor/ ./internal/runtime/executor/helps/`
- [x] `go test ./... -count=1`, `-race` on Claude/cache/helper/diagnostics, `go vet`, `gofmt -l`, `go build ./cmd/server`, `git diff --check`
- [x] Per-commit worktree build/test for all three commits
- [x] Reviewer: confirm cloaked multi-turn cache ownership change is intended — settled by the live measurement below
- [x] Reviewer: confirm API-key traffic no longer carries unpaired `ttl:1h` — verified by unit test + code path (`claudeCredentialUsesOAuth` gate, `TestClaudeExecutor_CacheTTLIsPairedWithExtendedCacheTTLBeta`), not live
- [x] Live: native Haiku helper through a real CPA keeps the measured wire — see "Live verification of the native helper wire" below

### Live verification of the cloaked cache-ownership question

The one substantive risk in commit 1 was dropping the `tools` breakpoint whenever a system exists: under cloak, `system[0]` is the `x-anthropic-billing-header:` block, and its bytes change on **every** request (the `cch` digits are an xxHash over the whole final body, and cloak's `cc_version` buildHash is `computeFingerprint` over the last user message). If that block were part of the prompt-cache key, every breakpoint after it would be poisoned and removing the tools breakpoint would take cloaked cache hits to zero.

Measured directly against `api.anthropic.com` with a Claude OAuth credential on `claude-haiku-4-5`, as a **read-only** probe against a pre-existing 4098-token cache entry (so cache-write throttling cannot confound it), with a no-change control before and after in the same run — reproduced across two independent runs:

| request | `cache_read_input_tokens` |
| --- | --- |
| baseline, no `system` field | 4098 |
| `+ system:[{billing block}]` | **4098** (same entry) |
| `+ system:[{"You are helpful."}]` (4 tokens) | 0 → creates/reads a separate 4102 entry |
| control, no `system` again | 4098 |

`system=[billing block]` is therefore **equivalent to sending no system at all**: Anthropic strips that block, and it contributes zero tokens to the prompt and to the cache key, while ordinary system content of only 4 tokens does change the key. The per-request `cch`/buildHash cannot poison any cache prefix, so leaving `tools` unstamped when a system exists is safe and matches the native builder.

Not established: per-section breakpoint efficacy (a `system` marker vs a message marker). Cache **creation** on this subscription credential is intermittently throttled — distinct large prefixes silently return `cache_creation_input_tokens=0` with no error — so that comparison was not measurable and is deliberately not claimed either way.

### Live verification of the native helper wire

Ran a debug CPA built from this branch (isolated state dir, its own port, OAuth credential with the refresh token stripped) and replayed both measured Haiku helper shapes through it, then read the upstream request log.

**Minimal helper** — upstream body is exactly `{model, max_tokens, messages, metadata}`: no `system`, so no billing block and no CCH; no `stream` key (the measured non-stream helper omits it rather than sending `false`); no `cache_control`, `context_management` or `diagnostics`; `Anthropic-Beta` forwarded verbatim with no `claude-code-20250219` and no `extended-cache-ttl-2025-04-11`; `Accept-Encoding: gzip` and no `X-Stainless-Async`, matching the minimal transport profile. HTTP 200.

**Structured helper** — upstream top-level keys are exactly the captured native order:

```
model, messages, system, tools, metadata, max_tokens, thinking, temperature, output_config, stream
```

with `temperature: 1` preserved, the caller's three `system` blocks untouched, no `cache_control`/`context_management`/`diagnostics` injected, `stream: true` and `X-Stainless-Async: async` preserved, and `Anthropic-Beta` forwarded verbatim. The billing block's `cch` is re-signed (`00000` → real digest) and `metadata.user_id` identity is remapped to the selected credential while `session_id` is preserved, both of which are the intended `ApplyClaudeCredentialMetadata` behaviour. HTTP 200 with a normal SSE stream.

Control in the same run: a plain non-native caller sending `temperature`/`top_p`/`top_k` still reaches upstream with `temperature` and `top_p` stripped and `top_k` kept — identical to the previous behaviour.

The measured shapes themselves were re-confirmed against real captured native traffic (316 `POST api.anthropic.com/v1/messages` flows from Claude Code `2.1.220`/`2.1.221` on OAuth). Across a spread of 24, only the structured helper carries `temperature`; main conversation turns, the minimal helper and model probes all omit it — which is exactly what the detector encodes.

### Live verification of the `context_management` rejection

Found while live-testing the helper wire, and fixed here rather than left in place because it is a hard upstream rejection.

Anthropic refuses the strategy CPA injects unless thinking is enabled or adaptive:

```
400 invalid_request_error:
`clear_thinking_20251015` strategy requires `thinking` to be enabled or adaptive
```

`injectClaudeCodeContextManagement` and `reconcileClaudeCodeContextManagement` both special-cased only the literal `{"type":"disabled"}`, so an **absent** `thinking` field read back as gjson's empty string and slipped through both. Two reachable paths produced the rejected body:

1. a cloaked caller that never sent `thinking` at all;
2. a caller whose `thinking` was enabled at injection time and then deleted by `disableThinkingIfToolChoiceForced`, which runs between injection and reconciliation.

Both now test for the accepted values through a shared `claudeThinkingAcceptsClearThinking` helper, so reconciliation also withdraws an object CPA injected itself once thinking disappears. Caller-owned and payload-rule-owned objects keep their existing precedence and are never withdrawn.

Probed directly against `api.anthropic.com` with `context_management` fixed and only `thinking` varying:

| `thinking` | result |
| --- | --- |
| absent | **400** `requires thinking to be enabled or adaptive` |
| `disabled` | **400** same |
| `enabled` | 200 |
| `adaptive` | reaches the model check (`adaptive thinking is not supported on this model`), i.e. accepted by the strategy |

Live through a real CPA on this branch: a cloaked no-thinking request now sends no `context_management` and returns 200 (previously 400), while a thinking-enabled request still carries it and returns 200.

The existing expectation that forced tool choice *retains* the automatic object pinned exactly the rejected shape, so it is inverted. Both halves of the fix were negative-control verified independently.

## Notes
- Related but distinct from #4731 (configurable default TTL): this PR aligns with native pairing (default no `1h`; OAuth-only upgrade + beta), not a free-form default TTL config.
- Scope is Claude executor / client detection only; no translator changes.

## Open reviewer question

Codex flagged (P2) that a markerless helper backed by an **API-key** credential now forwards the helper's OAuth-only beta profile (every allowed profile starts with `oauth-2025-04-20`) together with `x-api-key`, instead of being cloaked.

Verified: the mechanism is real, but it is **pre-existing** for confirmed native callers rather than introduced here. `baseBetas = incomingBetas` has always applied to any `confirmedClaudeCode` request regardless of credential kind — a standard (non-helper) native request on an API-key credential already emits `claude-code-20250219,oauth-2025-04-20,interleaved-thinking-2025-05-14` with `x-api-key` today. What commit 2 changes is only *which callers count as confirmed native*. The `!helperProfile` gate itself behaves correctly: `extended-cache-ttl-2025-04-11` is appended for OAuth std-native and for neither helper case.

So gating **only** the helper profile on OAuth would make helper and standard native traffic inconsistent on the same credential. The consistent alternatives are to leave this as-is, or to OAuth-gate confirmed-native beta passthrough as a whole — which is a broader behavior change than this PR's scope and touches the "confirmed native preserves its measured wire shape" invariant. Flagging for a maintainer decision rather than silently picking one.

### DIFF
--- a/internal/runtime/executor/caching_verify_test.go
+++ b/internal/runtime/executor/caching_verify_test.go
@@ -1,9 +1,12 @@
 package executor
 
 import (
+	"bytes"
 	"fmt"
+	"strings"
 	"testing"
 
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
 	"github.com/tidwall/gjson"
 )
 
@@ -36,8 +39,8 @@ func TestEnsureCacheControl(t *testing.T) {
 		}
 	})
 
-	// Test case 3: Tools are cached
-	t.Run("Tools Caching", func(t *testing.T) {
+	// Test case 3: Native Claude Code does not auto-stamp tools; system still caches.
+	t.Run("Tools Not Auto Cached", func(t *testing.T) {
 		input := []byte(`{
 			"model": "claude-3-5-sonnet",
 			"tools": [
@@ -49,21 +52,19 @@ func TestEnsureCacheControl(t *testing.T) {
 		}`)
 		output := ensureCacheControl(input)
 
-		// cache_control should only be on the LAST tool
-		tool0Cache := gjson.GetBytes(output, "tools.0.cache_control")
-		tool1Cache := gjson.GetBytes(output, "tools.1.cache_control.type")
-
-		if tool0Cache.Exists() {
-			t.Errorf("cache_control should NOT be on the first tool")
-		}
-		if tool1Cache.String() != "ephemeral" {
-			t.Errorf("cache_control not found on last tool. Output: %s", string(output))
+		if gjson.GetBytes(output, "tools.0.cache_control").Exists() || gjson.GetBytes(output, "tools.1.cache_control").Exists() {
+			t.Errorf("default ensureCacheControl must not stamp tools[*].cache_control: %s", string(output))
 		}
 
-		// System should also have cache_control
-		systemCache := gjson.GetBytes(output, "system.0.cache_control.type")
-		if systemCache.String() != "ephemeral" {
+		systemCache := gjson.GetBytes(output, "system.0.cache_control")
+		if systemCache.Get("type").String() != "ephemeral" {
 			t.Errorf("cache_control not found in system. Output: %s", string(output))
+		}
+		// The native constructor spreads ttl in only when a ttl is selected, so the
+		// default breakpoint carries none. upgradeClaudeCacheControlTTL adds it later
+		// for the credentials native uses the 1h pool on.
+		if systemCache.Get("ttl").Exists() {
+			t.Errorf("default system cache_control must not carry ttl. Output: %s", string(output))
 		}
 	})
 
@@ -94,24 +95,70 @@ func TestEnsureCacheControl(t *testing.T) {
 		}
 	})
 
-	// Test case 5: Only tools, no system
-	t.Run("Only Tools No System", func(t *testing.T) {
+	// Test case 5: tools without any system prompt. Native always sends a system
+	// prompt, so this shape only reaches CPA from OpenAI/Gemini translation where the
+	// caller supplied no system message. Without a tools breakpoint the sole marker
+	// would sit on the volatile final message and a stateless caller would rewrite
+	// the whole tools prefix on every request.
+	t.Run("Only Tools No System Falls Back To Tools Breakpoint", func(t *testing.T) {
 		input := []byte(`{
 			"model": "claude-3-5-sonnet",
 			"tools": [
-				{"name": "tool1", "description": "Tool", "input_schema": {"type": "object"}}
+				{"name": "tool1", "description": "Tool", "input_schema": {"type": "object"}},
+				{"name": "tool2", "description": "Tool", "input_schema": {"type": "object"}}
 			],
 			"messages": [{"role": "user", "content": "Hi"}]
 		}`)
 		output := ensureCacheControl(input)
 
-		toolCache := gjson.GetBytes(output, "tools.0.cache_control.type")
-		if toolCache.String() != "ephemeral" {
-			t.Errorf("cache_control not found on tool. Output: %s", string(output))
+		if gjson.GetBytes(output, "tools.0.cache_control").Exists() {
+			t.Errorf("only the last tool may host the fallback breakpoint: %s", string(output))
+		}
+		if got := gjson.GetBytes(output, "tools.1.cache_control.type").String(); got != "ephemeral" {
+			t.Errorf("missing tools fallback breakpoint when system is absent: %s", string(output))
+		}
+		if gjson.GetBytes(output, "tools.1.cache_control.ttl").Exists() {
+			t.Errorf("tools fallback breakpoint must not carry a default ttl: %s", string(output))
+		}
+		if got := gjson.GetBytes(output, "messages.0.content.0.cache_control.type").String(); got != "ephemeral" {
+			t.Errorf("rolling message breakpoint should still be present: %s", string(output))
 		}
 	})
 
-	// Test case 6: Many tools (Claude Code scenario)
+	t.Run("Empty System Still Falls Back To Tools", func(t *testing.T) {
+		for name, system := range map[string]string{
+			"empty array":  `"system": [],`,
+			"empty string": `"system": "",`,
+			"blank string": `"system": "   ",`,
+		} {
+			t.Run(name, func(t *testing.T) {
+				input := []byte(`{
+					"model": "claude-3-5-sonnet",
+					` + system + `
+					"tools": [{"name": "tool1", "description": "Tool", "input_schema": {"type": "object"}}],
+					"messages": [{"role": "user", "content": "Hi"}]
+				}`)
+				output := ensureCacheControl(input)
+
+				if got := gjson.GetBytes(output, "tools.0.cache_control.type").String(); got != "ephemeral" {
+					t.Errorf("an unusable system prompt must still yield a tools breakpoint: %s", string(output))
+				}
+				// Empty/blank string system must not be rewritten into a marked text
+				// block; that would double-stamp tools + a whitespace system host.
+				if bytes.Contains(output, []byte(`"text":""`)) || bytes.Contains(output, []byte(`"text":"   "`)) {
+					t.Errorf("unusable string system must stay unconverted: %s", string(output))
+				}
+				if gjson.GetBytes(output, "system.0.cache_control").Exists() {
+					t.Errorf("unusable system must not receive its own breakpoint: %s", string(output))
+				}
+				if countCacheControls(output) != 2 {
+					t.Errorf("want tools+message breakpoints only, got %d in %s", countCacheControls(output), string(output))
+				}
+			})
+		}
+	})
+
+	// Test case 6: Many tools (Claude Code scenario) — default skips tools.
 	t.Run("Many Tools (Claude Code Scenario)", func(t *testing.T) {
 		// Simulate Claude Code with many tools
 		toolsJSON := `[`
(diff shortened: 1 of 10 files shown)


sususu98:
Related issue: #4855 (OpenAI Responses to Claude: prompt caching stops advancing after cloaking).

Commit 1 is the direct fix for that report: cloaked first-user markers no longer freeze multi-turn cache placement.

Commit 2 is separate native Haiku helper recognition/passthrough work on the same Claude executor surface; no dedicated issue filed for that part.

Sewer56:
Thanks!
Ran into this yesterday.

Issue #4913, opened 2026-08-11
[Bug] Responses WebSocket retains ~1.3 GiB via gjson and causes 221 GiB transcript allocation churn


ramapitecusment:
**Is it a request payload issue?**

[ ] Yes
[x] No. This is process-level live-heap retention and allocation churn under sustained Responses WebSocket traffic.

## Summary

This is a follow-up to #4878 after the uTLS connection fix in `5d9b6299`.

The connection/goroutine accumulation was one part of that report. Forced-GC heap and cumulative-allocation profiles show two additional, independent memory problems in the Responses WebSocket path:

1. approximately **1.3 GiB of live heap** retained through `gjson.getBytes` and the WebSocket tool cache;
2. approximately **221.66 GiB of cumulative allocations** over 5.4 hours from repeatedly rebuilding and mutating a growing Responses transcript.

These paths remain unchanged in current `dev` at `934da237`.

## Environment

- macOS arm64, 64 GiB RAM
- CLIProxyAPI `v7.2.125`, built with Go 1.26.4
- Client: Codex Desktop
- Model: `gpt-5.6-sol`
- Transport: Responses WebSocket mode
- Long-running Codex task with a growing transcript
- `plugins.enabled: false`
- `request-log: false`

## Observed process state

After approximately 4.5–5.4 hours of sustained use:

- RSS: approximately **3.0 GiB**
- forced-GC live heap: **1472.54 MiB**
- cumulative allocations: **226,980.63 MiB (221.66 GiB)**
- allocation objects: approximately **62.9 million**

The cumulative value is allocation churn, not 221 GiB retained simultaneously. It still creates substantial GC and RSS pressure as the conversation grows.

## 1. Live heap retained through the WebSocket tool cache

Forced-GC heap profile:

```text
Total live heap:                   1472.54 MiB
github.com/tidwall/gjson.getBytes: 1296.08 MiB (88.02%)
bytes.Clone:                        127.38 MiB
Cached tool-call/output bodies:      ~10.90 MiB
```

Dominant retaining stack:

```text
OpenAIResponsesAPIHandler.ResponsesWebsocket
  -> responsesWebsocketToolCacheTurn.recordRequest
  -> gjson.GetBytes
```

Relevant current `dev` code:

[...]

PR #4921, opened 2026-08-12
perf(openai): detach websocket tool cache from transcripts

ramapitecusment:
## Summary

- detach cached Responses WebSocket `call_id` values from the parsed transcript backing allocation;
- scan request and response payloads through the repository's no-copy GJSON helper;
- keep independent copies of every value that survives the recording call;
- add a deterministic backing-storage regression test, a caller-buffer ownership test, and an 8 MiB allocation benchmark.

This addresses the live-heap retention mechanism reported in #4913. The separate full-transcript reconstruction churn from that issue is intentionally out of scope.

## Root cause

`gjson.GetBytes(payload, "input")` allocates a string copy of the complete input array. Each item's `call_id` can be a substring of that allocation. `recordItem` copied `item.Raw`, but stored the short `call_id` directly as a map key and in an ordering slice, keeping the full parsed input allocation reachable for the turn lifetime.

The recorder now reads from the immutable caller buffer without copying the full JSON subtree, then explicitly clones the short identifier and the raw tool item before either can escape the call.

## TDD regression

The regression creates an 8 MiB Responses input containing a large message and a small tool item. It records both `function_call` and `function_call_output` cases and inspects backing-storage ownership directly. On current `dev`, both cases fail because the stored ID points inside the complete parsed `input` allocation.

A second test calls `recordRequest`, overwrites the caller's payload after the function returns, and verifies that both the stored ID and raw tool item remain intact. This guards the lifetime contract introduced by the no-copy scan.

## Allocation result

`BenchmarkResponsesWebsocketToolCacheTurnRecordLargeRequest`, Apple M5 Pro, 8 MiB payload, `-benchtime=5x`:

| | Before | After |
| --- | ---: | ---: |
| bytes/op | 8,397,702 | 904 |
| allocs/op | 9 | 9 |

The payload-size allocation is removed; the remaining allocations are the turn maps/slices, detached identifier, and copied small tool item.

## Compatibility

Cache keys, eviction, TTL, commit behavior, and stored JSON bytes are unchanged. The no-copy results do not escape the recording methods, and the tests prove that retained data does not alias the caller buffer. This is inside the OpenAI Responses WebSocket handler and does not alter provider translation or request/response wire formats for Codex, Claude, Kimi, or Antigravity.

## Verification

- focused regression suite, `-count=20`
- focused regression suite under `-race`, `-count=3`
- 8 MiB benchmark with `-benchmem`
- `go vet ./sdk/api/handlers/openai`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket_toolcall_repair.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket_toolcall_repair.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/util"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 )
@@ -246,7 +247,7 @@ func (t *responsesWebsocketToolCacheTurn) recordRequest(payload []byte) {
 	if t == nil || len(payload) == 0 {
 		return
 	}
-	input := gjson.GetBytes(payload, "input")
+	input := util.GetGJSONBytesNoCopy(payload, "input")
 	if !input.Exists() || !input.IsArray() {
 		return
 	}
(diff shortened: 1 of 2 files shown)


ramapitecusment:
Superseded by c845ce15 on dev, which refactored the same request/tool-cache path and marked #4913 Fixed. The remaining response-buffer ownership and Codex continuation coverage is now isolated in #4924. Closing this conflicting duplicate rather than carrying the old implementation across the refactor.

PR #4922, opened 2026-08-12
perf(openai): bound websocket transcript reconstruction

ramapitecusment:
## Summary

- merge the previous input, previous response output, and new input into one validated JSON-array allocation instead of two decode/append/encode cycles;
- return unchanged transcripts directly from both dedupe passes when no duplicate exists;
- retain the established encoding/json path for the uncommon duplicate cases;
- apply small root-field mutations before installing the merged `input`, so `sjson` does not copy the complete transcript for `model`, `instructions`, and `stream`;
- use a no-copy view of the retained previous input while building the new owned result.

This addresses the cumulative transcript-reconstruction allocation churn in #4913. The independent live-heap tool-cache retention is handled separately by #4921.

## Allocation result

`BenchmarkNormalizeResponseSubsequentRequestLargeTranscript`, Apple M5 Pro, 8 MiB retained history plus assistant/tool output and a new tool result:

| | `dev` | This PR | Reduction |
| --- | ---: | ---: | ---: |
| time/op | 170.9 ms | 23.1–24.5 ms | 85.7–86.5% |
| bytes/op | 207,145,149 | 33,591,176–33,591,275 | 83.8% |
| allocs/op | 154 | 41–42 | 72.7–73.4% |

The remaining large allocations are intentional ownership boundaries: the merged array, the final request containing it, and the independent `lastRequest` state copy.

## TDD and compatibility coverage

The initial RED tests prove that both no-op dedupe passes allocate a new transcript on current `dev`. They now require the exact input string and backing storage to be reused when no duplicate exists.

Additional tests cover:

- first-occurrence behavior for duplicate tool `call_id` values;
- referenced tool-call selection for duplicate item IDs through the existing regression suite;
- ordered merging across arrays, empty parts, `null`, nested objects, escaped strings, primitives, and raw `<tag>` content;
- malformed/non-array input attribution, preserving whether the old caller reports `invalid previous response output` or `invalid request input`;
- complete non-passthrough WebSocket transcript merging.

The compaction, transcript-replacement, `previous_response_id`, pending tool-call, reasoning-item, and incremental-passthrough branches are unchanged. Duplicate cases still use the previous encoding/json implementation. The final normalized JSON values and item ordering remain the same; only redundant intermediate representations are removed.

This also follows the current [OpenAI GPT-5.6 persisted-reasoning guidance](https://developers.openai.com/api/docs/guides/latest-model): manual history must preserve every response output item, and `store: false`/ZDR callers must replay encrypted reasoning items. The optimized merge treats every array item as opaque validated JSON and writes its raw bytes in order; it does not decode, drop, or rewrite reasoning content.

This code is in the generic Responses WebSocket normalization path, so the full repository suite was used to cover Codex, Claude, Kimi, Antigravity, and all translator packages. No network credentials or live upstream calls are required.

## Verification

- focused semantic and integration tests, `-count=20`
- focused suite under `-race`, `-count=3`
- full `sdk/api/handlers/openai` package
- benchmark with `-benchmem`, `-benchtime=5x`, `-count=3`
- `go vet ./sdk/api/handlers/openai`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/sdk/api/handlers/openai/openai_responses_websocket_prewarm.go
+++ b/sdk/api/handlers/openai/openai_responses_websocket_prewarm.go
@@ -1,7 +1,6 @@
 package openai
 
 import (
-	"encoding/json"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


ramapitecusment:
Superseded by c845ce15 on dev, which replaced the transcript merge/repair path and marked #4913 Fixed. The remaining semantic regression coverage is now isolated in #4924. Closing this conflicting duplicate rather than rebasing an obsolete implementation.

PR #4932, opened 2026-08-12
fix(models): embed Grok 4.6 catalog entry

rluisr:
## Summary

Add `grok-4.6` to the embedded xAI model catalog so a fresh or offline CLIProxyAPI instance can resolve it before remote catalog refresh succeeds.

The remote `router-for-me/models` catalog already contains this entry, but the embedded fallback did not. That left xAI OAuth requests failing locally with `unknown provider for model grok-4.6`.

## Changes

- Add Grok 4.6 metadata, reasoning support, and text/image modalities to `internal/registry/models/models.json`.
- Add a regression test covering the embedded xAI model definition.

## Testing

- `go test ./internal/registry -count=1`
- `go build -o /var/folders/y8/vnh__wy973g2l_ckn6cfx2bns3fp19/T/opencode/cli-proxy-api-pr-final ./cmd/server`
- Validated the catalog JSON with `jq`.

Refs router-for-me/models#39.

### DIFF
--- a/internal/registry/model_definitions_test.go
+++ b/internal/registry/model_definitions_test.go
@@ -60,6 +60,26 @@ func TestWithXAIBuiltinsIncludesVideo15GAAndPreviewAlias(t *testing.T) {
 	}
 }
 
+func TestGetXAIModelsIncludesGrok46(t *testing.T) {
+	for _, model := range GetXAIModels() {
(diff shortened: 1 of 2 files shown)


rluisr:
@hkfires could you please review this?

PR #4955, opened 2026-08-14
fix(codex): honor negotiated protocol for ChatGPT requests

barry-ai:
I'm in a enterprise networks, TLS inspection gateways, the upstream may negotiate or return HTTP/1.1, CLIProxyAPI still parses the response as HTTP/2, it fails with:

```txt
http2: frame too large, note that the frame header looked like an HTTP/1.1 header
```

The ChatGPT/Codex upstream client establishes connections with a Chrome-style uTLS fingerprint. Previously, it always created an http2.ClientConn after the TLS handshake, regardless of the protocol actually negotiated through TLS ALPN.

This patch reads the negotiated ALPN protocol after the TLS handshake:
 - When h2 is negotiated, it keeps the existing HTTP/2 request path.
 - When http/1.1 is negotiated, or ALPN is omitted, it sends and receives HTTP/1.1 over the same TLS connection.
 - For unsupported protocols, it closes the connection and returns a clear error.

The patch does not change the TLS fingerprint or force a specific HTTP version. It makes the client honor the protocol actually negotiated with the upstream, preventing HTTP/1.1 responses from being parsed as HTTP/2 frames.

### DIFF
--- a/internal/runtime/executor/helps/utls_client.go
+++ b/internal/runtime/executor/helps/utls_client.go
@@ -37,6 +37,37 @@ type closeConnectionBody struct {
 	err             error
 }
 
+const (
+	maxHTTP1ResponseHeaderBytes = 10 << 20
+	maxInterimHTTP1Responses    = 10
+)
+
+var errHTTP1ResponseHeadersTooLarge = errors.New("utls: HTTP/1.1 response headers too large")
+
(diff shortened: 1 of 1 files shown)


barry-ai:
who will merge this patch？

PR #4956, opened 2026-08-14
fix(handlers): skip inactive before-auth interceptors

linger-alpha:
## What

Skip the before-auth request interceptor path when the plugin host reports that no request interceptors are active.

## Why

The server installs a non-nil plugin host even when no plugins are enabled. The after-auth path already uses `requestInterceptorsEnabled`, but the before-auth path only checked whether the host was nil.

As a result, a request with no active interceptors still passed through the empty plugin host. That path cloned the request body in the handler, cloned it again in the host, and then retained two handler-side copies returned from the no-op interception.

This was observed on a small-memory host handling a large Responses request. A forced-GC pprof `inuse_space` profile attributed about 110 MiB to `applyRequestInterceptorsBeforeAuth -> cloneBytes` for an approximately 55 MiB request body, despite no active plugins. 

## Behavior and compatibility

- Hosts reporting active request interceptors keep the existing before-auth behavior.
- Hosts that do not implement the detector keep the conservative existing behavior because `requestInterceptorsEnabled` defaults to true for unknown hosts.
- Only hosts explicitly reporting no active request interceptors take the zero-copy fast path.
- The change makes before-auth consistent with the existing after-auth guard.

## Checks

- `go test ./sdk/api/handlers -count=1`
- `go build -o /tmp/cpa-pr1-server-build ./cmd/server`

The regression test also checks that the interceptor is not called and the request body slices retain their original backing storage when request interceptors are inactive.

### DIFF
--- a/sdk/api/handlers/handlers_interceptors.go
+++ b/sdk/api/handlers/handlers_interceptors.go
@@ -425,7 +425,7 @@ func interceptStreamChunk(ctx context.Context, host PluginInterceptorHost, req p
 
 func (h *BaseAPIHandler) applyRequestInterceptorsBeforeAuth(ctx context.Context, handlerType, requestedModel, requestID string, req coreexecutor.Request, opts coreexecutor.Options, skipPluginID string) (coreexecutor.Request, coreexecutor.Options, *interfaces.ErrorMessage) {
 	host := h.interceptorHost()
-	if host == nil {
+	if !requestInterceptorsEnabled(host) {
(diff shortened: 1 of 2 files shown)


linger-alpha:
This optimization has since been independently implemented and merged in #4360.

The current dev branch already guards the before-auth path with requestInterceptorsEnabled and includes pointer-reuse regression coverage and allocation benchmarks, so this PR is now redundant.

Closing as superseded by #4360.

PR #4996, opened 2026-08-15
fix(auth): do not cooldown credentials on transport-level failures

windedge:
## Problem

A transport-level failure on a provider (for example open.bigmodel.cn via the openai-compatible provider returning "net/http: TLS handshake timeout") put the only credential into a 60s transient cooldown. Subsequent requests then failed first with a 500 "TLS handshake timeout" and afterwards with 503 "auth_unavailable: no auth available" until the cooldown expired, because non-quota cooldowns surface as blockReasonOther and produce a misleading auth_unavailable error.

## Fix

Classify transport-level failures as connection_lifecycle so they skip credential cooldown:

- net/http: TLS handshake timeout
- dial tcp: connection refused / connection reset
- broken pipe
- DNS no such host
- network is unreachable
- proxyconnect error
- i/o timeout

This is consistent with the existing design for client cancellation and WebSocket EOF disconnects, which already skip cooldown since they say nothing about credential health.

## Safety

The classification only applies to errors without an HTTP status. Status-bearing upstream errors (401/429/5xx response bodies) still cool down credentials exactly as before. A test covers this case.

## Tests

- Added TestManager_MarkResult_TransportFailureDoesNotCooldown in sdk/cliproxy/auth/connection_lifecycle_cooldown_test.go
- Added TestIsConnectionLifecycleError_TransportTextWithStatusStillCooldowns in sdk/cliproxy/auth/connection_lifecycle_cooldown_test.go
- Adjusted one pre-existing table case in sdk/cliproxy/auth/cooldown_backoff_test.go to use a non-transport message

`go test ./sdk/cliproxy/auth/ -count=1` passes.

## Note

This change does not add in-request retries for these errors; it prevents credential penalization so the next request retries naturally.

### DIFF
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -1369,15 +1369,20 @@ func vertexSALocation(auth *Auth) string {
 // (auth.Disabled or Status StatusDisabled) and peers currently blocked for
 // the model (quota cooldown, retry-after, or auth-level unavailability, as
 // judged by isAuthBlockedForModel) cannot serve requests, so they are ignored
-// when judging whether an alternative endpoint exists. When pollable peers
-// exist, a transport failure is specific to this auth's endpoint and rotation
-// can only reach a healthy credential if this auth cools down. Caller must
-// hold m.mu.
+// when judging whether an alternative endpoint exists. Peers that do not
+// support the route model (per authSupportsRouteModel, the same judgment
+// request selection applies) can never be picked for this model and are
+// ignored too, otherwise a healthy but model-ineligible peer would hide the
+// endpoint uniqueness and leave polling stuck on the failing auth. When
+// pollable peers exist, a transport failure is specific to this auth's
+// endpoint and rotation can only reach a healthy credential if this auth
+// cools down. Caller must hold m.mu.
 func (m *Manager) vertexTransportFailureIsAuthSpecificLocked(auth *Auth, modelKey string, now time.Time) bool {
 	loc := vertexSALocation(auth)
 	if loc == "" {
 		return false
 	}
+	registryRef := registry.GetGlobalRegistry()
 	hasPeer := false
 	for _, peer := range m.auths {
 		if peer == nil || peer.ID == auth.ID {
@@ -1391,6 +1396,9 @@ func (m *Manager) vertexTransportFailureIsAuthSpecificLocked(auth *Auth, modelKe
 		}
 		peerLoc := vertexSALocation(peer)
 		if peerLoc == "" {
+			continue
+		}
+		if !m.authSupportsRouteModel(registryRef, peer, modelKey) {
 			continue
 		}
 		hasPeer = true
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -743,7 +743,7 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 			}
 		} else {
 			if modelKey != "" {
-				if !shouldSkipCredentialCooldown(result.Error) {
+				if !shouldSkipCredentialCooldownForAuth(result.Error, auth) {
 					disableCooling := m.cooldownDisabledForAuth(auth)
 					state := ensureModelState(auth, modelKey)
 					state.Unavailable = true
@@ -1309,6 +1309,38 @@ func shouldSkipCredentialCooldown(err *Error) bool {
 	return isRequestScopedResultError(err) || isConnectionLifecycleResultError(err)
 }
 
+// shouldSkipCredentialCooldownForAuth reports failures that must not mark auth/model cooling.
+// Transport-level failures still cool down credentials that carry their own proxy
+// override, because those failures indicate auth-specific routing problems.
+func shouldSkipCredentialCooldownForAuth(err *Error, auth *Auth) bool {
+	if !shouldSkipCredentialCooldown(err) {
+		return false
+	}
+	if isTransportFailureResultError(err) && authHasProxyOverride(auth) {
+		return false
+	}
+	return true
+}
+
+// authHasProxyOverride reports whether the auth carries its own proxy configuration.
+// Transport failures through a per-auth proxy are auth-specific routing faults,
+// not shared network outages, so they should keep credential cooldown.
+func authHasProxyOverride(auth *Auth) bool {
+	return auth != nil && strings.TrimSpace(auth.ProxyURL) != ""
+}
+
+// isTransportFailureResultError reports whether the result error is classified as a
+// transport-level failure (as opposed to client cancellation or stream EOF), so
+// callers can restore cooldown when the failure is auth-specific. Only errors that
+// were classified connection-lifecycle, carry no HTTP status, and whose message
+// matches a transport pattern qualify.
+func isTransportFailureResultError(err *Error) bool {
+	if err == nil || statusCodeFromResult(err) != 0 {
+		return false
+	}
+	return hasTransportFailureMessage(err.Message)
+}
+
 // isConnectionLifecycleError reports transport/session lifecycle failures that must
 // not cool credentials: client cancellation, WebSocket close/EOF disconnects, and
 // transport-level dial/TLS/write failures that carry no credential signal.
@@ -1371,6 +1403,16 @@ func isConnectionLifecycleMessage(message string) bool {
 	}
 	// Transport-level failures (dial, DNS, TLS handshake, write) never indicate
 	// credential health, so they must not cool credentials either.
+	return hasTransportFailureMessage(lower)
+}
+
+// hasTransportFailureMessage reports whether a message matches a known transport-level
+// failure pattern (dial, DNS, TLS handshake, proxy connect, read/write timeouts).
+func hasTransportFailureMessage(message string) bool {
+	lower := strings.ToLower(strings.TrimSpace(message))
+	if lower == "" {
+		return false
+	}
 	for _, pattern := range transportFailureMessagePatterns {
 		if strings.Contains(lower, pattern) {
 			return true
@@ -1824,7 +1866,7 @@ func applyAuthFailureState(auth *Auth, resultErr *Error, retryAfter *time.Durati
 	if auth == nil {
 		return
 	}
-	if shouldSkipCredentialCooldown(resultErr) {
+	if shouldSkipCredentialCooldownForAuth(resultErr, auth) {
 		return
 	}
 	defer func() {
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -1327,13 +1327,16 @@ func shouldSkipCredentialCooldownForAuth(err *Error, auth *Auth) bool {
 // pool-aware Vertex regional-endpoint disambiguation. A transport failure on
 // a Vertex service-account auth whose regional endpoint is unique in the pool
 // (other Vertex service-account auths exist, none sharing its location) is an
-// auth-specific routing fault and must keep credential cooldown. Caller must
-// hold m.mu.
+// auth-specific routing fault and must keep credential cooldown. Failures in
+// the shared proxy dial layer never reach a regional endpoint, so they are
+// never attributed to one. Caller must hold m.mu.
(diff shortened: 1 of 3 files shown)


windedge:
Addressed: transport-classified failures now keep the transient cooldown when the selected auth carries its own Auth.ProxyURL, since a dead per-auth proxy is an auth-specific routing fault. Rotation moves off that credential while shared-network failures (global proxy or direct) still skip cooldown. Client cancellation and stream EOF lifecycle failures also still skip cooldown in all cases.

Note on base_url: deliberately not treated as a connection-specific override - openai-compatible credentials always carry Attributes["base_url"] (it is mandatory, shared by every credential of the provider), so including it would reintroduce single-credential lockout for the most common deployment while providing no rotation benefit when the shared endpoint is down.

Tests added for both the model-level and auth-level cooldown paths, plus a case confirming client cancellation still skips cooldown with a per-auth proxy.

windedge:
Addressed the review feedback: isTransportFailureResultError now requires the connection_lifecycle code explicitly before the per-auth proxy transport override applies, so request-scoped errors (Code request_scoped, e.g. a Claude Fast error wrapping a status-less TLS handshake timeout) always skip credential cooldown. Added a test covering request_scoped code + transport message + per-auth proxy (TestManager_MarkResult_RequestScopedTransportTextWithAuthProxySkipsCooldown); the earlier TestManager_MarkResult_TransportFailureWithAuthProxyStillCooldowns still passes.

windedge:
Declining this suggestion; base_url is not treatable as an auth-specific override in general.

For openai-compatible providers the base URL is provider-level shared configuration: OpenAICompatibilityAPIKey carries no per-key BaseURL, and the synthesizer copies the provider baseURL into Attributes["base_url"] of every credential under that provider (internal/watcher/synthesizer/config.go). A non-empty base_url is therefore the norm for every openai-compatible credential, not a per-credential override signal.

Including base_url in the override check would mean that any openai-compatible credential hitting a transient transport blip (for example the TLS handshake timeout in the issue that motivated this PR) gets cooled again - reintroducing exactly the single-credential lockout this PR fixes, now for every openai-compatible deployment. And when a shared endpoint is genuinely down, all credentials fail alike, so cooldown provides no rotation benefit, only lockout harm.

The scenario worth protecting - a Claude/Codex/Gemini key with a private BaseURL override while siblings use the default endpoint - is real, but the auth record alone cannot distinguish a per-key custom endpoint from a provider-shared value. Doing that correctly needs an explicit marker at synthesis time (or comparison against sibling credentials), which is a separate follow-up feature rather than a tweak to this predicate.

windedge:
Addressed in fdae3e5c: Vertex location is genuinely per-credential config (written at credential import time in internal/api/handlers/management/vertex_import.go), unlike the openai-compatible base_url case, so this point is valid and different from the base_url suggestion.

However, restoring cooldown for every Vertex service-account auth would regress the original fix, because location always defaults to us-central1 and single-credential Vertex users would go back to 60s lockouts on TLS timeouts. The implemented check is therefore pool-aware: a transport failure keeps cooldown only when other Vertex service-account auths exist in the pool and none shares the failing auth's location (i.e. its regional endpoint is unique, so rotation can only reach a healthy endpoint if this auth cools down). Shared-location pools and single-credential setups keep skipping cooldown, preserving the original transport-failure fix.

Added coverage:
- TestManager_MarkResult_TransportFailureWithUniqueVertexLocationStillCooldowns (model-level and auth-level paths keep cooldown; client cancellation still skips)
- TestManager_MarkResult_TransportFailureWithSharedVertexLocationSkipsCooldown (shared location and single-credential pools still skip)

windedge:
Both points addressed in 6b7ef6a2.

Proxy dial failures are no longer attributed to the location-derived endpoint: shouldSkipCredentialCooldownPoolAware now ignores proxy-connect-layer failures (message containing "proxyconnect tcp") when deciding vertex attribution, so a shared global proxy outage skips cooldown instead of draining the pool. Per-auth ProxyURL handling via authHasProxyOverride is unchanged and still restores cooldown. Covered by TestManager_MarkResult_TransportFailureWithProxyDialSkipsVertexAttribution, including a regression guard that direct endpoint failures (no such host) with a unique location still cool down.

transportFailureMessagePatterns now also recognizes "connection timed out", "operation timed out", and "server misbehaving" (net.DNSError), so these shared transport forms classify as connection_lifecycle and skip cooldown. Verified both layers that consume the list: isConnectionLifecycleMessage (classification) and isTransportFailureResultError (auth-specific restore gate); status-bearing errors remain coolable at both layers. Covered by TestManager_MarkResult_TransportFailureNewPatternsSkipCooldown.

windedge:
Both points addressed in e0278df9.

Regional attribution is now restricted to endpoint-attributable failures via a dedicated allowlist (vertexEndpointFailureMessagePatterns: no such host, connection refused, connection reset, broken pipe, i/o timeout, connection timed out, tls handshake timeout). Shared host-level forms such as "network is unreachable" and DNS "server misbehaving" no longer trigger vertex attribution and keep skipping cooldown, as does any "proxyconnect tcp" dial failure (excluded explicitly, since proxy errors often embed endpoint-style substrings like "connection refused"). isTransportFailureResultError and transportFailureMessagePatterns are untouched; they answer "is this a transport error" for the per-auth proxy path, which is a different question. Covered by TestManager_MarkResult_VertexSharedHostNetworkFailureSkipsCooldown, TestManager_MarkResult_VertexEndpointConnectionRefusedStillCooldowns, and the round-5 proxy test which still passes.

Disabled peers are now filtered in vertexTransportFailureIsAuthSpecificLocked using the same criterion the scheduler applies (peer.Disabled || peer.Status == StatusDisabled, per selector.go isAuthBlockedForModel and scheduler.go upsertAuthLocked), so a sole selectable credential is never cooled because of a disabled different-location peer. Covered by TestManager_MarkResult_VertexDisabledPeerSkipsCooldown.

windedge:
Addressed in 0513e940.

vertexTransportFailureIsAuthSpecificLocked now receives the model key and the failure timestamp, and filters peers with the exact criterion the scheduler uses for candidate selection (isAuthBlockedForModel in selector.go, which also handles model=="" by degrading to the auth-level availability check). A same-location peer that is currently cooling therefore no longer counts as a pollable alternative, so the failing credential restores cooldown and rotation fails over to the healthy different-location credential. Expired cooldowns still count as pollable, matching scheduler promotion behavior.

Covered by TestManager_MarkResult_VertexCoolingPeerStillCooldowns (failing EU auth + cooling EU peer + healthy US peer, model-level and auth-level) and TestManager_MarkResult_VertexExpiredCooldownPeerSkipsCooldown. The existing disabled-peer and unique-location tests were adjusted where the new filtering changes shared-fixture expectations.

windedge:
Addressed in a7e6cc32.

Peer attribution now applies the same model-support predicate request selection uses: peers are additionally filtered with authSupportsRouteModel, so a same-location peer that is not registered for the failing model no longer counts as a pollable alternative and the failing credential restores cooldown, letting rotation fall back to the healthy different-location credential. For the auth-level path (empty model key) the predicate is a no-op by design (authSupportsRouteModel returns true for empty route model), so behavior there is unchanged.

Covered by TestManager_MarkResult_VertexModelIneligiblePeerStillCooldowns (failing EU auth + ready model-ineligible EU peer + healthy US peer). The existing model-level Vertex tests were updated to register models explicitly, matching how the service layer registers credentials before selection (Manager.Register alone does not), so those tests now exercise the intended branches under production registration semantics.

windedge:
Did a self-review pass over the cumulative diff and fixed the findings in 189157b9:

- Major: SOCKS5 proxy dial failures ("socks connect tcp ...->...: connection refused", produced by golang.org/x/net/internal/socks via sdk/proxyutil/proxy.go) bypassed the "proxyconnect tcp" exclusion and would be misattributed to vertex regional endpoints, cooling every unique-location credential during a shared global socks5 proxy outage. "socks connect" is now excluded alongside "proxyconnect tcp". Covered by TestManager_MarkResult_TransportFailureWithSocksProxyDialSkipsVertexAttribution.
- Peer scan now checks vertexSALocation first so non-vertex peers skip the model-state/registry lookups inside the write lock; the standalone Disabled check was dropped since isAuthBlockedForModel already covers it (behavior unchanged).
- Test helper registerClientModelForTest now unregisters from the global registry via t.Cleanup, matching the existing pattern in this package.
- Added a regression test for the "connection timed out" attribution pattern.

Verified no other cooldown-writing path bypasses the pool-aware check (refresh, ephemeral/home results, availability-neutral records are intentionally unaffected), no lock-ordering issues (m.mu -> registry mutex, leaf lock), and no stale callers of the applyAuthFailureState signature.

windedge:
All three points verified against the code and addressed in 3782c01c.

Proxy mode: authHasProxyOverride now parses ProxyURL with proxyutil.Parse and only treats a real proxy endpoint (ModeProxy) as an override. "direct"/"none" (ModeDirect, documented in config.example.yaml) and empty values no longer restore cooldown, so a sole credential with proxy-url: direct keeps the transport-failure skip. Invalid/unparseable values still count as an override since a broken per-credential proxy config is an auth-specific fault. Covered by TestManager_MarkResult_TransportFailureWithDirectOrNoneProxySkipsCooldown (direct and none, model-level and auth-level).

Provider normalization: vertexSALocation now compares the provider trimmed and case-insensitively (EqualFold), matching executorKeyFromAuth's ToLower(TrimSpace(...)) used by selection and executor routing. Covered by TestManager_MarkResult_VertexProviderCaseNormalizedStillCooldowns.

Zero-weight peers: the attribution loop now consults the active selector. Only under WeightedRoundRobinSelector (where positiveWeightAuths and the scheduler's requirePositiveWeight both exclude weight<=0) do zero-weight peers no longer count as pollable; the default RoundRobinSelector ignores weight, so attribution correctly keeps counting them there. Covered by TestManager_MarkResult_VertexWeightedPeerZeroWeightStillCooldowns and TestManager_MarkResult_VertexRoundRobinIgnoresWeightSkipsCooldown.

windedge:
All three points verified and addressed in 401ddabb.

Shared proxy: a real per-auth ProxyURL still restores cooldown when it is credential-specific (sole credential, or no other selectable peer shares that URL), preserving TestManager_MarkResult_TransportFailureWithAuthProxyStillCooldowns. When another selectable peer (unblocked, and for the model path also registered for the model) has the same ProxyURL, the outage is treated as shared transport and cooldown is skipped, so a common company proxy dying no longer drains the whole pool. Covered by TestManager_MarkResult_TransportFailureWithSharedProxySkipsCooldown and TestManager_MarkResult_TransportFailureWithDedicatedProxyStillCooldowns.

Vertex peer eligibility now reuses the selector helpers instead of hand-mirroring more criteria. vertexPeerCandidatesLocked feeds candidates through availableAuthsForSelector (highest available priority tier, same as availableAuthsForRouteModel with allPriorities=false) and applies positiveWeightAuths when the active selector is weighted, including the SessionAffinitySelector.fallback wrapping that newRoutingSelector installs for weighted+affinity. Those helpers do not take m.mu, so calling them from MarkResult's write lock is safe.

Zero-weight peers behind an affinity-wrapped weighted selector and lower-priority same-location peers therefore no longer suppress cooldown. Covered by TestManager_MarkResult_VertexSessionAffinityWeightedZeroWeightStillCooldowns (both SessionAffinity constructors) and TestManager_MarkResult_VertexLowPrioritySameLocationPeerStillCooldowns.

windedge:
Addressed in 67e05895.

authHasSharedProxyPeerLocked now applies the same selector-eligibility pipeline as the vertex path: candidates (non-self, same provider via executorKeyFromAuth, unblocked per isAuthBlockedForModel, model-registered for the model path) pass through availableAuthsForSelector (highest available priority tier), and positiveWeightAuths is applied when the active selector is weighted, including the SessionAffinitySelector.fallback wrapping. A zero-weight peer sharing the proxy therefore no longer suppresses cooldown under weighted routing.

One deliberate behavior note: shared-proxy peers are now limited to the same provider (executorKeyFromAuth match), matching how request selection builds per-provider candidate sets. A cross-provider credential sharing the URL can never serve this credential's requests, so counting it would wrongly skip cooldown and keep rotation pinned to the dead proxy.

Covered by TestManager_MarkResult_TransportFailureWithSharedProxyWeightedZeroWeightPeerStillCooldowns (weighted, model and auth level) and TestManager_MarkResult_TransportFailureWithSharedProxyRoundRobinIgnoresWeightSkipsCooldown (default round-robin keeps skipping, since plain RR ignores weight).

PR #5015, opened 2026-08-16
fix(auth): share session-affinity binding across thinking-suffix variants and stop replaced routing selectors

### LINKED ISSUE #5016: [BUG]auth: session-affinity splits sessions across thinking-suffix model variants
- [ ] Yes, this is a request payload issue.
- [x] No, it's another issue. (Routing / session-affinity behavior, reproducible from configuration.)

**Describe the bug**

With `routing.session-affinity: true`, the affinity cache key in `SessionAffinitySelector.Pick`/`OnResult` embeds the raw route model, while round-robin cursors and cooldown states already canonicalize thinking suffixes via `canonicalModelKey`. A client that alternates between `claude-sonnet-4-5` and `claude-sonnet-4-5(high)` within one conversation gets two independent session bindings and can land on different credentials mid-conversation, silently breaking stickiness. `OnResult` failure cleanup can also miss the binding for the same reason, so a broken binding may not be released.

**CLI Type**
Any (observed with claude code / codex style clients behind the proxy).

**Model Name**
Any model using thinking-suffix variants, e.g. `claude-sonnet-4-5` vs `claude-sonnet-4-5(high)`.

**LLM Client**
Claude Code (any client that sends thinking-suffix model variants within one session).

**Request Information**
Not a payload issue; no request log needed. Deterministic given `routing.session-affinity: true`; reproduced with unit tests against `sdk/cliproxy/auth`.

**Expected behavior**
A session that switches between thinking-suffix variants of the same base model keeps its single credential binding, and failure cleanup releases it reliably.

**OS Type**
- OS: any (reproduced on Windows and Linux/WSL2)

**Additional context**
Regression tests and a fix are ready; linked PR below.

WilliamLambertCN:
Fixes #5016

## Summary

Two small fixes in the session-affinity routing area:

1. **Thinking-suffix variants split session bindings** (`selector.go`): the affinity cache key embedded the raw route model, while round-robin cursors (`RoundRobinSelector.Pick`) and cooldown states already canonicalize thinking suffixes via `canonicalModelKey`. A session sending `model` and `model(high)` was therefore split into two independent bindings and could land on different credentials, silently breaking stickiness; `OnResult` failure cleanup could also miss the binding for the same reason. Both `Pick` and `OnResult` now canonicalize the model when building cache keys, matching the rest of the routing pipeline.

2. **Routing config reloads leak the affinity selector's cleanup goroutine** (`service_config.go`): the Service's config path creates a fresh routing selector on every routing config change, and `SessionAffinitySelector` owns a `SessionCache` with a background cleanup goroutine — but the replaced selector was never stopped. Selector lifecycle now follows ownership: the Service tracks the routing selector it installed (`appliedRoutingSelector`) and stops it when a reload replaces it. `Manager.SetSelector` deliberately stays caller-agnostic and never stops anything, because an implicit stop there cannot see custom wrappers that retain the previous selector as a delegate. `SessionCache.Stop` is additionally made idempotent and concurrency-safe, since the Service stop and `StopAutoRefresh` can overlap — previously two concurrent stops could double-close `stopCh` and panic.

## Tests

New regression tests, all red before the fix and green after:

- `TestSessionAffinitySelector_ThinkingSuffixSharesSessionBinding`
- `TestSessionAffinityOnResult_ThinkingSuffixReleasesBinding`
- `TestManagerSetSelectorDoesNotStopReplacedSelector`
- `TestSessionCacheStopIsConcurrencySafe`
- `TestApplyManagerConfigStopsReplacedServiceAffinitySelector`
- `TestApplyManagerConfigTracksServiceInstalledSelector`

## Verification

- `go test ./sdk/cliproxy/...`: fully green.
- `go test -race ./sdk/cliproxy/ ./sdk/cliproxy/auth/` on **Linux (WSL2, go1.26.0 linux/amd64)**: green.

### DIFF
--- a/sdk/cliproxy/auth/conductor_set_selector_test.go
+++ b/sdk/cliproxy/auth/conductor_set_selector_test.go
@@ -1,0 +1,48 @@
+package auth
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+// SetSelector never stops the selector it replaces: selectors are owned by
+// whoever created them (the SDK caller or the Service's config path), and
+// implicitly stopping one breaks compositions that retain it, such as a
+// custom wrapper delegating to the previous selector.
+func TestManagerSetSelectorDoesNotStopReplacedSelector(t *testing.T) {
+	t.Parallel()
+
+	affinity := NewSessionAffinitySelector(&RoundRobinSelector{})
+	defer affinity.Stop()
+	manager := NewManager(nil, affinity, nil)
+
+	manager.SetSelector(&RoundRobinSelector{})
+	select {
+	case <-affinity.cache.stopCh:
+		t.Fatal("SetSelector stopped the caller-owned affinity selector")
+	default:
+	}
+	if current := manager.Selector(); current == Selector(affinity) {
+		t.Fatal("SetSelector did not swap the active selector")
+	}
+}
+
+// SessionCache.Stop must be safe to invoke concurrently: the Service's config
+// path, manager shutdown, and SDK caller cleanup can all stop the same
+// selector. A select-then-close idiom would double-close the channel and
+// panic.
+func TestSessionCacheStopIsConcurrencySafe(t *testing.T) {
+	t.Parallel()
+
+	cache := NewSessionCache(time.Minute)
+	var wg sync.WaitGroup
+	for i := 0; i < 8; i++ {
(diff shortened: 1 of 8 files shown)


WilliamLambertCN:
Splitting into two single-bug PRs to make review easier.

PR #5022, opened 2026-08-16
fix(interactions): stateful tool-calling cycle for antigravity agent (resume + structured results)

xitex:
## Problem

Google's **antigravity agent** (Interactions API) requires **stateful-only function calling** ([docs](https://ai.google.dev/gemini-api/docs/antigravity-agent#limitations)):

> Function calling is only supported in stateful mode. You must use `previous_interaction_id` to continue the turn; reconstructing history manually (stateless mode) is not supported.

Clients delegating through the OpenAI-compatible layer (`/v1/chat/completions`) had no way to satisfy this:

1. The proxy never captured `interaction_id` / `environment_id` from upstream responses, so a tool-result turn could not resume the interaction. Google rejected the replayed history with `400: Cannot specify tool calls outside of Turn items`.
2. Clients that send no explicit session marker (`Session-Id` header / `session_id` field) could not be keyed for continuation state at all.
3. Client tools named like the agent's built-in sandbox tools (`read_file`, `write_file`) caused `500 Unknown Error` — Google rejects re-declaring intrinsic tools.
4. OpenAI SDKs validate models via `GET /v1/models/{name}` before delegating; the missing endpoint made clients fall back to other providers.

### 1. Stateful continuation capture + rewrite
- Capture `interaction.id` + `environment_id` from every antigravity upstream response (stream and non-stream), keyed by conversation session.
- On a turn carrying `function_result`: attach `previous_interaction_id` + `environment` and strip input down to bare `function_result` steps (per docs' Turn-2 format).

### 2. Session key fallback for stateless clients
- New fallback key: canonical-JSON SHA-256 of the stable conversation prefix (messages up to and including the last user message).
- The tool-result turn shares this prefix with the initial turn, so both map to the same key and the continuation resolves.
- A new user message starts a fresh key. Canonical hashing keeps the key stable across clients that re-serialize history differently.

### 3. Tool-name collision round-trip
- Client tools colliding with antigravity built-ins are renamed with an `external_` prefix on the way upstream and restored to client-facing names on **every** response path (chat-completions stream/non-stream, responses translator: declarations, replay, output items, step events). Shared implementation in `internal/translator/common`.

### 4. `GET /v1/models/:model`
- Single-model endpoint so OpenAI-compatible SDKs can validate `antigravity-preview-05-2026`.

## Testing
- Unit tests: session-key derivation (incl. prefix fallback), continuation rewrite, name mapping round-trip.
- E2E verified against live API: 2-turn custom-function cycle via `/v1/chat/completions` returning `200 OK` with a correct `previous_interaction_id` + `environment` continuation body.

## Delegation guidance for antigravity via OpenAI-compatible endpoints

The agent runs in its own Google sandbox and does **not** see your local filesystem. For best throughput (fewest round-trips):

- **Pass content inline in the prompt** instead of asking the agent to read local files. A translation/edit task should complete in 1-2 turns, not dozens of read/write cycles.
- Keep instructions self-contained (rules, constraints, target format) in the prompt rather than requiring external skill/document lookups.
- Reserve custom function calling for actions that genuinely need the caller's environment (reading/writing caller-side files, calling private APIs).
- Note: free-tier input-token quota (100k/min/model) is shared across concurrent delegations; inline-content delegation also reduces token pressure.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,80 @@
 @AGENTS.md
+
+# Local Operational Notes
+
+## Two-Proxy Setup (2026-07-04)
+
+Two CLIProxyAPI instances run from the same project directory on different ports:
+
+### Port 8317 — Primary (config.yaml)
+- Auth dir: `~/.cli-proxy-api`
+- API keys: `local`, `opencode`
+- `force-model-prefix: true` — models must use prefix (e.g. `zen/deepseek-v4-flash`, `or/model-name`)
+- `routing.strategy: fill-first`
+- `disable-cooling: true`
+- Providers: antigravity (OAuth), codex (OAuth), openai-compatibility (opencode-zen, deepseek, nvidia, openrouter)
+- Used by: main agent, IDE integrations
+
+### Port 8318 — Sub-agent / Light Tasks (rr-config.yaml)
+- Auth dir: `~/.cli-proxy-api-2`
+- API keys: `local`, `opencode`
+- `force-model-prefix: true`
+- `routing.strategy: round-robin`
+- Providers: gemini-api-key (5 Google API keys), openai-compatibility (cerebras, groq)
+- Used by: sub-agents, small tasks
+- **NO deepseek/claude/codex/antigravity** — only gemini + cerebras + groq
+
+### Known Issues
+
+#### gemma-4-26b-a4b-it intermittent 500/503
+- Model is free, high demand → Google API returns 503 "high demand" or 500 "Internal error"
+- Streaming works more reliably than non-streaming during overload
+- All 6 gemini-api-keys get the same error (model-level overload, not key-level)
+- After all keys exhausted → proxy returns "auth_unavailable: no auth available"
+- This is a Google-side capacity issue, not a proxy bug
+
+#### Log collision
+- Both proxies write to `logs/main.log` because both run from same working directory
+- Request logs are separate files (per-request UUID), so those are fine
+- To fix: run second proxy from a different directory or set a custom log path
+
+#### DeepSeek on zen provider
+- `zen/deepseek-v4-flash` works on port 8317 (tested 2026-07-04)
+- Two API keys: `sk-9...BdZY` (direct), `sk-4...8iBU` (via socks5 proxy)
+- If client sees "No provider available" — check which port the client connects to
+- Port 8318 does NOT have deepseek configured
+
+#### gemma JSON response_format not translated (OPEN)
+- **Problem:** `response_format: {"type": "json_object"}` in OpenAI format is NOT translated to `responseMimeType: "application/json"` for Gemini
+- **Effect:** Gemma models wrap JSON in markdown fenced blocks (` ```json ``` `) even when client requests JSON mode
+- **Root cause:** `ConvertOpenAIRequestToGemini` (internal/translator/gemini/openai/chat-completions/gemini_openai_request.go:30) handles temperature, top_p, top_k, reasoning_effort, modalities, messages, tools — but NOT response_format
+- **Workaround exists:** Native Gemini format (`responseMimeType: application/json`) returns raw JSON. Plugin at `/Users/adidos/Projects/cliproxy-plugins/gemma-json-fix/` strips markdown but is incomplete (no model filter, no streaming)
+- **Gemma models missing from models.json:** `internal/registry/models/models.json` has 10 gemini models but NO gemma models (gemma-4-31b-it, gemma-4-26b-a4b-it). Works anyway because GeminiExecutor passes model names directly to Google API URL.
+
+## Request Flow Architecture
+
+```
+Client → Gin Router → AuthMiddleware (API key check)
+  → Handler (OpenAI/Gemini/Claude format detection)
+    → Pipeline.TranslateRequest() — format conversion
+      → Manager.selectAuth() — round-robin/fill-first credential selection
+        → ProviderExecutor.Execute()/ExecuteStream()
+          → Google/OpenAI/Anthropic API
+        ← response
+      ← Pipeline.TranslateResponse() — format conversion back
+    ← client
+```
+
+### Key Code Paths
+- **Gemini URL construction**: `internal/runtime/executor/gemini_executor.go:145`
+  - `url = baseURL/v1beta/models/{model}:{action}`
+  - Non-streaming: `:generateContent`, Streaming: `:streamGenerateContent?alt=sse`
+- **Auth selection**: `sdk/cliproxy/auth/scheduler.go` — cooldown-based, round-robin/fill-first
+- **Model registration**: `internal/registry/model_registry.go` — models registered from config at startup
+- **OpenAI-compat model mapping**: config `models[].name` → `models[].alias`, prefix stripped before routing
+
+### Error Response Patterns from CLIProxyAPI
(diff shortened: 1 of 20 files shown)


xitex:
Fixed in c7990c37: `agent_config.type = "antigravity"` is now set unconditionally in both the chat-completions and Responses converters, independent of token-limit presence. Client-supplied `max_tokens`/`max_output_tokens` are still dropped (antigravity truncates runs when `max_total_tokens` is sent). Added a regression test for the no-limit case.

xitex:
Addressed all review findings in 581fdf54:

1. **P1 partial coordinates** — capture now merges with cached state; a stream event carrying only one coordinate no longer blanks the other.
2. **P1 identical-prefix isolation** — the continuation key now prefers the conductor's derived session identity (`session.Enrich` → `DerivedSessionID`) before falling back to the conversation-prefix hash, so identical prefixes from different callers never share an upstream interaction.
3. **P1 auth-selection model** — switched to `gemini-3-flash`, which exists in the static Gemini catalog (unlike `gemini-3.7-flash`), so credential selection works under `--local-model`.
4. **P2 Responses rename gating** — `external_` renaming is now gated by target model in the Responses converter too; non-antigravity models keep client names.
5. **P2 slash IDs** — route is now `/v1/models/*model` and the handler strips the leading slash, so prefixed IDs like `teamA/gemini-3-pro-preview` resolve.
6. **P2 agent_config preservation** — both converters copy a client-supplied `agent_config` verbatim for non-antigravity Interactions models.

xitex:
Addressed the two new findings in ddbd585b:

1. **P1 nameless function_result** — the continuation rewrite now collects call_id → name from the replayed assistant function_call steps and populates missing names on retained function_result steps before dropping history (Chat Completions tool loops send role:"tool" with only tool_call_id + content).
2. **P2 client_version on detail route** — removed the shortcut; /v1/models/{id} always performs the normal single-record lookup so unknown IDs return 404. The Codex list payload remains available on the collection endpoint (/v1/models?client_version=...).

xitex:
Addressed both findings in c6a12d4b:

1. **P1 tool names for Responses continuations** — `AntigravityInteractionsState` now retains a call_id → function name map captured from upstream function_call steps (merged across events). The continuation rewrite falls back to this map when the incoming tool-result turn carries no name, so Responses-style loops (call_id + output only) still satisfy the function_result schema.
2. **P1 string tool outputs** — the Responses converter now normalizes antigravity-bound string outputs exactly like the Chat Completions path: valid JSON is sent raw, other text is wrapped in `{"result": ...}`.

xitex:
Addressed both findings in 5389c92a:

1. **P1 streamed step names** — `goframeCacheAntigravityInteractionsState` now captures `step.start` frames (`step.id` + `step.name`) even when the frame carries no interaction coordinates, and merges them into the cached name map alongside the top-level `steps` scan.
2. **P2 replayed call aliases** — `responsesFunctionCallToInteractions` takes the model flag; colliding names are aliased only for antigravity targets, so a non-antigravity request replaying a legitimate `read_file` keeps its declared name and matches its declaration.

xitex:
Addressed both findings in 026b8f74:

1. **P1 replayed Chat Completions names** — the converter now aliases replayed assistant `tool_calls` names and `role:tool` result names with the external_ prefix for antigravity targets, so the continuation rewrite no longer prioritizes an unaliased replayed name that mismatches the upstream call.
2. **P1 nested interaction.steps** — state capture scans both top-level `steps` and the nested `interaction.steps` envelope used by non-stream responses when building the call_id → name map.

xitex:
Addressed both findings in ceeb6efe:

1. **P1 all explicit session signals** — `antigravityExecSessionKey` now delegates to `cliproxyauth.ExtractSessionID` (the same extraction credential selection uses), so X-Claude-Code-Session-Id, Session-Id/Session_id, X-Session-ID, X-Session-Affinity, X-Client-Request-Id, body session_id/sessionId, conversation id, and prompt_cache_key all key the continuation state identically instead of the narrower hand-rolled list.
2. **P1 index by prior response ID** — captured state is additionally stored under a `response-id:<id>` alias entry. A standard Responses tool loop sends only previous_response_id + function_call_output (session.Enrich cannot re-derive the original identity from that input), so the continuation rewrite resolves the cached coordinates through that alias before giving up.

xitex:
Addressed the three findings from ceeb6efe (37f30d85 + ea037568):

1. **P2 generation controls** — `agent_config` copy and generation-control copy are now independent branches: a non-antigravity request can supply both.
2. **P2 structured results gating** — the Chat Completions converter now wraps string tool outputs only for antigravity targets; other Interactions models keep their original plain-string output.
3. **P2 Home lookup in execution** — `GetAntigravityInteractionsState` takes the caller's context so a continuation lookup is canceled with the request; the best-effort capture-side read uses a detached context so a slow store can never stall the stream read loop (this also fixes the pooled-connections test regression under -race).

xitex:
Addressed the P1 finding in 74a5545d: every continuation key — explicit session, derived identity, conversation-prefix fallback, and the response-id alias — is now namespaced by `CallerScopeMetadataKey`, so two downstream API keys sending the same explicit session value (e.g. "Session-Id: default") can no longer overwrite each other's captured interaction/environment state or resume a foreign upstream conversation.

xitex:
Addressed in e2747242: the response-ID alias is now written and resolved independently of the session key. A stateless /v1/responses follow-up (previous_response_id + function_call_output only) derives no session key, so gating the lookup on sessionKey != "" made the alias unreachable; the alias entry itself is caller-scoped at capture time, preserving isolation between callers.

xitex:
Addressed in ede61a85: dropped the hard-coded auth-selection model entirely. Native agent requests now force only the gemini-interactions provider and leave `AuthSelectionModel` empty, so the conductor's model-aware filtering is skipped and any available credential registered for that provider is eligible — including under `--local-model`, regardless of which ids the static catalog happens to contain. Test updated to register the agent id and verify the unfiltered pick.

xitex:
Addressed both findings in e17b34a8:

1. **P1 auth-selection model** — restored a catalog-backed id, now `gemini-3.6-flash`: it exists in the static Gemini catalog and is the documented antigravity default (the agent runs Gemini 3.6 Flash unless agent_config overrides it). Leaving the field empty made `authSelectionModelFromOptions` fall back to the agent id, which no catalog advertises — so the unfiltered-pick approach was wrong.
2. **P1 payload logging** — interactions request/response/continuation debug logs now emit metadata only (target, byte size, step/result counts). Full bodies were bypassing the configured request-recording path and persisting prompts and tool results in plain logs.

xitex:
Addressed in 6add3027: moved the interaction-state helpers (cache-key derivation, state capture, continuation rewrite) into `internal/runtime/executor/helps/` per the repository package layout, exporting `CacheAntigravityInteractionsState`, `ApplyAntigravityInteractionsContinuation`, and `AntigravityExecSessionKey`. Tests moved alongside. (`originalRequestRawJSON` stays in the executor package — it has other callers there.)

xitex:
Superseded by minimal splits (all green, guard passing): #5459 (GET /v1/models/:model) and #5460 (interactions session cache + state). Remaining translator-only piece (external_ tool-name round-trip) can't go through PRs due to translator-path-guard — tracked in #5190 item 2 for the maintenance team.

PR #5038, opened 2026-08-17
Budget-awareness: external supplier + remaining/recovery

### LINKED ISSUE #5034: Budget-awareness: per-auth remaining + optional recovery + external supplier
Filed from sparkfn/proxy-claude-v2#1765 item 7 / tracking sparkfn/proxy-claude-v2#1822.

## Ask

A generic budget-awareness mechanism so a host (or built-in poller) can mark an auth exhausted until recovery, without the conductor scraping provider consoles.

1. **Per-auth budget state** — remaining fraction or absolute amount, optional recovery time.
2. **Rule** — exhausted → ineligible until recovery (or until the next supplier refresh if there is no recovery time).
3. **External supplier interface** — upstream stays agnostic about how numbers are gathered. Hosts push or poll through that seam.
4. **Built-in pollers (adoption)** — DeepSeek `GET /user/balance` (https://api-docs.deepseek.com/api/get-user-balance) and windowed quota APIs (e.g. Zhipu) as first consumers.

We already park on collected window truth in the host (`acceptAuth`). We will **not** fork this repo; a host-side ParkSource should become a supplier later.

`Result.CredentialScope` is the wrong tool for prepaid `$` and for non-Anthropic window quotas — do not require hosts to set that flag on every 429.

jaaacki:
Fixes #5034.

Sketch + tests for a host-owned `BudgetSupplier` seam:

- Per-auth remaining (fraction or prepaid amount) + optional `RecoverAt`.
- Exhausted → ineligible until recovery, or until the next supplier refresh when there is no clock.
- Does **not** require hosts to set `Result.CredentialScope` (wrong tool for prepaid \$ and non-Anthropic windows).

Built-in DeepSeek / Zhipu pollers can adopt this later; this PR only lands the interface and `ApplyBudgetObservations`.

### DIFF
--- a/sdk/cliproxy/auth/budget_supplier.go
+++ b/sdk/cliproxy/auth/budget_supplier.go
@@ -1,0 +1,139 @@
+package auth
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// BudgetKind distinguishes a rolling time window from a prepaid remaining
+// balance. Hosts must not stuff a dollar amount into a window quota, and must
+// not invent a resets_at for prepaid balance.
+const (
+	BudgetKindWindow  = "window"
+	BudgetKindBalance = "balance"
+)
+
+// BudgetObservation is one auth's remaining budget as reported by a supplier.
+// Remaining is a fraction in [0,1] when known; Amount/Currency are optional
+// absolute prepaid fields. RecoverAt is zero when there is no clock (prepaid
+// $0 parks until the next supplier refresh).
+type BudgetObservation struct {
+	AuthID    string
+	Kind      string
+	Remaining *float64
+	Amount    string
+	Currency  string
+	Available *bool
+	RecoverAt time.Time
+}
+
+// BudgetSupplier is the host-owned seam for pushing collected usage into the
+// conductor. CLIProxyAPI stays agnostic about how numbers are gathered
+// (DeepSeek /user/balance, Zhipu quota API, a host ParkSource, etc.).
+//
+// A nil supplier is a no-op: selection stays on today's cooldown ladder.
+type BudgetSupplier interface {
+	Snapshot(ctx context.Context) ([]BudgetObservation, error)
+}
+
+// BudgetApplyConfig controls how observations become ineligibility.
+type BudgetApplyConfig struct {
+	// Threshold is the remaining fraction at or below which a window is
+	// exhausted. Default 0 (remaining <= 0).
+	Threshold float64
+	// Recheck is used when RecoverAt is zero (prepaid). Default 5m.
+	Recheck time.Duration
+	Now     func() time.Time
+}
+
+// ApplyBudgetObservations marks auths exhausted from supplier observations.
+// It never sets Result.CredentialScope — that flag is Anthropic-window
+// scoped and is the wrong tool for prepaid $ and non-Anthropic windows.
+//
+// Fail-open: missing remaining, unparseable amount, and absent observations
+// leave the auth eligible.
+func ApplyBudgetObservations(auths []*Auth, obs []BudgetObservation, cfg BudgetApplyConfig) {
+	if len(auths) == 0 || len(obs) == 0 {
+		return
+	}
+	now := time.Now()
+	if cfg.Now != nil {
+		now = cfg.Now()
+	}
+	recheck := cfg.Recheck
+	if recheck <= 0 {
+		recheck = 5 * time.Minute
+	}
+	byID := make(map[string]*Auth, len(auths))
+	for _, a := range auths {
+		if a != nil && a.ID != "" {
+			byID[a.ID] = a
+		}
+	}
+	for _, o := range obs {
+		a := byID[o.AuthID]
+		if a == nil {
+			continue
+		}
+		until, exhausted := observationExhausted(o, cfg.Threshold, recheck, now)
+		if !exhausted {
+			continue
+		}
+		a.Unavailable = true
+		a.Quota.Exceeded = true
+		a.Quota.Reason = "budget:" + normalizeBudgetKind(o.Kind)
+		a.Quota.NextRecoverAt = until
+	}
+}
+
+func observationExhausted(o BudgetObservation, threshold float64, recheck time.Duration, now time.Time) (time.Time, bool) {
+	kind := normalizeBudgetKind(o.Kind)
+	switch kind {
+	case BudgetKindBalance:
+		if o.Available != nil && !*o.Available {
+			return recoverOrRecheck(o.RecoverAt, recheck, now), true
+		}
+		if amt, ok := parseBudgetAmount(o.Amount); ok && amt <= 0 {
+			return recoverOrRecheck(o.RecoverAt, recheck, now), true
+		}
+		return time.Time{}, false
+	default:
+		if o.Remaining == nil {
+			return time.Time{}, false
+		}
+		if *o.Remaining > threshold {
+			return time.Time{}, false
+		}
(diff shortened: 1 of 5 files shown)


sdk/cliproxy/auth/conductor_lifecycle.go
```diff
@@ -125,7 +125,7 @@ func (m *Manager) Update(ctx context.Context, auth *Auth) (*Auth, error) {
 		if len(auth.ModelStates) == 0 && len(existing.ModelStates) > 0 {
 			auth.ModelStates = existing.ModelStates
 		}
-		if existing.Quota.Exceeded && existing.Quota.Reason == "credential_quota" && existing.Quota.NextRecoverAt.After(time.Now()) {
+		if existing.Quota.Exceeded && isCredentialWideQuotaReason(existing.Quota.Reason) && existing.Quota.NextRecoverAt.After(time.Now()) {
```
jaaacki: **[bug]** The advertised host hook cannot be applied to conductor state. `Manager.List` / `GetByID` return clones; `ApplyBudgetSupplier` / `ApplyBudgetObservations` only mutate the slice they are given; nothing in CLIP calls them. This PR then expands `Update` so an existing `budget:*` park with a future `NextRecoverAt` is copied back onto the incoming auth (same sticky path as Anthropic `credential_quota`). A host that does the only public sequence — `List()`, apply, `Update()` — can establish a first park, but a later healthy snapshot cannot unpark until `RecoverAt` (hours for a window; 5m default recheck for prepaid). Mutating live `m.auths` pointers is not possible through the public API and would race `MarkResult` (no lock). Parks also never set `NextRetryAfter`, so `authCooldownStateRecord` will not persist them.

**Suggestion:** Own this inside `Manager` (e.g. `SetBudgetSupplier` + a locked apply on live auths that sets `NextRetryAfter`, calls `updateAggregatedAvailability`, and persists cooldown). Keep `Update` sticky for reload, and unpark only on that locked path — do not tell hosts to `List` + apply + `Update`.

sdk/cliproxy/auth/budget_supplier.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		until, exhausted := observationExhausted(o, cfg.Threshold, recheck, now)
+		if !exhausted {
+			if isBudgetReason(a.Quota.Reason) {
+				a.Unavailable = false
+				a.Quota.Exceeded = false
+				a.Quota.Reason = ""
+				a.Quota.NextRecoverAt = time.Time{}
+			}
+			continue
```
jaaacki: **[bug]** Any observation that is not classified exhausted clears an existing `budget:*` park. `observationExhausted` returns false for missing window `Remaining`, unparseable/empty balance `Amount`, and a balance row with no `Available` / `Amount` / `Remaining`. The apply loop then treats that as recovery. A collector glitch or partial row therefore unparks a cred that was correctly parked. Absent IDs are actually a no-op (previous park stays), while a present incomplete row actively unparks. An empty `Snapshot` also returns without touching anyone, so a supplier that reports only exhausted auths, or "all healthy" as an empty list, can never unpark except by waiting out `RecoverAt`.

**Suggestion:** Unpark only on a positive healthy reading (window `Remaining > threshold`, or balance `Amount > 0` / `Available==true`). Incomplete rows and omitted IDs must no-op. If empty snapshot means "all healthy", say so and clear parks explicitly; otherwise document that every auth must be emitted every tick.

sdk/cliproxy/auth/budget_supplier.go
```diff
@@ -0,0 +1,186 @@
...
+	return now.Add(recheck)
+}
+
+func normalizeBudgetKind(kind string) string {
+	switch strings.ToLower(strings.TrimSpace(kind)) {
+	case BudgetKindBalance:
+		return BudgetKindBalance
+	default:
+		return BudgetKindWindow
+	}
```
jaaacki: **[bug]** `normalizeBudgetKind` maps every non-`balance` string, including `""` and `"both"`, to `window`. Prepaid exhaustion is then ignored: `Amount="0"` / `Available=false` with empty kind takes the window branch, sees `Remaining==nil`, and fails open (and, via the incomplete-row bug, would unpark a previous `budget:balance` park). Hosts already have `quota_kind=window|balance|both`; a supplier that forgets `Kind` or forwards `both` silently drops the prepaid signal.

**Suggestion:** Require a known kind (or treat unknown as a no-op). If CLIP should accept `both`, decide exhaustion from the fields that are present rather than coercing to window.

sdk/cliproxy/auth/budget_supplier.go
```diff
@@ -0,0 +1,186 @@
...
+		until, exhausted := observationExhausted(o, cfg.Threshold, recheck, now)
+		if !exhausted {
+			if isBudgetReason(a.Quota.Reason) {
+				a.Unavailable = false
+				a.Quota.Exceeded = false
+				a.Quota.Reason = ""
+				a.Quota.NextRecoverAt = time.Time{}
+			}
+			continue
+		}
```
jaaacki: **[suggestion]** Unpark forces `Unavailable=false` and does not call `updateAggregatedAvailability`. That can mark an auth eligible at the auth level while every `ModelStates` entry is still cooling, and it ignores a still-future `NextRetryAfter`. Park does not set `NextRetryAfter` either, so cooldown persistence and `ResetQuota` / restore stay inconsistent with `credential_quota`.

**Suggestion:** After apply, run the same aggregation the conductor already uses; set `NextRetryAfter` to the park deadline when exhausting.

sdk/cliproxy/auth/budget_supplier.go
```diff
@@ -0,0 +1,186 @@
...
+		if amt, ok := parseBudgetAmount(o.Amount); ok && amt <= 0 {
+			return recoverOrRecheck(o.RecoverAt, recheck, now), true
+		}
+		if o.Remaining != nil && *o.Remaining <= threshold {
+			return recoverOrRecheck(o.RecoverAt, recheck, now), true
+		}
+		return time.Time{}, false
+	default:
+		if o.Remaining == nil {
+			return time.Time{}, false
```
jaaacki: **[suggestion]** Balance exhaustion treats `Remaining <= threshold` as prepaid-empty. Host `decideBalancePark` only uses `BalanceAvailable` / `BalanceAmount`. A leftover `Remaining=0` on a funded account (`Amount="12"`, `Available=true`) would park. Threshold default 0 also means CLIP parks on remaining-fraction semantics while the host parks on utilization ≥ 1.0 — easy to invert when someone later wraps `ParkSource`.

**Suggestion:** For `balance`, ignore `Remaining` unless the supplier is documented to send a remaining fraction and nothing else. Do not reuse the window threshold for dollars.

sdk/cliproxy/auth/budget_supplier.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	obs, err := supplier.Snapshot(ctx)
+	if err != nil {
+		return err
+	}
+	ApplyBudgetObservations(auths, obs, cfg)
+	return nil
+}
+
+// ApplyBudgetObservations marks auths exhausted from supplier observations.
```
jaaacki: **[suggestion]** Several new comments narrate design history and what the code does not do (`Result.CredentialScope`, DeepSeek/Zhipu/ParkSource, "honest host hook") instead of a short why. Same pattern at the `BudgetKind` / `BudgetSupplier` doc comments.

**Suggestion:** Keep a one-line contract (nil supplier no-op; snapshot error fail-open; `budget:*` vs `credential_quota`). Drop the architecture sketch.

PR #5059, opened 2026-08-18
feat(compat): add opt-in fallback compaction for providers without compact endpoints

315zuicai:
## Summary

Codex clients send `/v1/responses/compact` and `compaction_trigger` items during long sessions. Official OpenAI/xAI/Codex endpoints implement that contract, but many OpenAI-compatible and third-party Responses providers only expose `/chat/completions` or `/responses`.

This adds an explicit `fallback-compaction` flag so those providers can still complete Codex remote compaction.

When the flag is enabled:

1. CPA recognizes `/v1/responses/compact` or a `compaction_trigger` item.
2. The conversation is summarized through `/chat/completions`.
3. The summary is sealed into a Codex-compatible `compaction` output item.
4. Later turns expand that capsule back into a developer message before the next request.

Default behavior is unchanged. Native `/responses/compact` passthrough remains the default for OpenAI-compatible providers and official Codex endpoints.

## Configuration

```yaml
openai-compatibility:
  - name: deepseek
    base-url: https://api.deepseek.com
    fallback-compaction: true

codex-api-key:
  - api-key: sk-...
    base-url: https://api.deepseek.com/v1
    fallback-compaction: true
```

DeepSeek is the validation scenario (`deepseek-v4-pro` via `openai-compatibility`, `deepseek-v4-flash` via third-party `codex-api-key`), but the implementation is not DeepSeek-specific.

## Notes

- Capsules are AES-GCM sealed from the provider API key. Later requests can recover the summary after credential rotation on the same provider/base URL.
- Tampered or foreign capsules are rejected.
- Streaming compact requests remain unsupported, matching the existing compact handler.
- Tests cover passthrough-by-default, fallback compact, multi-key capsule reuse, tamper rejection, and Codex-compatible endpoints.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -371,6 +371,7 @@ nonstream-keepalive-interval: 0
 #           - "context_window_exceeded"
 #         action: "stop-and-cooldown"
 #     base-url: "https://www.example.com" # use the custom codex API endpoint
+#     fallback-compaction: false # optional: synthesize compact responses through /chat/completions for third-party Responses endpoints without /responses/compact
 #     alpha-search: false # optional: allow this key to serve /v1/alpha/search via base-url + /alpha/search
 #     headers:
 #       X-Custom-Header: "custom-value"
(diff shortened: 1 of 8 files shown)


andyWang1688:
Same issue here — Codex 0.147 remote compaction v2 (`POST /v1/responses` with `reasoning.effort=max`) against DeepSeek official via `openai-compatibility` fails with either:

- `Error running remote compact task: Fatal error: remote compaction v2 expected exactly one compaction output item, got 0 from 2 output items`, or
- HTTP 400 `The reasoning_content in the thinking mode must be passed back to the API.`

Environment: CPA 7.2.140 (Homebrew), `deepseek-v4-pro` with `thinking.levels: [low, high, max]`, Codex `model_reasoning_effort = "max"`.

Debug log shows thinking level=max is applied to the upstream call on the compaction request, and the request hits plain `/v1/responses` (not `/v1/responses/compact`), so the existing compact special-handling never triggers. This PR is exactly the fix for this scenario — any chance it can get reviewed/merged soon? Happy to help test.

PR #5060, opened 2026-08-18
feat: add Command Code Go provider support

### LINKED ISSUE #3955: [Feature] Add Command Code (commandcode.ai) as a provider — enables the $1/mo Go plan
## Summary

Request to add [Command Code](https://commandcode.ai) (a coding agent + model gateway by Langbase) as a provider. The interesting part is the **Go plan**: **$1/month** with **$10 of included credits** and deal multipliers that push effective usage to ~$40 on selected OSS models ([pricing](https://commandcode.ai/pricing), [pricing limits](https://commandcode.ai/docs/resources/pricing-limits)). Roster includes DeepSeek V4 Pro, Kimi K2.7 Code, GLM-5.2, MiniMax M3, Qwen 3.7 Max, MiMo V2.5 Pro, Nemotron 3 Ultra, etc. (21 open-weight models, several with 1M context).

The Go plan is genuinely the best $1 in the space right now, and CLIProxyAPI is the obvious home for it — but there's a wrinkle that's worth flagging up front.

## The wrinkle: Go plan works via `/alpha/generate`, not `/provider/v1/*`

The gateway exposes three generation endpoints:

| Endpoint | Shape | Plan |
|----------|-------|------|
| `POST /provider/v1/messages` | Anthropic Messages | **Pro+ only** |
| `POST /provider/v1/chat/completions` | OpenAI Chat Completions | **Pro+ only** |
| `POST /alpha/generate` | CLI's own custom envelope | **Any plan, incl. Go** |

On the Go plan, both `/provider/v1/*` paths return:
```json
{ "error": { "message": "Your Go plan doesn't include API access. Upgrade to Pro or higher…", "code": "upgrade_required" } }
```

`/alpha/generate` is the endpoint the `cmd` CLI itself uses on every turn, so it is **not** plan-gated. The catch: it speaks a **custom envelope**, not OpenAI or Anthropic shape. So serving Go-plan users means a translator, exactly the kind of thing CLIProxyAPI does well.

Auth is a plain `user_...` API key (same key as the CLI, from https://commandcode.ai/settings/billing) — **no OAuth needed**, unlike Codex/Claude/Gemini.

## The protocol is already reverse-engineered and battle-tested

There's a mature reference implementation with a full wire-protocol spec:
- **Reference:** [`safzanpirani/pi-commandcode-provider`](https://github.com/safzanpirani/pi-commandcode-provider) — working OpenAI→envelope + NDJSON→SSE translator.
- **Wire protocol doc:** [`docs/wire-protocol.md`](https://github.com/safzanpirani/pi-commandcode-provider/blob/main/docs/wire-protocol.md) — strict `config` envelope, Vercel-AI-SDK `ModelMessage[]` schema, tool format, every NDJSON event type, usage shape.

### Request
```
POST https://api.commandcode.ai/alpha/generate
Authorization: Bearer user_…
x-cli-environment: production
x-command-code-version: 0.40.0   # must track the installed CLI; stale → rejected
x-session-id: <uuid>
```
```jsonc
{
  "config": { /* schema-strict; all fields required, neutral defaults OK */ },
  "memory": "", "taste": null, "skills": null, "permissionMode": "standard",
  "params": {
    "model": "deepseek/deepseek-v4-pro",
    "system": "…",
    "messages": [ /* Vercel AI SDK ModelMessage[] — NOT OpenAI, NOT Anthropic */ ],
    "tools": [ /* name + input_schema (JSON Schema) */ ],
    "max_tokens": 32000, "stream": true
  }
}
```

### Response = **newline-delimited JSON, NOT SSE**
```
{"type":"text-delta","id":"txt-0","text":"…"}
{"type":"tool-input-delta","id":"call_…","delta":"<json chunk>"}
{"type":"finish-step","finishReason":"stop|length|tool-calls","usage":{...}}
```

### Known sharp edges (verified from open issues in other routers)
- **`stream:false` is rejected** (`400/ECONNRESET`) — must force `stream:true` and buffer server-side. [decolua/9router#1840](https://github.com/decolua/9router/issues/1840), fixed in [#1843](https://github.com/decolua/9router/pull/1843).
- NDJSON lines on large tool outputs overflow `bufio.Scanner`'s default limit — needs an unbounded reader.
- The `config` envelope is schema-strict; the gateway's 400 usefully lists the exact missing field paths.

## Demand is real (this is not hypothetical)

A code/issue search of this repo for `commandcode` returns **zero** matches today. But the ecosystem is actively building it independently:

| Project | Stars | Status |
|---------|-------|--------|
| [`decolua/9router`](https://github.com/decolua/9router) | 18k | Active router; fixing Command Code stream bugs this week |
| [`can1357/oh-my-pi`](https://github.com/can1357/oh-my-pi) | 13.9k | PR [#1360](https://github.com/can1357/oh-my-pi/pull/1360) "add Command Code provider" (this week) |
| [`safzanpirani/pi-commandcode-provider`](https://github.com/safzanpirani/pi-commandcode-provider) | — | Reference protocol impl, resynced to CLI 0.39.1 four days ago |
| [`HaeMeto/Command-Code-AI-Proxy`](https://github.com/HaeMeto/Command-Code-AI-Proxy) | 10 | Go proxy on `/alpha/generate` |
| [`dev2k6/command-code-proxy-server`](https://github.com/dev2k6/command-code-proxy-server) | 8 | Go proxy, but Pro-only (`/provider/v1/*`) so it can't serve Go plan |

Same feature request also open elsewhere: [anomalyco/opencode#26338](https://github.com/anomalyco/opencode/issues/26338). The knowledge and the demand both exist; what's missing is a canonical home, which CLIProxyAPI is the natural fit for.

## Stability note (being candid)

`/alpha/generate` is an undocumented CLI-internal endpoint, and the `command-code` npm package is at `0.40.0` and moves frequently — the protocol drifts and the `x-command-code-version` header must track it. That's the **same risk class** as this project's existing OAuth/CLI-wrapping providers (Claude Code, Codex, Gemini CLI), and centralizing it here with CI is exactly what would keep consumers current instead of each fork re-deriving on every CLI bump.

## Proposed contribution

I've gathered a full evidence bundle (protocol spec, reference impl source, model roster, credit multipliers, stability assessment) and would like to open a PR. Two routes, and I'd like your steer on which you prefer:

1. **Example provider under `examples/`** — a Go Command Code provider built on the v7 SDK pattern already shown in [`examples/custom-provider/main.go`](https://github.com/router-for-me/CLIProxyAPI/blob/main/examples/custom-provider/main.go) (`sdktr.Register(openai.chat → commandcode, reqXform, respXform)` + an `Executor` hitting `/alpha/generate` + model registration in `OnAfterStart`). This path is open to contributors.

2. **Built-in provider** — I noticed [`pr-path-guard.yml`](https://github.com/router-for-me/CLIProxyAPI/blob/main/.github/workflows/pr-path-guard.yml) blocks `internal/translator/**` changes in PRs, so a built-in translator would need your team to land it. If you'd prefer it built-in, I'm happy to hand over a ready diff + tests and let you commit, or however you prefer to work.

My default plan is route **1** (example provider) unless you'd rather it be built-in.

---

**Question for maintainers:** preferred route (1 example provider vs 2 built-in), and any provider naming convention you'd like me to follow (e.g. `commandcode`, `command-code`, `cmd`)?

I'll open the PR against `dev` per the retarget workflow once you confirm direction.

cc maintainers

k176060444-lgtm:
### Summary

This PR adds **Command Code as a native provider** in the normal `cmd/server` build, giving the **Command Code Go subscription** a first-class, OpenAI-compatible API surface that Hermes and other agents can consume.

**The core problem this PR solves**: a Command Code Go subscription has **no standard Provider inference API permission**. The bridge therefore is a **protocol conversion**: it takes a Command Code credential and translates Claude/OpenAI-style requests through the authenticated `/alpha/generate` endpoint into an OpenAI-compatible API (`/v1/chat/completions`).

```
Command Code key  ->  Authorization: Bearer  ->  CLIProxyAPI  ->  /alpha/generate
```

This PR **never enables the standard Provider inference API for Go** — `/alpha/generate` remains the inference path. The value is the bridge, not new Provider entitlements.

Closes #3955

### Context

Implements the Command Code integration discussed in #3955 and #4007.

### What this adds

- Native `commandcode` executor and provider wiring in `cmd/server`
- OpenAI non-stream and SSE stream translation
- Tool-call / tool-result / parallel tool support
- **First-class API-key authentication** (see Authentication)
- Dynamic model catalog synchronized from the official provider catalog
- Model refresh integrated with the existing registry lifecycle
- Bounded NDJSON line and error-body handling
- Full `commandcode-api-key` Management API CRUD under `/v0/management/commandcode-api-key`

### Authentication — API key is the first-class credential

The real credential is the **Command Code API key**. `cmdc login` is just one way to obtain/save a key (it writes `~/.commandcode/auth.json`); it is a convenience/migration path, **not** a runtime dependency. CLIProxyAPI works with only a key — no node, npm, or `cmdc` binary required.

Explicit credential sources (in precedence order):

1. `config.yaml` → `commandcode-api-key` (Level 1)
2. CLIProxyAPI-managed `auth-dir` credential files, e.g. `auths/commandcode-*.json` (Level 2)
3. Environment variable `COMMANDCODE_API_KEY` / `COMMAND_CODE_API_KEY` (Level 3)
4. `~/.commandcode/auth.json` auto-import (Level 4) — **convenience / migration fallback only**

Semantics:

- **Different keys** across the explicit sources (config / managed auth-dir / env) coexist as **separate accounts** (each with independent prefix/priority/weight routing).
- **The same key across sources** is **deduplicated** to the highest-precedence source (config > managed auth-dir > env), so a duplicated credential never spawns duplicate clients/clients.
- `~/.commandcode/auth.json` is imported **only when no explicit credential is present**; it is never imported on top of an explicit one.
- All keys ultimately flow `Authorization: Bearer <key> -> /alpha/generate` in the same execution path.
- `402` from `/alpha/generate` is treated as a **per-model** plan/entitlement rejection (model-level cooldown only) — it does not globally disable the credential. `401` = invalid credential and `429` semantics are unchanged.

### Dynamic models

- **Primary source**: the official `GET https://api.commandcode.ai/provider/v1/models` catalog (anonymous, follows the configured CLIProxyAPI `proxy-url`, refreshed at startup and every 3h)
- **Cold-start / offline fallback only**: the locally installed `cmdc` CLI bundled catalog (`dist/bundled/command-code-knowledge/reference/models.md`), then last-known-good, then builtin bootstrap
- **No regression on transient remote failure**: once a working catalog is loaded, a remote failure preserves the current model set — it is never downgraded by an older local snapshot
- New and removed models are applied on refresh so existing auths re-register (no stale builtin-only pool)
- **Catalog depends on the external `/provider/v1/models` endpoint; catalog visibility does not imply plan/region entitlement** — the Command Code upstream (`/alpha/generate`) remains the final authority

### Streaming / tool fix

Command Code emits incremental tool-input events plus terminal full-input events. The translator prevents duplicate `function.arguments` emission while retaining a full-call fallback when no deltas were seen.

### Testing / validation

- Command Code executor tests (non-stream / stream / tool / NDJSON / error contract)
- Credential synthesizer/watcher tests covering the full precedence + cross-source dedup semantics
- `commandcode-api-key` Management API CRUD handler tests
- Dynamic catalog registry tests (remote success / failure fallback / LKG preservation / dynamic removal / casing / proxy transport)
- 402 per-model cooldown + 429 regression tests
- `go vet`, `go test`, `go build ./cmd/server`

Live validation on the real SYSTEM scheduled-task deployment (with the production binary rebuilt from this PR's HEAD):

- `/v1/models` serves the official remote catalog (Command Code Go models incl. `meta/muse-spark-*` and `deepseek/*`), not the builtin-only pool
- **Direct managed/config API key** (auth.json suppressed) → non-stream **HTTP 200**, SSE stream **HTTP 200**
- **auth.json-only fallback** (no explicit credential) → **HTTP 200**
- **Cross-source dedup** verified live (same key in config + auth-dir + env registers one client); distinct keys register as separate accounts
- Management API CRUD (`/v0/management/commandcode-api-key` GET/PUT/PATCH/DELETE) round-trips cleanly
- Hermes agent served through `127.0.0.1:8317` (cliproxy-cmdc provider)


first-storm:
I don't think hard code model list in the code is a good idea. ik the commandcode client does that but it doesn't mean we also need to do that. You can retrieve it from https://api.commandcode.ai/provider/v1/models. Also, you can actually get api key from commandcode webpage so a better way to get credentials is just ask the user to paste the key instead of import it from local cmdcode cli.

k176060444-lgtm:
Thanks for the review — agreed, and the hard-coded list is no longer the source of truth.

The model catalog now fetches from `https://api.commandcode.ai/provider/v1/models` (anonymous, proxy-aware via the configured CLIProxyAPI `proxy-url`, refreshed at startup and every 3h). The locally installed `cmdc` CLI catalog is kept only as a cold-start / offline fallback, and a transient remote failure never regresses the already-loaded model set (last-known-good is preserved). New/removed models are applied on refresh so existing auths re-register instead of staying on the builtin bootstrap list.

On credentials: the core purpose of this PR is supporting the **Go subscription, which has no Provider API entry point** — so the default path remains the `cmdc login` credential flowing to `/alpha/generate`. That is the only way a Go user can authenticate. An explicit `commandcode-api-key` config is an optional path that takes precedence when present and enters the same Bearer execution path; it is never required for the Go path. (We have not yet verified from our environment that a web-generated Provider API key is accepted by `/alpha/generate`, so we intentionally make no stronger claim.)

first-storm:
> Thanks for the review — agreed, and the hard-coded list is no longer the source of truth.
> 
> The model catalog now fetches from `https://api.commandcode.ai/provider/v1/models` (anonymous, proxy-aware via the configured CLIProxyAPI `proxy-url`, refreshed at startup and every 3h). The locally installed `cmdc` CLI catalog is kept only as a cold-start / offline fallback, and a transient remote failure never regresses the already-loaded model set (last-known-good is preserved). New/removed models are applied on refresh so existing auths re-register instead of staying on the builtin bootstrap list.
> 
> On credentials: the core purpose of this PR is supporting the **Go subscription, which has no Provider API entry point** — so the default path remains the `cmdc login` credential flowing to `/alpha/generate`. That is the only way a Go user can authenticate. An explicit `commandcode-api-key` config is an optional path that takes precedence when present and enters the same Bearer execution path; it is never required for the Go path. (We have not yet verified from our environment that a web-generated Provider API key is accepted by `/alpha/generate`, so we intentionally make no stronger claim.)

FYI, you can just generate an API key on the command code website. That key only accesses the Go endpoint, which is exactly what this pr is for. And one of the ways cmd CLI logs into Go subs is literally pasting that same key from the website, then saving it to disk as is. So the whole cmd login flow is a giant circle back to the exact key you could've copied in the first place. You don't need a whole cmd code binary just for getting the key man.

k176060444-lgtm:
Thanks for both pieces of feedback — both are now implemented in this PR.

**1. Model catalog → official remote source.** Hard-coding is gone. The model catalog is now fetched from `https://api.commandcode.ai/provider/v1/models` (anonymous, proxy-aware via the configured CLIProxyAPI `proxy-url`, refreshed at startup and every 3h). The locally installed `cmdc` CLI catalog is kept **only** as a cold-start / offline fallback; a transient remote failure never regresses the already-loaded model set (last-known-good preserved), and dynamic add/remove on refresh keeps auths re-registered.

**2. API key is the first-class credential — no `cmdc` binary needed.** You're right that generating/copying the key directly is simpler than the whole `cmdc login` circle. This PR now makes the **API key** the primary credential path: CLIProxyAPI can authenticate purely from a key (via `config.yaml` `commandcode-api-key`, or a CLIProxyAPI-managed `auth-dir` file, or `COMMANDCODE_API_KEY`/`COMMAND_CODE_API_KEY` env) with **no node/npm/cmdc installed**. Different keys coexist as separate accounts; the same key across sources is deduplicated by precedence.

`~/.commandcode/auth.json` is retained **only** as a convenience / migration / auto-import fallback (imported solely when no explicit credential is configured), because it is literally the website key saved to disk — never a runtime dependency.

To be precise about scope: this PR is a **protocol bridge** — it takes the Command Code Go credential and translates Claude/OpenAI-style requests through the authenticated `/alpha/generate` endpoint into an OpenAI-compatible API for Hermes/agents. It does **not** provide/claim the standard Provider inference API for Go; `/alpha/generate` remains the inference path and the final authority on per-model entitlement (402 treated per-model).

k176060444-lgtm:
Added `e543946` (`fix(openai): pass through context-window fields for commandcode models on /v1/models`) — a small follow-up that came up while validating the catalog sync end-to-end.

**Motivation**

The models handler intentionally narrows every registry entry down to id/object/created/owned_by — uniformly, across all channels and models. That keeps the listing clean and maximally compatible with strict OpenAI-style clients. One side effect of this design is that per-model context metadata held by the registry has no way to reach clients for *any* model. In practice this stays invisible whenever the client can resolve the window from its own metadata sources (well-known model families, third-party catalogs, etc.); it only surfaces for models those sources don't cover yet. That is exactly the case we hit: a local agent resolving `meta/muse-spark-1.2-contributor` via `/v1/models` could not determine its context window and fell back to a conservative 256K default, even though the true window is 1M — compacting conversations far earlier than necessary. Muse Spark is simply the first model we noticed it with, not the only one affected.

**What changed**

- `sdk/api/handlers/openai/openai_handlers.go`: adds a small helper, `appendCommandCodeContextFields()`. For entries where `owned_by == "commandcode"` only, it additionally emits `context_length`, `max_context_length`, and `max_completion_tokens` — and only when the value is positive. Every other entry keeps the exact historical four-field shape.
- `sdk/api/handlers/openai/openai_models_context_fields_test.go`: new test covering three cases — a commandcode entry exposes `context_length`; a zero-value entry omits the field entirely; non-commandcode entries do not leak any of the extra fields.

**Scope**

We deliberately kept the passthrough scoped to the Command Code channel and did not generalize it to other channels — since the underlying filter applies to all channels alike, whether richer model metadata should be exposed across the board seems like a design decision better made by the maintainers, so this change stays minimal unless there is interest in widening it.

**Verification**

`go vet` and the package test suite pass. Also verified against a live deployment: `/v1/models` now reports `"context_length": 1048576` for the Muse-family models, and downstream clients pick up the correct 1M window without any configuration changes.

PR #5069, opened 2026-08-19
fix(codex): skip image tool for Muse Spark

### LINKED ISSUE #5067: Codex executor injects image_generation into muse-spark-1.2; HasSuffix("spark") does not match and disable-image-generation is process-global
## Is it a request payload issue?

Yes. CLIProxyAPI's Codex executor appends a hosted `{ "type": "image_generation", "output_format": "png" }` tool to `/v1/responses` after payload rules. OpenCode Console Go Muse Spark 1.2 (`muse-spark-1.2`) rejects that tool:

```text
invalid_request_error: Unsupported tool type: 'image_generation'
```

The client (Grok Build on Responses) does **not** send hosted `image_generation`. It may send a local function named `image_gen`. The inject happens in CPA.

## Describe the bug

`ensureImageGenerationTool` skips inject when `strings.HasSuffix(baseModel, "spark")`. That was added for models such as `gpt-5.3-codex-spark` (see #2861).

`muse-spark-1.2` ends in `1.2`, not `spark`, so it is injected.

```go
if strings.HasSuffix(baseModel, "spark") {
    return body
}
```

Observed path: Grok Build `api_backend = responses` → new-api → CPA `codex-api-key` (OpenCode Console Go `/zen/go/v1`) → Muse Spark 1.2.

This process is mixed: official Codex **OAuth** (`auth-dir`) and Console Go API keys in one CPA.

## Why existing knobs are not usable on this process

`disable-image-generation` is process-global (`false` | `true` | `chat` | `passthrough`, PR #3832).

| Mode | Why it is not acceptable here |
|---|---|
| `passthrough` | After setting it, official Codex OAuth on the same process went down: Envoy `upstream connect error or disconnect/reset before headers` / `reset reason: remote`. Will not retry `passthrough` on that OAuth box. |
| `chat` | Stops inject for official Codex OAuth models that still rely on CPA to add the hosted image tool. |
| `true` | Also 404s `/v1/images/*`. |
| Muse-only `payload.filter` | Payload rules run **before** `ensureImageGenerationTool`. Filter is overwritten while mode is `false`. |

There is no per-model or per-`codex-api-key` skip.

## Expected

- Do not inject hosted `image_generation` for `muse-spark-1.2` / `muse-spark*` (and ideally any model whose name contains `spark`, not only a suffix).
- Keep inject (or client-supplied tool) for official Codex OAuth models on the same process.
- Prefer a per-model or per-key `disable-image-generation` so mixed OAuth + third-party Codex executors do not share one global flag.

## Related

- #3018 Codex image_generation injection may affect non-Codex harnesses (closed)
- #2861 Tool `image_generation` is not supported with `gpt-5.3-codex-spark` (closed; suffix skip)
- PR #3832 `passthrough` mode (merged)

## Environment

- CPA Codex executor: `ensureImageGenerationTool` after `ApplyPayloadConfig*`
- Model: `muse-spark-1.2` via OpenCode Console Go Responses (`@ai-sdk/openai`)
- Same CPA also serves official Codex OAuth

karlorz:
## Summary

- skip CPA's automatic hosted `image_generation` injection for the explicit `muse-spark-*` model family
- preserve the existing `*-spark` exclusion used by `gpt-5.3-codex-spark`
- keep automatic injection enabled for official Codex models such as `gpt-5.4` when `disable-image-generation` is Off

Fixes #5067.

## Why

OpenCode Console Go exposes `muse-spark-1.2` through the Responses API but rejects the hosted tool with `Unsupported tool type: 'image_generation'`. The existing suffix-only check does not match a versioned model name ending in `1.2`.

This is intentionally model-specific. It does not change the process-global `disable-image-generation` modes, so mixed Codex OAuth and Console Go credentials can remain on the same CPA process without disabling image generation for official Codex models.

## Tests

- `go test -count=1 -run '^(TestEnsureImageGenerationTool.*|TestCodexExecutorExecuteResponsesLiteHeaderDoesNotInjectImageGenerationTool|TestCodexExecutorExecuteStreamResponsesLiteHeaderForcesParallelToolCallsFalse)$' ./internal/runtime/executor`

`go test ./...` also ran. It reaches an unrelated existing failure in `TestCustomMagicHeaders_Claude` (`X-Claude-Code-Session-Id = "", want "claude-session-value"`). The same test fails on untouched `origin/main` at `c92f8bc6`.


karlorz:
Withdrawing this patch at the operator’s request. A hard-coded `muse-spark-*` exception is too narrow for the product: selective hosted-tool injection should be designed as a general per-model capability or per-credential feature, not a model-family special case. The proposed patch and test evidence have been preserved in the operator’s private design notes. Issue #5067 remains open for a broader solution.

PR #5076, opened 2026-08-19
fix: preserve async usage callbacks after request cancellation

haowang02:
Fixes https://github.com/router-for-me/CLIProxyAPI/issues/5244

## Problem

Usage records are delivered asynchronously, but the queue keeps the original request context. If the request finishes before a queued record is dispatched, that context is canceled. Consumers that honor context cancellation can then reject the callback, causing the usage record to be dropped.

## Fix

Detach the context from request cancellation when the record is queued. context.WithoutCancel preserves context values needed by usage consumers while preventing request completion from canceling asynchronous delivery.

## Test

Add a deterministic regression test that blocks dispatch, cancels the original request context, then verifies that the next usage consumer still receives the context value with no cancellation error.

- go test ./sdk/cliproxy/usage ./internal/pluginhost
- go test -race -run TestManagerPublishPreservesValuesAfterRequestCancellation ./sdk/cliproxy/usage

### DIFF
--- a/sdk/cliproxy/usage/manager.go
+++ b/sdk/cliproxy/usage/manager.go
@@ -313,6 +313,13 @@ func (m *Manager) Publish(ctx context.Context, record Record) {
 	if m == nil {
 		return
 	}
+	if ctx == nil {
+		ctx = context.Background()
+	} else {
+		// Usage delivery is asynchronous and commonly runs after the request ends.
+		// Preserve request values without letting request cancellation drop the record.
+		ctx = context.WithoutCancel(ctx)
+	}
 	// ensure worker is running even if Start was not called explicitly
 	m.Start(context.Background())
 	m.mu.Lock()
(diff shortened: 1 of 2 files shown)


haowang02:
维护人员您好，这个 bug 是在我尝试为 CPA 开发计费插件时发现的。我的计费插件需要在每次请求完成后接收 CPA 解析出的 usage，使用其中的 tokens 数据进行计费。

CPA 对 usage 的投递是异步的，但其把 HTTP 请求的 context 一起放入队列。响应结束后 context 随即被取消；如果 usage 还在等待处理，插件调用会因 context.Canceled 被直接跳过，造成相关 usage 不会被投递给计费插件，具体表现就是漏记一些请求。

该 PR 在 usage 入队前使用 context.WithoutCancel：保留请求中的模型、身份等上下文信息，同时解除与 HTTP 请求取消信号的绑定。这样即使响应已经结束，usage 仍能完整送达计费插件，避免账单缺失。

该 PR 仅添加数行代码，不会对现有 CLIProxyAPI 功能造成任何破坏性变更。
希望能尽快合并到仓库中，感谢审核🙏

PR #5120, opened 2026-08-20
[Feat] add a provider of codebuddy cn (both APY-KEY and OAuth login)

rizxfrog:
image

image

image

image

and I have tested (by new-api) and proven these models are avaliable

image


rizxfrog:
the frontend PR is https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/386

PR #5125, opened 2026-08-20
feat: expose current quota resets as an iCalendar feed

TheVerwalter:
## Summary
- expose per-model quota reset runtime state through the plugin host callback
- add the `quota-calendar` dynamic plugin example
- serve one stable-UID VEVENT per provider/model with the latest observed reset

## Design
The current plugin callback exposed only auth-level retry data, while CLIProxyAPI already tracks per-model state in `Auth.ModelStates`. This adds a non-secret `HostModelState` snapshot to `HostAuthFileEntry` and uses it from the plugin.

The feed uses deterministic CRLF serialization, UTC timestamps, stable UIDs, source-derived `DTSTAMP`/`LAST-MODIFIED`, and a 15-minute refresh hint. Expired or missing reset times are omitted.

## Tests
- `go test ./internal/pluginhost ./sdk/pluginapi ./sdk/cliproxy/auth`
- `go test ./...` in `examples/plugin/quota-calendar/go`
- `go build -buildmode=c-shared -o /tmp/quota-calendar.so .`
- `go build -o /tmp/cli-proxy-api ./cmd/server`
- `git diff --check`

## Security
The feed resource is unauthenticated under the existing plugin resource contract. The README explicitly recommends a private bind or reverse-proxy protection. No tokens or credential JSON are exposed by the new callback fields.

### DIFF
--- a/examples/plugin/Makefile
+++ b/examples/plugin/Makefile
@@ -1,4 +1,4 @@
-EXAMPLES := simple model auth frontend-auth executor protocol-format request-translator request-normalizer response-translator response-normalizer thinking usage cli management-api host-callback host-callback-auth-files host-model-callback claude-web-search-router
+EXAMPLES := simple model auth frontend-auth executor protocol-format request-translator request-normalizer response-translator response-normalizer thinking usage cli management-api host-callback host-callback-auth-files host-model-callback quota-calendar claude-web-search-router
 LANGUAGES := go c rust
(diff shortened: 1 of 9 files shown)


TheVerwalter:
Closing this PR: the requested solution must be plugin-only and must not modify the CLIProxyAPI core/plugin host or SDK. I will not pursue the current core-hook approach.

Issue #5085, opened 2026-08-19
5xx responses log no failure detail at info level — cooldown causes and failing legs are invisible without debug


the3asic:
## Environment

- Version: v7.2.111 (docker `eceasy/cli-proxy-api:v7.2.111`)
- Setup: models served by mixed legs — a Claude-compat provider + an OpenAI-compat provider for one model, codex-api-key entries + codex OAuth auth-files for another. `debug: false` in production.

## Symptom 1: 503 `auth_unavailable` with no server-side trace

When every auth for a model is in cooldown, the handler returns 503 in ~7ms (request never reaches an upstream). At info level the only log is the gin line:

`"`"`
[info ] 503 | 7ms | POST /v1/responses
`"`"`

The actual reason — `auth_unavailable: no auth available (providers=...)` — is only in the response body the client sees. On the server we cannot answer *which* auths were cooling, *why*, or *how long is left* without flipping debug on.

## Symptom 2: 500 on upstream TTFB hang, no failing-leg trace

One request hung at the upstream (response headers never arrived), CLIProxyAPI returned 500 after 7.3s. No warn/error line at info identified the provider or upstream error; we had to correlate client-side timestamps against the upstream shim's own logs to find the leg.

## Related: request-scoped 400s cool the whole credential

`isRequestInvalidError()` (internal/.../conductor_cooldown.go) only exempts 400s whose text contains `invalid_request_error` / `bad_request_error` / `INVALID_ARGUMENT` / `FAILED_PRECONDITION`. A provider 400 with a different shape — e.g. Zhipu's `{"error":{"code":"1210","message":"...please use low, high, or max"}}` (an invalid reasoning-effort level, deterministic per request) — falls into the default transient branch and puts the credential into the 1-minute cooldown. For a single-key leg that means one bad request parameter takes the whole leg offline for a minute; combined with an unrelated cooldown on the sibling leg it produced the 503 above.

## Requests

1. On `auth_unavailable` 503: log one warn line listing each cooling auth, the cooldown reason, and remaining time.
2. On upstream failure (5xx/timeout/disconnect): log one warn line with provider id, model, upstream error summary, and duration — independent of debug.
3. Consider classifying request-scoped 400s (deterministic parameter errors) so they don't cool the credential — e.g. extend `isRequestInvalidError()` to recognize provider error-code bodies.

Per-provider `disable-cooling` already mitigates case 3 for us (thank you for that knob) — the main ask is visibility for 1 and 2.

PR #5135, opened 2026-08-21
feat(api): log why auth selection failed (#5085)

YogaSakti:
Addresses request 1 of #5085.

## Problem

When every credential for a model is blocked, the request is rejected before it reaches an upstream. At info level the only trace is the gin access line:

```
[info ] 503 | 7ms | POST /v1/responses
```

The reason — `auth_unavailable: no auth available (providers=..., model=...)` — exists only in the body the client receives. Which credentials were blocked, why, and for how long is reachable only by enabling debug, which is not an option in production.

## Change

- `sdk/cliproxy/auth/selector.go`: `getAvailableAuthsWithPriorityMode` already walks every candidate through `isAuthBlockedForModel` to decide availability. It now reuses that walk to build a compact summary — `codex-a=cooldown for 1m0s (429 quota), codex-b=disabled` — capped at 5 entries with a `+N more` suffix so a large pool cannot flood a log line. The three labels follow the existing `blockReason` classification, so `cooldown` means quota-backed cooldown and `unavailable` means a non-quota block, exactly as the selector already distinguishes them.
- `sdk/api/handlers/handlers_errors.go`: `enrichAuthSelectionError` is already the single point where the client-facing `(providers=..., model=...)` detail is assembled, and every rejected request passes it exactly once. It now emits one `Warnf` through `helps.LogWithRequestID(ctx)`, so the line carries `request_id` like the rest of the codebase and survives `debug: false`. It takes a `ctx` parameter for that; the four call sites are updated.

The summary travels beside `auth.Error` rather than inside it. `TestErrorLegacyUnkeyedLiteralCompatibility` pins that struct to four fields as an unkeyed literal, and keeping the text outside also makes it structurally impossible to serialize toward a client. `errors.As` still reaches the underlying `*Error` through `Unwrap`.

## Verification

```
go test ./...
go build -o test-output ./cmd/server
```

`TestEnrichAuthSelectionErrorLogsBlockedCredentials` asserts exactly one warn entry carrying the provider, model and per-credential reasons, and that the returned client error does not carry them. Dropping the log line to `Debugf` fails it, which is the regression the issue reports. `TestGetAvailableAuthsReportsWhyEveryAuthIsBlocked` covers the three block reasons and that the client-facing message stays clean.

## Not included

Request 2 (a warn line for upstream 5xx/timeout failures) is left out on purpose. That path has no chokepoint: 30 call sites across the executors each emit `helps.LogWithRequestID(ctx).Debugf("request error, error status: %d, ...")`, and the level is chosen at each one. `helps.RecordAPIResponseError` is not an alternative — it writes into the request-log capture and only when `request-log: true`, never to logrus. So it is either a new shared helper plus a 30-site swap, or per-site edits. Happy to send that as a follow-up in whichever shape you prefer.

Request 3 (classifying deterministic provider 400s so they do not cool a credential) changes cooldown semantics and the issue itself raises it as a question, so it is left for a maintainer decision.

### DIFF
--- a/sdk/api/handlers/handlers_auth_selection_log_test.go
+++ b/sdk/api/handlers/handlers_auth_selection_log_test.go
@@ -1,0 +1,55 @@
+package handlers
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	coreauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
+)
+
+func TestEnrichAuthSelectionErrorLogsBlockedCredentials(t *testing.T) {
+	hook := test.NewGlobal()
+	t.Cleanup(hook.Reset)
+	previousLevel := log.GetLevel()
+	log.SetLevel(log.InfoLevel)
+	t.Cleanup(func() { log.SetLevel(previousLevel) })
+
+	now := time.Now()
+	auths := []*coreauth.Auth{
+		{
+			ID:            "codex-a",
+			Quota:         coreauth.QuotaState{Exceeded: true, Reason: "credential_quota", NextRecoverAt: now.Add(time.Minute)},
+			StatusMessage: "429 quota",
+		},
+		{ID: "codex-b", Disabled: true},
+	}
+	selector := &coreauth.FillFirstSelector{}
+	_, selectionErr := selector.Pick(context.Background(), "codex", "gpt-5.6-terra", cliproxyexecutor.Options{}, auths)
+	if selectionErr == nil {
+		t.Fatal("expected selection to fail when every auth is blocked")
+	}
+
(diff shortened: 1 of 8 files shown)


YogaSakti:
Withdrawing this — it duplicates work that already landed.

`48749717` ("fix(cliproxy): add warn-level diagnostics for auth cooldown and upstream execution failures", Closes: #5085) is on `dev` and already adds `warnLogAuthUnavailable` wired into eight manager selection paths, with `authCoolingSummary` rendering identity, reason and remaining time per credential. That covers request 1, and the same commit covers request 2 as well. I missed it because #5085 is still shown as open; my mistake for not checking the branch log before writing code.

The bot review was right on both counts, and for the same underlying reason: my summary sat in the selector wrapper, which the manager paths bypass, and it read the auth-level `StatusMessage` where `authCoolingSummary` already prefers the matching model state's message.

One narrow gap does seem to remain in the landed version, in case it is useful: `warnLogAuthUnavailable` only collects candidates whose block reason is `blockReasonCooldown`, and logs nothing when `coolingSummaries` is empty. A pool where every candidate is blocked by a non-quota `Unavailable` state therefore still produces no warn line. Happy to send a few-line follow-up against `warnLogAuthUnavailable` for that if you consider it worth covering — otherwise please disregard.

Closing.

PR #5140, opened 2026-08-21
fix(auth): lower transient error cooldown default to 10s

warelik:
## Summary

Lowers the effective default transient-error cooldown from 60 seconds to 10 seconds.

A single transient 5xx/408 blip was sidelining a genuinely-live account or model for a full minute. With several auths sharing the same upstream model, one gateway hiccup could cause a service blackout. The legacy `transientErrorCooldown` constant is now `10s`, and the example config documents the new default. The `transient-error-cooldown-seconds` knob remains fully configurable for the normal transient-error path.

## Note on `ErrorCodeForceCooldown`

`transientErrorCooldown` also backs the two `ErrorCodeForceCooldown` fallbacks in `conductor_cooldown.go` (model state and auth state: `now.Add(transientErrorCooldown)`). These enforce request-scoped `stop-and-cooldown` / `continue-and-cooldown` rules. That path uses the same constant directly and is intentionally not gated by `transient-error-cooldown-seconds`, so those forced cooldowns also drop from 60s to 10s. This is consistent with the blackout-avoidance rationale: a force-cooldown rule should not sideline a live account for a full minute either. A code comment has been added to `conductor_refresh.go` to make this explicit.

## Cross-link

Mirror PR in `CLIProxyAPIPlus`: #205

## Related work

- Quota-429 ladder hardening is handled in #5130 (this repo) and #198 (`CLIProxyAPIPlus`). Those PRs are related but not in conflict with this change.

## Test plan

- `sdk/cliproxy/auth/conductor_overrides_test.go` pins the new default at ~10s.
- `go test ./sdk/cliproxy/auth/...` and `go build ./...` pass.

### DIFF
--- a/sdk/cliproxy/auth/conductor_refresh.go
+++ b/sdk/cliproxy/auth/conductor_refresh.go
@@ -32,7 +32,7 @@ const (
 	refreshIneffectiveBackoff = 30 * time.Second
 	quotaBackoffBase          = time.Second
 	quotaBackoffMax           = 30 * time.Minute
-	transientErrorCooldown    = time.Minute
+	transientErrorCooldown    = 10 * time.Second
 )
 
 // StartAutoRefresh launches a background loop that evaluates auth freshness
--- a/sdk/cliproxy/auth/conductor_refresh.go
+++ b/sdk/cliproxy/auth/conductor_refresh.go
@@ -32,7 +32,12 @@ const (
 	refreshIneffectiveBackoff = 30 * time.Second
 	quotaBackoffBase          = time.Second
 	quotaBackoffMax           = 30 * time.Minute
-	transientErrorCooldown    = 10 * time.Second
+	// transientErrorCooldown is the default for 408/500/502/503/504 transient
+	// errors and the fallback enforced cooldown for request-scoped
+	// stop-and-cooldown / continue-and-cooldown rules. The latter use this
+	// constant directly and do not consult the transient-error-cooldown-seconds
+	// knob.
+	transientErrorCooldown = 10 * time.Second
 )
 
 // StartAutoRefresh launches a background loop that evaluates auth freshness
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -162,7 +162,7 @@ disable-cooling: false
 save-cooldown-status: false
 
 # Cooldown duration in seconds for transient upstream errors (408/500/502/503/504).
-# Set to 0 to keep the legacy 60-second cooldown; set to -1 to disable transient error cooldowns.
+# Set to 0 to keep the legacy 10-second cooldown; set to -1 to disable transient error cooldowns.
 transient-error-cooldown-seconds: 0
 
 # When true, globally disable Claude request cloaking (the Claude Code CLI disguise and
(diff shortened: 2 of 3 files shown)


warelik (commented):
**Review verdict: APPROVED** (posted as COMMENT; GitHub rejects APPROVE on your own PR).

The PR lowers the built-in transient-error cooldown default from 60s to 10s by changing the single `transientErrorCooldown` constant, which feeds all three use sites: the `nextTransientErrorRetryAfter` default path (conductor_cooldown.go:94) and the two `ErrorCodeForceCooldown` fallbacks (conductor_cooldown.go:892 and :1966). The `transient-error-cooldown-seconds` knob semantics (0 = built-in default, -1 = disable, N = explicit seconds) are untouched. The one duration-sensitive test is updated with a ±5s window around the new 10s target; the measured value lands just under 10s, so no flake risk. I found no other code, tests, or docs pinning the old 60s value. I did not re-run the test plan; the diff is consistent with it.

Non-blocking notes:
1. See the inline comment on conductor_refresh.go: the request-scoped `stop-and-cooldown` / `continue-and-cooldown` force-cooldown path shares this constant and ignores the config knob, so those rules drop from 60s to 10s with no operator override. Worth confirming as intended.
2. config.example.yaml now says `keep the legacy 10-second cooldown` — 10s is the new default rather than a legacy behavior, so `use the default 10-second cooldown` would read better (it does at least match the `SetTransientErrorCooldownSeconds` doc comment terminology).

Mirror PR kaitranntt/CLIProxyAPIPlus#205 is identical; the same review applies there.

sdk/cliproxy/auth/conductor_refresh.go
```diff
@@ -32,7 +32,7 @@ const (
 	refreshIneffectiveBackoff = 30 * time.Second
 	quotaBackoffBase          = time.Second
 	quotaBackoffMax           = 30 * time.Minute
-	transientErrorCooldown    = time.Minute
+	transientErrorCooldown    = 10 * time.Second
```
warelik: This constant also backs the two ErrorCodeForceCooldown fallbacks in conductor_cooldown.go (lines 892 and 1966: now.Add(transientErrorCooldown)), which enforce request-scoped stop-and-cooldown / continue-and-cooldown rules. That path ignores the transient-error-cooldown-seconds knob entirely, so those rules silently drop from a 60s to a 10s enforced cooldown with no operator override. Presumably intended given the blackout-avoidance rationale, but worth calling out explicitly in the PR description.

warelik (commented):
**Review verdict: APPROVED** — re-review of head 37be828 (posted as COMMENT; GitHub rejects APPROVE on your own PR).

The new docs commit fully addresses my previous inline feedback, in two places:

1. The code comment on `transientErrorCooldown` documents exactly the dual role I flagged: default for 408/500/502/503/504 transient errors (via `nextTransientErrorRetryAfter`) and the enforced fallback for request-scoped stop-and-cooldown / continue-and-cooldown rules, with the latter not consulting the `transient-error-cooldown-seconds` knob. I verified this against conductor_cooldown.go (:94, :892, :1966) — the comment is accurate. Const-block formatting stays gofmt-clean (the comment line starts a new alignment run, so the single-space `=` is correct).
2. The PR description now has an explicit Note on ErrorCodeForceCooldown section calling out the 60s-to-10s behavior change on that path and stating it is intentional — exactly the call-out I asked for.

Nothing else changed since my prior approval. The remaining notes from that review (the config.example.yaml `legacy` wording, and the force-cooldown path having no operator override) were non-blocking and stand as recorded. Ready to merge. Mirror PR kaitranntt/CLIProxyAPIPlus#205 carries the identical change and review.

warelik:
Closing as possibly no longer relevant: I have not used this project for a while, and much of this has likely been addressed in the meantime. Happy to reopen if any of it is still useful.

PR #5143, opened 2026-08-21
fix(auth): rotate API-key model pools for all configured providers

warelik:
Fixes Gap 5 from failover-auth-audit.md.

Previously only openai-compatibility resolved multi-model alias pools; Claude, Gemini, Codex, xAI, and Vertex API-key providers used only the first resolved model. This change resolves the API-key alias pool for every provider whose config carries Models [] and reuses the existing `nextModelPoolOffset` rotation, while keeping `filterExecutionModels` dropping blocked members.

- Adds `resolveAPIKeyUpstreamModelPool` and `modelPoolKey`.
- Updates `executionModelCandidates` and `executionModelCandidatesWithAlias`.
- Adds `conductor_models_test.go` with multi-model rotation, suffix preservation, blocked-member filtering, and single-model regression tests.

Cross-link: https://github.com/kaitranntt/CLIProxyAPIPlus/pull/208

### DIFF
--- a/sdk/cliproxy/auth/conductor_models.go
+++ b/sdk/cliproxy/auth/conductor_models.go
@@ -88,6 +88,22 @@ func openAICompatModelPoolKey(auth *Auth, requestedModel string) string {
 	return strings.ToLower(strings.TrimSpace(auth.ID)) + "|" + openAICompatProviderKey(auth) + "|" + strings.ToLower(base)
 }
 
+func apiKeyModelPoolKey(auth *Auth, requestedModel string) string {
+	base := strings.TrimSpace(thinking.ParseSuffix(requestedModel).ModelName)
+	if base == "" {
+		base = strings.TrimSpace(requestedModel)
+	}
+	provider := strings.ToLower(strings.TrimSpace(auth.Provider))
+	return strings.ToLower(strings.TrimSpace(auth.ID)) + "|" + provider + "|" + strings.ToLower(base)
+}
+
+func modelPoolKey(auth *Auth, requestedModel string) string {
+	if isConfiguredOpenAICompatAuth(auth) {
+		return openAICompatModelPoolKey(auth, requestedModel)
+	}
+	return apiKeyModelPoolKey(auth, requestedModel)
+}
+
 func (m *Manager) nextModelPoolOffset(key string, size int) int {
 	if m == nil || size <= 1 {
 		return 0
@@ -156,6 +172,28 @@ func resolveOpenAICompatUpstreamModelPool(cfg *internalconfig.Config, auth *Auth
 	return resolveModelAliasPoolFromConfigModels(requestedModel, asModelAliasEntries(entry.Models))
 }
 
+func (m *Manager) resolveAPIKeyUpstreamModelPool(auth *Auth, requestedModel string) []string {
+	return resolveAPIKeyUpstreamModelPool(m.loadAPIKeyModelRouting().config, auth, requestedModel)
+}
+
+func resolveAPIKeyUpstreamModelPool(cfg *internalconfig.Config, auth *Auth, requestedModel string) []string {
+	if !isConfiguredModelRoutingAuth(auth) {
+		return nil
+	}
+	requestedModel = strings.TrimSpace(requestedModel)
+	if requestedModel == "" {
+		return nil
+	}
+	if cfg == nil {
+		cfg = &internalconfig.Config{}
+	}
+	models := configuredModelAliasEntries(cfg, auth)
+	if len(models) == 0 {
+		return nil
+	}
+	return resolveModelAliasPoolFromConfigModels(requestedModel, models)
+}
+
 func preserveRequestedModelSuffix(requestedModel, resolved string) string {
 	return preserveResolvedModelSuffix(resolved, thinking.ParseSuffix(requestedModel))
 }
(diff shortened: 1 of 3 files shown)


sdk/cliproxy/auth/conductor_models.go
```diff
@@ -289,11 +327,11 @@ func (m *Manager) executionModelCandidatesWithAlias(auth *Auth, routeModel strin
...
 	}
 	if len(candidates) == 0 {
-		if pool := resolveOpenAICompatUpstreamModelPool(routing.config, auth, upstreamModel); len(pool) > 0 {
+		if pool := resolveAPIKeyUpstreamModelPool(routing.config, auth, upstreamModel); len(pool) > 0 {
 			if len(pool) == 1 {
 				candidates = pool
 			} else {
-				offset := m.nextModelPoolOffset(openAICompatModelPoolKey(auth, upstreamModel), len(pool))
+				offset := m.nextModelPoolOffset(modelPoolKey(auth, upstreamModel), len(pool))
 				candidates = rotateStrings(pool, offset)
```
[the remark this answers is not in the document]
warelik: Fixed in cb43791. predictedHomeConcurrencyModel now uses resolveAPIKeyUpstreamModelPool and returns ("", false) when the resolved pool contains multiple members, matching the existing OpenAI-compat behavior. Added tests for multi-model pool rejection and single-model acceptance.

sdk/cliproxy/auth/conductor_models.go
```diff
@@ -156,6 +172,28 @@ func resolveOpenAICompatUpstreamModelPool(cfg *internalconfig.Config, auth *Auth
 	return resolveModelAliasPoolFromConfigModels(requestedModel, asModelAliasEntries(entry.Models))
 }
 
+func (m *Manager) resolveAPIKeyUpstreamModelPool(auth *Auth, requestedModel string) []string {
```
warelik, line 175: This new generalized helper also needs to replace the one remaining `resolveOpenAICompatUpstreamModelPool` call site: `predictedHomeConcurrencyModel` in `sdk/cliproxy/auth/conductor_home.go:375`. That function predicts which upstream model execution will use, and for openai-compat multi-model pools it deliberately declines (`return "", false`) because the executed member is unpredictable. After this PR, claude/gemini/codex/xai/vertex multi-model pools rotate at execution time, but the prediction path still falls through to `applyAPIKeyModelAlias`, which always resolves the first pool member — so `selection.accountedModel` (used when Home returns no lease envelope) mis-attributes rotated requests to the first member. One-line swap to `m.resolveAPIKeyUpstreamModelPool(auth, upstreamModel)`: behavior for openai-compat auths is identical because `configuredModelAliasEntries` resolves the same config entry for them.

warelik (commented):
**Requesting changes** — one finding.

The pool generalization in `conductor_models.go` itself is correct: `resolveAPIKeyUpstreamModelPool` resolves the same config entry for openai-compat auths as the old helper (via the `default` branch of `configuredModelAliasEntries`), the `modelPoolKey` dispatch preserves existing rotation-state keys for openai-compat, and single-model / no-alias-match requests produce exactly the old results. The new tests cover rotation, suffix preservation, blocked-member filtering, and the single-model regression.

One gap: `predictedHomeConcurrencyModel` (sdk/cliproxy/auth/conductor_home.go:375) still resolves pools via `m.resolveOpenAICompatUpstreamModelPool`. With this change, claude/gemini/codex/xai/vertex multi-model pools now rotate at execution time, but the Home prediction path falls through to `applyAPIKeyModelAlias`, which always resolves the first pool member. For openai-compat multi-model pools that function deliberately declines to predict (`return "", false`) precisely because the executed member is unpredictable. `selection.accountedModel` (used when Home supplies no lease envelope, conductor_home.go:461) will therefore mis-attribute every rotated request that lands on a non-first member. Swapping that one call site to `m.resolveAPIKeyUpstreamModelPool` keeps openai-compat behavior identical and fixes the mismatch.

sdk/cliproxy/auth/conductor_models.go
warelik: Already done in this PR — predictedHomeConcurrencyModel in sdk/cliproxy/auth/conductor_home.go:375 uses m.resolveAPIKeyUpstreamModelPool (see the diff hunk in conductor_home.go). Tests for multi-model pool rejection and single-model acceptance were added in conductor_models_test.go.

warelik (commented):
**Approving** — the follow-up commit resolves the previous finding.

`predictedHomeConcurrencyModel` (sdk/cliproxy/auth/conductor_home.go:375) now uses `resolveAPIKeyUpstreamModelPool`, so multi-model API-key pools on claude/gemini/codex/xai/vertex decline prediction (`return "", false`) exactly as openai-compat pools already did, and single-model pools still resolve to the configured upstream name. Behavior for openai-compat auths is unchanged: `configuredModelAliasEntries` resolves the same compat entry as the old helper. The two new tests (multi-model decline, single-model accept) match `validCanonicalHomeConcurrencyModelKey` semantics.

Rest of the change stands as previously reviewed: pool resolution and rotation generalization is behavior-preserving for single-model and openai-compat paths, with good coverage of rotation, suffix preservation, and blocked-member filtering. LGTM.

warelik:
Closing as possibly no longer relevant: I have not used this project for a while, and much of this has likely been addressed in the meantime. Happy to reopen if any of it is still useful.

PR #5144, opened 2026-08-21
fix(openai): fallback to reasoning field in non-stream responses conversion

warelik:
Stock port of kaitranntt/CLIProxyAPIPlus#191.

Extends non-stream OpenAI Responses conversion to fall back from
`choices.0.message.reasoning_content` to `choices.0.message.reasoning`.

repos in sync.


warelik:
Closing this PR because  blocks any  changes for READ-only accounts. Filed #5147 with the goal, rationale, and intended implementation so the maintenance team can apply the stock port.

warelik:
This PR touches internal/translator/ and therefore fails the translator-path-guard workflow, which blocks any PR in that path from a READ-only account. I have opened #5147 with the defect description, root cause file:line, and proposed patch. This PR is left open as a ready reference implementation for the maintenance team.

warelik:
Stock path-guard makes internal/translator PR unmergeable. Corresponding issue #5147 is now closed. Implementation remains retained upstream in Plus #191 and in warelik/mission-integration.

Issue #193, opened 2025-10-31
claude code with copilot


ahmedrageh:
it use this config 
      {
            "id": "claude-sonnet-4-5-20250929",
            "configId": "thinking",
            "owned_by": "claude",
            "temperature": 0,
            "thinking": {
                "type": "enabled"
            }
        },

with claude code 
but i have problem 

Sorry, your request failed. Please try again. Request id: 8aa9f936-a3de-4148-921e-c1f88f0a3183

Reason: OAI Compatible API error: [400] Bad Request {"type":"error","error":{"type":"invalid_request_error","message":"temperature and top_p cannot both be specified for this model. Please use only one."},"request_id":"req_011CUeYKpaRJEviPWSmX87RJ"}

PR #5145, opened 2026-08-21
fix(claude): write non-stream reasoning to canonical reasoning_content

warelik:
Stock port of kaitranntt/CLIProxyAPIPlus#193.

Ensures Claude → OpenAI non-stream conversion writes thinking output to
`choices.0.message.reasoning_content` instead of the legacy `reasoning`
field, matching the streaming path.

repos in sync.


warelik:
Closing because the workflow blocks any internal/translator/ changes for READ-only accounts. Filed #5148 with the intended implementation.

warelik:
This PR touches internal/translator/ and therefore fails the translator-path-guard workflow, which blocks any PR in that path from a READ-only account. I have opened #5148 with the defect description, root cause file:line, and proposed patch. This PR is left open as a ready reference implementation for the maintenance team.

warelik:
Stock path-guard makes internal/translator PR unmergeable. Corresponding issue #5148 is now closed. Implementation remains retained upstream in Plus #193 and in warelik/mission-integration.

PR #5146, opened 2026-08-21
fix(gemini): deterministic tool-call IDs with collision avoidance

warelik:
Stock port of kaitranntt/CLIProxyAPIPlus#194.

Replaces random/prefixed Gemini tool/call ID generation in
`claude/gemini` and `codex/gemini` request translators with deterministic
sequential IDs (`toolu_*` / `call_*`) that pre-scan explicit IDs and skip
collisions, matching the openai/gemini translator pattern.

repos in sync.

### DIFF
--- a/internal/translator/claude/gemini/claude_gemini_request.go
+++ b/internal/translator/claude/gemini/claude_gemini_request.go
@@ -86,12 +86,41 @@ func ConvertGeminiRequestToClaude(modelName string, inputRawJSON []byte, stream 
 		return ids
 	}
 
+	usedToolIDs := make(map[string]bool)
+	if contents := root.Get("contents"); contents.Exists() && contents.IsArray() {
+		contents.ForEach(func(_, content gjson.Result) bool {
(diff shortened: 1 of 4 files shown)


warelik:
Closing because the workflow blocks any internal/translator/ changes for READ-only accounts. Filed #5149 with the intended implementation.

warelik:
This PR touches internal/translator/ and therefore fails the translator-path-guard workflow, which blocks any PR in that path from a READ-only account. I have opened #5149 with the defect description, root cause file:line, and proposed patch. This PR is left open as a ready reference implementation for the maintenance team.

warelik:
Stock path-guard makes internal/translator PR unmergeable. Corresponding issue #5149 is now closed. Per luispater: in the native Gemini protocol specification, tool and call IDs do not exist in upstream requests and are generated entirely by the system during translation; under standard protocol usage, sequential ID generation will not cause any conflicts. Implementation remains retained upstream in Plus #194 and in warelik/mission-integration.

PR #5201, opened 2026-08-24
feat: support provider-level proxy for OpenAI-compatible providers

wnzzer:
## Summary
- add an optional provider-level `proxy-url` to OpenAI-compatible providers
- fall back to provider proxy when a matching API key has no per-key proxy
- preserve per-key proxy precedence
- document the new configuration field

## Resolution order
`api-key-entries[].proxy-url` > `openai-compatibility[].proxy-url` > global `proxy-url`

## Validation
- `git diff --check` passed
- Added unit tests for provider fallback and per-key override
- Go tests could not run in this environment because Go/gofmt are not installed.


wnzzer:
Closing temporarily to run complete core and integration tests before resubmitting.

PR #5225, opened 2026-08-25
feat(kimi): add native kimi-api-key config and executor routing

### LINKED ISSUE #5097: Feature request:unify Kimi key configuration and support Kimi Code API keys
## Current situation

The current Kimi provider configuration in Management Center only offers
Kimi Open Platform API-key configuration. It asks users to add an
OpenAI-compatible key and an Anthropic key as separate entries.

Both protocols use the same Kimi Open Platform API key, so users currently
have to enter the same key twice if they want to use both protocols.

The form also does not support Kimi Code / Coding Plan API keys. Kimi Code can
currently be used through the existing OAuth login flow, but the Kimi provider
configuration has no Kimi Code / Coding Plan API-key entry.

## Requested design

Keep a single Kimi provider in Management Center. 

When adding a Kimi key, let the user select the service:

- Kimi Open Platform
- Kimi Code / Coding Plan

When Kimi Open Platform is selected, require a region selection before the
API-key field:

- Domestic
- International

Kimi Code does not need a region selection.

The Kimi form should no longer ask users to choose between separate
OpenAI-compatible and Anthropic protocol entries. Users should enter the API
key once, and CPA should route each request automatically based on its inbound
protocol.

The form should select the Kimi service, not the inbound protocol.

## Kimi Open Platform requests

The user should enter one Open Platform API key only once.

Open Platform API-key requests should use the same KimiExecutor request-routing
and translation flow currently used by Kimi Code OAuth credentials. The
selected region determines the Open Platform endpoints.

Within this shared flow:

- OpenAI Chat Completions requests should use Kimi's OpenAI-compatible endpoint.
- Anthropic Messages requests should use Kimi's Anthropic-compatible endpoint.
- Other supported CPA inbound formats should continue to use the existing
  translation behavior.

The user should not need to enter the same Open Platform key once for each
protocol or provide a custom base URL.

## Kimi Code API keys

Add support for Kimi Code / Coding Plan API keys in the same Kimi provider form.

A Kimi Code API key should use the same KimiExecutor request routing and model
handling as the existing Kimi Code OAuth credential. Existing Kimi OAuth login
and token refresh behavior must remain unchanged.

## Kimi Code quota

Kimi Code API keys can also query the account's current Coding Plan quota, just
like Kimi Code OAuth credentials.

Kimi provides this information through:

https://api.kimi.com/coding/v1/usages

When a Kimi Code API key is configured, it should appear in the Kimi section of
the quota management page and show the same quota card as a Kimi Code OAuth
credential.

Only Kimi Code credentials should use this endpoint. Kimi Open Platform
credentials should not trigger a Coding quota request.

## Expected result

The Kimi provider should have one key form with the following choices:

- Kimi Open Platform
  - Region
  - API key
- Kimi Code / Coding Plan
  - API key

Each Kimi key entry should also support an optional custom display name. When
multiple Kimi keys are configured, the name should be shown in the provider and
quota management views so users can distinguish the keys easily.

The user selects the Kimi service and enters the key once. CPA handles endpoint
selection, protocol routing, and request translation automatically.

YIXIE233:
## Summary
Adds `kimi-api-key` so a Kimi API key is a normal Kimi credential. You pick Open Platform or Coding Plan, enter the key once, and Chat Completions / Messages go to the matching Kimi endpoints. Same executor as Kimi OAuth.

Fixes #5097

## Related
This is the same change as the Management Center UI: https://github.com/router-for-me/Cli-Proxy-API-Management-Center/pull/393

## Why
Kimi OAuth already works as one provider. API keys don't.

If you want both Chat Completions and Messages, you have to add the same key twice: `openai-compatibility` and `claude-api-key`. That's also what the Management Center Kimi form does today. Coding Plan API keys aren't in that form at all (only OAuth).

Those two entries register the same model name on two different providers. When a client calls `k3` (or whatever you configured), CPA mixed-schedules across both. It does not look at whether the call was `/v1/chat/completions` or `/v1/messages`:

- a chat request can hit the Claude key and get translated to Messages
- a Messages request can hit the OpenAI-compat key and get translated to chat

#5097 wants one Kimi form, one key, service + region (Open Platform) or just the key (Coding Plan), and CPA picking the Kimi URL from the request.

### Config
```yaml
kimi-api-key:
  - api-key: "sk-..."
    service: open-platform    # or coding-plan
    region: domestic          # open-platform only: domestic | international
    name: "optional label"
```

Also supports the usual per-key fields: weight, prefix, proxy-url, headers, models, excluded-models, priority, disable-cooling.

Hosts come from `service` / `region`. No custom base-url:

- coding-plan → `https://api.kimi.com/coding`
- open-platform domestic → `https://api.moonshot.cn`
- open-platform international → `https://api.moonshot.ai`

### Routing
One key becomes provider `kimi`. The existing Kimi executor then:

- Chat Completions → Kimi's OpenAI-compatible `/v1/chat/completions`
- Anthropic Messages → Kimi's Anthropic endpoint (`…/anthropic` on Open Platform, coding host otherwise)
- other CPA request formats keep the current translation

Open Platform calls don't send Kimi Code `X-Msh-*` headers. OAuth login and refresh are untouched.

### Models
If you set `models`, aliases and thinking on that key work like other API keys. If you leave `models` empty, the built-in Kimi catalog is used (same idea as xAI). In Management Center, saving a key with no model names fetches `/v1/models` and fills the upstream names only (no auto aliases).

### Management / TUI
- `GET/PUT/PATCH/DELETE /v0/management/kimi-api-key` (includes `auth-index`)
- TUI Keys tab shows them read-only, same as xAI/Claude keys

### Quota
Coding Plan keys use the same usages URL as Kimi Code OAuth:

https://api.kimi.com/coding/v1/usages

They show up on the quota page in the Kimi section, with the same card as an OAuth file.

`name` is optional. If you set it, the provider list and quota cards use it. If you don't, keys are still distinguishable the same way as Claude/xAI: masked API key, then `#1`, `#2`, …

Open Platform keys never call the Coding usages API and don't appear on that Kimi quota card.

## Tests
- yaml parse/sanitize, host helpers, matching by api-key + service + region
- synthesizer writes `service`, `region`, `base_url`, `models_hash`
- management CRUD, weight, disable-cooling
- Open Platform URL/header switch
- alias routing and thinking on the selected key
- `gofmt`, `go test` on those packages, `go build -o test-output ./cmd/server`

## Notes
An API key with empty `models` advertises the full built-in Kimi catalog, including coding-oriented IDs. Put the upstream names in `models` if you only want those.


YIXIE233:
@codex review

PR #5228, opened 2026-08-25
fix(antigravity): remove cross-endpoint fallback (#5209)

### LINKED ISSUE #5209: [Bug/Architecture] Antigravity consumer accounts must exclusively use daily-cloudcode-pa to prevent 429 cooldown cascades
## 📌 Problem & Symptom

In CLIProxyAPI (CPA), Antigravity OAuth accounts (Google AI Pro, Ultra, One, and Free tier) frequently encounter cascading `429 Too Many Requests / RESOURCE_EXHAUSTED` errors, which causes CPA to mark healthy credentials as unavailable and place them into **Cooldown**. Over time, this leads to large-scale account pool starvation.

---

## 🔬 Root Cause Analysis

We investigated this through both **binary disassembly of the official `agy` client (`~/.local/bin/agy`)** and **live API probing across 24 active consumer accounts**.

### 1. Official `agy` Routing Architecture
In `codeassistclient.(*CLIAuthProvider).ResolveAuthMode` and `UpdateEndpointURL`:
1. `agy` queries `POST /v1internal:loadCodeAssist` to determine user entitlements.
2. For personal Google accounts (Free, Pro, Ultra, Google One where `gcpManaged == false`), it resolves to `AuthMode: personal_consumer`.
3. In `UpdateEndpointURL`, `personal_consumer` accounts are **strictly and exclusively bound to `https://daily-cloudcode-pa.googleapis.com`**.
4. **Zero Cross-Tier Fallback:** The official `agy` client **never** falls back to `cloudcode-pa.googleapis.com` for consumer accounts. The two backends are strictly partitioned.

### 2. Live Probe Verification (24 Active Accounts)
Using identical request payloads, headers, and OAuth access tokens:
* `https://daily-cloudcode-pa.googleapis.com`: **100% (24/24) return HTTP 200 OK** with smooth streaming generation.
* `https://cloudcode-pa.googleapis.com`: **100% (24/24) return HTTP 429 RESOURCE_EXHAUSTED** because Google gates this endpoint to enterprise/GCP-licensed accounts.

### 3. The Fallback Cascading Cooldown Trap in CPA
Currently, in `internal/runtime/executor/antigravity_executor_request.go`:
```go
var antigravityBaseURLFallbackOrder = func(auth *cliproxyauth.Auth) []string {
    if base := resolveCustomAntigravityBaseURL(auth); base != "" {
        return []string{base}
    }
    return []string{
        antigravityBaseURLDaily,
        antigravityBaseURLProd, // https://cloudcode-pa.googleapis.com
    }
}
```
1. When a consumer account experiences transient latency or single-request concurrency throttling on `daily`, CPA retries against `prod` (`cloudcode-pa`).
2. `cloudcode-pa` **always answers 429** for consumer accounts.
3. CPA interprets this as genuine quota exhaustion and puts the healthy credential into **Cooldown**.
4. Traffic rolls over to the next credential, triggering the same failure loop until the entire account pool is locked out.

---

### 1. Remove `cloudcode-pa` from Default Fallback
In `internal/runtime/executor/antigravity_executor_request.go`, restrict the default slice to `antigravityBaseURLDaily` only:

```diff
--- a/internal/runtime/executor/antigravity_executor_request.go
+++ b/internal/runtime/executor/antigravity_executor_request.go
@@ -439,7 +439,6 @@ var antigravityBaseURLFallbackOrder = func(auth *cliproxyauth.Auth) []string {
 	}
 	return []string{
 		antigravityBaseURLDaily,
-		antigravityBaseURLProd,
 		// antigravitySandboxBaseURLDaily,
 	}
 }
```

*(Note: Custom endpoints can still be explicitly configured per-credential via `base_url` in auth JSON / attributes).*

---

## 📋 Additional Observation: Quota RPC
Official `agy` defines `POST /v1internal:retrieveUserQuotaSummary` (`RetrieveUserQuotaResponse` with `BucketInfo` containing `remaining_fraction`, `remaining_amount`, and `reset_time`), which can be leveraged for proactive quota accounting and intelligent credential scheduling in the future.

hkfires:
Fixes #5209

### DIFF
--- a/internal/runtime/executor/antigravity_executor_buildrequest_test.go
+++ b/internal/runtime/executor/antigravity_executor_buildrequest_test.go
@@ -13,6 +13,34 @@ import (
 	sdktranslator "github.com/router-for-me/CLIProxyAPI/v7/sdk/translator"
 )
 
+func TestResolveAntigravityRequestBaseURL(t *testing.T) {
+	t.Run("default uses daily endpoint", func(t *testing.T) {
+		if got := resolveAntigravityRequestBaseURL(&cliproxyauth.Auth{}); got != antigravityBaseURLDaily {
+			t.Fatalf("base URL = %q, want %q", got, antigravityBaseURLDaily)
+		}
+	})
+
+	t.Run("custom attribute endpoint remains supported", func(t *testing.T) {
+		auth := &cliproxyauth.Auth{Attributes: map[string]string{"base_url": "https://enterprise.example.com/"}}
+		if got := resolveAntigravityRequestBaseURL(auth); got != "https://enterprise.example.com" {
+			t.Fatalf("base URL = %q, want custom endpoint", got)
+		}
+	})
+
+	t.Run("custom auth file endpoint remains supported", func(t *testing.T) {
+		auth := &cliproxyauth.Auth{Metadata: map[string]any{"base_url": "https://enterprise.example.com/"}}
+		if got := resolveAntigravityRequestBaseURL(auth); got != "https://enterprise.example.com" {
+			t.Fatalf("base URL = %q, want custom auth file endpoint", got)
+		}
+	})
+}
+
+func TestAntigravityLoadCodeAssistBaseURLRemainsProdByDefault(t *testing.T) {
+	if got := antigravityLoadCodeAssistBaseURL(&cliproxyauth.Auth{}); got != antigravityBaseURLProd {
+		t.Fatalf("loadCodeAssist base URL = %q, want %q", got, antigravityBaseURLProd)
+	}
+}
+
 func TestAntigravityBuildRequest_SanitizesGeminiToolSchema(t *testing.T) {
 	body := buildRequestBodyFromPayload(t, "gemini-2.5-pro")
 
--- a/internal/runtime/executor/antigravity_executor_credits.go
+++ b/internal/runtime/executor/antigravity_executor_credits.go
@@ -525,55 +525,8 @@ func (e *AntigravityExecutor) updateAntigravityCreditsBalance(ctx context.Contex
 		return
 	}
 }
-func antigravityShouldRetryNoCapacity(statusCode int, body []byte) bool {
-	if statusCode != http.StatusServiceUnavailable {
-		return false
-	}
-	if len(body) == 0 {
-		return false
-	}
-	msg := strings.ToLower(string(body))
-	return strings.Contains(msg, "no capacity available")
-}
-
-func antigravityShouldRetryTransientResourceExhausted429(statusCode int, body []byte) bool {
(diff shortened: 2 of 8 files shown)


sususu98:
## Follow-up: enterprise / PayGo endpoint binding (do not land in this PR)

This change is the right consumer fix for #5209: generate / stream / countTokens now bind to `daily-cloudcode-pa` with **zero cross-endpoint fallback**. Live check on `0854cb2d` agreed: CPA request-logs showed exactly one `https://daily-cloudcode-pa.googleapis.com` call per path, and the same consumer token still gets `429 RESOURCE_EXHAUSTED` on `https://cloudcode-pa.googleapis.com`.

Official `agy` (`CLIAuthProvider`) does **not** fallback either. It classifies, then hard-binds:

1. Explicit override wins (`CLOUD_CODE_URL` / `configCloudCodeURL`) — CPA equivalent is per-auth `base_url`.
2. `ResolveAuthMode` from `LoadCodeAssist` / entitlement:
   - `personal_consumer` (Free / Google AI Pro / Ultra / One) → **daily only**
   - `business_licensed_enterprise` / `business_agent_platform_paygo` → **prod only**
   - entitlement failure / pending / unknown → consumer / daily
3. `UpdateEndpointURL` writes the chosen host in memory. `agy` persists `userTier`, **not** the resolved URL as `base_url`.

### What not to do

Do **not** auto-write the inferred host into auth JSON `base_url`.

- `base_url` is already the operator override. Once written, it is indistinguishable from a human pin and will freeze a wrong choice across refresh / upgrade / downgrade.
- `paidTier` and `project_id` are **not** enterprise signals. Pro/Ultra consumer accounts have `paidTier`; `onboardUser` still mints a Cloud Code `project_id`. Using either to pick prod would recreate the #5209 cooldown cascade.

Until we have a confirmed enterprise / PayGo `loadCodeAssist` payload, keep fail-closed: unknown → daily. A missed enterprise account can set `base_url` to prod. The reverse (consumer locked on prod) is the outage.

### Suggested follow-up (separate PR)

Reuse the `loadCodeAssist` CPA already issues on prod (login / `PrepareRequestAuth` / credits hint). Parse `gcpManaged` plus the real enterprise / PayGo entitlement fields (need a live enterprise sample — do not guess from `paidTier`). Cache `auth_mode` in memory; optionally persist diagnostic-only keys such as `gcp_managed` / `antigravity_auth_mode` / `user_tier`. Resolver order:

`explicit base_url` → `auth_mode in {enterprise, paygo} ? prod : daily`

Generation stays single-host. `loadCodeAssist` can remain on prod for classification. `sdk/cliproxy` `fetchAvailableModels` still has a `[daily, prod]` loop; that path does not cooldown, but it should be aligned in the same follow-up.

PR #5235, opened 2026-08-25
feat(wsrelay): accept stable provider name via provider_name query parameter

### LINKED ISSUE #5234: feat: allow AI Studio websocket clients to register a stable provider name (avoid aistudio-XXXX churn)
**Is a request payload issue?**
No.

**Problem**
AI Studio websocket providers (registered via `/v1/ws`) always get a random `aistudio-XXXX` name generated server-side in `internal/wsrelay/manager.go` (`randomProviderName()`). Because these runtime-only credentials are also deleted on disconnect (`wsOnDisconnected` -> `AuthUpdateActionDelete`), every browser restart produces a brand-new provider entry and the old one disappears:

- Provider / usage statistics are aggregated by `Metadata["email"]`, which is just the random channel ID - so per-account stats are impossible to keep attached to one identity.
- After a restart you cannot tell which new provider corresponds to which Google account, so anomalies cannot be traced back to a specific account.

**Proposal**
Wire the existing-but-unused `wsrelay.Options.ProviderFactory` hook so a client can claim a stable name via a `provider_name` query parameter on the websocket upgrade request, e.g.:

```
wss://proxy.example.com/v1/ws?auth_token=...&provider_name=user1
```

With a stable name, the manager's existing "replaced by new connection" path kicks in on reconnect: the old session of the same provider is replaced and `wsOnDisconnected` skips the registry delete, so the provider entry and its usage statistics survive restarts. Clients that do not send the parameter keep the current random-name behavior (fully backwards compatible).

Naming rule (for traceability): the value is trimmed and lowercased by the manager, and the `aistudio-` prefix is enforced in the factory because `wsOnConnected` only registers channels carrying that prefix.

I have a working patch (16 lines in `sdk/cliproxy/service_auth.go`) and will open a PR against `dev` referencing this issue.

**Use case**
Multi-account AI Studio keep-alive setups (e.g. one headless browser per Google account): each browser instance injects its own account identifier as `provider_name`, so the CLIProxyAPI dashboard shows one stable entry per account.

**OS Type**
- OS: Linux (Docker)

xyonium:
## Summary

Wire the existing-but-unused `wsrelay.Options.ProviderFactory` so websocket clients can claim a stable provider identity via an optional `provider_name` query parameter on the `/v1/ws` upgrade request:

```
wss://proxy.example.com/v1/ws?auth_token=...&provider_name=user1  ->  provider "aistudio-user1"
```

Fixes #5234

## Why

AI Studio websocket providers currently always get a random `aistudio-XXXX` name generated in `internal/wsrelay/manager.go`. Since these runtime-only credentials are deleted on disconnect, every client restart deletes the old provider entry and registers a brand-new one. Usage statistics (aggregated by `Metadata["email"]` = the random channel ID) become fragmented across restarts, and after a restart there is no way to tell which provider entry corresponds to which Google account.

With a stable name the manager's existing "replaced by new connection" path handles reconnects: the old session of the same provider is replaced, `wsOnDisconnected` skips the registry delete for the replacement case, and the provider entry with its usage statistics survives restarts.

## Changes

- `sdk/cliproxy/service_auth.go` - `ensureWebsocketGateway()` now passes a `ProviderFactory` to `wsrelay.Options`:
  - reads `provider_name` from the upgrade request query (trimmed);
  - enforces the `aistudio-` prefix (required by `wsOnConnected`, which only registers prefixed channels);
  - returns an empty string when absent -> the manager falls back to its random name, so behavior is fully backwards compatible.

No changes to `internal/wsrelay/` - the factory hook was already there, it was simply never wired up.

## Naming rule (traceability)

| client sends | registered provider |
|---|---|
| (nothing) | `aistudio-<random16>` (unchanged) |
| `provider_name=user1` | `aistudio-user1` |
| `provider_name=Aistudio-Bob` | `aistudio-bob` (manager lowercases) |

Server-side lowercasing is pre-existing manager behavior (`s.provider = strings.ToLower(name)`).

## Verification

- `gofmt` clean, `go build ./cmd/server` passes on top of current `dev` (998dcfeb).
- Runtime-tested in production deployment (multi-account AI Studio keep-alive, one headless browser per Google account): each account's client now appends its own `provider_name`; after browser restarts the provider list stays stable and per-account usage statistics remain attached to one identity.

## Security note

Anyone holding a valid `auth_token` could claim/steal an existing provider name (the previous session of that name gets replaced). This is the same trust boundary as today - any valid token can already register a provider, just under a random name - so no new privilege is introduced; only the display/aggregation identity becomes attacker-choosable.

### DIFF
--- a/internal/wsrelay/manager.go
+++ b/internal/wsrelay/manager.go
@@ -185,11 +185,13 @@ func (m *Manager) handleSessionClosed(s *session, cause error) {
 	m.sessMutex.Lock()
 	if cur, ok := m.sessions[key]; ok && cur == s {
 		delete(m.sessions, key)
+	} else if ok {
+		cause = errors.New("replaced by new connection")
 	}
-	m.sessMutex.Unlock()
 	if m.onDisconnected != nil {
 		m.onDisconnected(s.provider, cause)
 	}
+	m.sessMutex.Unlock()
 }
 
 func randomProviderName() string {
--- a/sdk/cliproxy/service_auth.go
+++ b/sdk/cliproxy/service_auth.go
@@ -2,6 +2,7 @@ package cliproxy
 
 import (
 	"context"
+	"net/http"
 	"strings"
 	"time"
 
@@ -197,7 +198,21 @@ func (s *Service) ensureWebsocketGateway() {
 		return
 	}
 	opts := wsrelay.Options{
-		Path:           "/v1/ws",
+		Path: "/v1/ws",
+		// Accept an optional stable provider name from the websocket client via the
+		// "provider_name" query parameter. With a stable name, reconnects replace the
+		// previous session of the same provider instead of registering a new random
+		// aistudio-XXXX provider on every browser restart, keeping provider and usage
+		// statistics attached to one identity. The "aistudio-" prefix is enforced here
+		// because wsOnConnected only registers channels carrying it. An empty name
+		// falls back to the manager's random name (previous behavior).
+		ProviderFactory: func(r *http.Request) (string, error) {
+			name := strings.TrimSpace(r.URL.Query().Get("provider_name"))
+			if name != "" && !strings.HasPrefix(strings.ToLower(name), "aistudio-") {
+				name = "aistudio-" + name
+			}
+			return name, nil
+		},
 		OnConnected:    s.wsOnConnected,
 		OnDisconnected: s.wsOnDisconnected,
 		LogDebugf:      log.Debugf,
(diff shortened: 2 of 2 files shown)


xyonium:
Closing this one. It grew into three separate concerns (one feature + two race fixes), and one of them deletes a guard in `wsOnConnected` that was added deliberately — which I never explained. As a single diff it can't be reviewed.

Splitting it up:

- The `provider_name` feature is resubmitted on its own: no `internal/wsrelay/` changes, no behavior change when the parameter is absent.
- The two reconnect races are filed as a separate bug report. They exist on current `dev` regardless of whether the feature lands.

Thanks for looking at it.

PR #5245, opened 2026-08-25
fix(codex): reduce Claude code-mode context overhead

DerekCounihan:
## Problem
Codex code-mode requests embed the full nested connector registry inside the `functions.exec` description when translated to Claude. With a large app catalogue this consumed roughly 650k tokens before the first user message and caused compaction to appear ineffective.

## Solution
- advertise compact Codex metadata for Claude models, including Responses Lite, code-mode-only tools, and an 850k auto-compaction threshold
- replace only the expanded `functions.exec` description with concise `ALL_TOOLS` discovery instructions while preserving the executable tool and every nested connector
- add regression coverage for both model metadata and tool-description compaction

## Verification
- `go test -count=1 ./...`
- `go build -o /private/tmp/cli-proxy-api-pr-check ./cmd/server`
- live Fable smoke test reduced first-request input from 654,197 to 48,654 tokens
- live tool-discovery smoke test still found all 46 Outlook Email tools without calling Outlook

## Screenshots / Recordings
Not applicable: backend translation and model metadata only.

Generated with an AI coding agent

### DIFF
--- a/internal/client/codex/models/models.go
+++ b/internal/client/codex/models/models.go
@@ -231,12 +231,14 @@ func applyCodexClientSearchToolSupport(entry map[string]any, id string, template
 
 func applyCodexClientModelMetadata(entry map[string]any, id string, model map[string]any, optimizeMultiAgentV2 bool) {
 	info := registry.LookupModelInfo(id)
+	isClaudeModel := strings.EqualFold(stringModelValue(model, "type"), "claude")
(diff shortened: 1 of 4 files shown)


DerekCounihan:
Closing this upstream PR. The change will be maintained as a local CLIProxyAPI patch instead.

PR #5291, opened 2026-08-27
feat(registry): overridable remote model catalog URLs (env)

WaldemarFech:
## What

Lets users override where CLIProxyAPI fetches the remote model catalogs from, via two environment variables:

- `CPA_REMOTE_MODELS_URL` — models.json (provider routing + /v1/models)
- `CPA_REMOTE_CODEX_CLIENT_MODELS_URL` — codex_client_models.json (client view)

When set, a comma-separated list of custom URLs is tried **before** the built-in mirrors; built-in defaults stay appended as fallback so a broken custom source never disables catalog updates.

## Why

Today, serving a model slot that the bundled/remote catalogs do not know yet (e.g. `gpt-daybreak-blue-latest`, visible in the official Codex CLI cache but absent from router-for-me/models) requires either waiting for a catalog release or patching the source and rebuilding. With this change it becomes a one-line env var against any fork/raw URL.

It also makes catalog PRs trivially testable: point the env at your branch's raw URL, start the server, verify. We used exactly this method to prove [models#44](https://github.com/router-for-me/models/pull/44) end-to-end through the real updater (log line + completion probe).

## Implementation notes

- Small shared resolver in `internal/registry/remote_urls.go`; both updaters (`model_updater.go`, `codex_client_models_updater.go`) swap their loop source to `resolveRemoteURLs(...)`. Defaults untouched otherwise.
- Unit tests cover unset/empty env and list trimming; added doc block to config.example.yaml including how to discover correct slugs (`~/.codex/models_cache.json`, `/v1/models`) with ready-to-paste PowerShell/bash commands.
- `.env` files continue to work since values are read via os.Getenv at fetch time.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -2,6 +2,26 @@
 # Use "127.0.0.1" or "localhost" to restrict access to local machine only.
 host: ""
 
+# ── Remote model catalog sources (env-based) ───────────────────────────────
+# The catalogs that feed model routing and /v1/models are refreshed from
+# remote URLs. Custom URLs set via these environment variables are tried FIRST,
+# official mirrors follow as fallback (comma-separated list, whitespace trimmed):
+#
+#   CPA_REMOTE_MODELS_URL                models.json            (provider routing)
+#   CPA_REMOTE_CODEX_CLIENT_MODELS_URL   codex_client_models.json (client view)
+#
+# Discovering correct model slugs for your account:
+#   - Official Codex CLI cache lists every slot your account can see:
+#       PowerShell: Select-String 'slug' "$env:USERPROFILE\.codex\models_cache.json"
+#   - Once a slug is routed, the active catalog is visible at:
+#       curl http://127.0.0.1:8317/v1/models -H "Authorization: Bearer <api-key>"
+#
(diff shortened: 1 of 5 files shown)


WaldemarFech:
Extension pushed to this branch (second commit 5f14797e): source entries that are not http(s) URLs are now read from disk - plain paths including Windows ones like C:\catalogs\my-models.json, relative paths, and file:// URIs.

This removes the hosting requirement from the override mechanism entirely:

```powershell
$env:CPA_REMOTE_MODELS_URL = 'C:\catalogs\my-models.json'
```

Unit tests cover resolver prepending, scheme detection table cases and local reads via t.TempDir; AGENTS.md build verification ran clean.

WaldemarFech:
Verified the local-file source extension end-to-end against a real Pro account:

```
startup model refresh completed from
  C:\Users\walde\...\test-catalog\models.json,
  changes detected for providers: [codex]
GET /v1/models contains gpt-daybreak-blue-latest (sourced only from that local file)
chat completion on the slot returned HTTP 200 (probe echoed LOCALFILE_OK)
```

The tested binary was built from HEAD of this branch with a catalog that deliberately did NOT contain the slot embedded, so the positive result can only come from the new local-file read path.

PR #5303, opened 2026-08-28
feat(executor): add opt-in response cache for OpenAI-compatible providers

kantar0:
## Summary

This PR introduces an optional, configurable response cache for OpenAI-compatible providers.

### Rationale & Problem Solved
Many upstream providers and aggregator relays (such as New API, One API, custom reverse proxies, or endpoints lacking native prompt caching) charge the full input and output token price on every request, including byte-identical retries, repeated agent checks, and subagent validations.

By enabling `response-cache`, CLIProxyAPI can serve identical requests locally from a bounded in-memory LRU cache with a customizable TTL. This provides:
- **Zero latency** on repeat queries.
- **Zero token cost / quota usage** on upstreams that do not support prompt caching.
- **Full streaming support:** SSE streams are cached on successful completion (`[DONE]`) and replayed faithfully through the translator pipeline.

### Changes
1. **Cache Primitive (`internal/cache/response_cache.go`):**
   - Thread-safe LRU eviction with TTL expiration.
   - Configurable `MaxEntries` (default 256), `MaxEntryBytes` (default 2MiB), and `TTL` (default 5m).
   - Optional model allowlist.

2. **Executor Integration (`internal/runtime/executor/openai_compat_executor.go`):**
   - Intercepts both non-streaming and streaming requests before the HTTP call.
   - Non-streaming: Serves cached body and translates according to client format.
   - Streaming: Replays raw SSE frames through `replayCachedStream`, preserving token tracking and client compatibility. Incomplete or aborted streams are never cached.

3. **Config Definition (`internal/config/config_types.go` & `config.example.yaml`):**
   - Added `ResponseCacheConfig` under `openai-compatibility[].response-cache`.
   - Disabled by default (`enabled: false`) to preserve existing behavior.

4. **Lifecycle & Helpers (`internal/runtime/executor/helps/response_cache.go`):**
   - Provider cache registry outlives config hot-reloads so valid entries are preserved across reloads.

### Verification
- Added comprehensive unit tests covering cache expiration, LRU eviction, stream serialization/replay, error non-caching, and allowlists.
- Verified with live A/B testing on test servers (both non-stream and SSE stream replayed identically).
- Passed `gofmt` and package tests (`go test ./internal/config/ ./internal/cache/ ./internal/runtime/executor/...`).

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -616,6 +616,12 @@ nonstream-keepalive-interval: 0
 #     support-prompt-cache-key: false # optional: derive prompt_cache_key for requests from all input protocols
 #     disable-cooling: false # optional provider override: true disables cooling, false enables it; omit to inherit global
 #     request-retry: 3 # optional per-provider override; 0 disables additional rounds; omit or set < 0 to inherit global
+#     response-cache: # optional: reuse upstream responses for byte-identical requests (useful for aggregators that bill every retry and offer no prompt caching)
+#       enabled: false # default false; only exact same-payload requests hit the cache
+#       ttl: "5m" # optional: entry lifetime; default 5m
+#       max-entries: 256 # optional: retained responses; default 256
+#       max-entry-bytes: 2097152 # optional: per-response size limit; default 2 MiB
+#       models: [] # optional: restrict caching to these upstream model names; empty caches every model
 #     request-scoped-errors: # optional: custom rules to classify upstream errors by status and body patterns
 #       - status: 400
 #         match:
--- a/internal/api/handlers/management/config_auth_index.go
+++ b/internal/api/handlers/management/config_auth_index.go
@@ -48,6 +48,7 @@ type openAICompatibilityWithAuthIndex struct {
 	Models                []config.OpenAICompatibilityModel        `json:"models,omitempty"`
 	Headers               map[string]string                        `json:"headers,omitempty"`
 	SupportPromptCacheKey bool                                     `json:"support-prompt-cache-key,omitempty"`
+	ResponseCache         *config.ResponseCacheConfig              `json:"response-cache,omitempty"`
 	DisableCooling        *bool                                    `json:"disable-cooling,omitempty"`
 	RequestRetry          *int                                     `json:"request-retry,omitempty"`
 	RequestScopedErrors   []config.RequestScopedErrorRule          `json:"request-scoped-errors,omitempty"`
@@ -309,6 +310,7 @@ func (h *Handler) openAICompatibilityWithAuthIndex() []openAICompatibilityWithAu
 			Models:                entry.Models,
 			Headers:               entry.Headers,
 			SupportPromptCacheKey: entry.SupportPromptCacheKey,
+			ResponseCache:         &entry.ResponseCache,
 			DisableCooling:        entry.DisableCooling,
 			RequestRetry:          entry.RequestRetry,
 			RequestScopedErrors:   entry.RequestScopedErrors,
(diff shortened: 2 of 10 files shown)


kantar0:
Addressed all review feedback in commit :

1. **Custom header expansion:** Restored  parameter in  across request executors.
2. **Responses usage details:** Applied  to non-stream responses on cache hit and miss paths.
3. **Response headers preservation:** Stored and returned cloned upstream headers on both stream and non-stream cache hits.
4. **Early stream memory bounding:** Bounded SSE frame accumulation during streaming using  to avoid buffering oversized streams.
5. **Management API support:** Added  field projection and patch handling to  and .

kantar0:
Addressed all review feedback in commit `25067d9b`:

1. **Custom header expansion:** Restored `opts.Headers` parameter in `ApplyCustomHeadersFromAttrs` across request executors.
2. **Responses usage details:** Applied `helps.EnsureResponsesUsageDetails` to non-stream responses on cache hit and miss paths.
3. **Response headers preservation:** Stored and returned cloned upstream headers on both stream and non-stream cache hits.
4. **Early stream memory bounding:** Bounded SSE frame accumulation during streaming using `MaxEntryBytes` to avoid buffering oversized streams.
5. **Management API support:** Added `response-cache` field projection and patch handling to `openAICompatibilityWithAuthIndex` and `PatchOpenAICompat`.

kantar0:
Addressed the remaining second-round review findings in `73b9d859` after rebasing onto current `origin/dev`:

- Added effective upstream headers to the cache signature using deterministic, case-insensitive normalization. This prevents cross-tenant/session cache collisions when dynamic headers differ; header values are only hashed and are not stored in the cache key.
- Replaced newline-delimited SSE frame storage with a length-prefixed binary encoding. Multiline/pretty-printed JSON events now round-trip safely, and malformed/truncated entries are rejected rather than partially replayed.
- Cache hits still preserve the original `usage` in the downstream response, but no longer publish it as a fresh upstream request in internal provider accounting.
- Added regression tests for header isolation/determinism, multiline SSE events, and truncated cached streams.

Verification after rebase:
- `go test ./...` ✅
- `go build -o test-output ./cmd/server` ✅
- `git diff --check` ✅

kantar0:
Addressed both new lifecycle findings in `9da740b1`.

I agreed with both comments, but fixed them by removing the process-global cache registry rather than adding global pruning bookkeeping:

- Each `OpenAICompatExecutor` now owns a concurrency-safe `ResponseCacheRegistry`.
- Concrete provider configurations are keyed by executor provider ID plus `config_index`, so duplicate display names with different cache profiles no longer replace each other.
- When configuration reload replaces the executor, the old registry becomes unreachable and its cached payloads are eligible for GC. Removed or renamed providers therefore cannot accumulate globally for the process lifetime.
- Cache contents are intentionally not preserved across config reloads; bounded lifecycle and payload retirement are preferable to retaining stale responses.
- Added regression coverage for duplicate-name provider configurations and retained existing profile-change behavior tests.

Verification:
- `go test ./...` ✅
- `go build -o test-output ./cmd/server` ✅
- `git diff --check` ✅

kantar0:
Addressed the stream-memory finding in `d6e0d8da`.

The review was correct: limiting only payload bytes did not account for per-frame collection overhead or the second allocation performed during final encoding.

The stream path now builds the final uint64 length-prefixed cache representation incrementally:

- Before appending each frame, it checks the complete encoded size (`current + 8-byte prefix + frame`) against `max-entry-bytes`.
- If the next frame would exceed the limit, the buffer is discarded and caching is disabled for that stream.
- There is no intermediate `[]string` collection and no second full-size encoding allocation.
- `Store` receives the already encoded, bounded buffer.

Added tests for exact-bound acceptance and prefix-aware overflow rejection.

Verification:
- `go test ./...` ✅
- `go build -o test-output ./cmd/server` ✅
- `git diff --check` ✅

Issue #3663, opened 2026-06-01
Sanitizer leaves signed thinking blocks with empty text intact, causing Anthropic    400 "thinking blocks cannot be modified"


norandom:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[  ] No, it's another issue.

**If it's a request payload issue, you MUST know**
Our team doesn't have any GODs or ORACLEs or MIND READERs. Please make sure to attach the request log or curl payload.

See below. I have a fix.

**Describe the bug**

Summary                                                                                 

When a client round-trips an assistant message and strips the thinking text but keeps the signature field, the Claude-upstream sanitizer (SanitizeClaudeMessagesForClaudeUpstream) preserves the block unchanged. Anthropic then rejects the request with:        
                                                  
`400 invalid_request_error: messages.X.content.Y: `thinking` or `redacted_thinking` `

blocks in the latest assistant message cannot be modified. These blocks must remain as  they were in the original response.   


**CLI Type**
What type of CLI account do you use?  (gemini-cli, gemini, codex, claude code or openai-compatibility)

`opencode` with Anthropic Opus 4.8 Thinking

**Model Name**

Opus 4.8 Thinking

**LLM Client**

`opencode` 

**Request Information**

Repro                                                                                   
                                                                                      
Send /v1/messages to a Claude upstream with an assistant turn containing:             
  
{"type": "thinking", "thinking": "", "signature": "<valid-looking Claude E-prefix signature ~1KB>"}                                                                      

Observed with opencode 2.x using @ai-sdk/anthropic against claude-opus-4-8(xhigh) after a tool-use turn.                                                                        
                                                                                     
                                                                       
                 
 In internal/signature/claude_messages_sanitize.go::SanitizeClaudeMessagesSignaturesForTarget, the loop checks isEmptyClaudeThinkingPlaceholder (which requires both empty signature AND empty text), then jumps to DecideSignatureCompatibility which sees a  valid-shaped Claude signature and returns SignatureActionPreserve. The asymmetric case -  signature present + text empty — is never caught, so the broken block is forwarded as-is. Anthropic's server-side signature validation then fails because the signature does not validate against empty text.


**Expected behavior**

```
Patch (against 3a54fb7f):                                                               
  rawSignature := part.Get("signature").String()
  // A signed thinking block whose text has been stripped (e.g. by a                      
[...]

Issue #3624, opened 2026-05-29
[Bug] claude-opus-4-8 multi-turn tool use fails with 400 "thinking blocks ... cannot be modified" — effort-2025-11-24 thinking blocks dropped by signature sanitizer


Torusioadapt:
### Summary
With `claude-opus-4-8` over the Claude OAuth upstream, multi-turn tool-use
conversations fail after several turns with:

> messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest
> assistant message cannot be modified. These blocks must remain as they were
> in the original response.

Single-shot requests work fine; the error only appears once the conversation
has accumulated assistant turns with thinking blocks and tool calls.

### Environment
- CLIProxyAPI: v7.1.30 (also reproduced on v7.1.27 and v7.1.29 — identical failure)
- Client: opencode 1.15.x (`@ai-sdk/anthropic`, native `/v1/messages`)
- Client header: `Anthropic-Beta: effort-2025-11-24`
- Model: `claude-opus-4-8`, Claude OAuth (Claude Code identity) upstream
- Routing: `round-robin`, `session-affinity: true`

### Reproduction
1. Use opencode with a `claude-opus-4-8` provider pointed at CLIProxyAPI.
2. Run an agentic task with several tool calls (interleaved thinking).
3. After ~15–35 messages the upstream returns HTTP 400 with the error above.

### What I verified it is NOT
I captured the failing request body and replayed it deterministically:
- **Not multi-account round-robin**: `session-affinity` was active and the whole
  conversation used a single OAuth account (confirmed in selector logs).
- **Not message restructuring**: captured the upstream `=== API REQUEST 1 ===`
  via `request-log` — it has the exact same message array as the client sent
  (same count, same per-message block counts). CPA does not merge/reorder.
- **Not a version regression**: v7.1.27 / v7.1.29 / v7.1.30 all fail identically.
- **Not the thinking param**: fails with `{type:adaptive}`, `{type:enabled,
  budget_tokens:N}`, and with the param removed.
- The client sends every `thinking` block with full text **and** a valid
  `signature`.

### The only thing that fixes it
Removing all `thinking` / `redacted_thinking` blocks from the message history
before forwarding makes the request succeed (thinking can stay enabled).

### Likely root cause
`internal/signature/claude_messages_sanitize.go`
(`SanitizeClaudeMessagesForClaudeUpstream`) runs strict signature validation on
[...]

PR #5318, opened 2026-08-28
fix(claude): merge adjacent thinking blocks in a replayed Responses conversation

### LINKED ISSUE #5317: [Bug] Two consecutive reasoning items become adjacent thinking blocks and permanently kill a Codex session on Claude

## Summary

A Codex session talking to a Claude model through the Responses endpoint can be killed **permanently and silently** by one ordinary turn. Once the conversation contains two consecutive `reasoning` items, every later request replays them as two adjacent `thinking` blocks and Anthropic rejects the whole request:

```
400 invalid_request_error
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as they were in the original response.
```

This is not a transient error. The offending pair is now permanent history, so the session cannot answer even a one-word prompt again. Codex runs compaction pre-sampling on every turn, so the user never gets a working turn back and never sees a recoverable state; the session is simply dead.

I lost two real sessions this way, worth 1,094,153 and 488,300 tokens.

## Root cause

Anthropic never emits two adjacent `thinking` blocks. When a replayed assistant message carries them, the server reads it as a modified block and rejects it.

An agent reaches that shape on a completely ordinary turn: it thinks, thinks again, then acts. The Responses history then holds two consecutive `reasoning` items, and `convertOpenAIResponsesRequestToClaude` maps each one to its own `thinking` block inside the same assistant message.

The same shape appears whenever reasoning is interleaved with tool calls, because the tool_use blocks are hoisted to the end of the message and the thinking blocks collapse together behind them.

## Measured reproduction

Probes sent to `/v1/responses` with real `reasoning` items harvested from a rollout file, model `claude-opus-5`:

| shape | result |
| --- | --- |
| user, reasoning, assistant, user | 200 |
| user, **reasoning, reasoning**, assistant, user | **400 cannot be modified** |
| user, reasoning, text, reasoning, assistant, user | 200 |
| user, **reasoning, tool_call, reasoning, tool_call**, tool_output, tool_output, user | **400 cannot be modified** |

Adjacency is the whole discriminator. Separate the two reasoning items with any other block and the identical request succeeds.

This is a different defect from #3663 and #3624. Those are the native `/v1/messages` signature sanitizer producing an empty or dropped block. This one is the Responses -> Claude request translator producing a block *count* and *ordering* Anthropic never emits, with every signature intact. Fixing the sanitizer does not fix this.

## Fix

Fold a run of adjacent `thinking` blocks in one assistant message into the first block of the run: concatenate the text, keep the first signature. Anthropic does not verify thinking text against the signature, which is what makes the merge lossless for the reasoning the model reads back. `redacted_thinking` carries opaque data that cannot be concatenated, so it is left untouched and only breaks a run.

Every shape that already returns 200 keeps returning 200, because a message with no adjacent pair is returned unchanged.

PR follows.

LilMGenius:
Fixes #5317

## Why this must land

One ordinary turn permanently bricks a Codex session on Claude. An agent that thinks twice before acting emits two consecutive `reasoning` items; the translator turns them into two adjacent `thinking` blocks; Anthropic rejects the request with `"thinking" or "redacted_thinking" blocks in the latest assistant message cannot be modified`. The pair is now permanent history, so **every** later request fails the same way. Codex runs compaction pre-sampling on every turn, so the session cannot answer even a one-word prompt and never surfaces a recoverable state.

Two of my real sessions died this way, worth 1,094,153 and 488,300 tokens.

## The change

`mergeAdjacentClaudeThinkingBlocks` folds a run of adjacent `thinking` blocks in one assistant message into the first block of the run: text concatenated with a blank line, first signature kept. Anthropic does not verify thinking text against the signature, so the merge is lossless for the reasoning the model reads back. `redacted_thinking` carries opaque data that cannot be concatenated, so it is left untouched and only breaks a run.

Called from `flushPendingMessage` for assistant messages, before the existing tool_use combine step.

## Blast radius

A message with no adjacent pair is returned unchanged, so every shape that already returns 200 keeps returning 200. Verified against Anthropic through a locally built binary:

| shape | before | after |
| --- | --- | --- |
| reasoning, reasoning, assistant, user | 400 | **200** |
| reasoning, tool_call, reasoning, tool_call, outputs, user | 400 | **200** |
| reasoning, assistant, user (control) | 200 | 200 |

## Tests

Two regressions added, both verified red before the fix:

- `TestConsecutiveReasoningItemsDoNotProduceAdjacentThinkingBlocks` expects `thinking,text`
- `TestReasoningBetweenToolCallsStaysReplayable` expects `thinking,tool_use,tool_use`

`go test ./internal/translator/claude/... -count=1` passes, `go build ./...` exits 0. `TestInterleavedThinkingAndSearchSurvivesRoundTrip` and `TestConvertOpenAIResponsesRequestToClaude_KeepsToolUseAdjacentToToolResult` still pass unchanged.

## Not a duplicate of #3663 / #3624

Those are the native `/v1/messages` signature sanitizer emptying or dropping a block. This is the Responses translator emitting a block ordering Anthropic never produces, with every signature intact.

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_request.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_request.go
@@ -180,11 +180,14 @@ func convertOpenAIResponsesRequestToClaude(modelName string, inputRawJSON []byte
 		}
 
 		parts := pendingParts
-		if pendingRole == "assistant" && len(pendingToolUseParts) > 0 {
-			combined := make([][]byte, 0, len(pendingParts)+len(pendingToolUseParts))
-			combined = append(combined, pendingParts...)
-			combined = append(combined, pendingToolUseParts...)
-			parts = combined
+		if pendingRole == "assistant" {
+			parts = mergeAdjacentClaudeThinkingBlocks(parts)
+			if len(pendingToolUseParts) > 0 {
+				combined := make([][]byte, 0, len(parts)+len(pendingToolUseParts))
+				combined = append(combined, parts...)
+				combined = append(combined, pendingToolUseParts...)
+				parts = combined
+			}
 		}
 		if len(parts) > 0 {
 			msg := []byte(`{"role":"","content":[]}`)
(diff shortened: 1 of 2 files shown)


LilMGenius:
Superseded by 07d8156 — thanks, @luispater.

Verified before closing: applying only this PR's tests to HEAD, the replayed assistant message now comes back as `thinking, text` with no adjacent pair, so the 400 is gone.

Your `appendReasoning` is the better call. Flushing `pendingToolUseParts` when a later reasoning item arrives promotes the tool call into a real separator, so `thinking, tool_use, thinking, tool_use` survives with the original order intact — my version merged that history away. Replacing the block outright also keeps each signature paired with its own text, where my concatenation kept the first block's signature over merged text and would break the day that pairing is checked.

Closing.

Issue #5279, opened 2026-08-26
[Bug] OpenAI Responses to Claude Fable forwards unsupported trailing assistant prefill


thianphat:
## Problem

A Codex or OpenCodex `/v1/responses` continuation can replay history whose final input item is a plain assistant message. The Responses to Claude translator converts that item into the final Anthropic `messages` entry unchanged. Claude Fable rejects this request with HTTP 400:

```json
{"error":{"type":"invalid_request_error","message":"This model does not support assistant message prefill. The conversation must end with a user message."}}
```

This occurred after a valid user continuation and many successful tool cycles. The last replayed item was an assistant progress message with no following tool call. Therefore the client request arriving at CLIProxyAPI ended with an assistant message, and the translator forwarded an unsupported prefill shape.

## Affected path

- Endpoint: `/v1/responses`
- Translator: `internal/translator/claude/openai/responses/claude_openai-responses_request.go`
- Model: `claude-fable-5`
- The non-Anthropic compatibility conversion should remain unchanged.

## Expected behavior

For Claude models that do not support assistant prefill, the translated Anthropic conversation must end with a user message. Existing behavior for models that support assistant prefill must not change.

## Proposed implementation

After `flushPendingMessage()` and the existing final-message sanitizers, add a model-gated helper that removes the final assistant message for Fable. If that leaves no conversational message, append the same minimal empty user message already used for system-only input. For example:

```go
func dropUnsupportedFableAssistantPrefill(modelName string, messages [][]byte) [][]byte {
    normalized := strings.ToLower(strings.TrimSpace(modelName))
    if !strings.Contains(normalized, "fable") || len(messages) == 0 {
        return messages
    }
    last := gjson.ParseBytes(messages[len(messages)-1])
    if !strings.EqualFold(strings.TrimSpace(last.Get("role").String()), "assistant") {
        return messages
    }
    return messages[:len(messages)-1]
}
```

Call this only in the Anthropic path, after trailing-thinking cleanup. Then ensure the result has a user message when the slice becomes empty. A capability flag would be preferable to a model-name check if the registry already exposes assistant-prefill support.

## Regression coverage

Add focused tests proving:

1. A Fable Responses input ending in a plain assistant `output_text` message is translated to a conversation ending in the preceding user message.
2. A Fable input already ending in a user or tool-result message is unchanged.
3. A non-Fable Claude model retains current assistant-prefill behavior.
4. The compatibility path remains unchanged.

[...]

PR #5320, opened 2026-08-28
fix(claude): drop trailing assistant prefill for every model that rejects it

### LINKED ISSUE #5319: [Bug] Trailing assistant prefill is only dropped for Fable, so claude-opus-5 and claude-sonnet-4-6 still 400

## Summary

Any Responses conversation whose history ends on an assistant turn fails on `claude-opus-5` and `claude-sonnet-4-6`:

```
400 invalid_request_error
This model does not support assistant message prefill. The conversation must end with a user message.
```

The translator already knows about this constraint. `dropUnsupportedFableAssistantPrefill` drops the trailing assistant message, but it matches only the string `fable`, so the two most-used Claude families still send the rejected shape.

This is fatal rather than annoying because Codex runs compaction pre-sampling on every turn. Once a session reaches this state, the user gets a 400 for every prompt, including a one-word one, and the session is dead with no recoverable state shown.

## Measured

Probes to `/v1/responses` with input `[user, assistant]`:

| model | result |
| --- | --- |
| claude-opus-5 | **400 prefill** |
| claude-sonnet-4-6 | **400 prefill** |
| claude-fable-5 | handled today (message dropped) |
| claude-haiku-4-5 | 200, prefill accepted |

Reasoning effort is not the discriminator. The same trailing shape on `claude-opus-5` returns 400 at effort `none`, `minimal` and `high` alike. It is a model-family capability.

## Why the trailing message can just be dropped

The Responses API has no prefill concept. A trailing assistant item there is replayed history, not an instruction to continue mid-sentence. Dropping it loses nothing the next request needs, which is exactly the reasoning the existing Fable branch already relies on.

#5279 fixed this for Fable only and is the incomplete predecessor.

#5094 reports the same error string, but for the Antigravity `compaction_trigger` marker on a different route, and proposes an executor-level fallback. This issue is narrower and orthogonal: the Claude Responses translator's model list is wrong regardless of how the trailing assistant turn got there.

## Fix

Generalize the Fable-only match into a measured model-family list. A new family is one entry.

PR follows.

LilMGenius:
Fixes #5319

## Why this must land

`dropUnsupportedFableAssistantPrefill` already encodes the right idea, but it matches only `fable`. `claude-opus-5` and `claude-sonnet-4-6` reject the identical shape and are not on the list, so every Responses conversation ending on an assistant turn returns:

```
This model does not support assistant message prefill. The conversation must end with a user message.
```

Codex runs compaction pre-sampling on every turn, so a session in this state answers nothing at all, not even a one-word prompt, and the failure looks permanent to the user.

## The change

- `dropUnsupportedFableAssistantPrefill` becomes `dropUnsupportedAssistantPrefill`, driven by `assistantPrefillRejectingModels`.
- The list holds the measured families: `fable`, `opus-5`, `sonnet-4-6` plus their unhyphenated spellings. Adding a family later is one entry.
- Behavior for Fable is byte-identical to before.

## Measured evidence in the doc comment

| model | trailing assistant message |
| --- | --- |
| claude-opus-5 | rejected |
| claude-sonnet-4-6 | rejected |
| claude-fable-5 | rejected (already handled) |
| claude-haiku-4-5 | accepted |

Reasoning effort is not the discriminator: `claude-opus-5` returns 400 at effort `none`, `minimal` and `high` alike.

## Blast radius

Verified against Anthropic through a locally built binary: `[user, assistant]` on `claude-opus-5` goes 400 -> **200**, and a conversation that already ends with a user turn is unaffected.

One existing test changed. `TestConvertOpenAIResponsesRequestToClaude_SystemLevelInputsBecomeSeparateSystemBlocks` incidentally used `claude-opus-5` while asserting system-block routing, and its fixture ends with an assistant message that Anthropic would now reject. It is pointed at `claude-haiku-4-5-20251001`, a prefill-accepting model, so the test keeps measuring exactly what it was written to measure.

## Tests

`TestTrailingAssistantPrefillDroppedForModelsThatRejectIt`, table-driven over fable / opus-5 / sonnet-4-6 (dropped) and haiku-4-5 (kept). Verified red before the fix.

`go test ./internal/translator/claude/... -count=1` passes, `go build ./...` exits 0.

## Scope

#5279 covered Fable only. #5094 reports the same error string on the Antigravity `compaction_trigger` route and needs a different, executor-level fix; this PR is orthogonal to it and correct regardless of how the trailing assistant turn arrived.

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_assistant_prefill_test.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_assistant_prefill_test.go
@@ -1,0 +1,36 @@
+package responses
+
+import "testing"
+
+// Anthropic reads a trailing assistant message as prefill, and the current
+// Claude generation rejects it: "This model does not support assistant message
+// prefill. The conversation must end with a user message." The Responses API
+// has no prefill concept, so a trailing assistant item is replayed history,
+// not a directive. Measured against Anthropic: claude-opus-5, claude-sonnet-4-6
+// and claude-fable-5 reject it, claude-haiku-4-5 accepts it. Only the Fable
+// name was handled, so every Codex compaction turn on Opus and Sonnet failed.
+func TestTrailingAssistantPrefillDroppedForModelsThatRejectIt(t *testing.T) {
+	raw := responsesRequestFromItems(
+		`{"type":"message","role":"user","content":[{"type":"input_text","text":"hi"}]}`,
+		`{"type":"message","role":"assistant","content":[{"type":"output_text","text":"answer"}]}`,
+	)
+
+	for _, tc := range []struct {
+		model       string
+		wantDropped bool
+	}{
+		{"claude-fable-5", true},
+		{"claude-opus-5", true},
+		{"claude-sonnet-4-6", true},
+		{"claude-haiku-4-5-20251001", false},
+	} {
+		out := ConvertOpenAIResponsesRequestToClaude(tc.model, raw, false)
+		last := lastClaudeMessageRole(t, out)
+		if tc.wantDropped && last == "assistant" {
+			t.Fatalf("%s: conversation still ends with an assistant message. Output: %s", tc.model, string(out))
+		}
+		if !tc.wantDropped && last != "assistant" {
+			t.Fatalf("%s: prefill was dropped for a model that accepts it. Output: %s", tc.model, string(out))
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


LilMGenius:
Superseded by 6f25b9a — thanks, @luispater.

Verified before closing: this PR's test passes unmodified against HEAD, including the `claude-haiku-4-5` control that must *keep* its trailing assistant message, so the widening did not become a blanket strip.

`claudeModelRejectsAssistantPrefill` is the shape I was reaching for, and naming it off the family list reads better than the alias soup I had — I was carrying `opus5` and `sonnet4-6` spellings on speculation rather than on anything I had actually seen arrive.

Closing.

PR #5322, opened 2026-08-28
fix(claude): drop a trailing thinking block from the final assistant message

### LINKED ISSUE #5321: [Bug] A history ending on a reasoning item sends a trailing thinking block, which Anthropic rejects

## Summary

A Responses conversation whose history ends on a `reasoning` item becomes an assistant message whose last content block is `thinking`. Anthropic rejects it:

```
400 invalid_request_error
messages.N: The final block in an assistant message cannot be `thinking`.
```

This is reachable on a model that accepts assistant prefill, so it is a separate defect from the prefill list. Measured on `claude-haiku-4-5`, which happily accepts a trailing assistant *text* message but rejects a trailing assistant *thinking* message.

## Measured

Probes to `/v1/responses`, model `claude-haiku-4-5-20251001`:

| input | result |
| --- | --- |
| user, assistant(text) | 200 |
| user, assistant(text), reasoning | 200 |
| user, reasoning, user | 200 |
| **user, reasoning** | **400 final block cannot be thinking** |

Only the trailing position matters. A thinking block in an earlier turn is the replayed chain of thought and is accepted.

## Why the block can be dropped

A `thinking` block cannot act as prefill. It carries nothing the next request needs, and Anthropic will not let it stand as the final block under any circumstances. Removing it from the trailing message only is lossless, and the earlier turns keep their full reasoning history.

## Fix

Strip trailing `thinking` and `redacted_thinking` blocks from the final assistant message, and drop the message entirely if nothing else remains.

PR follows.

LilMGenius:
Fixes #5321

## Why this must land

When a replayed Responses history ends on a `reasoning` item, the translator emits an assistant message whose last block is `thinking`, and Anthropic returns:

```
messages.N: The final block in an assistant message cannot be `thinking`.
```

This fires on models that accept assistant prefill, so it survives the prefill fix and needs its own handling. Measured on `claude-haiku-4-5`: a trailing assistant *text* message is accepted, a trailing assistant *thinking* message is not.

## The change

`dropTrailingThinkingBlocks` removes `thinking` and `redacted_thinking` blocks from the end of the **trailing** assistant message, and drops the message entirely when no other block remains.

A thinking block cannot serve as prefill, so it carries nothing the request needs. Earlier turns are untouched, so the replayed chain of thought is preserved in full.

## Blast radius

Only the last message is examined, and only when it is an assistant message ending in a thinking block. Every other shape is returned byte-identical. Verified against Anthropic through a locally built binary:

| shape | model | before | after |
| --- | --- | --- | --- |
| user, reasoning | haiku-4-5 | 400 | **200** |
| user, assistant, reasoning | opus-5 | 400 | **200** |
| user, reasoning, user (control) | opus-5 | 200 | 200 |

## Tests

`TestTrailingThinkingOnlyAssistantMessageIsNotSent`, verified red before the fix with the exact rejected payload in the failure message.

`go test ./internal/translator/claude/... -count=1` passes, `go build ./...` exits 0.

### DIFF
--- a/internal/translator/claude/openai/responses/claude_openai-responses_request.go
+++ b/internal/translator/claude/openai/responses/claude_openai-responses_request.go
@@ -453,6 +453,7 @@ func convertOpenAIResponsesRequestToClaude(modelName string, inputRawJSON []byte
 	hadMessages := len(messageBlocks) > 0
 	if !preserveEmptyThinkingBlocks {
 		messageBlocks = dropUnsupportedFableAssistantPrefill(modelName, messageBlocks)
+		messageBlocks = dropTrailingThinkingBlocks(messageBlocks)
 	}
 	// Preserve a minimal conversational turn for system-only inputs or when messages became empty
 	// so downstream validation still sees a Claude-shaped request.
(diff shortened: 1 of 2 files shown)


LilMGenius:
Superseded by c350d3f — thanks, @luispater.

Verified before closing: this PR's test passes unmodified against HEAD.

Two things yours does better. Running `stripTrailingClaudeThinkingBlocks` *before* the prefill drop is the safer order — mine ran after, so on a prefill-rejecting model the strip was already looking at a message that had been removed. And folding a lone surviving `text` block back to the string form (guarded on `cache_control`/`citations`) keeps the output shape consistent with the rest of the translator, which mine did not bother with.

Closing.

PR #5326, opened 2026-08-28
fix(codex): always emit cache_creation_input_tokens in Claude usage

f4llenz:
## Problem

The Claude usage object built from a Codex Responses reply never sets
`cache_creation_input_tokens`, so the field is simply absent on every translated gpt
response. Anthropic's own API always includes it, so any consumer written against the shape
this translator imitates reads a missing key and has to guess whether that means "zero" or
"unknown". Either guess corrupts a spend estimate, in opposite directions.

## Why there is no value to map

**No mapping exists, and that is a property of the API rather than a gap in the translator.**
The Responses `usage` object carries `input_tokens`, `input_tokens_details.cached_tokens`,
`output_tokens` and `output_tokens_details.reasoning_tokens` — and no cache-write counter,
because OpenAI prompt caching is implicit and carries no write premium. Nothing upstream
separates "input tokens written to cache this turn" from "input tokens that missed".

Deriving the value from the uncached remainder would be wrong twice over: those tokens are
already counted in `input_tokens`, so any consumer summing the fields double-counts them, and
they carry no write premium the way Anthropic's `cache_creation_input_tokens` does.

## Change

Both emission sites — the streaming `message_delta` and the non-streaming message — now write
an explicit `0`, and `extractResponsesUsage` documents why the field is *unmappable* rather
than unimplemented. A reader gets a known zero instead of an absent key.

Consumers should read `cache_creation_input_tokens: 0` on this path as "this API cannot report
it", not as "no cache was written", and price cache-write cost from
`input_tokens` minus `cache_read_input_tokens` instead. That is what the added comment says,
so the constraint travels with the code.

## Compatibility

Behaviorally inert for the proxy's own accounting: `usage_helpers.go` already read the absent
field as `0` via gjson `.Int()`. The change is additive to the emitted JSON and moves the
translated shape closer to the Anthropic format it imitates.

## Verification

`go build ./...`, `go vet ./internal/...`, `go test ./...` and `gofmt -l internal/` are all
clean on this branch against current `main`.

### DIFF
--- a/internal/translator/codex/claude/codex_claude_response.go
+++ b/internal/translator/codex/claude/codex_claude_response.go
@@ -160,6 +160,9 @@ func ConvertCodexResponseToClaude(_ context.Context, _ string, originalRequestRa
 		inputTokens, outputTokens, cachedTokens := extractResponsesUsage(responseData.Get("usage"))
 		template, _ = sjson.SetBytes(template, "usage.input_tokens", inputTokens)
 		template, _ = sjson.SetBytes(template, "usage.output_tokens", outputTokens)
+		// Always emitted, always zero: the Responses API reports no cache-write counter.
+		// See extractResponsesUsage for why this is unmappable rather than unimplemented.
+		template, _ = sjson.SetBytes(template, "usage.cache_creation_input_tokens", 0)
 		if cachedTokens > 0 {
 			template, _ = sjson.SetBytes(template, "usage.cache_read_input_tokens", cachedTokens)
 		}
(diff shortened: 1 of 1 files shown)


f4llenz:
Closing this: `dev` now maps the cache-write counter directly (`input_tokens_details.cache_write_tokens` / `cache_creation_tokens`, 893abbab and #5956), so the field is populated whenever upstream reports it. That makes this PR's "unmappable" premise obsolete. Thanks for the fix, @luispater @sususu98.

PR #5337, opened 2026-08-29
feat(executor): add provider request rate limiting

kantar0:
## Summary

Add an opt-in, per-credential request rate limiter for OpenAI-compatible upstream providers.

## Motivation

Some relay providers enforce strict requests-per-minute windows and count rejected attempts against the same quota. Bursty local traffic can therefore trigger a self-sustaining sequence of HTTP 429 responses even when average throughput is below the advertised limit.

## Changes

- Add `openai-compatibility[].requests-per-minute` (`0`/omitted disables limiting).
- Smooth requests with `golang.org/x/time/rate` using burst 1.
- Keep independent state per concrete provider configuration and credential.
- Apply the gate before opening upstream connections for chat, streaming chat, image, and streaming image requests.
- Observe HTTP 429 responses and pause the same credential for `Retry-After` (seconds or HTTP-date), with a one-minute fallback when absent/invalid.
- Scope limiter state to the executor lifecycle so hot reloads retire removed provider state.
- Expose the field through management API GET/PATCH; reject negative values.

## Safety

- Disabled by default; existing configurations are unaffected.
- Waiting is context-cancellable.
- No timeout or deadline is introduced after the upstream connection is established.
- API keys and request bodies are not retained by limiter state.

## Verification

- `go test ./...`
- `git diff --check`

All pass locally.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -614,6 +614,8 @@ nonstream-keepalive-interval: 0
 #     prefix: "test" # optional: require calls like "test/kimi-k2" to target this provider's credentials
 #     base-url: "https://openrouter.ai/api/v1" # The base URL of the provider.
 #     support-prompt-cache-key: false # optional: derive prompt_cache_key for requests from all input protocols
+#     requests-per-minute: 0 # optional: smooth requests to this provider; 0 disables limiting
+#     # An upstream 429 also pauses the same credential until Retry-After, or for one minute when absent.
 #     disable-cooling: false # optional provider override: true disables cooling, false enables it; omit to inherit global
 #     request-retry: 3 # optional per-provider override; 0 disables additional rounds; omit or set < 0 to inherit global
 #     request-scoped-errors: # optional: custom rules to classify upstream errors by status and body patterns
--- a/go.mod
+++ b/go.mod
@@ -35,6 +35,7 @@ require (
 	golang.org/x/oauth2 v0.30.0
 	golang.org/x/sync v0.22.0
 	golang.org/x/sys v0.47.0
+	golang.org/x/time v0.14.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	gopkg.in/yaml.v3 v3.0.1
 )
@@ -55,7 +56,6 @@ require (
 	github.com/rogpeppe/go-internal v1.15.0 // indirect
 	github.com/wlynxg/anet v0.0.5 // indirect
 	go.uber.org/atomic v1.11.0 // indirect
-	golang.org/x/time v0.14.0 // indirect
 )
 
 require (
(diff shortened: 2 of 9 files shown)


kantar0:
Addressed the queued-request race in `b9a84d4f`.

The review was correct: callers already queued on the token bucket could retain an old `blockedUntil` snapshot and proceed after a later request published a 429 pause.

`Wait` now:
1. waits for a limiter token;
2. re-reads the shared `blockedUntil` under the registry mutex;
3. waits when a newer pause exists;
4. loops through the limiter and checks again before release.

Added a regression test that queues a caller first, publishes `Retry-After` afterward, and verifies the queued call remains blocked.

Verification: `go test ./...`, required server build, and `git diff --check` all pass.

kantar0:
Addressed both new findings in `84bfccdf`.

- `HttpRequest` (used by raw OpenAI-compatible routes such as alpha search) now uses the same concrete provider/config/credential limiter key, waits before `Do`, and observes upstream 429 headers.
- `Retry-After` parsing now returns both delay and validity. Legal `Retry-After: 0` and elapsed valid HTTP dates mean immediate retry; the one-minute fallback applies only to absent, malformed, or negative values.

Added regression coverage for valid zero versus missing Retry-After.

Verification: full tests, required server build, and diff check pass.

kantar0:
Addressed in `b16199fa`.

`PutOpenAICompat` now validates `requests-per-minute` on every normalized entry before replacing/persisting the list. Negative values return HTTP 400 with the indexed field name, matching PATCH semantics (`0` remains the supported disabled value).

Added a PUT regression test for `requests-per-minute: -1`.

Full tests, required server build, and diff check pass.

DatUET:
[kantar0](https://github.com/kantar0) Can you resolve the conflict? I think your approach could help many users right now. Thank you.

kantar0:
Hi @DatUET — thanks for the nudge! The PR is rebased on the current `dev` and all conflicts are resolved. Unfortunately my GitHub token lacks the `workflow` scope so the force-push is rejected by GitHub even though none of my commits touch `.github/workflows`.

The two conflict resolutions are minimal and easy to apply manually if you'd like to do the merge from your side:

**`go.mod`** — move `golang.org/x/time` from `indirect` to the main `require` block (it is now a direct dependency of the rate limiter):

```diff
 	golang.org/x/text v0.40.0
+	golang.org/x/time v0.14.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
```

and remove it from the `indirect` block:

```diff
-	golang.org/x/time v0.14.0 // indirect
```

**`internal/api/handlers/management/config_auth_index.go`** — add `RequestsPerMinute` back to the `openAICompatibilityWithAuthIndex` struct (removed in the upstream rebase) in two places:

```diff
 	SupportPromptCacheKey bool                                     `json:"support-prompt-cache-key,omitempty"`
+	RequestsPerMinute     int                                      `json:"requests-per-minute,omitempty"`
 	DisableCooling        *bool                                    `json:"disable-cooling,omitempty"`
```

and in the mapping function:

```diff
 		SupportPromptCacheKey: entry.SupportPromptCacheKey,
+		RequestsPerMinute:     entry.RequestsPerMinute,
 		DisableCooling:        entry.DisableCooling,
```

Everything else merges cleanly. If you can rebase the branch from your side or grant the `workflow` scope, I'm happy to push the resolved state directly.

PR #5338, opened 2026-08-29
fix(executor): recover truncated streamed tool calls

kantar0:
## Summary

Add an opt-in workaround for OpenAI-compatible upstreams that truncate tool-call arguments in streaming mode.

## Problem

Some aggregator gateways intermittently emit a tool name with an empty `function.arguments` string and immediately finish the stream with `tool_calls`. The identical request sent with `stream: false` returns complete arguments, leaving streaming agent clients with an unusable call.

## Changes

- Add `openai-compatibility[].non-stream-tool-calls` (disabled by default).
- When enabled and a streaming request declares non-empty `tools`:
  1. Send the translated upstream request with `stream: false` and remove `stream_options`.
  2. Validate the complete JSON response.
  3. Synthesize equivalent OpenAI `chat.completion.chunk` frames.
  4. Feed those frames through the existing stream translation pipeline.
- Requests without tools preserve native upstream streaming.
- Preserve dynamic custom headers and upstream response headers.
- Expose the option through management API GET/PATCH.

Synthesized frames preserve role, content, `reasoning_content`, complete tool calls, finish reason, usage (once), and `[DONE]`. The normal translation pipeline provides downstream OpenAI, Anthropic Messages, and Responses API compatibility.

## Safety

- Opt-in only; no behavior change by default.
- No timeout or deadline is introduced after upstream connection establishment.
- Invalid/non-convertible upstream bodies return HTTP 502 rather than producing a partial tool call.
- Request cancellation stops downstream frame delivery.

## Verification

- `go test ./...`
- `git diff --check`

All pass locally.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -614,6 +614,9 @@ nonstream-keepalive-interval: 0
 #     prefix: "test" # optional: require calls like "test/kimi-k2" to target this provider's credentials
 #     base-url: "https://openrouter.ai/api/v1" # The base URL of the provider.
 #     support-prompt-cache-key: false # optional: derive prompt_cache_key for requests from all input protocols
+#     # Workaround for upstreams that truncate streamed tool-call arguments. Tool-bearing
+#     # requests are sent non-streamed and converted back to downstream SSE locally.
+#     non-stream-tool-calls: false
 #     disable-cooling: false # optional provider override: true disables cooling, false enables it; omit to inherit global
 #     request-retry: 3 # optional per-provider override; 0 disables additional rounds; omit or set < 0 to inherit global
 #     request-scoped-errors: # optional: custom rules to classify upstream errors by status and body patterns
--- a/internal/api/handlers/management/config_auth_index.go
+++ b/internal/api/handlers/management/config_auth_index.go
@@ -48,6 +48,7 @@ type openAICompatibilityWithAuthIndex struct {
 	Models                []config.OpenAICompatibilityModel        `json:"models,omitempty"`
 	Headers               map[string]string                        `json:"headers,omitempty"`
 	SupportPromptCacheKey bool                                     `json:"support-prompt-cache-key,omitempty"`
+	NonStreamToolCalls    bool                                     `json:"non-stream-tool-calls,omitempty"`
 	DisableCooling        *bool                                    `json:"disable-cooling,omitempty"`
 	RequestRetry          *int                                     `json:"request-retry,omitempty"`
 	RequestScopedErrors   []config.RequestScopedErrorRule          `json:"request-scoped-errors,omitempty"`
@@ -309,6 +310,7 @@ func (h *Handler) openAICompatibilityWithAuthIndex() []openAICompatibilityWithAu
 			Models:                entry.Models,
 			Headers:               entry.Headers,
 			SupportPromptCacheKey: entry.SupportPromptCacheKey,
+			NonStreamToolCalls:    entry.NonStreamToolCalls,
 			DisableCooling:        entry.DisableCooling,
 			RequestRetry:          entry.RequestRetry,
 			RequestScopedErrors:   entry.RequestScopedErrors,
--- a/internal/api/handlers/management/config_lists.go
+++ b/internal/api/handlers/management/config_lists.go
@@ -801,6 +801,7 @@ func (h *Handler) PatchOpenAICompat(c *gin.Context) {
 		Models                *[]config.OpenAICompatibilityModel  `json:"models"`
 		Headers               *map[string]string                  `json:"headers"`
 		SupportPromptCacheKey *bool                               `json:"support-prompt-cache-key"`
+		NonStreamToolCalls    *bool                               `json:"non-stream-tool-calls"`
 		RequestRetry          *int                                `json:"request-retry"`
 		RequestScopedErrors   *[]config.RequestScopedErrorRule    `json:"request-scoped-errors"`
 	}
@@ -877,6 +878,9 @@ func (h *Handler) PatchOpenAICompat(c *gin.Context) {
 	}
 	if body.Value.SupportPromptCacheKey != nil {
 		entry.SupportPromptCacheKey = *body.Value.SupportPromptCacheKey
+	}
+	if body.Value.NonStreamToolCalls != nil {
+		entry.NonStreamToolCalls = *body.Value.NonStreamToolCalls
 	}
 	if body.Value.RequestScopedErrors != nil {
 		entry.RequestScopedErrors = append([]config.RequestScopedErrorRule(nil), *body.Value.RequestScopedErrors...)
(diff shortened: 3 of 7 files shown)


kantar0:
Addressed both findings in `7b017905`.

- The blocking upstream `Do`/`io.ReadAll` now runs inside the stream producer goroutine. `ExecuteStream` returns its `StreamResult` immediately, allowing the HTTP streaming handler to bootstrap the response and emit configured keep-alives while the non-stream upstream is generating.
- Upstream connection/read/status failures are delivered as `StreamChunk.Err` and recorded through failure accounting.
- Once the complete upstream response is available, usage publication is deferred around synthesized downstream delivery. A downstream cancellation therefore still records the provider-billed request and tokens.

The tradeoff is that upstream response headers cannot be attached synchronously before the producer connects; payload/tool-call behavior and accounting remain correct.

Verification: `go test ./...`, required server build, and `git diff --check` all pass.

kantar0:
Addressed all three new findings in `920f1538`.

- Upstream `Do` and status handling remain synchronous, preserving bootstrap retries and making upstream headers available in `StreamResult` for passthrough.
- After a successful status, the producer immediately translates and emits one valid assistant-role bootstrap chunk before reading the full non-stream body. This releases `readStreamBootstrap`, allowing the handler's normal streaming keep-alives to run during buffered generation. The later synthesized frames omit the duplicate role opening.
- Non-streaming `message.refusal` is now preserved as a streaming `delta.refusal`, including replies with null/empty content.
- Deferred usage finalization from the prior fix remains in place.

This keeps the workaround local to the executor and existing translation pipeline; no central conductor or translator code was changed.

Verification: full tests, required server build, and diff check pass.

kantar0:
Addressed in `ed6481f5`.

The bootstrap now receives a generated stable `chatcmpl-<uuid>` ID and Unix creation timestamp. After the buffered upstream response arrives, those same values are applied to the body before synthesizing all remaining chunks. This keeps one consistent stream identity from the initial role chunk through terminal events and prevents `/v1/responses` translation from initializing with an empty ID or `created_at: 0`.

Added assertions that bootstrap frames include non-empty/stable `id`, `model`, and `created` metadata.

Full tests, required server build, and diff check pass.

kantar0:
Addressed both findings in `334b0415`.

- Removed the pre-body semantic bootstrap. The producer now fully reads and validates the HTTP 200 body with `SynthesizeOpenAIStreamFrames` before emitting its first payload. A malformed response or missing usable `choices` therefore reaches `readStreamBootstrap` as an error and preserves credential/model fallback instead of prematurely committing HTTP 200.
- The synthesizer now emits `delta.role: assistant` at the start of every returned choice, so `n > 1` streams introduce choices 0..N correctly.
- Since metadata is now available before the first frame, the temporary generated bootstrap ID/timestamp override was removed and upstream `id`/`created` are retained.

The current stream contract cannot simultaneously send a downstream pre-bootstrap heartbeat and preserve fallback: `readStreamBootstrap` treats any non-empty payload as successful selection, while the HTTP handler only starts keep-alives after `ExecuteStream` returns. Solving that requires a separate central conductor/handler contract change rather than a semantic fake chunk in this executor-scoped workaround.

Full tests, required server build, and diff check pass.

kantar0:
Addressed in `7faff30e`.

The synthesizer now preserves a non-null top-level `service_tier` on every generated chat-completion chunk. This keeps the provider's actual tier visible to downstream OpenAI clients and allows `StreamUsageBuffer.ObserveOpenAIStream` to retain it in usage accounting.

Added a regression test using `service_tier: "priority"` that verifies every synthesized non-`[DONE]` frame carries the tier.

Full tests, required server build, and diff check pass.

kantar0:
Addressed all three findings in `e241d79f`.

- Preserves a non-null top-level `system_fingerprint` on every synthesized chunk.
- Preserves `choices[].logprobs` on the corresponding content chunk; if the response has logprobs without content, emits one standalone choice chunk with an empty delta and the logprobs instead of dropping them.
- Keeps `reasoning_content` as the preferred field and falls back to `message.reasoning` when it is absent, preserving the alternate field name for the existing Responses converters.

Added regression coverage combining alternate reasoning, token log probabilities, and system fingerprint metadata.

Full tests, required server build, and diff check pass.

kantar0:
Addressed both findings in `a4e84bea`.

- Synthesized tool calls now start from the upstream call object, so provider metadata such as Gemini's `tool_calls[].extra_content.google.thought_signature` survives the rewrite; only `index`, `id`, `type`, and the complete `function.name`/`function.arguments` are overridden.
- Each choice is validated before any payload is emitted. An empty `choices` array, a missing/non-object message, or a placeholder entry such as `{"choices":[{}]}` now returns no frames, so the executor reports a bootstrap error and credential/model fallback still applies. A genuinely empty assistant reply is still accepted when the upstream provides a terminal `finish_reason`.

Added regression tests for the thought signature passthrough, rejection of unusable and mixed choice arrays, and acceptance of an empty reply with a terminal reason.

Full tests, required server build, and diff check pass.

kantar0:
Addressed both findings in `e5770a89`.

- Choice validation now inspects the nested calls: a reply whose `tool_calls` contains a non-object entry or a call without a usable `function.name` produces no frames, so the executor raises a bootstrap error and credential/model fallback still applies instead of fabricating an empty function call.
- Content deltas now carry `message.annotations` (for example URL citation offsets) when present, matching the native OpenAI passthrough behavior for Chat Completions clients.

Added regression tests for `tool_calls:[{}]`, an unnamed function call, and annotation preservation alongside content.

Full tests, required server build, and diff check pass.

kantar0:
Addressed in `3ed6799a`.

A non-null `choices[].message.audio` object is now replayed as a `delta.audio` chunk, so audio data, transcript, and identifiers reach Chat Completions clients even when `message.content` is null. The audio chunk also counts as emitted content, so a logprobs-only frame is not produced alongside it.

Audio is additionally treated as usable assistant output during choice validation, independent of the finish reason.

Added a regression test asserting the audio object, transcript, and data survive a content-null reply.

Full tests, required server build, and diff check pass.

kantar0:
Addressed both findings in `f03c44ca`.

- The body must now pass `gjson.ValidBytes` before any synthesis. A truncated document or one with trailing garbage produces no frames, so the executor reports a bootstrap error and credential/model fallback still applies instead of replaying a partial document.
- Tool call validation now also requires `function.arguments` to be non-empty and valid JSON. Empty or partial arguments are exactly the upstream truncation this option exists to recover from, so such a reply now triggers fallback rather than being replayed as an unusable call.

Added regression tests for a truncated body, trailing garbage, empty arguments, and malformed arguments.

Full tests, required server build, and diff check pass.

ArshansGithub:
bump, unable to use GLM 5.3 using Cloudflare responses API due to this tool calling behavior bug.

ArshansGithub:
Validated on a `dev` + this PR build: `cf-glm53` streamed `tool_choice` call now emits `input_json_delta` with `{"city":"Paris"}`; before, `input:{}`. Carried on the fork branch `feat/claude-code-cache-tooling` as a dependency. Note the fix needs `non-stream-tool-calls: true` on the CF provider entry, which the config example should say.

PR #5351, opened 2026-08-30
feat(management): expose the routing prefix in the auth-files listing

gonzalonicolasr:
`GET /v0/management/auth-files` reports everything needed to identify a
credential except the one field that determines how a request reaches it.

With `force-model-prefix` enabled, `personal/claude-opus-5` and
`ag/gemini-3.5-flash` are answered by different credentials. The console has no
way to show which card owns which prefix, because the only way to read it today
is `GET /auth-files/:name` and parsing the raw credential — which pulls the
access and refresh tokens into the browser. That is fine for a details sheet the
user opened deliberately; it is not fine for a listing that renders every
credential on page load.

Both branches of the handler are covered: the in-memory one reads `auth.Prefix`,
the directory-scan fallback reads `prefix` from the file with `gjson`, mirroring
how `project_id` is already handled a few lines above. The field is omitted when
empty, so existing consumers see no change.

The console side is a separate PR against
`router-for-me/Cli-Proxy-API-Management-Center`, and it degrades cleanly on
servers that do not send the field.

## Testing

- `gofmt` clean, `go build ./cmd/server` OK, `go test ./internal/api/handlers/management/...` passes.
- Built from `v7.2.142` with only this change and run against a copy of a real
  auth dir: the listing reports `prefix` for the credentials that set one and
  omits it for the rest, and `GET /v1/models` returns a byte-identical model list
  (83 models across 5 prefixes), confirming routing is untouched.

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -251,6 +251,9 @@ func (h *Handler) listAuthFilesFromDisk(c *gin.Context) {
 				if projectID := strings.TrimSpace(gjson.GetBytes(data, "project_id").String()); projectID != "" {
 					fileData["project_id"] = projectID
 				}
+				if prefix := strings.TrimSpace(gjson.GetBytes(data, "prefix").String()); prefix != "" {
+					fileData["prefix"] = prefix
+				}
 				if pv := gjson.GetBytes(data, "priority"); pv.Exists() {
 					switch pv.Type {
 					case gjson.Number:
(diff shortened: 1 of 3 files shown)


gonzalonicolasr:
Good catch — confirmed and fixed in fe1bdf9.

The disk fallback trimmed whitespace only, while `synthesizeFileAuths` also
strips surrounding slashes and rejects an interior one. So `/team/` was reported
as `/team/` although requests reach it as `team/...`, and `team/sub` was reported
despite not being routable by prefix at all. For a field whose whole purpose is
to tell the user how a credential is addressed, describing a route that does not
exist is the worst possible failure.

Rather than copy the four lines, I extracted the rule into
`synthesizer.NormalizeCredentialPrefix` and call it from both places, so the
reported value cannot drift from the routing it describes. `prefix_test.go`
pins the behaviour, including the two cases above.

The in-memory branch needed no change: it reads `auth.Prefix`, which synthesis
had already normalized.

PR #5381, opened 2026-09-01
fix(pluginhost): preserve Prefix and Disabled in host.auth.save upsert

viccom:
## Problem

When a plugin persists credential JSON through the `host.auth.save` RPC, the host rebuilds the in-memory `Auth` from that payload via `buildAuthFromFileData` and replaces the existing record via `upsertAuthRecord` → `Manager.Update` (whole-record replacement).

The rebuilt auth **ignored the file's `prefix` and `disabled` metadata**: `Prefix` stayed empty and `Status` was hardcoded to active. This produces two concrete failures for any plugin that rewrites its auth files on a schedule:

1. **Model prefix loss** — every scheduled rewrite (daily check-in at 09:00, token keepalive at 22:00, credit lifecycle reconcile) cleared `auth.Prefix`, so `applyModelPrefixes` stopped emitting `<prefix>/<model>` IDs and `/v1/models` silently fell back to bare model IDs until the next host restart (the watcher's re-synthesis goes through the plugin's `ParseAuth`, which restores `Prefix`).
2. **Disabled flag flip** — an auth file with `disabled: true` rewritten through `host.auth.save` came back as an active in-memory auth.

We hit (1) in production with a plugin whose daily check-in rewrites every auth file: all provider-prefixed model IDs disappeared every morning.

## Fix

- `buildAuthFromFileData`: parse `metadata["prefix"]` and `metadata["disabled"]` exactly the way the watcher's `synthesizer.SynthesizeAuthFile` already does (same trimming/validation rules), and carry both into the rebuilt `Auth` (`Prefix`, `Status`, `Disabled`).
- `upsertAuthRecord`: when updating an existing record and the payload carries no `prefix` (legacy auth files written before the field existed), inherit the live record's `Prefix` instead of clearing it.

## Testing

- New regression test `TestHostAuthSaveCallbackPreservesPrefixAndDisabled`:
  - legacy payload (no `prefix` field) → inherits the seeded live `Prefix`
  - payload with `prefix` + `disabled: true` → file is authoritative (`Prefix` updated, `Status` disabled)
  - fresh file with `prefix` → registered with the file's prefix
- Verified the test fails on unpatched code with `Prefix after legacy save = ""` (the exact production symptom).
- `go test ./internal/pluginhost/` passes; `go build ./...` clean.

### DIFF
--- a/internal/pluginhost/auth_callbacks.go
+++ b/internal/pluginhost/auth_callbacks.go
@@ -326,6 +326,29 @@ func (h *Host) buildAuthFromFileData(path string, data []byte) (*coreauth.Auth, 
 	if email, ok := metadata["email"].(string); ok && strings.TrimSpace(email) != "" {
 		label = strings.TrimSpace(email)
 	}
+	// Mirror the watcher's file synthesizer (synthesizeFileAuths): both
+	// "prefix" and "disabled" are part of the physical auth-file contract, so
+	// rebuilding the in-memory auth from the file payload must carry them over.
+	// Without this, every host.auth.save from a plugin drops Prefix and the
+	// model registry loses the "<provider>/<model>" IDs until the watcher
+	// re-synthesizes the file (and re-flips Disabled back to active).
+	prefix := ""
+	prefixKeyPresent := false
+	if rawPrefix, ok := metadata["prefix"]; ok {
+		prefixKeyPresent = true
+		if rawString, okString := rawPrefix.(string); okString {
+			trimmed := strings.TrimSpace(rawString)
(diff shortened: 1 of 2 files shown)


viccom:
Addressed the P2: prefix inheritance now keys off metadata key *presence* instead of emptiness. `buildAuthFromFileData` inherits the live `Prefix` only when the payload has no `"prefix"` key at all (legacy files); an explicitly present key stays authoritative even when it normalizes to empty, so `"prefix": ""` clears the prefix. Regression test extended with the explicit-empty case. Rebased onto `dev` (f9a5e562).

@codex review

PR #5393, opened 2026-09-01
feat(codex): dedicated /backend-api/codex/models catalog

congee949:
## Summary
- Add `GET /backend-api/codex/models` so Codex CLI with `chatgpt_base_url` gets a catalog that does not negotiate by User-Agent.
- When `optimize-multi-agent-v2` is off, strip template `multi_agent_version=v2` and leave `v1`/missing alone.

## Test plan
- [ ] `go test ./internal/api ./internal/client/codex/models`
- [ ] Confirm `/backend-api/codex/models` returns the same body for `claude-cli` and browser User-Agents
- [ ] Confirm sol/terra omit `v2` when the optimize flag is false

### DIFF
--- a/internal/api/server_codex_models_test.go
+++ b/internal/api/server_codex_models_test.go
@@ -1,0 +1,36 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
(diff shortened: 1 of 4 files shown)


congee949:
Closing as upstream maintainers noted no current plans for this endpoint in #5781.

PR #5398, opened 2026-09-01
feat(responses): add remote compaction v2 for Codex and opt-in compat

congee949:
## Summary
- Keep `compaction_trigger` on POST `/responses` and require exactly one `compaction` + `encrypted_content` output item (v2). Failures are request-scoped.
- Native Codex always speaks v2; OpenAI-compat is opt-in via `remote-compaction-v2`. Origin xAI still strips the trigger and uses v1 `/responses/compact`.

## Test plan
- [ ] `go test ./internal/runtime/executor ./internal/runtime/executor/helps ./sdk/cliproxy/auth ./internal/config ./internal/watcher/synthesizer ./internal/api/handlers/management`
- [ ] Codex trigger request stays on `/responses` and rejects chat-style output
- [ ] xAI-only pools still compact via origin's v1 adapter
- [ ] Compat providers stay off unless `remote-compaction-v2: true`

### DIFF
--- a/internal/runtime/executor/compaction_support.go
+++ b/internal/runtime/executor/compaction_support.go
@@ -1,0 +1,136 @@
+package executor
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/runtime/executor/helps"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+	"github.com/tidwall/gjson"
+)
+
+// remoteCompactionV2ProtocolError identifies a failure in the compaction
+// response contract. It deliberately remains status-bearing for downstream
+// clients while preventing credential rotation and cooldown side effects.
+type remoteCompactionV2ProtocolError struct {
+	statusErr
+}
+
+func (remoteCompactionV2ProtocolError) IsRequestScoped() bool { return true }
+
+func (e remoteCompactionV2ProtocolError) Unwrap() error { return e.statusErr }
+
+func newRemoteCompactionV2ProtocolError(code int, message string) error {
+	return remoteCompactionV2ProtocolError{statusErr: statusErr{code: code, msg: message, requestScoped: true}}
+}
+
+func rejectUnsupportedCompaction(opts cliproxyexecutor.Options, payloads ...[]byte) error {
+	if opts.Alt == "responses/compact" {
+		return statusErr{code: http.StatusNotImplemented, msg: "/responses/compact not supported"}
+	}
+	if helps.ResponsesHasCompactionTrigger(payloads...) {
+		return newRemoteCompactionV2ProtocolError(http.StatusNotImplemented, "remote compaction v2 is not supported for this provider")
+	}
+	return nil
+}
+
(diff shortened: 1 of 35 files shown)


internal/runtime/executor/openai_compat_executor_compaction.go
```diff
@@ -0,0 +1,326 @@
...
+	if httpResp.StatusCode < 200 || httpResp.StatusCode >= 300 {
+		b, _ := io.ReadAll(httpResp.Body)
+		helps.AppendAPIResponseChunk(ctx, e.cfg, b)
+		helps.LogWithRequestID(ctx).Debugf("request error, error status: %d, error message: %s", httpResp.StatusCode, helps.SummarizeErrorBody(httpResp.Header.Get("Content-Type"), b))
+		return resp, statusErr{code: httpResp.StatusCode, msg: string(b)}
+	}
+
+	body, errRead := io.ReadAll(httpResp.Body)
+	if errRead != nil {
+		helps.RecordAPIResponseError(ctx, e.cfg, errRead)
```
cookerpapa, line 47: When a compatibility upstream returns SSE to `Execute`, `aggregateResponsesCompactionPayload` unwraps the terminal event to a plain response object, but `eventType` remains `response.completed`. `validateRemoteCompactionV2Response` therefore reads `response.status` from an object whose completed status is now at `status`, rejecting a valid result with HTTP 502 (`remote compaction v2 response is missing completed status`). I reproduced this at `853e0d23` with a mock HTTP 200 `text/event-stream` response containing:

```text
data: {"type":"response.completed","response":{"id":"resp_1","status":"completed","output":[{"type":"compaction","encrypted_content":"opaque"}]}}
```

The existing JSON-response control passes. Please keep the validated event envelope through aggregation, or validate before unwrapping, while preserving the tests that reject a missing nested status.

internal/runtime/executor/compaction_support.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if helps.ResponsesHasCompactionTrigger(payloads...) {
+		return newRemoteCompactionV2ProtocolError(http.StatusNotImplemented, "remote compaction v2 is not supported for this provider")
+	}
+	return nil
+}
+
+func remoteCompactionV2MissingItemErr(payload []byte) error {
+	typed, total := helps.ResponsesOutputItemCounts(payload, "compaction")
+	if typed == 1 {
```
cookerpapa, line 41: Counting `type: compaction` alone accepts a completed response with `output: [{"type":"compaction"}]`, or with `encrypted_content: null` / a number. I verified all of those return nil from `validateRemoteCompactionV2Response` at this head. The client then receives a successful compaction without usable state to replay, defeating the new validation boundary. The [official response item schema](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_compaction_item.py) requires string `encrypted_content`. Please validate the sole item's nonempty encrypted content as well as its count, and cover missing, empty, null and non-string values.

cookerpapa (commented):
I reproduced two compaction contract failures on head `853e0d23` while checking Responses compatibility. The checks use local synthetic upstream responses; the existing JSON compaction control and server build pass. Details and minimal payloads are inline. Reviewed with Codex assistance.

PR #5412, opened 2026-09-02
fix(claude): align beta assembly and helper transport with the measured 2.1.258

HYEONMIN94:
On top of the 2.1.258 support in `dev`, this aligns the parts that still differed from the measured wire behavior.

### Changes

- `advanced-tool-use-2025-11-20`: previously attached whenever the request had tools. Now sent only while an advanced tool-use feature such as tool search is actually in use (`tool_search_tool_*`, `defer_loading`, `input_examples`, `allowed_callers`), or when the caller requests it. 2.1.258 did not attach this beta even to a request with 158 plain inline tools.
- `afk-mode-2026-01-31`: forwarded at its measured position between `fast-mode` and `extended-cache-ttl` when the caller sends it, and included as an insertion boundary for `advisor-tool`.
- Haiku helper transport: 2.1.258 helpers send the same `Accept-Encoding: gzip, deflate, br, zstd` as the main thread and never send `X-Stainless-Async`. `x-client-request-id` is attached only when the base URL is api.anthropic.com, so the check accepts an empty value or a valid UUID and rejects only a malformed one.

### Verification

- `gofmt -l .` clean, `go build ./cmd/server`, `go test ./...` pass on current `dev`.
- Built from this branch and run with `--local-model` and a Claude OAuth credential:
  - a `claude-fable-5-1` request returns `200`
  - a real Claude Code 2.1.258 pointed at the proxy through `ANTHROPIC_BASE_URL` responds normally in both `-p` and interactive mode, every request `200`
  - the request logs show no `advanced-tool-use` on requests with only inline tools, and `afk-mode` placed right before `extended-cache-ttl`
  - the helper request from the interactive session is recognized as native and its betas are kept as sent

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -588,7 +588,8 @@ nonstream-keepalive-interval: 0
 # Anthropic-Beta is assembled per request rather than sent as a fixed list, matching
 # Claude Code: context-1m sits right after claude-code, mid-conversation-system
 # is added only for models that accept a role=system turn, advanced-tool-use only when
-# the request declares tools, and server-side-fallback / fallback-credit /
+# the request uses tool search (deferred tools) or another advanced tool-use feature,
+# afk-mode sits between fast-mode and extended-cache-ttl, and server-side-fallback / fallback-credit /
 # structured-outputs trail effort. On direct api.anthropic.com a caller may only ask for
 # betas real Claude Code also sends, and they are placed at their observed positions;
 # anything else is dropped so the outgoing set stays one a real client could produce.
--- a/internal/runtime/executor/claude_executor_native_helper_test.go
+++ b/internal/runtime/executor/claude_executor_native_helper_test.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
 	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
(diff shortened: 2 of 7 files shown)


realies:
errors out with
> ● API Error: 400 messages.4: role 'system' must precede an 'assistant' message or end the array; the directive-only form (content: [] with output_config) is accepted at any position

mardausdennis:
chicken noodle soup chicken noodle

realies:
Follow-up on the 400 I reported here: I traced it to base behaviour in the signature sanitizer, not to this PR. Claude Fable 5.1 emits a thinking-signature generation the CAIS validator does not recognise, so every thinking block is dropped, and a thinking-only assistant turn vanishes, which breaks the system-message placement rule. The fix is in a separate PR against dev: #5422. This PR is not at fault and is not blocked by it; the two were exercised together on a private build and work as expected.

KJ7LNW:
How does this pull request compare to #5456? 

It looks like there's some overlap here. Should either learn from the other?

HYEONMIN94:
@KJ7LNW No real overlap; they cover different layers.

**#5412** aligns what the proxy emits on every request with the measured 2.1.258:
- the beta assembly
- Haiku helper detection
- count_tokens
- a legacy switch for operators pinned below 2.1.258

**#5456** covers billing-header continuity and the Fable 5.1 request body, which #5412 does not touch.

Since sususu98 maintains this area and landed the baseline bump on dev, #5456 is the right one to build on. What is in #5412 is easy to fold in there; I am happy to rebase it on top of #5456 or close it, whichever is preferred.

KJ7LNW:
I'm hoping this pull request 5412 finds its way in. Your alignment changes are beneficial in terms of correctness and it looks like you've put a lot of thought into it. 

so  probably rebase when @sususu98 lands #5456 and keep this one open.

sususu98:
Thanks a lot @HYEONMIN94 for your meticulous measurements and great work on this!

Your findings on the 2.1.258 wire behavior—especially pruning `advanced-tool-use` for plain inline tools, adding `afk-mode`, and aligning the helper compression profile—are spot on and very valuable for correctness.

One minor detail from our transparent proxy and live wire captures: when Claude Code runs behind a transparent gateway / MITM proxy (where the client still targets `api.anthropic.com`), native 2.1.258 helper requests do still carry a valid UUID `x-client-request-id`. To prevent rejecting genuine helpers in those setups, we should probably allow `requestID == "" || isValidUUID(requestID)`.

Since I will be continuously maintaining this fingerprint and model parity area, I plan to land #5456 first to establish the foundational P0 billing header continuity state machine and runtime behavior on `dev`. Once #5456 is merged, we can incorporate these alignment refinements (along with the helper check adjustment) cleanly on top.

Thanks again for the awesome contribution!

HYEONMIN94:
@sususu98 Fixed as you suggested, and this PR is now narrowed to the items you called out.

**request-id**: you are right, I had missed it; my captures went through ANTHROPIC_BASE_URL, so that case never showed up. The check now accepts an empty value or a valid UUID.

**What stays** (everything else follows #5456):
- the advanced-tool-use condition
- afk-mode
- the helper compression profile
- the request-id relaxation

**Next**: I will push the rebased version once #5456 lands.

Since you maintain this area, please handle it whichever way is easiest for you. I would appreciate the chance to contribute this through this PR. Thanks for the review.

KJ7LNW:
fyi, #5456 merged into the  `router-for-me:dev`  branch if you'd like to rebase on top of that: 

https://github.com/router-for-me/CLIProxyAPI/pull/5456#event-30538617275

HYEONMIN94:
@sususu98 @KJ7LNW Rebased on current `dev` now that #5456 has landed.

The PR is now a single commit with only the items called out in review: the advanced-tool-use condition, afk-mode, the helper compression profile, and the request-id check.

Tests pass on `dev`, and a real Claude Code 2.1.258 pointed at the proxy responds normally in both `-p` and interactive mode.

I would appreciate your review. Thank you.

HYEONMIN94:
@sususu98 Addressed the request-id point Codex raised.

Making the `x-client-request-id` check optional opened a path where a helper that arrived without the header received a newly generated UUID on a custom upstream. It is now attached only for a first-party upstream or when the caller actually sent one, so a helper still carries its own request id through as before.

Added a regression test; the full suite passes on `dev`.

Issue #5419, opened 2026-09-02
Tool arguments lost when openai-compat upstream sends per-chunk usage


flobernd:
## Summary

In the OpenAI → Claude streaming translator, a `tool_use` content block is closed on the very first tool-call chunk whenever that chunk also carries a `usage` object. Upstreams that include `usage` on every streaming chunk therefore lose every `function.arguments` delta. The client receives a `tool_use` block with `input: {}` and every tool call fails. Claude Code reports "Invalid tool parameters".

Plain text streaming is unaffected, and the non-streaming path returns the correct arguments.

## Environment

- CLIProxyAPI v7.2.147 (commit 17a65ee), image `eceasy/cli-proxy-api:v7.2.147`, linux/amd64
- Provider: an `openai-compatibility` entry pointing at a self-hosted vLLM server
- Client: Anthropic Messages API, `stream: true`

## Reproduction

1. Configure an `openai-compatibility` provider whose server emits a `usage` object on every SSE chunk.
2. POST to `/v1/messages` with `stream: true` and one tool definition.
3. Ask the model to call that tool.

What the upstream sends:

```
chunk 1  delta.tool_calls[0]  id + function.name          usage.completion_tokens 5
chunk 2  delta.tool_calls[0]  function.arguments '{"skill": "stop-s'
chunk 3  delta.tool_calls[0]  function.arguments 'lop"}'  finish_reason tool_calls
```

What the proxy emits downstream:

```
event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"...","name":"Skill","input":{}}}

event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"tool_use"},"usage":{"input_tokens":191,"output_tokens":5}}

event: message_stop
```

Expected: one or more `content_block_delta` events of type `input_json_delta` carrying `{"skill": "stop-slop"}` before `content_block_stop`.

Actual: no `input_json_delta` at all. The reported `output_tokens: 5` matches the upstream's first chunk, which shows the stream was finalized there.

## Analysis

`internal/translator/openai/claude/openai_claude_response.go`:

```go
if !param.MessageDeltaSent && (param.FinishReason != "" || param.SawToolCall) {
    usage := root.Get("usage")
    if usage.Exists() && usage.Type != gjson.Null {
        finalizeOpenAIAnthropicContentBlocks(param, &results)
[...]

PR #5430, opened 2026-09-02
fix(translator): keep tool arguments when openai-compat upstreams send per-chunk usage

ggbdpq:
## Summary

- OpenAI-to-Claude streaming translation finalized the entire stream on the first tool-call chunk whenever that chunk carried `usage` (the `|| param.SawToolCall` shortcut from 677dbe1d, #5308). Upstreams that attach usage to every chunk (self-hosted vLLM etc.) therefore closed the `tool_use` block while the arguments accumulator was still empty, and every later `function.arguments` delta was dropped. Clients received `input: {}` and every tool call failed (report + wire-level analysis in #5419).
- Fix: decouple "usage seen" from "stream terminal". While `finish_reason` is still empty, the latest usage values are buffered on the conversion params instead of finalizing; they are emitted when `finish_reason` arrives, or at `[DONE]` as a fallback.
- This preserves the #5308 behavior of reporting real usage when an upstream omits `finish_reason`, and restores full tool arguments for per-chunk-usage upstreams.

## Verification

| Check | Result |
| --- | --- |
| New `TestStreamingTool_PerChunkUsageKeepsToolArguments` (3-chunk vLLM-style sequence from #5419) | fails before fix (`tool arguments lost: got ""`), passes after |
| New `TestStreamingTool_PerChunkUsageWithoutFinishReasonUsesBufferedUsage` | fails before fix (`output_tokens = 5`), passes after |
| Existing #5308 tests (`TestStreamingTool_UsageWithoutFinishReasonEmitsMessageDelta`, `TestStreamingTool_OmittedFinishReasonEmitsMessageDeltaOnDone`, `TestStreamingText_OmittedFinishReasonEmitsEndTurnOnDone`) | pass unchanged |
| `go test ./internal/translator/openai/claude/` | pass |
| `go test ./internal/translator/... ./sdk/...` | pass |
| `go test -p 2 ./...` | all packages pass except pre-existing xai executor flakes that fail identically on unmodified `main` on this Windows machine (parallel compile load also crashed the Go compiler 3 times there, incl. once on a clean `main` tree) |
| `gofmt` | clean |

Notes on the flake disclaimer: `TestXAIExecutorExecuteImagesUsesImagesEndpointAndPublishesUsage`, `TestXAIExecutorExecuteVideosCreate`, `TestXAIWebsocketsExecuteStreamSendsResponseCreateWithPreviousResponseID`, and `TestAntigravityAuthHasCreditsRequiredHomeBalanceUsesKV` each failed only under full-suite parallel load on this machine and pass individually; a `-count=3` run of the executor package on unmodified `main` reproduced failures across a wider set. None of them touch this translator. Upstream CI is the authoritative gate.

## AI use

Implemented by GLM-5.3-Flash (ZCode) from the root-cause analysis in #5419, reviewed and verified locally by a human.

## Checklist

- [x] Regression tests included (failing before the fix)
- [x] Existing behavior covered by tests preserved (#5308)
- [x] No unrelated changes; translator-only diff

### DIFF
--- a/internal/translator/openai/claude/openai_claude_response.go
+++ b/internal/translator/openai/claude/openai_claude_response.go
@@ -46,6 +46,15 @@ type ConvertOpenAIResponseToAnthropicParams struct {
 	ContentBlocksStopped bool
 	// Track if message_delta has been sent
 	MessageDeltaSent bool
+	// Usage buffered from the most recent chunk seen before the stream
+	// terminates. Upstreams that attach usage to every chunk must not end the
+	// stream on a non-terminal chunk; the buffered values are emitted when
+	// finish_reason arrives, or at [DONE] as a fallback.
+	LastUsageSeen        bool
+	LastInputTokens      int64
+	LastOutputTokens     int64
+	LastCachedTokens     int64
+	LastCacheWriteTokens int64
 	// Track if message_start has been sent
 	MessageStarted bool
 	// Track if message_stop has been sent
@@ -97,6 +106,11 @@ func ConvertOpenAIResponseToClaude(_ context.Context, _ string, originalRequestR
 			FinishReason:                "",
 			ContentBlocksStopped:        false,
 			MessageDeltaSent:            false,
+			LastUsageSeen:               false,
+			LastInputTokens:             0,
+			LastOutputTokens:            0,
+			LastCachedTokens:            0,
+			LastCacheWriteTokens:        0,
 			ToolCallBlockIndexes:        make(map[int]int),
 			TextContentBlockIndex:       -1,
 			ThinkingContentBlockIndex:   -1,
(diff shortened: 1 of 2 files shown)


ggbdpq:
Fixed in 2974d857: the finish_reason chunk now closes the stream immediately, flushing the buffered values when its own usage is absent. One guard from the original suggestion: when neither is available, the translator keeps waiting for the usage-only chunk that OpenAI's include_usage mode sends after finish_reason (locked in by the existing `TestStreamingUsage_PreservesCacheWriteTokens`, which has usage arriving after finish_reason). Added `TestStreamingTool_FinishReasonChunkFlushesBufferedUsage` covering the per-chunk-usage + separate finish_reason sequence, asserting message_delta/message_stop land on the terminal chunk and [DONE] re-emits nothing.

ggbdpq:
Heads-up on the failing ❌: this PR intentionally touches `internal/translator/` to fix #5419, so the `ensure-no-translator-changes` path guard fails by design — the build and all tests pass. I don't have write access, so per AGENTS.md the analysis and plan are on the issue first; this PR is a concrete, tested proposal (both Codex review suggestions are already addressed in 2974d857). If the maintenance team prefers to apply the change differently, feel free to use the patch or close this — happy to adjust.

ggbdpq:
Closing: the patch landed on dev via cherry-pick `14ee0c84` (same title, authorship preserved), and #5419 is closed as Fixed. Thanks for applying it directly.

Issue #5429, opened 2026-09-02
OpenAI→Claude: cache_control on a tool message content part is emitted inside tool_result.content[0], Anthropic 400s


thomaswillner:
**Is it a request payload issue?**
[X] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[ ] No, it's another issue.

**Describe the bug**

On `/v1/chat/completions` → Claude, when a **tool** message carries `cache_control` on one of its content parts, the OpenAI→Claude translator emits `cache_control` in **two** places:

- on the `tool_result` block (correct), and
- **nested inside `tool_result.content[0]`** (rejected by Anthropic).

Anthropic returns:

```
400 invalid_request_error
messages.2.content.0.tool_result.content.0.cache_control: cache_control may not be specified
within `tool_result.content`. Instead, place it directly on `tool_result`
```

The translator already knows the correct position — it sets one there — but it also carries the client's content-part marker through into the nested array instead of hoisting or dropping it.

**Minimal reproduction (cURL, no agent involved)**

```bash
curl -s -w '\nHTTP=%{http_code}\n' \
  -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  http://127.0.0.1:8317/v1/chat/completions -d '{
  "model": "claude-opus-5",
  "max_tokens": 128,
  "messages": [
    {"role":"user","content":[{"type":"text","text":"Use calc for 2+2.","cache_control":{"type":"ephemeral"}}]},
    {"role":"assistant","tool_calls":[{"id":"call_1","type":"function",
      "function":{"name":"calc","arguments":"{\"expr\":\"2+2\"}"}}]},
    {"role":"tool","tool_call_id":"call_1",
     "content":[{"type":"text","text":"4","cache_control":{"type":"ephemeral"}}]}
  ],
  "tools":[{"type":"function","function":{"name":"calc","description":"calc",
    "parameters":{"type":"object","properties":{"expr":{"type":"string"}},"required":["expr"]}}}]
}'
```

→ **HTTP 400** with the error above.

**Control:** remove `cache_control` from the **tool** message's content part only, change nothing else → **HTTP 200**. A single field is the trigger.

**Observed translation (from CLIProxyAPI's own error dump)**

Inbound `=== REQUEST BODY ===` (OpenAI wire, from the client):

```
.messages[0].content[0].cache_control = {"type":"ephemeral"}
[...]

PR #5432, opened 2026-09-02
fix(translator): hoist tool-result content part cache_control to the block

ggbdpq:
## Summary

- Anthropic only accepts `cache_control` on the `tool_result` block itself; a marker inside `tool_result.content[*]` is rejected with `400 invalid_request_error` (report + wire dump in #5429).
- The OpenAI-to-Claude request translator ran tool message content parts through the same converter as regular message parts, which copies part-level `cache_control` onto every converted part. Clients that attach `cache_control` to tool output parts got the marker emitted in two places: on the `tool_result` block (correct) and nested inside `tool_result.content[0]` (rejected), so every tool-calling turn 400'd on Claude models.
- Fix: strip the marker from parts converted inside `tool_result.content`, and hoist the first source part's `cache_control` onto the `tool_result` block before `AttachMessageCacheControl` runs, so a part-level marker still wins over a message-level one, matching the existing priority. Regular message parts keep their part-level `cache_control` (unchanged behavior).

## Verification

| Check | Result |
| --- | --- |
| New `TestConvertOpenAIRequestToClaude_ToolResultPartCacheControlHoisted` (minimal 3-message sequence from #5429) | fails before fix (`tool_result block cache_control.type = ""`, inner marker present), passes after |
| User text part keeps its part-level `cache_control` (asserted in the same test) | pass |
| `go test ./internal/translator/claude/openai/chat-completions/` (all existing tests incl. cache_control cases) | pass |
| `go test ./internal/translator/... ./sdk/...` | pass |
| `gofmt` | clean |

## AI use

Implemented by GLM-5.3-Flash (ZCode) from the reproduction in #5429, reviewed and verified locally by a human.

## Checklist

- [x] Regression test included (failing before the fix)
- [x] Regular-part `cache_control` behavior preserved (asserted)
- [x] No unrelated changes; translator-only diff

### DIFF
--- a/internal/translator/claude/openai/chat-completions/claude_openai_request.go
+++ b/internal/translator/claude/openai/chat-completions/claude_openai_request.go
@@ -208,7 +208,7 @@ func convertOpenAIRequestToClaude(modelName string, inputRawJSON []byte, stream,
 					contentBlocks = append(contentBlocks, part)
 				} else if contentResult.Exists() && contentResult.IsArray() {
 					contentResult.ForEach(func(_, part gjson.Result) bool {
-						claudePart := convertOpenAIContentPartToClaudePart(part)
+						claudePart := convertOpenAIContentPartToClaudePart(part, true)
 						if claudePart != "" {
 							contentBlocks = append(contentBlocks, []byte(claudePart))
 						}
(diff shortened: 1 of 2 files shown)


ggbdpq:
Heads-up on the failing ❌: this PR intentionally touches `internal/translator/` to fix #5429, so the `ensure-no-translator-changes` path guard fails by design — the build and all tests pass. I don't have write access, so per AGENTS.md the analysis and plan are on the issue first; this PR is a concrete, tested proposal. If the maintenance team prefers to apply the change differently, feel free to use the patch or close this — happy to adjust.

ggbdpq:
Rebased onto current dev (the response_format tests added to this file upstream are kept alongside the hoist test) — the proposal still applies cleanly and the package tests pass. The remaining ❌ is the `ensure-no-translator-changes` path guard, which is expected for a translator-only change without write access; see the earlier note. Issue #5429 is still open, so this remains available as a reviewed proposal whenever the team wants it.

ggbdpq:
Authorization status: `GET /repos/...` reports `push: false` for this account, so per the AGENTS.md rule the sanctioned path was followed — the goal, rationale, and analysis live on issue #5429, and this PR is the concrete, tested implementation for the maintainers to apply (directly, via cherry-pick, or as a reference), exactly as happened with #5430 / commit `14ee0c84`. If the team prefers this PR be closed and the change tracked on the issue alone, say the word and I will close it.

Issue #5170, opened 2026-08-22
fix(home): map user_credits_insufficient to 402 and user_period_limit_exceeded to 429


sususu98:
### Problem Description

When running in Home mode, CLIProxyAPI forwards dispatch requests to CLIProxyAPIHome via RESP `RPOP auth`. When a request is rejected by Home due to billing or period limits, Home returns structured error envelopes:
- `{"error": {"type": "user_credits_insufficient", "message": "insufficient user credits"}}`
- `{"error": {"type": "user_period_limit_exceeded", "message": "period limit exceeded (...)"}}`

### Current Behavior

In `sdk/cliproxy/auth/home_concurrency.go` -> `decodeHomeDispatchError`:
```go
result := &Error{Code: code, Message: message, Retryable: detail.Retryable, HTTPStatus: http.StatusBadGateway}
switch strings.ToLower(code) {
case "model_not_found":
    result.HTTPStatus = http.StatusNotFound
case "authentication_error", "unauthorized", "no_credentials", "invalid_credential":
    result.HTTPStatus = http.StatusUnauthorized
case "credential_concurrency_exceeded", "credential_model_concurrency_exceeded":
    result.HTTPStatus = http.StatusTooManyRequests
    return newHomeConcurrencyBusyError(result, time.Duration(detail.RetryAfterMS)*time.Millisecond)
case "auth_not_found", "auth_unavailable", "refresh_temporarily_unavailable", "home_unavailable",
    "concurrency_protocol_required", "concurrency_tracker_unavailable", "concurrency_node_unavailable":
    result.HTTPStatus = http.StatusServiceUnavailable
}
return result
```

Because neither `user_credits_insufficient` nor `user_period_limit_exceeded` is handled in the `switch`, both fall through to the default `http.StatusBadGateway` (`502`), returning an `internal_server_error` envelope to downstream clients.

### Expected Behavior

- `user_credits_insufficient` should map to HTTP **`402 Payment Required`** (`http.StatusPaymentRequired`).
- `user_period_limit_exceeded` should map to HTTP **`429 Too Many Requests`** (`http.StatusTooManyRequests`).

### Proposed Solution

Update `decodeHomeDispatchError` in `sdk/cliproxy/auth/home_concurrency.go` to handle these error codes:
```go
case "user_credits_insufficient":
    result.HTTPStatus = http.StatusPaymentRequired
case "user_period_limit_exceeded":
[...]

PR #5433, opened 2026-09-02
fix(auth): map home billing errors to 402 and 429

ggbdpq:
## Summary

- Home rejects dispatch requests for billing and period limits with structured error envelopes (`user_credits_insufficient`, `user_period_limit_exceeded`), but `decodeHomeDispatchError` in `sdk/cliproxy/auth/home_concurrency.go` has no case for either code, so both fall through to the default `HTTP 502` and downstream clients see an `internal_server_error` envelope for what is a billing/limit problem (analysis and proposed switch cases in #5170).
- Map `user_credits_insufficient` to `402 Payment Required` and `user_period_limit_exceeded` to `429 Too Many Requests`, as proposed in the issue. No concurrency-busy semantics are added for the 429 case — that stays specific to the `credential_concurrency_exceeded` family.

## Verification

| Check | Result |
| --- | --- |
| New `TestHomeBillingErrorsMapTo402And429` (table-driven, both codes) | fails before fix (both 502), passes after |
| `go test ./sdk/cliproxy/auth/` (all existing home dispatch/concurrency tests) | pass |
| `gofmt` | clean |

## AI use

Implemented by GLM-5.3-Flash (ZCode) from the analysis in #5170, reviewed and verified locally by a human.

## Checklist

- [x] Regression test included (failing before the fix)
- [x] Existing error-code mappings untouched
- [x] No unrelated changes

### DIFF
--- a/sdk/cliproxy/auth/home_concurrency.go
+++ b/sdk/cliproxy/auth/home_concurrency.go
@@ -253,6 +253,10 @@ func decodeHomeDispatchError(raw []byte) error {
 	case "credential_concurrency_exceeded", "credential_model_concurrency_exceeded":
 		result.HTTPStatus = http.StatusTooManyRequests
 		return newHomeConcurrencyBusyError(result, time.Duration(detail.RetryAfterMS)*time.Millisecond)
+	case "user_credits_insufficient":
+		result.HTTPStatus = http.StatusPaymentRequired
+	case "user_period_limit_exceeded":
+		result.HTTPStatus = http.StatusTooManyRequests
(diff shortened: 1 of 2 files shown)


ggbdpq:
Closing in favor of upstream `f416175f` ("fix(home): map user_credits_insufficient to 402 and user_period_limit_exceeded to 429"), which landed on dev with the same 402/429 mapping and its own tests (`TestHomeUserBillingAndPeriodLimitErrors`). The conflict on rebase was exactly this duplication. Thanks for merging the fix upstream.

PR #5446, opened 2026-09-03
feat(claude-code): agent-aware prompt-cache keepalive

ArshansGithub:
Branch: `feat/claude-code-cache-keepalive` — 13 commits off `dev` (18e01a76), HEAD `29bffbeb`.

## Summary

With `promptCacheTtl: "1h"`, an orchestrator session that blocks on a subagent for longer than the TTL pays a full cache write of its whole context on the return turn instead of a read. On the 1h pool that write costs roughly ten times the read it replaces, so refreshing the entry with one minimal request per window is close to free. A client-side hook cannot do this, because it cannot guarantee which account the refresh lands on; the proxy holds the last request body, the credential and the session-to-account binding, so its probe warms the same per-account entry the next real request will hit. The feature is opt-in and off by default, and only requests from a confirmed Claude Code client that carried an explicit `cache_control` ttl of `1h` are ever probed. A 5m session is never probed: on that pool the thirteen reads an hour cost more than the single write they avoid.

Independent of the cache-stats and version-floor branches. It reads the cache-miss diagnostics itself rather than importing the cache-stats helper; see the note below.

## Diff

26 files, +3,637 / −0.

| File | + |
| --- | --- |
| `internal/runtime/keepalive/keepalive.go` (new) | 928 |
| `internal/runtime/keepalive/keepalive_test.go` (new) | 796 |
| `docs/cache-keepalive.md` (new) | 241 |
| `sdk/cliproxy/service_cache_keepalive.go` (new) | 232 |
| `internal/runtime/keepalive/liveness_claude_code.go` (new) | 221 |
| `internal/runtime/keepalive/liveness_claude_code_test.go` (new) | 217 |
| `internal/config/claude_cache_keepalive.go` (new) | 179 |
| `internal/runtime/executor/helps/claude_cache_keepalive_test.go` (new) | 133 |
| `internal/config/claude_cache_keepalive_test.go` (new) | 122 |
| `internal/api/handlers/management/cache_keepalive_test.go` (new) | 111 |
| `internal/runtime/executor/helps/claude_cache_keepalive.go` (new) | 109 |
| `internal/runtime/keepalive/liveness_claude_code_symlink_test.go` (new) | 71 |
| `sdk/cliproxy/service_cache_keepalive_test.go` (new) | 70 |
| `config.example.yaml` | 65 |
| `sdk/cliproxy/auth/selector_session_binding.go` (new) | 41 |
| `internal/runtime/keepalive/default.go` (new) | 28 |
| `internal/api/handlers/management/cache_keepalive.go` (new) | 26 |
| `internal/runtime/executor/claude_executor_execute.go` | 15 |
| `internal/runtime/executor/claude_executor_stream.go` | 12 |
| remaining wiring (`service_lifecycle.go`, `service_config.go`, `config_load.go`, `parse.go`, `sdk_config.go`, `server_management.go`, `README.md`) | 20 |

No deletions; the two executor files gain one Observe call and a start timestamp each.

## Behaviour

- Session-keyed scheduler. On each confirmed Claude Code request carrying `cache_control.ttl: "1h"`, the body is stored and a timer is armed for `ttl − before-expiry`.
- On fire the probe is skipped if a real request superseded it, if `max-probes` consecutive probes were already sent, if no agent is live, or if the auth binding was lost or moved. Otherwise the stored body is sent with `max_tokens: 1` and with `thinking` and the thinking-dependent `context_management` edit removed, since neither is in the cached prefix and the API rejects one without the other. Tools, system, messages, `cache_control` and betas go out byte-identical.
- The probe goes through `Manager.Execute` with `pinned_auth_id`, so translation, headers, retries and token refresh are the same path a real request takes. The probe marks its own context so it cannot observe itself and rearm.
- Liveness `claude-code-tasks` looks for any `<output-dir>/<project-slug>/<session>/tasks/*.output` (symlinks resolved) modified within `agent-idle-window`. The `~/.claude/tasks` todo list is only a secondary OR, since it is not subagent state. Documented caveat: a silent process does not advance its output file, so the window must exceed the expected silence. `liveness: always` skips the check entirely, at a cost of up to `max-probes` reads per idle session.
- Observability: a `cache-keepalive:` prefix on every schedule, probe and skip, with a reason. `probe MISSED` at WARN carrying `cache_miss_reason` and `cache_missed_input_tokens` when the probe's read is zero or below half the observed baseline. `GET /v0/management/cache-keepalive` reports per-session `next_probe_at`, `probes_sent`, `last_probe` and `retired_reason` plus global counters. Retired sessions stay listed under a cap so idle and broken are distinguishable; stored bodies are dropped on retirement.
- Nothing is written to disk. Probe bodies are held in memory only.

## Config

```yaml
claude-code:
  cache-keepalive:
    # Opt-in. Default false.
    enabled: false

    # Fire the probe at (request start + ttl - before-expiry). Default 5m.
    before-expiry: 5m

    # Probe only while a task of that session is still running. Default true.
    only-when-agents-active: true

    # claude-code-tasks (default) reads Claude Code's on-disk task state.
    # always skips the liveness check.
    liveness: claude-code-tasks

    # How recently a task output file must have been touched. Default 10m.
    agent-idle-window: 10m

    # Stop after this many consecutive probes without an intervening real
    # request. Bounds an abandoned session to max-probes reads. Default 6.
    max-probes: 6

    # max_tokens the probe body carries. Default 1.
    max-tokens: 1
```

Applied on hot reload; disabling the block cancels every timer and logs `cache-keepalive: disabled`.

## Verification

```
go build ./...   ok
go vet ./...     ok
go test ./...    ok
```

123 packages: 93 with tests, all passing; 30 with no test files; 0 failures. 63 tests cover the scheduler, the liveness reader including the symlink case, the config, the management handler and the diagnostics reader.

Behaviour verified on a scratch instance against api.anthropic.com with a real `claude` session (haiku, `--strict-mcp-config`) and a ticking background task: probes fired while the task ran, reporting `status=hit` with `cache_read_input_tokens` equal to the real request's, then a `no-live-agents` skip once the task ended and the window passed. A second run took two hits and stopped at `max-probes`, with the endpoint reflecting both the live and the end state. Probing a real body is what surfaced the `context_management` rejection that the final probe shape fixes.

`config.example.yaml` documents only the `claude-code.cache-keepalive` block on this branch; no `usage-cache-stats` keys are present.

## Note for reviewers

`claudeCacheMissReason` in `sdk/cliproxy/service_cache_keepalive.go` reads `diagnostics.cache_miss_reason` from both the non-streaming body and the streaming `message_start` event. It duplicates `applyClaudeCacheMissReason` from the per-session cache-statistics branch, which is not a dependency of this one. The comment in the file points at the other implementation so the two can be folded into a shared helper once both land.

Why: see https://github.com/ArshansGithub/claude-code-fable-usage

Context: https://github.com/ArshansGithub/claude-code-fable-usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ PackyCode provides special discounts for our software users: register using <a h
 - Grok Build multi-account load balancing
 - OpenAI-compatible upstream providers via config (e.g., OpenRouter)
 - Reusable Go SDK for embedding the proxy (see `docs/sdk-usage.md`)
+- Agent-aware prompt-cache keepalive for Claude Code sessions on the 1h cache pool (see `docs/cache-keepalive.md`)
 
 ## Getting Started
 
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ PackyCode provides special discounts for our software users: register using <a h
 - Grok Build multi-account load balancing
 - OpenAI-compatible upstream providers via config (e.g., OpenRouter)
 - Reusable Go SDK for embedding the proxy (see `docs/sdk-usage.md`)
-- Agent-aware prompt-cache keepalive for Claude Code sessions on the 1h cache pool (see `docs/cache-keepalive.md`)
+- Agent-aware prompt-cache keepalive for Claude Code sessions, on the 1h cache pool and on the 5m pool for models with cheap cache reads (see `docs/cache-keepalive.md`)
 
 ## Getting Started
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -190,6 +190,10 @@ claude-code:
 
   # Agent-aware prompt-cache keepalive.
   #
+  # State is visible at GET /v0/management/cache-keepalive and in the log, where
+  # every line is prefixed "cache-keepalive:". A probe that itself found nothing
+  # cached logs at WARN: that is the malfunction signal.
+  #
   # An orchestrator session that blocks on a subagent for longer than the prompt cache
   # TTL pays a full cache write on the return turn instead of a read. On the 1h pool
   # that write costs roughly ten times the read it replaces, so refreshing the entry
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -188,6 +188,55 @@ claude-code:
   # When true, return original model IDs in Anthropic model list responses instead of cloaked IDs.
   disable-cloaking-model-list: false
 
+  # Agent-aware prompt-cache keepalive.
+  #
+  # An orchestrator session that blocks on a subagent for longer than the prompt cache
+  # TTL pays a full cache write on the return turn instead of a read. On the 1h pool
+  # that write costs roughly ten times the read it replaces, so refreshing the entry
+  # with one minimal request per window is close to free.
+  #
+  # The proxy holds the last request body, the credential, and the session-to-account
+  # binding, so its probe warms the same per-account entry the next real request hits.
+  # A client-side hook cannot guarantee the account.
+  #
+  # Only requests from a confirmed Claude Code client that carried an explicit
+  # cache_control ttl of "1h" are ever probed. A 5m request is never probed: on that
+  # pool the thirteen reads an hour cost more than the single write they avoid.
+  cache-keepalive:
+    # Opt-in. Default false.
+    enabled: false
+
+    # Fire the probe at (request start + ttl - before-expiry). Default 5m.
+    before-expiry: 5m
+
+    # Probe only while a task of that session is still running. A session idling on
+    # human input is never probed: that wait is unbounded, and the guarantee that
+    # another turn is coming is what makes the probe pay for itself. Default true.
+    only-when-agents-active: true
+
+    # Liveness strategy.
+    #   claude-code-tasks  read Claude Code's on-disk task state (default)
+    #   always             no liveness check; for clients whose agent state the proxy
+    #                      cannot see. Costs up to max-probes reads per idle session.
+    liveness: claude-code-tasks
+
+    # Stop after this many consecutive probes without an intervening real request.
+    # Bounds the cost of an abandoned session to max-probes reads of its prefix.
+    # Default 6.
+    max-probes: 6
+
+    # max_tokens the probe body carries. Default 1.
+    max-tokens: 1
+
+    # Where the claude-code-tasks liveness check looks. Both default to the paths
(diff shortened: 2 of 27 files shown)


ArshansGithub:
Follow-up pushed: the keepalive now covers 5-minute sessions, not just the 1h pool.

The 1h-only rule assumed cache reads cost 0.1x base input, where the ~12 reads an hour a 5m window needs (1.2x the context) are a wash against the 1.25x write they avoid. On `claude-fable-5-1` and `claude-mythos-5-1` reads are 0.025x base ($0.25/MTok against $10/MTok input), so those same 12 reads cost ~0.3x against that 1.25x write: probing is about 4x cheaper than letting the entry expire. Anthropic's prompt-caching guidance says the same from the other side, recommending a keepalive on the 5m tier over paying the 1h TTL premium on these models unless pauses regularly approach an hour.

New keys under `claude-code.cache-keepalive`, all with working defaults:

- `probe-5m: auto | always | never` (default `auto`) probes a 5m session only when the request model is on the cheap-cache-read list, which `probe-5m-models` replaces (default `claude-fable-5-1`, `claude-mythos-5-1`, matched case-insensitively as substrings so provider-prefixed and suffixed spellings resolve). `never` is the previous behaviour.
- `before-expiry-5m: 45s` is the 5m tier's lead time, and `max-probes-5m: 30` its budget (~2h of idle, where 6 probes buy 6 hours on the 1h pool but only 25 minutes here). The TTL clock runs from the request's start and generation counts against it, so both tiers schedule from request start, not response end.

The probe body, the liveness gate and the binding check are the 1h path unchanged; the liveness gate is still what bounds the spend. `ttl_tier` and the probe-5m decision (`model-auto`, `always`, `skipped-never`, `skipped-model`, `n/a`) now appear in the `cache-keepalive:` log lines and in `GET /v0/management/cache-keepalive`, and the two skips are counted under `skipped_by_reason` so a policy that drops every 5m session is never silent.

`go build ./... && go vet ./... && go test ./...` all clean: 93 packages with tests passing, 30 with no test files, 0 failures. New tests cover the 5m scheduling time, `auto` taking fable and skipping opus-5, `always`/`never`, the 5m reschedule lead time and budget, and the tier fields in the management payload.

KJ7LNW:
See also #4731 

Note to reviewers, this 5446 PR is also useful even after the one hour version lands. People might wish to auto-refresh even in one-hour intervals.

Issue #4497, opened 2026-07-21
codex/claude translator: canary omission of forced strict=false for function tools


jingjingboy:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. I am using Claude Code through the Claude Messages compatibility endpoint and the Codex translator rewrites the tool definitions.
[ ] No, it's another issue.

**Describe the bug / canary request**

The Claude-to-Codex request translator unconditionally writes `strict: false` for every function tool after converting Claude's `input_schema` to Responses API `parameters`:

```go
tool, _ = sjson.SetRawBytes(tool, "parameters", []byte(normalizeToolParameters(toolResult.Get("input_schema").Raw)))
tool, _ = sjson.DeleteBytes(tool, "input_schema")
tool, _ = sjson.DeleteBytes(tool, "parameters.$schema")
tool, _ = sjson.DeleteBytes(tool, "cache_control")
tool, _ = sjson.DeleteBytes(tool, "defer_loading")
tool, _ = sjson.SetBytes(tool, "strict", false)
```

I would like to run a narrowly scoped canary that omits the `strict` field instead of forcing `false`. The first canary should **not** set `strict: true`, because some Claude Code schemas may not satisfy all strict JSON Schema requirements and could be rejected with HTTP 400.

The goal is to determine whether field omission improves tool-argument schema adherence for Codex-backed models while remaining compatible with Claude Code tool declarations.

**CLI Type**

Codex OAuth account, accessed through the Claude Code compatibility translator.

**Model Name**

`gpt-5.6-sol`, reasoning effort `xhigh`.

**LLM Client**

Claude Code 2.1.177.

**Request Information**

The relevant input is a standard Claude Messages request with function tools, for example:

```json
{
  "model": "gpt-5.6-sol",
  "messages": [{"role": "user", "content": "Read a file and report its first line."}],
  "tools": [
    {
      "name": "Read",
      "description": "Read a file",
      "input_schema": {
        "type": "object",
        "properties": {
          "file_path": {"type": "string"},
          "offset": {"type": "integer"},
          "limit": {"type": "integer"}
        },
        "required": ["file_path"]
      }
    }
  ]
}
```

No credentials, authorization headers, or production request logs are included here.

**Expected behavior**

For the first canary, translated function tools should preserve normalized `parameters` but omit the `strict` property unless it was explicitly and safely supplied by a compatible upstream schema policy.

Intended implementation:

```diff
[...]

PR #5457, opened 2026-09-03
fix(translator/codex): downgrade text.format.strict when schema cannot satisfy strict mode

ahoo:
Claude `output_config.format` with `type: json_schema` is translated to Codex `text.format` with `strict: true` by default, unless the caller explicitly passes `strict: false`. When the schema declares optional properties (missing from `required`), strict backends reject the request with HTTP 400 — e.g. prompt Stop-hook structured output (`cli_proxy_structured_output`).

This change downgrades `strict` to `false` when the schema (recursively) has any declared property missing from its sibling `required` list, instead of emitting an unsatisfiable strict schema. Fully-required schemas keep `strict: true`; explicit `strict: false` is unchanged.

Mirrors the precedent of #4497 (forced `strict: false` for function tools).

Tests: extends `TestConvertClaudeRequestToCodex_OutputConfigFormat` with optional-property downgrade + fully-required keeps-strict cases; full package suite passes.

### DIFF
--- a/internal/translator/codex/claude/codex_claude_request.go
+++ b/internal/translator/codex/claude/codex_claude_request.go
@@ -438,6 +438,12 @@ func convertClaudeRequestToCodex(modelName string, inputRawJSON []byte, _ bool, 
 		if s := format.Get("strict"); s.Exists() && s.Type == gjson.False {
 			strict = false
 		}
+		// OpenAI strict mode requires every declared property to be listed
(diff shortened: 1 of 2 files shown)


ahoo:
Closing in favor of an issue per the translator guard policy (external PRs may not touch internal/translator). Will file the reproduction + suggested fix as an issue for the maintenance team.

Issue #5190, opened 2026-08-23
feat(interactions): session fallback, tool-name collision, models endpoint (from blocked PR #5022)


xitex:
## Context

Google's **Interactions API** (`/v1beta/interactions`) powers the **Antigravity Agent** (`antigravity-preview-05-2026`). The naming overlap with Google's Antigravity OAuth product causes confusion — this issue is about the **agent**, not OAuth.

**Related:** #4995 (original bug report), PR #5022 (blocked by `translator-path-guard`), commit `10afcc8c` (partial fix merged by maintainer).

## What #10afcc8c already fixed

✅ Environment context propagation (`previous_interaction_id` ↔ `previous_response_id`, `environment_id`)  
✅ `agent_config.type = "antigravity"` forwarding  
✅ `max_output_tokens` → `agent_config.max_total_tokens` routing  
✅ Stream state preservation for `environment_id`

### 1. Session key fallback for stateless clients
Clients that don't send `Session-Id` header or explicit session marker can't be keyed for continuation state.  
**Fix:** Canonical-JSON SHA-256 of the stable conversation prefix (messages up to last user message) as fallback key.

### 2. Tool-name collision round-trip (`external_` prefix)
Client tools named like Antigravity's built-in sandbox tools (`read_file`, `write_file`, `execute_code`) cause `500 Unknown Error` — Google rejects re-declaring intrinsic tools.  
**Fix:** Rename colliding tools with `external_` prefix upstream, restore on every response path (chat-completions stream/non-stream, responses translator). Shared helper in `internal/translator/common/antigravity_tools.go`.

### 3. `GET /v1/models/:model` endpoint
OpenAI SDKs validate models via this endpoint before delegating. Missing endpoint causes fallback to other providers.  
**Fix:** Single-model endpoint for `antigravity-preview-05-2026`.

## Files affected (from PR #5022)
- `internal/cache/antigravity_interactions_session.go` — session caching
- `internal/runtime/executor/helps/antigravity_interactions_state.go` — state management  
- `internal/translator/common/antigravity_tools.go` — tool-name collision
- `sdk/api/handlers/gemini/interactions_handlers.go` — handlers
- `sdk/api/handlers/openai/openai_handlers.go` — models endpoint

## Why this matters
Without these fixes, agents delegating through the OpenAI-compatible layer can't use local tools with Antigravity — the whole point of the proxy integration.

## Suggestion
[...]

PR #5459, opened 2026-09-03
feat: GET /v1/models/:model endpoint (from #5190)

xitex:
Single-model endpoint so OpenAI SDK clients validate models via detail route instead of 404 + provider fallback.

Item 3 of #5190. Split from #5022. Zero translator changes: 2 files, +29.


xitex:
@codex review — addressed: detail route now dispatches through the Home catalog when Home.Enabled (same as collection), local registry otherwise.

Issue #5451, opened 2026-09-03
perf(pluginhost): HistoryChunks is cloned and re-encoded for every payload stream chunk


VictorSun92:
## Summary

`Host.InterceptStreamChunk` clones the delivered-chunk window into the payload of every stream-chunk interceptor call. The window is bounded at 64 chunks or 1 MiB, so a plugin that never reads `HistoryChunks` still pays a `cloneByteSlices` of it plus a base64 JSON encode, once per SSE frame, and the host holds the ring for the life of every in-flight stream.

Schema v3 already omits `OriginalRequest`/`RequestBody` on payload chunks for exactly this reason. `HistoryChunks` has no such gate.

## Environment

- CLIProxyAPI v7.2.147 (commit 17a65ee), image `eceasy/cli-proxy-api:latest`, linux/arm64
- Plugin: a native c-shared plugin declaring `StreamChunkInterceptor`, schema 3
- Client: Gemini native API, `streamGenerateContent`, `alt=sse`

## Reproduction

Decode a realistic host payload — one small Gemini text chunk plus a saturated 64-chunk window, ~10.9 KB — into the plugin's `StreamChunkInterceptRequest` equivalent. Go 1.24, arm64:

| plugin struct declares | ns/op | B/op | allocs/op |
| --- | --- | --- | --- |
| every payload field | 133,052 | 13,624 | 103 |
| only the 4 fields it reads | 98,009 | 480 | 9 |
| same, window not attached | **6,856** | 480 | 9 |

Declaring fewer fields stops the decoder allocating for the window, but cannot stop the lexer walking it: the remaining 98 us is a scan of bytes the plugin discards. Dropping the window from the payload is a 14x reduction on that path — for a 500-chunk answer, ~49 ms of plugin-side CPU per response versus ~3.4 ms.

## Analysis

`internal/pluginhost/adapters_interceptors.go`:

```go
// Schema v3+ omits request bodies on payload chunks to avoid re-sending multi-MB
// prompts across cgo/JSON for every frame. Legacy plugins still receive them.
if req.ChunkIndex != pluginapi.StreamChunkHeaderInitIndex && streamChunkOmitsRequestBodies(record.plugin.SchemaVersion) {
    ...
}
nextReq.HistoryChunks = cloneByteSlices(req.HistoryChunks)   // no gate
```

[...]

PR #5479, opened 2026-09-04
pluginhost: let stream-chunk interceptors opt out of the history window

bytemain:
## Problem

The host retains a bounded history window (up to 64 chunks / 1 MiB) of already-delivered stream chunks and deep-clones it into `StreamChunkInterceptRequest.HistoryChunks` for **every** stream-chunk interceptor on **every** payload frame (and JSON-marshals it again for RPC plugins). Interceptors that never read `HistoryChunks` still pay that cost on both sides, and because the window grows with the stream, the total cost approaches O(n²) over a long stream.

This is not theoretical: we run a capture/observability interceptor in production that only consumes the current chunk. On long streams the per-frame history clone/marshal/decode dominated CPU (a mostly-idle gateway climbed to several fully busy cores; profiles attributed the majority of time to this path).

## Change

Add an explicit **opt-out capability** — feature detection rather than a schema-version ratchet, so it stays orthogonal to future contract versions:

- A plugin that declares `stream_chunk_omit_history: true` in its registration capabilities no longer receives `HistoryChunks` on payload stream-chunk calls (per-plugin gate in `InterceptStreamChunk*`).
- `Host.StreamChunkPayloadIncludesHistory()` reports whether any active stream-chunk interceptor still consumes the window. Both streaming handlers consult it **per chunk** (not snapshotted), so a mid-stream plugin reload takes effect immediately: accumulation stops once the last consumer opts out, and a consumer activated mid-stream receives history from that chunk on (the window is not reconstructed retroactively).
- Plugins that do not declare the capability see byte-identical behavior; hosts that don't know the flag simply ignore it, so plugins can declare it unconditionally.

If you'd rather evolve this as a schema-version step (like `schema_version >= 3` omitting per-chunk request bodies), the change converts directly — happy to rework to that form.

## Benchmarks

Per delivered frame against a 64-chunk / 1 MiB history window (`BenchmarkInterceptStreamChunkHistory*`):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| legacy | ~256,266 | 1,051,176 | 70 |
| opt-out | ~739 | 800 | 5 |

## Tests

- `TestStreamChunkHistoryPolicyByCapability`: per-plugin omission (a legacy plugin keeps receiving history alongside an opted-out plugin that doesn't) and the host-level accumulation policy flip.
- `TestRegisterRPCPluginParsesStreamChunkOmitHistory`: the registration JSON capability propagates through the real RPC decode.
- `TestExecuteStreamHistoryAccumulationAcrossRoutesAndReloads`: drives both real streaming routes (auth-manager and plugin-executor) through both mid-stream policy transitions and asserts the exact history the interceptor sees on the third frame.

## Field experience

This exact capability form has been running in our production deployment for a while with the expected CPU reduction and no capture/data regressions.

### DIFF
--- a/internal/pluginhost/adapters_interceptors.go
+++ b/internal/pluginhost/adapters_interceptors.go
@@ -277,7 +277,14 @@ func (h *Host) InterceptStreamChunkExcept(ctx context.Context, req pluginapi.Str
 			nextReq.RequestBody = bytes.Clone(req.RequestBody)
 		}
 		nextReq.Body = bytes.Clone(current.Body)
-		nextReq.HistoryChunks = cloneByteSlices(req.HistoryChunks)
+		// Plugins that opt out via StreamChunkOmitHistory never receive the delivered-chunk
+		// history window: no interceptor consumes it, and cloning/marshaling up to 1 MiB of
+		// it per frame is near O(n^2) on long streams. Others still get it unchanged.
+		if record.plugin.Capabilities.StreamChunkOmitHistory {
+			nextReq.HistoryChunks = nil
+		} else {
+			nextReq.HistoryChunks = cloneByteSlices(req.HistoryChunks)
+		}
 		nextReq.Metadata = cloneInterceptorMetadata(req.Metadata)
 		if resp, ok := h.callStreamChunkInterceptor(ctx, record, interceptor, nextReq); ok {
 			current.Headers = mergeHeaders(current.Headers, resp.Headers, resp.ClearHeaders)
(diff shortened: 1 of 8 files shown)


bytemain:
Withdrawing — dev already implements exactly this via 649a8bdb (schema v5, closes #5451), which supersedes this PR. That's precisely the outcome we were after; we'll adopt schema_version 5 in our plugin once it ships in a release. Thanks!

PR #5507, opened 2026-09-04
fix(codex): preserve required empty reasoning levels

### LINKED ISSUE #5500: Codex 0.153 rejects model catalog when Antigravity Claude entries omit supported_reasoning_levels
## Summary

Codex Desktop 26.901.20858 (bundled `codex-cli 0.153.0-alpha.5`) rejects the entire CLIProxyAPI model catalog when Antigravity exposes `claude-opus-4-6-thinking` or `claude-sonnet-4-6` without a `supported_reasoning_levels` field.

This makes unrelated models, including all Gemini models, disappear from the Codex model picker because Codex falls back to its built-in catalog.

## Environment

- CLIProxyAPI: `v7.2.149` (`2a6b87a`)
- Codex Desktop: `26.901.20858`
- Bundled Codex CLI: `0.153.0-alpha.5`
- Provider with affected entries: Antigravity

## Reproduction

1. Configure an Antigravity OAuth credential that exposes both affected Claude models.
2. Request the Codex catalog:

```bash
curl -H 'Authorization: Bearer <key>' \
  'http://localhost:8317/v1/models?client_version=0.153.0-alpha.5'
```

3. Point Codex at CLIProxyAPI and run:

```bash
codex debug models
```

## Actual behavior

Codex reports:

```text
failed to refresh available models: failed to decode models response: missing field `supported_reasoning_levels`
```

The response contains 21 models, but only these two entries omit the required field:

- `claude-opus-4-6-thinking`
- `claude-sonnet-4-6`

Codex rejects the complete response and falls back to its built-in model list. Gemini inference still works when its model ID is supplied explicitly, which confirms that this is a catalog decoding issue rather than a routing or authentication failure.

## Likely cause

The affected Antigravity registry entries declare dynamic thinking support but do not provide `thinking.levels`. The Codex catalog builder therefore omits `supported_reasoning_levels`. New Codex clients deserialize that field as required.

The relevant logic is in `internal/client/codex/models/models.go`, around `applyCodexClientThinkingMetadata` / `codexClientThinkingSupport`.

## Expected behavior

Every model returned in the Codex-specific `models` response should contain a valid `supported_reasoning_levels` array, even if it is empty, so one incomplete provider entry cannot invalidate the entire catalog.

## Suggested fix

When `supported_reasoning_levels` is absent, add a narrowly scoped fallback for the affected Claude models, without overwriting future upstream metadata. For example:

- levels: `low`, `medium`, `high`, `max`
- default: `medium`

Alternatively, emit an empty array when a model has no explicit levels. A regression test should cover both the current client and the legacy-client level filter.

A local build using the narrow fallback was verified with:

- all model tests passing;
- 21 total models and 8 Gemini models preserved;
- both Claude entries containing the field;
- `codex debug models` exiting successfully with no decode error.

## Configuration workaround

Until fixed upstream, excluding the two entries avoids a custom build and preserves all Gemini models:

```yaml
oauth-excluded-models:
  antigravity:
    - claude-opus-4-6-thinking
    - claude-sonnet-4-6
```

This workaround hides the two Claude models and is not a complete fix.

Vurliy:
## Summary

- keep validated template reasoning levels for models that declare budget-based thinking without discrete levels
- emit `supported_reasoning_levels: []` when a model declares no reasoning capability or client-version filtering removes every level
- omit `default_reasoning_level` whenever the resulting level list is empty
- cover Claude OAuth, Antigravity, explicit empty capability, and legacy-client filtering paths

## Background

Codex deserializes `supported_reasoning_levels` as a required field. Since `cdda333c` (`fix(codex): clear unsupported reasoning levels`), CLIProxyAPI deletes this field when a model has no explicit reasoning levels or when client-version filtering removes all levels.

A single affected model then causes Codex to reject the entire remote model catalog and fall back to its bundled catalog.

The two zero-level cases have different semantics. A model with a nonzero thinking budget still supports reasoning, so it keeps the validated default template levels and selector. A model with no declared capability, or whose levels are all incompatible with the requesting client, must not inherit unsupported choices; it receives an empty array instead. This preserves Codex's required response shape without advertising unsupported levels.

`default_reasoning_level` remains omitted whenever no selectable level exists.

Fixes #5500

## Testing

- `go test ./internal/client/codex/models`
- `go test ./...`
- `go build ./cmd/server`

---

### 摘要

- 对于使用 token budget、但没有离散推理等级的模型，保留经过验证的模板推理等级
- 当模型没有声明推理能力，或客户端版本过滤掉所有等级时，返回 `supported_reasoning_levels: []`
- 结果等级列表为空时，不返回 `default_reasoning_level`
- 覆盖 Claude OAuth、Antigravity、显式空能力以及旧客户端等级过滤路径

### 背景

Codex 在反序列化时将 `supported_reasoning_levels` 视为必填字段。自提交 `cdda333c`（`fix(codex): clear unsupported reasoning levels`）以来，当模型没有明确的推理等级，或客户端版本过滤后没有剩余等级时，CLIProxyAPI 会删除该字段。

只要模型目录中有一个模型受到影响，Codex 就会拒绝整个远程模型目录，并回退到内置模型目录。

两种“零等级”情况的语义并不相同。具有非零 thinking budget 的模型仍支持推理，因此保留经过验证的默认模板等级和选择器。没有声明推理能力的模型，或其等级与请求客户端全部不兼容的模型，不应继承不受支持的选项；这两种情况改为返回空数组。这样既能满足 Codex 要求的响应结构，也不会声明模型不支持的等级。

没有可选等级时，`default_reasoning_level` 仍保持不返回。

对应问题：#5500

### 测试

- `go test ./internal/client/codex/models`
- `go test ./...`
- `go build ./cmd/server`

### DIFF
--- a/internal/client/codex/models/models.go
+++ b/internal/client/codex/models/models.go
@@ -398,7 +398,7 @@ func applyCodexClientThinkingMetadata(entry map[string]any, thinking *registry.T
 		})
 	}
 	if len(levels) == 0 {
-		delete(entry, "supported_reasoning_levels")
+		entry["supported_reasoning_levels"] = []any{}
 		delete(entry, "default_reasoning_level")
 		return
 	}
@@ -434,7 +434,7 @@ func sanitizeCodexClientReasoningMetadata(entry map[string]any, clientVersion st
 	}
 
 	if len(levels) == 0 {
-		delete(entry, "supported_reasoning_levels")
+		entry["supported_reasoning_levels"] = []any{}
 		delete(entry, "default_reasoning_level")
 		return
 	}
--- a/internal/client/codex/models/models.go
+++ b/internal/client/codex/models/models.go
@@ -377,6 +377,11 @@ func applyCodexClientThinkingMetadata(entry map[string]any, thinking *registry.T
 	if thinking == nil {
 		return
 	}
+	// Budget-based thinking has no discrete registry levels. Keep the validated
+	// template levels so Codex clients retain their effort selector.
+	if len(thinking.Levels) == 0 && (thinking.Min > 0 || thinking.Max > 0) {
+		return
+	}
 
 	levels := make([]any, 0, len(thinking.Levels))
 	defaultLevel := ""
--- a/internal/client/codex/models/models_test.go
+++ b/internal/client/codex/models/models_test.go
@@ -549,7 +549,7 @@ func TestCodexClientModelsResponseDoesNotInheritUnsupportedReasoningLevels(t *te
 	}
 }
 
-func TestCodexClientModelsResponseKeepsRequiredEmptyReasoningLevels(t *testing.T) {
+func TestCodexClientModelsResponseKeepsTemplateReasoningLevelsForBudgetThinking(t *testing.T) {
 	for _, modelID := range []string{"claude-opus-4-20250514", "claude-opus-4-6-thinking"} {
 		t.Run(modelID, func(t *testing.T) {
 			resp := BuildResponse([]map[string]any{{"id": modelID}}, nil, false)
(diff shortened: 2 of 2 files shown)


zesming:
Heads-up in case it saves you a round: this needs more than a rebase now.

Applying just the early-return to a clean `dev` still fails the package suite, because 5208aec added a `budget-only model` case to `TestCodexClientModelsResponseDoesNotInheritUnsupportedReasoningLevels` that asserts `[]` for `Min: 1024, Max: 64000` — the exact behaviour we both want to change:

```
--- FAIL: ...DoesNotInheritUnsupportedReasoningLevels/budget-only_model
    supported_reasoning_levels JSON = [{low},{medium},{high},{xhigh}], want []
```

So that expectation has to be reconciled either way. I moved it into the budget-only test rather than deleting it, since those levels are supported rather than unsupported and the enclosing test is about levels a client cannot use.

I have #5523 open with the same fix, rebased and green. No preference on which one lands — yours got here first and I'm happy to close mine if you'd rather carry it. Just wanted the test conflict to be visible before you spend time on the rebase.

PR #5511, opened 2026-09-04
fix(thinking): forward requested reasoning effort verbatim for assumed default levels

ggbdpq:
## Summary

OpenAI-compatibility models whose configuration declares no `thinking` block get a proxy-fabricated default level set `["low", "medium", "high"]` (`buildOpenAICompatibilityConfigModels` / `compileOpenAICompatibleModelCapabilities`). `mapConfiguredHighIntent` treated that fabricated set as a real model contract and remapped a requested `reasoning_effort: "xhigh"` down to `high`, which upstreams that accept `xhigh` but not `high` (e.g. SGLang chat templates with `xhigh/medium/low`) then reject with a 400 reporting the wrong level.

This marks the fabricated level sets with a new `ThinkingSupport.LevelsAssumed` flag and skips the high-intent remap plus the level-support enforcement for them, so the explicitly requested effort is forwarded verbatim for the upstream to accept or reject. Explicitly configured `thinking.levels` keep the existing remap/clamp behavior (including `TestApplyThinkingWithModelInfoKeepsSameFamilyValidationStrict`).

## Verification

| Claim | Command | Result |
| --- | --- | --- |
| Requested `xhigh` no longer downgraded for assumed-level models (fails before: got `high`) | `go test ./internal/thinking/ -run TestApplyThinkingWithModelInfoKeepsAssumedLevelRequestsVerbatim -v` | red before fix, green after |
| Explicit levels keep remap/validation behavior | `go test ./internal/thinking/` | ok (incl. locked strict-validation and remap cases) |
| Dependent packages | `go test -p 2 ./sdk/cliproxy/... ./internal/...` | ok; `internal/runtime/executor` xai Images/Videos flaky in combined runs on this machine (pre-existing, single-run green, documented in prior PRs) |

## AI use

## Checklist

- [ ] Behavior change to existing flags/endpoints: default-level fabrication no longer clamps requested effort (see Summary)

### DIFF
--- a/internal/registry/model_registry.go
+++ b/internal/registry/model_registry.go
@@ -116,6 +116,11 @@ type ThinkingSupport struct {
 	// Levels defines discrete reasoning effort levels (e.g., "low", "medium", "high").
 	// When set, the model uses level-based reasoning instead of token budgets.
 	Levels []string `json:"levels,omitempty" yaml:"levels,omitempty"`
+	// LevelsAssumed marks Levels as a proxy default fabricated for models whose
+	// configuration declares no thinking support at all. The declared set cannot
+	// be trusted, so requested effort levels must be forwarded verbatim for the
+	// upstream to validate instead of being mapped or clamped against them.
+	LevelsAssumed bool `json:"levels_assumed,omitempty" yaml:"levels-assumed,omitempty"`
 }
 
 // ModelRegistration tracks a model's availability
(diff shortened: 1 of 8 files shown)


ggbdpq:
Fixed in 8deb49a4: for assumed level sets the none/auto fallbacks are skipped, so `reasoning_effort: "none"` is forwarded as `none` (the openai applier writes it for assumed supports) and `auto` leaves the body unchanged for the upstream default. Regression tests cover both sentinels; explicitly configured levels keep the existing disable/auto semantics, including `TestApplyThinkingWithModelInfoKeepsSameFamilyValidationStrict`.

ggbdpq:
Fixed in the latest push: the OpenAI applier now serializes `ModeAuto` for assumed level sets as `reasoning_effort: "auto"`, so the `(auto)` suffix path — which has no body field to preserve — forwards the explicit request instead of dropping it. The sentinel test gained the suffix case (fails before with an empty effort, passes after) and the body-borne auto assertion was updated from “unchanged” to “serialized as auto”.

PR #5514, opened 2026-09-04
fix(auth): cool single upstream 404s for minutes instead of 12 hours

ggbdpq:
## Summary

A single upstream HTTP 404 set `NextRetryAfter = now + 12h`, taking a Codex credential out of rotation for half a day. 404s on these routes are frequently transient (streamed `response.failed` events map to 404 via `codexTerminalFailureStatus`, upstream blips, model rollouts), and transient 5xx errors only cool for ~1 minute. With a single credential configured, every subsequent request returned `503 auth_unavailable` without reaching the upstream until a manual disable/enable toggle or restart.

This splits the 404 handling:

- raw 404s that do not explicitly identify the requested model as unsupported cool for **10 minutes** (new `notFoundCooldown`);
- 404s that explicitly name the model (existing `isExplicitModelNotFoundError` classification, e.g. `Code: model_not_found`) keep the **12h** model-support cooldown, now named `modelSupportCooldown` alongside the existing 400/422 model-support path.

The credential-level path (`applyAuthFailureState`) has no model context, so its 404s use the short cooldown; model-specific persistence is enforced on the model-state path where the model is known.

## Verification

| Claim | Command | Result |
| --- | --- | --- |
| Single 404 (no model identification) cools minutes at both credential and model level (fails before: `12h0m0s`) | `go test ./sdk/cliproxy/auth/ -run TestManager_MarkResult_SingleNotFoundCooldownsMinutes -v -count=1` | red before fix, green after |
| Explicit model-not-found 404 keeps ~12h | `go test ./sdk/cliproxy/auth/ -run TestManager_ExecuteCount_ExplicitModelNotFoundSuspendsModel -count=1` | pass (existing locked behavior) |
| auth package | `go test -p 1 ./sdk/cliproxy/auth/ -count=1` | ok |

## AI use

## Checklist

- [ ] Behavior change: single raw 404 cooldown 12h -> 10min (see Summary)

### DIFF
--- a/sdk/cliproxy/auth/conductor.go
+++ b/sdk/cliproxy/auth/conductor.go
@@ -50,6 +50,10 @@ type Result struct {
 	Provider string
 	// Model is the upstream model identifier used for the request.
 	Model string
+	// UpstreamModel is the alias-resolved upstream name the provider actually
+	// received when it differs from Model. Structured not-found errors name
+	// this identifier, so 404 classification matches against both.
+	UpstreamModel string
 	// RouteModel is the requested logical route model before alias resolution.
 	RouteModel string
 	// Success marks whether the execution succeeded.
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -859,9 +859,11 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 						case 404:
 							if disableCooling {
 								state.NextRetryAfter = time.Time{}
-							} else if isExplicitModelNotFoundError(result.Error, thinking.ParseSuffix(modelKey).ModelName) {
+							} else if isExplicitModelNotFoundForState(result.Error, modelKey, result.UpstreamModel) {
 								// A 404 that explicitly names this model as unsupported is
-								// demonstrably persistent and keeps the long cooldown.
+								// demonstrably persistent and keeps the long cooldown. The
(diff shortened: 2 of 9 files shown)


ggbdpq:
Fixed in 4cf81d80: the transient branch now keeps the existing deadline whenever it is still live and later than the proposed cooldown, so an in-flight generic 404 completing after an explicit model-not-found result no longer shortens the 12h window to 10 minutes. Regression test `TestManager_MarkResult_Transient404KeepsLongerModelDeadline` covers the ordering (fails before, passes after). Note the same invariant is enforced for the auth-level writer in #5519; whichever PR merges second will rebase onto the other.

ggbdpq:
Both findings addressed (rebased onto current dev, commit c3839530):

1. **Auth-level 404 vs 401** — the credential-level write is now covered by the extend-never-shorten clamp that landed with `1c22598d`: a generic 404 following a live 401 deadline keeps the 30-minute window (`TestManager_MarkResult_Transient404KeepsLongerCredentialDeadline` locks this).
2. **Alias-resolved upstream name** — `Result` now carries `UpstreamModel` (set at the execution result sites where the resolved model is known) and 404 classification matches the structured error against both the route model and the upstream model (`isExplicitModelNotFoundForState`), so `model upstream-name was not found` receives the 12-hour model-support cooldown instead of the 10-minute transient window (`TestManager_MarkResult_ExplicitNotFoundMatchesUpstreamModel`).

Issue #5501, opened 2026-09-04
Cooldown: writers don't always keep the longest live deadline, and scheduling can disagree with the client projection


realies:
Reproduced against a pinned, unmodified checkout of this repository in a clean container. Three independent findings, all in `sdk/cliproxy/auth`: two are cases where a later write to cooldown state overwrites a longer, still-live deadline instead of only ever extending it; the third is a case where the client-facing availability projection disagrees with what the scheduler actually enforces.

## Environment

- Commit: `4a5ab534f827991a86dac70a300b6aa3ef8389ae` (`dev`, unmodified)
- Container: `golang:1.26-bookworm`, digest `sha256:9fdc884aacc3bec89b20ffc69f4bb369c78210e3e4f600387b5128b12c199f81`
- Command: `docker run --rm -v $(pwd):/repo -w /repo golang:1.26-bookworm go test ./sdk/cliproxy/auth/... -run 'TestI2' -v`
- The test file below lives in package `auth` (test-only, not proposed for inclusion) so it can call a few unexported functions directly: `isAuthBlockedForModel`, `clientModelProjectionForAuth`, `existingModelState`, `canonicalModelKey`.

**How the preconditions below are established:** no test sets a cooldown field (`NextRetryAfter`, `ModelStates`, `Quota`, `Unavailable`) directly — every state transition goes through this package's own unmodified writer, `Manager.MarkResult`, called with a `Result` value, and the two functions being checked, `isAuthBlockedForModel` and `clientModelProjectionForAuth`, are the same unexported production functions, called directly because the test lives in the same package. Two of the `Result` values additionally set `CredentialScope` or `RetryAfter` to select the credential-wide path (Finding 1's second write, Finding 2's short-429 sub-case); those are set by the test itself, and whether real call sites populate them the same way from a live provider response was not verified here.

## Read this before the results below

One of the three findings (the same-model writer sequences, both sub-cases) reports its verdict only through a `t.Logf` line, not a test assertion. Both sub-cases **pass regardless of whether the behaviour is present** — `--- PASS` there is not evidence the behaviour is absent; the log line is the only trustworthy signal. The other two findings use `t.Errorf`, so for those, `--- FAIL` correctly means the behaviour reproduces.

## Finding 1 — a write on one model can shorten a longer, unrelated model's live deadline

Model B gets an explicit 404 (12h deadline). Model A, on the same credential, then gets a shorter, credential-scoped 429. Model B's own deadline moves down.

[...]

PR #5519, opened 2026-09-04
fix(auth): keep the longest live cooldown across failure writes and align the model projection

ggbdpq:
## Summary

Fixes the three gaps reported in #5501 (the issue proposes no design; the invariant chosen here is that a failure write may only extend a still-live cooldown, never shorten it):

1. **Cross-model propagation** — a credential-scoped 429 on model A overwrote sibling model B's longer live deadline (repro: 30m 401 on B, then 5min credential-scoped 429 on A, B dropped to 5m). The propagation loop now takes the longest of the incoming deadline, the sibling's quota window, and the sibling's live `NextRetryAfter` (same for the credential-level write in that loop).
2. **Same-model sequences** — a later short 429 or transient 500 overwrote the model's longer live deadline (repro: 30m 401 → 2m/1m). Failure writes to `state.NextRetryAfter` and `auth.NextRetryAfter` are now clamped to keep a still-live earlier deadline. A deliberate zero write (`disableCooling`) still clears.
3. **Projection vs scheduling** — `clientModelProjectionForAuth` never consulted the credential-wide cooldown, reporting `Suspended=false` while `isAuthBlockedForModel` blocked the model (repro: credential-wide 401, no per-model state). The projection now applies the credential-wide cooldown exactly where scheduling does: only when the auth has no per-model states at all — with states present, unmatched models stay schedulable by design (locked by `TestManager_NoForkAlias_PerModelIsolation_NoAggregationFallbackFlip`).

## Verification

| Claim | Command | Result |
| --- | --- | --- |
| Finding 1: sibling deadline preserved (fails before: shortened to 5m) | `go test ./sdk/cliproxy/auth/ -run TestManager_MarkResult_CredentialScopeKeepsLongerSiblingDeadline -v -count=1` | red before, green after |
| Finding 2: later shorter writes preserve 30m deadline (fails before: 2m/1m) | `go test ./sdk/cliproxy/auth/ -run TestManager_MarkResult_LaterShorterFailureKeepsLongerModelDeadline -v -count=1` | red before, green after |
| Finding 3: projection agrees with scheduling (fails before: Suspended=false) | `go test ./sdk/cliproxy/auth/ -run TestManager_ClientModelProjection_ReflectsAuthLevelCooldown -v -count=1` | red before, green after |
| No regression incl. per-model isolation lock | `go test -p 1 ./sdk/cliproxy/auth/ -count=1` | ok (first pass caught the over-broad projection condition via that lock; narrowed to the `len(ModelStates)==0` fallback) |

## AI use

## Checklist

### DIFF
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -794,6 +794,7 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 					state.Unavailable = true
 					state.Status = StatusError
 					state.UpdatedAt = now
+					prevModelRetryAfter := state.NextRetryAfter
 					if result.Error != nil {
 						state.LastError = cloneError(result.Error)
 						state.StatusMessage = result.Error.Message
@@ -873,6 +874,11 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 										if otherState.Quota.Exceeded && otherState.Quota.NextRecoverAt.After(otherNext) {
 											otherNext = otherState.Quota.NextRecoverAt
 										}
+										// Propagation only extends a sibling's still-live
+										// per-model deadline; it never shortens one.
+										if !otherState.NextRetryAfter.IsZero() && otherState.NextRetryAfter.After(otherNext) {
+											otherNext = otherState.NextRetryAfter
+										}
 										otherState.NextRetryAfter = otherNext
 										applyCooldownFields(&otherState.Quota, QuotaState{
 											Exceeded:      true,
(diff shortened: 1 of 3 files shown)


ggbdpq:
Fixed in 7a4d2dd3: the clamp now runs in a `defer`, so the Cloudflare-challenge and invalid-grant early returns keep a still-live longer deadline too. Regression test `TestManager_MarkResult_CloudflareKeepsLongerCredentialDeadline` fails before (12h shortened to 10s) and passes after; zero writes from `disableCooling` still clear.

ggbdpq:
Closing: upstream `1c22598d` ("fix(auth): preserve active cooldown deadlines on subsequent failures", Closes #5501) implements the same extend-never-shorten invariant — including the credential-level clamp on the Cloudflare/invalid-grant paths and the projection alignment, with regression tests covering both. This PR's review fix (clamp via defer covering the early-return branches) is subsumed by those tests; adapting this branch on top of dev would only duplicate the merged behavior. Thanks for taking it directly to dev.

Issue #5392, opened 2026-09-01
fix(wsrelay): a late Delete from the outgoing session can drop the auth of the live session


xyonium:
Two orderings in the websocket connect/disconnect path can leave a connected provider with no auth entry, making it invisible to routing until it reconnects again. Found while running stable provider names (see the companion feature request, #5234); both are reachable on current `dev`, they are just much easier to hit with a stable name because the same provider key is reused across reconnects.

### Race A — clean disconnect, fast reconnect

```
t0  session1's run() returns -> cleanup(errClosed) -> handleSessionClosed
    cur == s1, so the map entry is deleted and cause stays errClosed
    -> wsOnDisconnected -> queued Delete
t1  session2 connects -> handleWebsocket installs it -> wsOnConnected
    guard does GetByID, sees the auth still active (the Delete is still queued)
    -> returns early, no Add is emitted
t2  the queued Delete is consumed -> auth removed
```

Result: session2 is connected and has no auth. It stays invisible until it disconnects and reconnects.

The guard is the reason this window is not self-correcting. It is at `sdk/cliproxy/service_auth.go`, in `wsOnConnected`:

```go
if existing, ok := s.coreManager.GetByID(channelID); ok && existing != nil {
    if !existing.Disabled && existing.Status == coreauth.StatusActive {
        return
    }
}
```

It uses auth state as a proxy for "already registered by a live session", but that state is stale by construction during t1 — the Delete that would have invalidated it is still in the queue.

**Emitting the Add unconditionally is safe, which is the part I failed to say in #5235.** The Add is idempotent: `prepareCoreAuthForModelRegistration` routes an existing ID to `Update`, not `Register` (`sdk/cliproxy/service_auth.go:284-296`), and `Update` preserves `CreatedAt`, `Success`, `Failed`, `LastRefreshedAt` and `ModelStates` (`sdk/cliproxy/auth/conductor_lifecycle.go:133-140`). So an unconditional Add cannot reset usage counters — only `Register` would, and it is not reached. The cost is one redundant registration in the duplicate case.

That said, the guard is your code and it is clearly deliberate, so I would rather you pick the shape. Removing it is one option; making it check session identity instead of auth state is another. Both close the window.

### Race B — overlapping sessions

```
t1  session2 installs itself under the provider key and emits its Add
t2  session1's cleanup runs; the map now holds session2, so cur != s1
    -> the map entry is left alone, but cause is still the original error
[...]

PR #5520, opened 2026-09-04
fix(wsrelay): keep reconnects registered across disconnect/reconnect races

ggbdpq:
## Summary

Fixes the two connect/disconnect orderings from #5392 that could leave a live websocket provider without an auth entry:

- **Race A (clean disconnect, fast reconnect)** — `wsOnConnected` skipped the auth Add whenever the stored auth still looked active, but that state is stale by construction while the previous session's Delete is still queued; consuming the Delete then left the replacement invisible to routing until it reconnected again. The Add is now emitted unconditionally. It is idempotent: `prepareCoreAuthForModelRegistration` routes an existing ID to `Update`, which preserves `CreatedAt`, `Success`/`Failed`, `LastRefreshedAt`, and `ModelStates`. The issue deferred the shape to the maintainers; removing the guard is the only workable shape at this site because the service layer has no session identity to check.
- **Race B (overlapping sessions)** — when the disconnecting session no longer owns the provider key, `handleSessionClosed` still reported the original error, so `wsOnDisconnected` emitted a real Delete that dropped the replacement's auth. The disconnect is now reported as `replaced by new connection` in that case — a message `wsOnDisconnected` already handles by returning early. As flagged in the issue, the callback now runs under `sessMutex` so the disconnect is dispatched before a replacement can install itself and emit its Add; `emitAuthUpdate`'s synchronous fallback path does not touch `wsrelay.Manager` state, so no lock inversion is introduced.

## Verification

| Claim | Command | Result |
| --- | --- | --- |
| Race B: replacement reported instead of original cause (fails before: cause = original error) | `go test ./internal/wsrelay/ -run TestHandleSessionClosed -v -count=1` | red before fix, green after |
| Race B: owner disconnect keeps cause and deletes entry | `go test ./internal/wsrelay/ -run TestHandleSessionClosed_OwnerKeepsCauseAndDeletes -v -count=1` | pass |
| Race A: Add emitted for already-active channel auth (fails before: no update) | `go test -p 1 ./sdk/cliproxy/ -run TestService_WsOnConnected_AlwaysEmitsAuthAdd -v -count=1` | red before fix, green after |
| Packages | `go test ./internal/wsrelay/ -count=1` and `go test -p 1 ./sdk/cliproxy/ -count=1` | ok |

## AI use

## Checklist

### DIFF
--- a/internal/wsrelay/manager.go
+++ b/internal/wsrelay/manager.go
@@ -185,11 +185,17 @@ func (m *Manager) handleSessionClosed(s *session, cause error) {
 	m.sessMutex.Lock()
 	if cur, ok := m.sessions[key]; ok && cur == s {
 		delete(m.sessions, key)
+	} else if ok {
+		// A newer session owns the key: this disconnect must not emit a Delete
+		// that would drop the replacement's auth. Reporting under the lock
+		// queues the disconnect before the replacement's onConnected Add can
+		// be emitted.
+		cause = errors.New("replaced by new connection")
 	}
-	m.sessMutex.Unlock()
 	if m.onDisconnected != nil {
 		m.onDisconnected(s.provider, cause)
 	}
+	m.sessMutex.Unlock()
 }
 
 func randomProviderName() string {
--- a/internal/wsrelay/manager.go
+++ b/internal/wsrelay/manager.go
@@ -20,10 +20,11 @@ type Manager struct {
 	upgrader  websocket.Upgrader
 	sessions  map[string]*session
 	sessMutex sync.RWMutex
-	// disconnectDrains tracks one channel per provider key whose disconnect
-	// callback is still running. A replacement connection waits on it so its
-	// onConnected Add can never precede the previous session's Delete.
-	disconnectDrains map[string]chan struct{}
+	// keyLocks serializes auth emissions per provider key: an install's
+	// onConnected Add and an owner disconnect's onDisconnected Delete each
+	// validate ownership and emit while holding the key's lock, so a stale
+	// emission can never land after a newer session's Delete.
+	keyLocks map[string]*sync.Mutex
 
 	providerFactory func(*http.Request) (string, error)
 	onConnected     func(string)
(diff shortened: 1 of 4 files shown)


ggbdpq:
Both findings addressed in 56a84b83:

1. **Stale onConnected** — the handler now emits the connect event only while its session still owns the provider key, so a middle connection resuming after a third one replaced it can no longer add a phantom auth after the third session's Delete.
2. **Callback under the session lock** — the disconnect callback now runs outside `sessMutex` (re-entry via `Send`/session lookups can no longer self-deadlock, and inline auth cleanup no longer blocks lookups). The Delete-before-Add ordering that motivated the original lock scope is preserved with a per-key drain channel: the owner close registers it under the lock, a replacement waits on it before emitting its Add, and it is closed when the callback completes — including on panic.

`TestHandleWebsocket_OrdersReplacementConnectAfterDisconnectCallback` drives a blocking disconnect callback through a real websocket reconnect and asserts the ordering; the original red-green tests for the underlying races are unchanged.

ggbdpq:
All three findings addressed by replacing the drain channels with a per-key callback lock (commit c2b8bcbb, force-pushed):

1. **Ownership check-then-emit window** — installs and owner-disconnect emissions are now serialized under the same per-key lock, and the ownership validation happens under that lock immediately before emission. A third connection cannot install between the check and the emit, so the stale Add can never follow a newer session's Delete; the last emission for a key always belongs to its live session.
2. **Overlapping owner-disconnect drains** — the drain map is gone. Each owner disconnect holds its key's lock for the duration of its callback; a replacement install waits for the lock before it can emit anything, so callbacks can no longer overwrite or bypass one another.
3. **Timed assertion** — the replacement handler's progress is now established by a provider-factory barrier (deterministic) before the bounded wait; the positive ordering (disconnect-done before the replacement's Add) remains asserted on a single ordered channel.

The callback still runs under the per-key lock rather than the global session lock, so same-key manager calls from a callback are the only re-entrancy restriction and other providers are never blocked. `-race` could not be run locally (no cgo toolchain on this Windows machine); the package tests pass with 5x repetition and CI coverage applies.

PR #5534, opened 2026-09-05
fix(payload): apply parameters in deterministic order

realies:
Payload parameters were applied in Go map iteration order, so identical inputs could apply or skip an interacting content override. Apply parameter keys in lexicographic order because declaration order is not retained in the decoded Params map; rule-list order remains unchanged. Check: `go test ./internal/runtime/executor/helps -run ^TestApplyPayloadConfigParameterOrderDeterministic$ -count=1 -v` verifies the expected serialized body on all 512 calls; the full helps package also passes.


realies:
@codex review

PR #5536, opened 2026-09-05
fix(openai-compat): reject out-of-range Retry-After seconds instead of overflowing

realies:
A numeric Retry-After one second past the largest representable time.Duration overflowed to a negative delay. Treat out-of-range values as an invalid header, so the caller takes its existing invalid-header handling (including the TPM fallback) instead of a negative wait. This is robustness against hostile or broken upstream input, not a case seen in ordinary traffic. Check: `go test ./internal/runtime/executor -run '^TestOpenAICompatRetryAfterSecondsBoundary$' -count=1 -v`


realies:
@codex review

PR #5537, opened 2026-09-05
fix(responses): preserve escapes after invalid UTF-16 surrogate pairs

realies:
The truncated custom-tool input fallback consumed two Unicode escapes even when they did not form a valid UTF-16 surrogate pair, losing the second escape's character. Consume both escapes only when pair decoding succeeds; otherwise decode the second escape independently. The check forces the fallback path and covers a valid pair, a high surrogate followed by an escaped letter, and two high surrogates.


realies:
@codex review

PR #5583, opened 2026-09-07
fix(claude): forward the per-turn-control beta requested by the caller

realies:
## Defect

Claude Code 2.1.261 sends a per-turn `{"role":"system","content":[],"output_config":{...}}` turn inside `messages`, declared by the `per-turn-control-2026-07-01` beta. `claudeCodeCLIBetas` rebuilds the `Anthropic-Beta` header from the 2.1.258 baseline, which has no entry for that beta, so the turn is forwarded while its beta is stripped and Anthropic rejects the request:

```
400 invalid_request_error: messages.3.output_config: Extra inputs are not permitted
```

Measured on a gateway serving 2.1.261 clients: 77 such rejections across 4 sessions in 12 h (one session retried 67 times).

## Mechanism

On the direct-Anthropic rebuild path (`!preserveCallerFingerprint`) caller betas only reach upstream through the known-beta flags in `claudeCodeCLIBetas`; unknown betas are dropped by design (`TestClaudeCodeCLIBetas_MatchesObservedClientMatrix`, "unknown caller beta is not smuggled into the baseline"). `per-turn-control-2026-07-01` was unknown.

## Fix

One constant, and forward it when the caller requests it, in the position the client uses (right after `mid-conversation-system-2026-04-07`). Unknown betas are still dropped.

## Check

- `TestApplyClaudeHeaders_PerTurnControlBetaFollowsCaller`: fails on the previous baseline, passes with this change, and asserts an unknown caller beta still does not reach the upstream header.
- Replay of a captured 2.1.261 request carrying the per-turn turn and the client's 12 betas, through a single OAuth credential: 400 (message above) on `dev` @ fa01468e, 200 on `dev` + this commit.
- Side effect observed while replaying: forwarding the beta changes Anthropic's prompt-cache identity for histories that already contain `role: system` turns (one cache miss, +15 tokens on a 320k-token prefix). The native client sends the beta itself, so this matches what unproxied sessions already get.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `per-turn-control-2026-07-01` beta capability when requested by callers.
  - Preserves the expected ordering of supported beta capabilities in requests.

- **Tests**
  - Added coverage verifying per-turn control handling and filtering of unknown beta capabilities.

### DIFF
--- a/internal/runtime/executor/claude_executor_beta_policy_test.go
+++ b/internal/runtime/executor/claude_executor_beta_policy_test.go
@@ -523,3 +523,42 @@ func TestApplyClaudeHeaders_StructuredHelperBetaOrderPreservedWithAdvisor(t *tes
 		t.Fatalf("helper Anthropic-Beta =\n got:  %q\n want: %q", got, helperBeta)
 	}
 }
+
+// Claude Code 2.1.261 sends a per-turn {"role":"system"} turn that carries
+// output_config, declared by per-turn-control-2026-07-01. The rebuilt baseline
+// must forward that beta when the caller asks for it, in the client's position
+// right after mid-conversation-system, while still dropping unknown betas.
+func TestApplyClaudeHeaders_PerTurnControlBetaFollowsCaller(t *testing.T) {
+	incoming := http.Header{}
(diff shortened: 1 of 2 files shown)


realies:
@codex review

realies:
Superseded by d9b8fdb7 (#5738), which forwards unmanaged caller betas on both the messages and count_tokens paths; verified by replay against dev.

PR #5585, opened 2026-09-07
fix(auth): tolerate suspend in OAuth refresh loop and selection

### LINKED ISSUE #5584: OAuth credentials permanently blocked after Linux system suspend
[ *] No, it's another issue.

**Describe the bug**
When running CLIProxyAPI as a local daemon (e.g. systemd user service) on a Linux machine that suspends/sleeps (such as a laptop), OAuth credentials become permanently blocked upon system resume.

Go timers freeze during sleep: `auto_refresh_loop.go` uses standard `Go time.Timer` (`CLOCK_MONOTONIC`), which stops ticking while Linux is suspended. The background refresh loop therefore misses its deadline and does not fire upon wake.

`selector.go` pre-filters expired tokens: When a request arrives, `isAuthBlockedForModel()` sees the access token expired and marks the credential as blocked. Because it gets filtered out before execution, the request handler never gets a chance to use the refresh_token or trigger a refresh retry.

**CLI Type**
What type of CLI account do you use?  (gemini, codex, claude code or openai-compatibility)
Gemini

**Model Name**
What model are you using? (example: gemini-2.5-pro, claude-sonnet-4-20250514, gpt-5, etc.)
gemini-3.8-flash-high

**LLM Client**
What LLM Client are you using? (example: roo-code, cline, claude code, etc.)
Hermes

**Expected behavior**

CLIProxyAPI should detect that the access token expired while asleep and execute a token refresh.

**OS Type**
 - OS: NixOS
 - Version 26.05

Murat65536:
Closes #5584 

Use a CLOCK_BOOTTIME-backed timer on Linux so background token refresh deadlines fire after system suspend, and allow expired credentials with valid refresh tokens to remain selectable for on-demand refresh.

### DIFF
--- a/sdk/cliproxy/auth/auto_refresh_loop.go
+++ b/sdk/cliproxy/auth/auto_refresh_loop.go
@@ -120,14 +120,9 @@ func (l *authAutoRefreshLoop) rebuild(now time.Time) {
 }
 
 func (l *authAutoRefreshLoop) loop(ctx context.Context) {
-	timer := time.NewTimer(time.Hour)
-	if !timer.Stop() {
-		select {
-		case <-timer.C:
-		default:
-		}
-	}
-	defer timer.Stop()
+	timer := newRefreshDeadlineTimer(time.Hour)
+	timer.Stop()
+	defer timer.Close()
 
 	var timerCh <-chan time.Time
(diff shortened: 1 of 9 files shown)


Murat65536:
@codex review

Issue #5476, opened 2026-09-04
A single upstream 404 blocks a Codex credential for 12 hours


lingkai1:
## Summary

A single upstream HTTP 404 on a Codex credential sets `NextRetryAfter = now + 12h`, taking the credential out of rotation for half a day. With only one credential configured, every subsequent request fails with:

```
503 auth_unavailable: no auth available (providers=codex, model=...; last upstream error: status 404)
```

...without the proxy even attempting the upstream, and there is no recovery path except restarting the process or disabling/re-enabling the auth via the Management API.

## Observed behavior

- One upstream 404 at 22:45 (likely a transient upstream blip) marked the credential `not_found`.
- All requests for the next ~12 hours returned 503 immediately; the per-credential request counters showed 0 attempts in every 10-minute bucket, confirming requests were never dispatched.
- Direct upstream verification with the same token showed the endpoint and model working fine (HTTP 200) the whole time.
- Toggling `disabled` true→false via `PATCH /v0/management/auth-files/status` cleared the stuck state immediately and traffic recovered.

## Root cause

`sdk/cliproxy/auth/conductor_cooldown.go` (the `case 404:` branch):

```go
case 404:
    auth.StatusMessage = not_found
    if disableCooling {
        auth.NextRetryAfter = time.Time{}
    } else {
        auth.NextRetryAfter = now.Add(12 * time.Hour)
    }
```

12 hours is far too aggressive for 404s, which are frequently transient in this context (e.g. a streamed `response.failed` event mapped to 404 by `codexTerminalFailureStatus`, upstream blips, model rollouts). Transient 5xx errors only get ~60s, while a one-off 404 — no more clearly permanent — gets 12h.

## Suggested fix

- Reduce the 404 cooldown to minutes (e.g. 5–10 min), or
- apply escalating backoff across repeated 404s instead of a fixed 12h on first occurrence, or
- only apply the long cooldown to 404s that are demonstrably persistent (e.g. consecutive failures), and keep single occurrences in a short transient window.

ggbdpq:
Reproduced against current dev: both the credential-level (`applyAuthFailureState`) and model-level 404 branches set a fixed 12h cooldown. Fix per your option 3: raw 404s cool 10 minutes, while 404s that explicitly name the model as unsupported (existing `isExplicitModelNotFoundError` classification) keep the 12h cooldown. PR: #5514

lingkai1:
[...]

PR #5594, opened 2026-09-07
fix(auth): use transient cooldown for HTML 404 error pages

hlwhl:
HTML 404 pages currently trigger the same 12-hour model/credential cooldown as other 404 responses. This PR proposes using the existing configurable transient cooldown (one minute by default) for HTML pages, so a recovered upstream route can be retried sooner. This is a recovery-policy improvement: HTML alone does not prove that a failure is transient, and a persistent HTML 404 will be retried more often.

The change applies at both model and credential scope. Plain/structured 404s and explicit model-not-found codes retain their existing behavior. Cloudflare challenges keep their separate classification; disabled-cooldown settings and longer live deadlines remain respected. HTML tag-boundary matching includes space, tab, newline, carriage return, and form feed, addressing the automated review.

Related to #5476 and a narrower alternative to #5514. This proposal only changes HTML error pages, rather than every unclassified 404, and includes the current monotonic cooldown protections. Both alternatives need not be merged.

### Incident clarification

The user subsequently identified their network proxy as the cause of the reported outage and confirmed recovery. The original upstream HTTP status was not retained. A minimal request succeeded after clearing local routing state during investigation, but that did not establish the cause of the user's overall failure. This PR must not be treated as a confirmed fix for that incident. Its justification is the independently reproducible 12-hour HTML-404 cooldown and the proposed recovery-policy tradeoff, subject to maintainer agreement.

### Validation

- Regression tests demonstrate the existing 12-hour HTML/doctype cooldown at model and credential scope and the proposed one-minute behavior.
- Review regression cases for newline/tab/CR/form-feed tag boundaries fail before the follow-up and pass after it; non-HTML tag prefixes remain unchanged.
- Tests cover plain/structured 404s, custom/disabled transient cooldowns, and preservation of longer deadlines.
- `go test ./sdk/cliproxy/auth -count=1` passes.
- `go build -o test-output ./cmd/server` passes.
- Changed Go files formatted with gofmt; `git diff --check` passes.

Implemented and verified with Codex assistance.

### DIFF
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -835,7 +835,10 @@ func (m *Manager) MarkResult(ctx context.Context, result Result) {
 								state.NextRetryAfter = next
 							}
 						case 404:
-							if disableCooling {
+							if isHTMLNotFoundResultError(result.Error) {
+								state.NextRetryAfter = recoverableFailureRetryAfter(now, disableCooling)
+								state.Unavailable = !state.NextRetryAfter.IsZero()
+							} else if disableCooling {
 								state.NextRetryAfter = time.Time{}
 							} else {
 								next := now.Add(12 * time.Hour)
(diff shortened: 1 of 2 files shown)


hlwhl:
Closing because the reported outage was traced to the user’s network proxy, and the available evidence does not establish a CPA bug as its cause. The proposed cooldown change is a recovery-policy adjustment rather than a confirmed fix for that incident.

Issue #5535, opened 2026-09-05
bug(wsrelay): stream_end silently evicts a buffered chunk when the request queue is full


etwk:
**Is it a request payload issue?**
- [x] No, this is an internal WebSocket relay delivery issue.

**Describe the bug**

When a pending request's 64-slot channel is full, `deliverTerminal` removes the oldest queued frame to make room for `stream_end`. `dispatch` then closes the request channel normally. The consumer receives a successful end-of-stream message despite missing part of the response.

This is a focused follow-up to #4709, which is closed as fixed. The implementation added by [02c02cda](https://github.com/router-for-me/CLIProxyAPI/commit/02c02cda50b71251d3a10e381e0c462453045d0b) addresses ordinary chunk delivery but still loses a queued chunk at terminal delivery under buffer pressure.

Verified against upstream `main` and `dev`, both at `5208aec703b5ce7e3445f6e9d91cc13b3e78003a` (v7.2.151) when checked.

**Affected code**

- [`deliverTerminal`](https://github.com/router-for-me/CLIProxyAPI/blob/5208aec703b5ce7e3445f6e9d91cc13b3e78003a/internal/wsrelay/session.go#L82-L109): the full-channel path executes `<-pr.ch`, discarding one frame, then enqueues the terminal message.
- [`dispatch`](https://github.com/router-for-me/CLIProxyAPI/blob/5208aec703b5ce7e3445f6e9d91cc13b3e78003a/internal/wsrelay/session.go#L259-L282): treats that terminal delivery as completion and closes the channel.

**Reproduction / request information**

No credentials, HTTP request payload, or external model are needed. The existing `TestSession_Dispatch_TerminalDeliveredWhenBufferFullAndSessionClosing` already constructs the failing state:

1. Queue 64 `stream_chunk` messages with sequence numbers 0 through 63 without draining the channel.
2. Dispatch `stream_end`.
3. Drain the channel.

Replace its `if len(msgs) == 0` assertion with:

```go
if len(msgs) != pendingChannelBuffer+1 {
    t.Fatalf("frame loss: expected %d frames, got %d (first chunk seq=%v, last type=%s)",
        pendingChannelBuffer+1, len(msgs), msgs[0].Payload["seq"], msgs[len(msgs)-1].Type)
}
```

Run:

```sh
go test ./internal/wsrelay -run '^TestSession_Dispatch_TerminalDeliveredWhenBufferFullAndSessionClosing$' -count=1 -v
```

[...]

Issue #4709, opened 2026-08-01
Tracking: fix empirically reproduced concurrency, lifecycle, and accounting bugs


sususu98:
## Summary

A focused mock/race audit against commit [`4731e2e`](https://github.com/router-for-me/CLIProxyAPI/commit/4731e2e89561e838f9c2aec6e9c45a14d11ddac7) reproduced several independent correctness defects in request routing and lifecycle code.

This is a consolidated tracking issue. The findings can be fixed in separate PRs, but the `internal/wsrelay` items should be handled first because they include a process panic, a real data race, unbounded per-request goroutine retention, and silent protocol-frame loss.

All stress counts below demonstrate reachability and repeatability only; they are not estimates of production incidence.

### 1. Critical: `internal/wsrelay` channel lifecycle, map race, goroutine retention, and frame loss

Affected code: `internal/wsrelay/session.go` (`dispatch`, `request`, and `cleanup`).

Four independent scenarios reproduced three times each:

1. **Concurrent send/close can panic.**
   - 4,000 iterations using the same `pendingRequest` send/close operations recovered 3,907 / 3,934 / 3,930 `send on closed channel` panics.
   - A focused `go test -race` also reported concurrent channel send/close and exited with status 1.
   - `dispatch` can retain a `*pendingRequest` while request cancellation or cleanup closes its channel.

2. **`cleanup` races with active `sync.Map` operations.**
   - One goroutine repeatedly performed `Store`, `Load`, and `Delete` while the real `session.cleanup()` ran.
   - The race detector reported the write at `internal/wsrelay/session.go:182`, where `s.pending = sync.Map{}` replaces a map that other goroutines may still access.

3. **Successful requests retain cancellation goroutines until the whole WebSocket closes.**
   - A real local WebSocket pair completed 64 requests using `context.Background()`.
   - All three runs retained exactly 64 `session.request.func1` goroutines after the responses completed.
   - The goroutines exited only after closing the entire session.

4. **A slow consumer loses protocol frames silently.**
   - With the request channel capacity of 8 and no consumer, dispatching 9 non-terminal frames retained only 8.
   - The ninth frame was dropped without an error or backpressure signal in all three runs.

Suggested direction:

- Give `pendingRequest` one synchronized lifecycle API for delivery and close; never send directly to a channel that another goroutine may close.
- Do not replace an active `sync.Map`; drain with `Range` plus `LoadAndDelete`.
[...]

PR #5601, opened 2026-09-07
fix(wsrelay): wait for buffer space instead of evicting a queued frame

YogaSakti:
## Problem

`deliverTerminal` evicts one queued frame when the pending request channel is full, then enqueues the terminal message in its freed slot:

```go
default:
    select {
    case <-pr.ch:   // oldest queued response frame is discarded
    default:
    }
```

`dispatch` treats that delivery as normal completion and closes the request channel, so the consumer receives `stream_end` and a successful end-of-stream while one response chunk is gone. Ordinary chunk delivery is not affected: `deliver` already waits for buffer space.

Checked on base commit `82f4f370df45f7a5a4be52f8101d4294f656ff02` (`dev`):

- `internal/wsrelay/session.go:82` — `deliverTerminal`, the eviction path.
- `internal/wsrelay/session.go:53` — `deliver`, which blocks on a select with `done` / session-closed / request-context escape hatches.
- `internal/wsrelay/session.go:261` — `dispatch`, which closes the request after terminal delivery.

The eviction path was introduced with the buffered-delivery hardening in `02c02cda` (the fix for #4709); that change covered ordinary chunks but left terminal delivery lossy under buffer pressure.

## Changes

`internal/wsrelay/session.go`

- `deliverTerminal` now waits for buffer space using the same select as `deliver`, so a terminal message never replaces response data. The `done`, session-closed and request-context cases are preserved, so a stalled consumer cannot pin the session read loop — the same escape hatches `deliver` relies on, already covered by `TestSession_SlowConsumer_UnblocksOnContextCancel` and `TestSession_SlowConsumer_UnblocksOnSessionClose`.
- `pr.terminal` is set only once the message is actually queued. Previously it was set even when delivery failed, which suppressed the explicit error frame `cancelWithError` injects for a request that never received its terminal message.

`internal/wsrelay/session_test.go`

- `TestSession_Dispatch_TerminalDeliveredWhenBufferFullAndSessionClosing` now dispatches the terminal message from a goroutine against a saturated buffer and drains with a concurrent consumer, asserting all 64 chunks arrive in sequence order followed by `stream_end`. The previous assertion only checked that some messages arrived and that the last one was `stream_end`, which the eviction path satisfied.
- `TestSession_Dispatch_TerminalUnblocksOnSessionClose` is added so a saturated terminal delivery is proven to unblock on session cleanup rather than hanging.

## Verification

Red state on the unmodified base, with the test change applied:

```
$ go test ./internal/wsrelay -run '^TestSession_Dispatch_Terminal' -count=1
--- FAIL: TestSession_Dispatch_TerminalDeliveredWhenBufferFullAndSessionClosing (0.01s)
    session_test.go:279: frame loss: expected 65 frames, got 64
FAIL
```

This matches the failure reported in #5535.

Green state after the fix:

```
$ go test ./internal/wsrelay -count=1
ok  	github.com/router-for-me/CLIProxyAPI/v7/internal/wsrelay	0.528s

$ go test ./internal/wsrelay -count=5 -race
ok  	github.com/router-for-me/CLIProxyAPI/v7/internal/wsrelay	1.891s

$ gofmt -l internal/wsrelay/
$ go build -o test-output ./cmd/server
$ go test ./...
```

`go test ./...` passes with no failures. `gofmt` reports nothing for the changed files.

The regression the new assertions catch: a terminal message that is delivered by discarding buffered response data, reported downstream as a successful stream.

## Scope and omissions

- `cancelWithError` uses the same evict-then-enqueue shape when it injects an error frame during cleanup. That path is a failure path where the stream is already being torn down, and #5535 does not cover it, so it is left unchanged rather than folded into this fix.
- Production occurrence frequency is not measured; the reproduction is the in-repo unit test, which needs no credentials, model or client.

### DIFF
--- a/internal/wsrelay/session.go
+++ b/internal/wsrelay/session.go
@@ -89,22 +89,25 @@ func (pr *pendingRequest) deliverTerminal(sessClosed <-chan struct{}, msg Messag
 	if pr.closed {
 		return false
 	}
-	pr.terminal = true
 
+	var ctxDone <-chan struct{}
+	if pr.reqCtx != nil {
+		ctxDone = pr.reqCtx.Done()
+	}
+
+	// Wait for buffer space like deliver does instead of evicting a queued
+	// frame: dropping response data to make room for the terminal message
+	// reports a successful stream while part of the response is missing.
 	select {
+	case <-pr.done:
+		return false
+	case <-sessClosed:
+		return false
+	case <-ctxDone:
+		return false
 	case pr.ch <- msg:
+		pr.terminal = true
 		return true
-	default:
-		select {
-		case <-pr.ch:
-		default:
-		}
-		select {
-		case pr.ch <- msg:
-			return true
-		default:
-			return false
-		}
 	}
 }
 
--- a/internal/wsrelay/session_test.go
+++ b/internal/wsrelay/session_test.go
@@ -247,23 +247,84 @@ func TestSession_Dispatch_TerminalDeliveredWhenBufferFullAndSessionClosing(t *te
 		})
 	}
 
-	// Dispatch terminal message
-	s.dispatch(Message{
-		ID:   reqID,
-		Type: MessageTypeStreamEnd,
-	})
+	enteredDispatch := make(chan struct{})
+	go func() {
+		close(enteredDispatch)
+		// Dispatch terminal message against a saturated buffer
+		s.dispatch(Message{
+			ID:   reqID,
+			Type: MessageTypeStreamEnd,
+		})
+	}()
 
-	var msgs []Message
-	for msg := range req.ch {
-		msgs = append(msgs, msg)
+	<-enteredDispatch
+	// Small yield to ensure the goroutine is in deliverTerminal's select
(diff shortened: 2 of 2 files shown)


YogaSakti:
Closing this as a duplicate.

`7871a5a9` ("fix(wsrelay): preserve queued frames during terminal message delivery") landed on `dev` at 12:45 UTC, eight minutes before this PR was opened, and resolves the same defect. My triage checks against `dev` were clean about an hour earlier and I did not re-check upstream before pushing.

That commit is also a superset of this branch:

- `dispatch` keeps the `deliverTerminal` result and cancels the request with the appropriate cause when delivery is interrupted, instead of closing it normally. This branch left that path closing normally, which the review correctly flagged as P1.
- `setOnBlocked` gives the tests a deterministic handshake for a saturated buffer, replacing the `time.Sleep` ordering used here — also the second P1, and the practice AGENTS.md:59 rules out.

Nothing in this branch adds anything on top of it, so there is no remaining delta worth rebasing. Sorry for the noise.

PR #5613, opened 2026-09-08
feat(management): add per-session affinity invalidation

RoySRose:
## Summary

- add `SessionCache.InvalidateMatching`, `MerklePrefixMatcher.InvalidateSession`, and `SessionAffinitySelector.InvalidateSession` so a single caller can be released from its bound credential
- expose it as `DELETE /v0/management/session-affinity?session_id=...`, returning `removed` plus per-layer counts
- match subagent identities (`claude:<uuid>:agent:<id>`) exactly, so releasing a parent session leaves the subagents underneath it bound

Session affinity has only one release mechanism today: `InvalidateAuth`, which moves *every* session off a credential. Moving one caller currently means either minting a new session id — which throws away that conversation's upstream prompt cache — or restarting the proxy, which throws away every binding and all the warm state with it. This adds the per-session counterpart alongside the existing auth-wide one.

Cache removal is group-wide rather than alias-wide: every alias in a group names the same logical session, so dropping only the requested alias would let the next request rediscover a surviving sibling and re-bind the caller to the same auth. The existing `Invalidate` remains for callers that want the surgical, sibling-preserving removal.

The endpoint returns 200 whether or not a binding existed, following `CancelAuthSession` — a caller cannot distinguish an unknown session from one whose TTL just lapsed, and the request is a pure removal, so making the no-op case an error would only break retries and fire-and-forget scripts. Authentication and 400 handling stay with the shared management middleware and `NormalizeExplicitID`.

## Tests

- `go build ./...`, `go test ./...` — 0 failures
- new unit tests cover the cache (group-wide removal, sibling isolation, ids containing `::`), the LCP matcher, the selector delegating to both layers, and the handler (200 with and without an existing binding, 400 on a missing id, 401 unauthenticated)
- exercised against a throwaway instance with two stub credentials: a session pinned to one credential across repeated requests, `DELETE` reported `removed=true`, the next request logged `cache miss, new binding` and re-selected from the pool, and a second session's binding was untouched

### DIFF
--- a/internal/api/handlers/management/session_affinity.go
+++ b/internal/api/handlers/management/session_affinity.go
@@ -1,0 +1,58 @@
+package management
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	coreauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+	coresession "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/session"
+)
+
+// sessionAffinityInvalidator is implemented by selectors that can release the
+// affinity bindings of a single session. It is declared here rather than
+// referencing the concrete selector so that pools configured with a plain
+// round-robin selector simply report the capability as unavailable.
+type sessionAffinityInvalidator interface {
+	InvalidateSession(sessionID string) coreauth.SessionInvalidation
+}
+
+// DeleteSessionAffinity releases the affinity bindings of one session so its
+// next request is re-selected from the pool.
+//
+// Operators otherwise have only two ways to move a single caller onto another
+// credential: give the caller a new session id, which throws away the upstream
+// prompt cache for that conversation, or restart the proxy, which throws away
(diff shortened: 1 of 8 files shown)


RoySRose:
Thanks — fixed in ea4c9059.

You are right that the key cannot be decoded: both the session id and the model may contain `::`, so taking the model from the last separator misreads a model that contains it, and taking it from the first would misread a session id that does.

Rather than change the key representation (which is written and read in several places), the invalidation path no longer decodes at all. The caller already knows the session id it wants released, so the match is now a prefix test of that id against the remainder after the provider — exact for both shapes, since the session id is the first field after the provider.

One case stays inherently ambiguous: `claude::s1::foo::bar` is a valid key for session `s1` with model `foo::bar` **and** for session `s1::foo` with model `bar`. It is now owned by both. Removal is the safe resolution — a spurious release costs one re-selection, while a missed release leaves the caller pinned, which is the failure this endpoint exists to fix. That case is covered by a test.

PR #5614, opened 2026-09-08
feat(routing): per-model credential weights via model_weights

RoySRose:
## Summary

- add an optional `model_weights` map to a credential's auth file: `{"model_weights": {"claude-opus-5": 0, "claude-fable-5-1": 10000}}`
- `authWeightForModel(auth, model)` resolves the table for the requested model and falls back to the existing account-wide `weight` when the model is absent, so files without the field behave exactly as before
- a value of `0` removes that credential from the pool **for that model only**, leaving it eligible for every other model
- expose the field through the management API: `GET /v0/management/auth-files` reports it, `PATCH /v0/management/auth-files/fields` accepts an object or `null` to clear
- picked up by the existing auth-file watcher, so operators can retune weights without a restart

Today `weight` is per credential, so an operator who wants one account to absorb most of the traffic for one model has no way to say so without changing that account's share for every other model. That is the common case when different models on the same account have separate quota windows: a credential can be nearly exhausted on one model family while still having plenty of room on another, and the single knob forces a choice between starving the healthy family or over-driving the exhausted one.

Keys are matched on the lowercase canonical model id (the thinking-suffix form is normalised first, so `claude-opus-5[1m]`-style variants resolve to their base entry). An unparsable `model_weights` value causes that auth file to be skipped with a warning rather than being silently treated as unweighted — a malformed routing table should be loud, not quietly ignored.

## Tests

- `go build ./...`, `go test ./...` — 0 failures
- new unit tests cover resolution and fallback, `0` excluding one model while others stay eligible, case and thinking-suffix normalisation, malformed values, and the management read/patch/clear round trip
- exercised against a throwaway instance with two stub credentials: with `model_weights` steering one model to the second credential, 20/20 requests for that model landed there while 20/20 requests for another model still landed on the first, and editing the file at runtime flipped the routing without a restart

### DIFF
--- a/internal/api/handlers/management/auth_files.go
+++ b/internal/api/handlers/management/auth_files.go
@@ -275,6 +275,11 @@ func (h *Handler) listAuthFilesFromDisk(c *gin.Context) {
 						}
 					}
 				}
+				if mv := gjson.GetBytes(data, coreauth.AttributeModelWeights); mv.Exists() && mv.IsObject() {
+					if modelWeights, errModelWeights := coreauth.ParseModelWeights(mv.Raw); errModelWeights == nil && len(modelWeights) > 0 {
+						fileData[coreauth.AttributeModelWeights] = modelWeights
+					}
+				}
 				if nv := gjson.GetBytes(data, "note"); nv.Exists() && nv.Type == gjson.String {
 					if trimmed := strings.TrimSpace(nv.String()); trimmed != "" {
 						fileData["note"] = trimmed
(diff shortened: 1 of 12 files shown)


RoySRose:
Thanks — fixed in 7c6ccec8.

Colliding entries now fail the whole table with an error naming both spellings and the key they share, instead of letting map iteration order pick a winner. The flip you describe (0 vs a positive weight moving the credential between excluded and active on reload, with nothing changing on disk) is exactly why silently keeping one was wrong; an ambiguous routing rule has no correct resolution at load time.

Covered by tests for the three real collision sources: case (`Claude-Opus-5`), surrounding whitespace, and the thinking suffix (`gpt-5.6-sol(high)` vs `gpt-5.6-sol`).

Small correction to the example: the bracketed context form (`claude-opus-5[1m]`) does **not** normalize to the base model, so it is not a source of collisions.

PR #5618, opened 2026-09-08
feat(sdk): observe final HTTP attempts without replacing provider transports

### LINKED ISSUE #5616: SDK: expose a non-replacing outbound HTTP transport decorator for safe attempt observation
Embedding consumers need secret-safe per-attempt telemetry of final provider request controls after translation, without enabling full request logging or replacing provider-specific transports. Current Manager.SetRoundTripperProvider replaces transports: NewUtlsHTTPClient replaces Chrome/Anthropic fingerprints when a context RoundTripper is supplied, proxy configuration may bypass it, and Antigravity owns credential-scoped HTTP/1.1 pools. Therefore it is not a safe universal observer seam.

Please expose a request-context-scoped transport decorator applied AFTER each helper has chosen its actual transport, preserving proxy priority, uTLS, credential pools and protocol pins. A callback receiving the selected http.RoundTripper and returning a wrapping http.RoundTripper is sufficient; consumers can capture only allow-listed model/thinking fields and status, keeping prompts/auth/raw errors out of logs. It must not enable RequestLog or change retry/selection behavior. Factories: helps.NewProxyAwareHTTPClient, helps.NewUtlsHTTPClient, newAntigravityHTTPClient (after HTTP/1.1 selection); composition must avoid double wrapping. WebSocket payload observation is a separate scope and must not be misrepresented as HTTP coverage.

Acceptance: nil callback preserves exact transport identity; wrapping preserves fingerprints/proxy/H1/pooling; request context isolation and race tests; final payload and stream lifecycle visible to wrapper. I can contribute a focused PR for this extension.

jaaacki:
Closes #5616.

Adds sdk/httptransport.WithDecorator + Decorate. Embedders can wrap the executor-selected RoundTripper after proxy/TLS/pool selection rather than replacing it through the conductor override. Nil decorators preserve exact existing client identity. Client copies preserve timeouts/redirect/jar policy and reuse the original transport/pool. No request logging is enabled; callbacks are context-local and documented as concurrency-safe.

Applied once at the end of plain proxy-aware, uTLS, and Antigravity HTTP client construction. Antigravity suppresses intermediate helper decoration before final HTTP/1.1/credential-pool policy is applied. WebSocket payload observation is explicitly outside this HTTP seam.

Validation (Docker golang:1.26.6; no host builds; no real provider calls):
- go test -race ./sdk/httptransport ./internal/runtime/executor/helps -run "TestDecorator|TestClientDecorators" — PASS
- go test ./internal/runtime/executor -run TestAntigravityDecorator — PASS
- go build -o /out/cli-proxy-api ./cmd/server — PASS
- gofmt and git diff --check — PASS

Tests pin nil behavior, selected transport identity, protected TLS family retention, proxy selection, HTTP/1.1 and credential pool reuse, and exactly-once decorator application.

### DIFF
--- a/internal/runtime/executor/antigravity_executor.go
+++ b/internal/runtime/executor/antigravity_executor.go
@@ -22,6 +22,7 @@ import (
 	antigravityclaude "github.com/router-for-me/CLIProxyAPI/v7/internal/translator/antigravity/claude"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/util"
 	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+	"github.com/router-for-me/CLIProxyAPI/v7/sdk/httptransport"
 	"github.com/router-for-me/CLIProxyAPI/v7/sdk/proxyutil"
 	sdktranslator "github.com/router-for-me/CLIProxyAPI/v7/sdk/translator"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


jaaacki:
Additional embedding integration finding before adoption: public HTTP handlers call GetContextWithCancel(..., context.Background()). That helper copies selected request metadata but does not currently carry arbitrary values from c.Request.Context(). Therefore WithDecorator on an incoming request context alone would not reach these helpers. The SDK API should explicitly inherit the decorator into the execution parent (while preserving an explicitly supplied parent decorator/cancellation/deadlines), with a GetContextWithCancel regression test. Direct Manager.Execute users already preserve the passed context; HTTP-handler embedders need this additional bridge. Our consumer rollout is not claiming universal SDK coverage until the hook plus this propagation path are reviewed/released.

PR #5625, opened 2026-09-08
fix(plugin-store): reduce release checks and honor shared GitHub rate-limit cooldowns

### LINKED ISSUE #5608: 插件商店对每个插件获取版本导致api限流被打满
[x] No, it's another issue.

**Describe the bug**
查看现有插件商店时，会进行大量查询，直接打满GitHub匿名api单ip60次每小时的限制

**Expected behavior**
对还未安装的插件不再查询版本或使用缓存，遵循GitHub返回的限流冷却

LTbinglingfeng:
## Summary

Fixes #5608.

Reduce unnecessary GitHub API requests when browsing the plugin store, coalesce release checks across concurrent listings, and honor GitHub rate-limit cooldowns across plugin-store clients. Also propagate proxy/egress identity through the SDK and Home plugin-sync paths so one exit IP's cooldown does not incorrectly block another exit IP.

## Problem

Previously, listing the plugin store queried the latest GitHub release for every catalog entry, including uninstalled plugins and sources that could not update an installed plugin. A sufficiently large catalog could exhaust GitHub's unauthenticated limit of 60 requests per hour per IP just by being opened.

The existing per-repository cache did not coalesce in-flight requests or bound concurrency across listings. Failed lookups used a fixed retry interval instead of GitHub's cooldown information, allowing subsequent listings or installations to keep hitting an already rate-limited API.

### 1. Only check releases that can update installed plugins

- Query latest releases only for installed plugins whose catalog source is allowed to update them.
- Skip uninstalled plugins, direct-download plugins, and unrelated or unverifiable update sources.
- Preserve catalog ordering and positional version results.
- Use the registry version when no successful release lookup is available. Previously cached successful versions can still be displayed without issuing another request, including for uninstalled plugins.

### 2. Share and throttle release lookups

- Coalesce concurrent lookups for the same repository, effective credentials, and network scope.
- Limit active release checks to **two per management handler**, shared across catalog requests.
- Cache successful lookups for **one hour**; retry ordinary failures after **30 seconds**. Start these intervals when the lookup finishes, not when it enters the queue.
- Retain the last successful version when a refresh fails rather than replacing it with an empty result.
- Handle cancellation without poisoning the failure cache: a canceled follower does not cancel the leader, and a live follower can take over after a canceled leader.
- Bind the cache identity and initial release request to the same credential snapshot, including when environment-provided credentials change while a request is queued.

### 3. Honor shared GitHub API cooldowns

- Share cooldowns across clients, repositories, release metadata, and API-hosted asset downloads for the same effective credentials and network scope.
- Recognize HTTP 429 and rate-limit-related HTTP 403 responses using `Retry-After`, `X-RateLimit-Remaining`, `X-RateLimit-Reset`, and recognized GitHub rate-limit messages.
- Use exponential backoff from **one minute up to one hour** when a rate-limit rejection has no usable retry time.
- Record successful responses that consume the final request, while still returning their data. Late successful responses do not clear an active cooldown.
- Reject new requests locally during a cooldown, including queued release checks. Ordinary non-rate-limit 403 responses and non-GitHub hosts do not establish a GitHub cooldown.
- Apply header-based cooldowns before reading an error body, and avoid exposing authenticated response bodies or credentials in rate-limit errors.
- Prune inactive cooldown identities after a grace period.
- Return HTTP **429** with `Retry-After` and `plugin_store_rate_limited`, `retry_after`, and `retry_at` fields when installation encounters a rate limit, including during registry resolution.
- Stop alternate release-tag installation attempts immediately on rate-limit errors or request cancellation.

### 4. Preserve proxy/egress isolation in SDK and Home clients

- Add the backward-compatible `Client.WithNetworkScope()` SDK method; existing constructor signatures remain unchanged.
- Pass the normalized proxy URL as the network scope in both Home client creation paths: configured authentication and temporary resolved authentication.
- Keep cooldowns shared for the same credentials and exit configuration, but isolated between distinct proxy scopes and the direct-connection scope.
- `WithNetworkScope()` identifies the egress; it does **not** configure the HTTP transport. SDK embedders using custom proxies should set a matching scope explicitly.

## User-visible behavior

- Browsing an uninstalled catalog no longer triggers per-plugin GitHub release checks; fetching the configured registries is still required.
- Installed plugins retain update checks, restricted to the applicable source and subject to caching and concurrency limits.
- Release-check failures or cooldowns do not discard previously known versions.
- Installation receives actionable retry timing instead of repeatedly attempting requests during a known cooldown.
- Switching Home's configured proxy uses the corresponding cooldown identity rather than inheriting another exit configuration's state.

## Validation

Completed locally:

- [x] `gofmt -w .`
- [x] `go test ./...`
- [x] `go test -race ./sdk/pluginstore ./internal/homeplugins ./internal/pluginstore ./internal/api/handlers/management`
- [x] `git diff --check`

Regression coverage includes catalog filtering, source matching, successful-version retention, completion-based cache intervals, concurrent lookup coalescing, cancellation handoff, shared concurrency limits, credential snapshots and expiry, GitHub cooldown parsing/backoff/recovery, queued-request suppression, installation retry responses, and SDK/Home network-scope isolation and sharing.

The Home network-scope regression test was also run against the pre-fix Home implementation using a Go overlay: both client paths failed by issuing proxy requests instead of observing the existing scoped cooldown. The same test passes with this fix.

Tests use mocked responses or a local test proxy; they do not intentionally exhaust a live GitHub API quota.

### DIFF
--- a/internal/api/handlers/management/handler.go
+++ b/internal/api/handlers/management/handler.go
@@ -58,8 +58,7 @@ type Handler struct {
 	configReloadHook        func(context.Context, *config.Config)
 	pluginStoreRegistryURL  string
 	pluginStoreHTTPClient   pluginstore.HTTPDoer
-	pluginReleaseCacheMu    sync.Mutex
-	pluginReleaseCache      map[string]pluginReleaseCacheEntry
+	pluginReleases          pluginReleaseCache
 }
 
 type configReloadSnapshot struct {
--- a/internal/api/handlers/management/handler.go
+++ b/internal/api/handlers/management/handler.go
@@ -58,6 +58,7 @@ type Handler struct {
 	configReloadHook        func(context.Context, *config.Config)
 	pluginStoreRegistryURL  string
 	pluginStoreHTTPClient   pluginstore.HTTPDoer
+	pluginStoreRateLimiter  *pluginstore.GitHubRateLimiter
 	pluginReleases          pluginReleaseCache
 }
 
--- a/internal/api/handlers/management/plugin_store.go
+++ b/internal/api/handlers/management/plugin_store.go
@@ -9,8 +9,6 @@ import (
 	"net/http"
 	"runtime"
 	"strings"
-	"sync"
-	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
@@ -22,20 +20,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
-
-const (
-	// pluginReleaseCacheTTL bounds how long a resolved latest release version is
-	// reused before the GitHub API is queried again.
-	pluginReleaseCacheTTL = 10 * time.Minute
-	// pluginReleaseFailureCacheTTL throttles retries after a failed lookup so a
-	// rate-limited or unreachable API is not hammered on every listing.
-	pluginReleaseFailureCacheTTL = 30 * time.Second
-)
-
-type pluginReleaseCacheEntry struct {
-	version   string
-	expiresAt time.Time
-}
 
 type pluginStoreListResponse struct {
 	PluginsEnabled bool                   `json:"plugins_enabled"`
@@ -186,6 +170,8 @@ func (h *Handler) ListPluginStore(c *gin.Context) {
 		storeVersion := plugin.Version
 		if latestVersions[index] != "" {
 			storeVersion = latestVersions[index]
+		} else if cachedVersion := h.pluginReleases.cached(client, plugin); cachedVersion != "" {
+			storeVersion = cachedVersion
 		}
 		entries = append(entries, pluginStoreListEntry{
 			StoreID:             htmlsanitize.String(item.source.ID + "/" + plugin.ID),
@@ -542,13 +528,13 @@ func (h *Handler) newPluginStoreClient(proxyURL string, registryURL string, stor
 		registryURL = pluginstore.DefaultRegistryURL
 	}
 	if httpClient != nil {
-		return pluginstore.Client{HTTPClient: httpClient, RegistryURL: registryURL, Auth: storeAuth}
+		return pluginstore.Client{HTTPClient: httpClient, NetworkScope: strings.TrimSpace(proxyURL), RegistryURL: registryURL, Auth: storeAuth}
 	}
 	client := &http.Client{}
 	if strings.TrimSpace(proxyURL) != "" {
 		util.SetProxy(&sdkconfig.SDKConfig{ProxyURL: strings.TrimSpace(proxyURL)}, client)
 	}
-	return pluginstore.Client{HTTPClient: client, RegistryURL: registryURL, Auth: storeAuth}
+	return pluginstore.Client{HTTPClient: client, NetworkScope: strings.TrimSpace(proxyURL), RegistryURL: registryURL, Auth: storeAuth}
 }
 
 func (h *Handler) fetchSourcedPlugins(ctx context.Context, proxyURL string, storeAuth []pluginstore.AuthConfig, sources []pluginstore.Source) ([]sourcedPlugin, []pluginStoreSourceErr) {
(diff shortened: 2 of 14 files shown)


sususu98:
Thank you for this comprehensive PR. It significantly improves GitHub rate-limit handling, request coalescing, and network scope isolation across the plugin store.

During an in-depth review of the changes, we identified a few concrete issues that are worth addressing:

### 1. `ListPluginStore` does not propagate `RateLimitError` on registry failure
- **Location**: `internal/api/handlers/management/plugin_store.go:131-136`
- **Problem**: In `InstallPluginFromStore` (and `findPluginStoreInstallTarget`), rate limit errors encountered during registry fetching or release installation are handled via `writePluginStoreRateLimit(c, ...)` to return HTTP 429 with `Retry-After`. However, in `ListPluginStore`, when registry fetching fails, it unconditionally returns HTTP 502 `plugin_store_registry_failed` without checking for rate limit errors:
  ```go
  plugins, sourceErrors := h.fetchSourcedPlugins(c.Request.Context(), proxyURL, storeAuth, sources)
  if len(plugins) == 0 && len(sourceErrors) > 0 {
      c.JSON(http.StatusBadGateway, gin.H{"error": "plugin_store_registry_failed", "message": sourceErrors[0].Message})
      return
  }
  ```
- **Suggestion**: Check `sourceErrors` with `writePluginStoreRateLimit` so that clients browsing the catalog receive actionable retry timing instead of a generic 502:
  ```go
  if len(plugins) == 0 && len(sourceErrors) > 0 {
      for _, srcErr := range sourceErrors {
          if writePluginStoreRateLimit(c, srcErr.cause) {
              return
          }
      }
      c.JSON(http.StatusBadGateway, gin.H{"error": "plugin_store_registry_failed", "message": sourceErrors[0].Message})
      return
  }
  ```

---

### 2. Multi-source error scanning only checks `sourceErrors[0]` for rate limiting
- **Location**: `internal/api/handlers/management/plugin_store.go:626-630`
- **Problem**: When multiple sources are configured and none match the target plugin, the error handling only checks `sourceErrors[0].cause`:
  ```go
  if len(matches) == 0 {
      if len(plugins) == 0 && len(sourceErrors) > 0 {
          if writePluginStoreRateLimit(c, sourceErrors[0].cause) {
              return pluginstore.Source{}, pluginstore.Plugin{}, pluginstore.Client{}, false
          }
          c.JSON(http.StatusBadGateway, gin.H{"error": "plugin_store_registry_failed", "message": sourceErrors[0].Message})
          return pluginstore.Source{}, pluginstore.Plugin{}, pluginstore.Client{}, false
      }
  ```
  If `sourceErrors[0]` failed due to a regular network issue (e.g. DNS or connection error) but a subsequent source was rate-limited by GitHub (`RateLimitError`), the 429 status is masked and the endpoint returns 502 instead.
- **Suggestion**: Iterate through `sourceErrors` to prioritize any `RateLimitError` before falling back to 502:
  ```go
  for _, srcErr := range sourceErrors {
      if writePluginStoreRateLimit(c, srcErr.cause) {
          return pluginstore.Source{}, pluginstore.Plugin{}, pluginstore.Client{}, false
      }
  }
  ```

---

### 3. `sdk/pluginstore` does not export `RateLimitError` for external consumers
- **Location**: `sdk/pluginstore/pluginstore.go` & `internal/pluginstore/github_rate_limit.go:14-19`
- **Problem**: `RateLimitError` is defined in `internal/pluginstore`. External consumers importing the public SDK `github.com/router-for-me/CLIProxyAPI/v7/sdk/pluginstore` cannot import the `internal` package due to Go's visibility rules (notice `sdk/pluginstore/network_scope_test.go:48` had to import `internalpluginstore`). External callers using `client.FetchLatestRelease` or `client.InstallManifest` have no exported type to do `errors.As(err, &rateLimit)` to access `RetryAt` or `RetryAfterSeconds()`.
- **Suggestion**: Export a type alias in `sdk/pluginstore/pluginstore.go`:
  ```go
  // RateLimitError reports a GitHub API cooldown.
  type RateLimitError = pluginstore.RateLimitError
  ```

---

### 4. `pluginReleaseCache.entries` map has no expiration/eviction mechanism
- **Location**: `internal/api/handlers/management/plugin_store_release.go:32-38, 159-164`
- **Problem**: `GitHubRateLimiter` has an hourly prune (`pruneLocked`) to drop stale entries. In contrast, `pluginReleaseCache.entries` only ever grows (`cache.entries[key] = entry`) and has no prune or TTL eviction. Because cache keys incorporate `(repository + token + networkScope)`, switching proxies, rotating tokens, or inspecting many plugins will accumulate entries indefinitely in long-running processes.
- **Suggestion**: Add a periodic or lazy cleanup for `cache.entries` to evict entries whose `nextCheckAt` is well in the past (e.g. older than 24 hours).

PR #5636, opened 2026-09-08
fix(auth): cool explicit model_not_found failures

### LINKED ISSUE #5635: Codex streamed model_not_found bypasses per-model cooldown
## Summary

A structured Codex failure with `error.type=invalid_request_error` and `error.code=model_not_found` is classified as request-scoped before the auth manager records the result.

That classification causes `shouldSkipCredentialCooldown` to return true, so the existing per-credential/per-model cooldown is skipped. Session affinity is also preserved, allowing later requests for the same session to keep selecting the credential that cannot execute the model.

## Reproduction

Use a Codex OAuth credential that lists a model but rejects execution with a terminal stream error such as:

```json
{"error":{"type":"invalid_request_error","code":"model_not_found","message":"The model gpt-5.5 does not exist or you do not have access to it."}}
```

The error may be exposed with HTTP-like status 400 or 404. After the failure, the credential/model pair remains eligible instead of entering the existing model-support cooldown.

## Expected behavior

An explicit structured `model_not_found` should:

- retain the machine-readable `model_not_found` classification;
- not be treated as a caller-attributed request fault;
- enter the existing long per-model cooldown for the selected credential;
- allow session affinity to release the failed credential.

Generic 404 responses without an explicit model-not-found identifier should keep their separate transient handling.

## Proposed fix

Recognize `isExplicitModelNotFoundError` before generic request-fault classification, preserve the error code, and treat it as a model-support result for the existing cooldown path.

This is complementary to #5477 and #5514: those address generic/transient 404 cooldown duration, while this issue covers explicit `model_not_found` responses that currently bypass cooldown entirely.

eshao731:
## Summary

- preserve the `model_not_found` code before generic request-fault classification;
- keep explicit model-not-found responses eligible for the existing per-credential/per-model cooldown;
- prevent `invalid_request_error + model_not_found` from being treated as request-scoped;
- add regression coverage for both 400 and 404 representations of the streamed error.

## Problem

Codex can return a terminal stream error with `type=invalid_request_error` and `code=model_not_found`. The generic request-fault classifier currently wins, causing `shouldSkipCredentialCooldown` to skip model cooling and preserve session affinity. A credential that cannot execute the requested model can therefore remain selected repeatedly.

The model-not-found identifier is more specific than the generic error type and should use the existing model-support cooldown path.

This does not change generic 404 handling and is complementary to #5514.

## Verification

- `go test ./sdk/cliproxy/auth -count=1`
- `go build -o /tmp/cli-proxy-api-upstream-pr-check ./cmd/server`
- `git diff --check origin/main...HEAD`

Fixes #5635

### DIFF
--- a/sdk/cliproxy/auth/conductor_cooldown.go
+++ b/sdk/cliproxy/auth/conductor_cooldown.go
@@ -1434,6 +1434,8 @@ func resultErrorFromError(err error) *Error {
 		resultErr.HTTPStatus = statusCodeFromError(err)
 	}
 	switch {
+	case isExplicitModelNotFoundError(err, ""):
+		resultErr.Code = "model_not_found"
 	case isRequestScopedError(err) || isRequestInvalidError(err):
 		// Prefer true request-scoped faults (including Claude OAuth cancellation)
 		// over the broader connection-lifecycle classification.
@@ -1660,6 +1662,9 @@ func isInvalidGrantResultError(err *Error) bool {
 func isModelSupportResultError(err *Error) bool {
 	if err == nil {
 		return false
+	}
+	if isExplicitModelNotFoundError(err, "") {
+		return true
 	}
 	status := statusCodeFromResult(err)
 	if status != http.StatusBadRequest && status != http.StatusUnprocessableEntity {
@@ -1976,6 +1981,9 @@ func isRequestInvalidError(err error) bool {
 	}
 	if isRequestScopedError(err) {
 		return true
+	}
+	if isExplicitModelNotFoundError(err, "") {
+		return false
 	}
 	if isCloudflareChallengeError(err) {
 		return false
--- a/sdk/cliproxy/auth/model_not_found_cooldown_test.go
+++ b/sdk/cliproxy/auth/model_not_found_cooldown_test.go
@@ -1,0 +1,68 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+	"testing"
+	"time"
+)
+
+type modelNotFoundStatusError struct {
+	status int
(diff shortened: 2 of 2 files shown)


surpaimb:
I prepared a tested variant and additional regressions in [surpaimb/CLIProxyAPI@fc9b5c5](https://github.com/surpaimb/CLIProxyAPI/commit/fc9b5c563cc93c3fa0bfb6e64f1d222c58269344) to help fold the remaining guards into this PR, without opening a duplicate.

At the current head (6f389f9), two classification boundaries need protection:

1. As the existing review notes, check the executor's explicit typed RequestScopedError marker before examining model_not_found in a message or wrapped cause. It must remain availability-neutral and preserve session affinity.
2. Use structured code/type identifiers for this override. The existing isExplicitModelNotFoundError helper also matches message text, so an unrelated request error such as {"error":{"type":"invalid_request_error","code":"invalid_value","message":"Unsupported enum value: model_not_found"}} can incorrectly enter model-support cooldown.

The variant preserves typed request-scope precedence, then checks identifiers in known error envelopes and wrapped/joined errors. Its tests cover 400/404 representations, the existing 12-hour model cooldown, sibling-model availability, actual failed-affinity removal, typed/wrapped request errors, misleading messages/metadata, generic 404s, and previous_response_not_found. No quota policy changes.

The new regressions fail on unmodified v7.2.155 and pass with the variant. The complete auth package, server build, go vet, and formatting/diff checks pass. Combined with the independent capacity fix in #5655, all 92 tested packages and the focused bootstrap/auth race suite pass.

PR #5653, opened 2026-09-09
feat(codex): support GPT Image 2.5 Flare and Sunburst

dinhkarate:
## Summary

- add `gpt-image-2.5-flare` and `gpt-image-2.5-sunburst` as Codex image-only built-in models
- route both models through the existing Codex Responses API `image_generation` tool path
- preserve direct `/images/*` routing for `gpt-image-1.5` and `gpt-image-2`
- hide image-only entries from the Codex chat model picker
- strip provider prefixes from the upstream image tool model while retaining them for main-model routing

## Upstream verification

OpenAI documents both model IDs for Image API and Responses API image generation:

- https://developers.openai.com/api/docs/guides/image-generation
- https://developers.openai.com/api/docs/models/gpt-image-2.5-flare
- https://developers.openai.com/api/docs/models/gpt-image-2.5-sunburst

Both IDs were also verified against the ChatGPT Codex Responses upstream with an OAuth account:

- `gpt-image-2.5-flare`: HTTP 200 with an image generation event
- `gpt-image-2.5-sunburst`: HTTP 200 with an image generation event

The Codex `/models` endpoint currently omits image tool models, matching the existing reason `gpt-image-1.5` and `gpt-image-2` are injected as built-ins.

## Tests

- `go test ./...`
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o /tmp/cli-proxy-api-gpt-image-2.5 ./cmd/server`


lowlandsheperd:
不考虑默认模型改为gpt-5.6-luna吗

PR #5675, opened 2026-09-09
fix(antigravity): keep thinking block open on empty text parts in claude stream

### LINKED ISSUE #5674: fix(antigravity): empty text part while thinking is open emits content_block_stop for an index that was never started (Claude Code: Content block not found)
[ ] Yes, this is a request payload issue.
[x] No, it's another issue.

**Describe the bug**

Streaming `/v1/messages` through the Antigravity executor with a gemini model: Claude Code aborts the turn with `API Error: Content block not found` after a long thinking phase. CLIProxyAPI logs the request as `200`. Seen 5 times since 2026-08-24 on gemini-3.7-flash-high / gemini-3.8-flash-high, always on a response that only produced thinking so far, always after the stream had been idle for 15-30s.

Root cause is in `internal/translator/antigravity/claude/antigravity_claude_response.go` (text part branch, around line 324). For a non-thought text part with no `finishReason`, the translator closes the active block and increments `ResponseIndex`, but only opens a new text block when the text is non-empty. Gemini sends `{"text":""}` parts while it is still thinking, so after such a part `ResponseIndex` has moved on while `ResponseType` still says thinking. The trailing `content_block_stop` from `appendFinalEvents` (or the next `thinking_delta`) then targets an index that never got a `content_block_start`, and Claude Code throws `RangeError: Content block not found`.

Reproducer as a unit test (fails on main):

```go
func TestConvertAntigravityResponseToClaudeStream_EmptyTextPartKeepsThinkingBlockOpen(t *testing.T) {
	requestJSON := []byte(`{"model":"gemini-3-flash-agent"}`)
	thinkingChunk := []byte(`{"response":{"candidates":[{"content":{"role":"model","parts":[{"text":"Thinking","thought":true}]}}],"responseId":"resp-1"}}`)
	emptyTextChunk := []byte(`{"response":{"candidates":[{"content":{"role":"model","parts":[{"text":""}]}}]}}`)
	finishChunk := []byte(`{"response":{"candidates":[{"content":{"role":"model","parts":[{"text":""}]},"finishReason":"STOP"}],"usageMetadata":{"promptTokenCount":100,"thoughtsTokenCount":20,"totalTokenCount":120}}}`)
	// feed the four chunks (incl. "[DONE]") and assert every content_block_delta/stop index had a content_block_start
}
```

Output on main:

```
content_block_start   index 0  thinking
content_block_delta   index 0
content_block_stop    index 0
content_block_stop    index 1   <- never started
message_delta / message_stop
```

Intended fix (only close the active block when a text block will actually follow):

```go
						} else if partText != "" {
							if params.ResponseType != 0 {
								appendEvent("content_block_stop", fmt.Sprintf(`{"type":"content_block_stop","index":%d}`, params.ResponseIndex))
								params.ResponseIndex++
							}
							appendEvent("content_block_start", fmt.Sprintf(`{"type":"content_block_start","index":%d,"content_block":{"type":"text","text":""}}`, params.ResponseIndex))
							data, _ := sjson.SetBytes([]byte(fmt.Sprintf(`{"type":"content_block_delta","index":%d,"delta":{"type":"text_delta","text":""}}`, params.ResponseIndex)), "delta.text", partText)
							appendEvent("content_block_delta", string(data))
							params.ResponseType = 1
							params.HasContent = true
						}
```

With that change the new test passes and `go test ./...` is green. The `gemini/claude` translator does not have this problem because it always opens a new text block after closing one.

**CLI Type**
claude code (Antigravity OAuth upstream)

**Model Name**
gemini-3.8-flash-high, gemini-3.7-flash-high

**LLM Client**
claude code 2.1.238 via Agent SDK

**Request Information**
Any streaming `/v1/messages` request where the model thinks long enough for Gemini to emit an empty text part before the answer. Not deterministic from the client side.

**Expected behavior**
Every `content_block_delta` / `content_block_stop` refers to an index that was started.

**OS Type**
- OS: Linux (docker `eceasy/cli-proxy-api:latest`, v7.2.155 commit 7fac6b1)

riba2534:
## Summary

Claude Code aborts a streamed `/v1/messages` response through the Antigravity executor with `API Error: Content block not found` when Gemini sends an empty `{"text":""}` part while a thinking block is still open.

## Why

In `ConvertAntigravityResponseToClaude`, a non-thought text part with no `finishReason` closes the active block and bumps `ResponseIndex`, but only opens a new text block when the text is non-empty. For an empty part the index moves on while `ResponseType` still says "thinking", so the next `content_block_stop` (from `appendFinalEvents`) or `thinking_delta` targets an index that never received `content_block_start`:

```
content_block_start   index 0  thinking
content_block_delta   index 0
content_block_stop    index 0
content_block_stop    index 1   <- never started
message_delta / message_stop
```

Gemini emits these empty text parts during long thinking (observed on gemini-3.7/3.8-flash with the stream idle for 15-30s before the failure). The fix skips the close when no text block will follow. The `gemini/claude` translator is not affected because it always opens a new text block after closing.

## Verification

- New `TestConvertAntigravityResponseToClaudeStream_EmptyTextPartKeepsThinkingBlockOpen` fails on main at the `content_block_stop for index 1` assertion and passes with this change.
- `go test ./...` and `go build -o test-output ./cmd/server` pass.

Closes #5674

### DIFF
--- a/internal/translator/antigravity/claude/antigravity_claude_response.go
+++ b/internal/translator/antigravity/claude/antigravity_claude_response.go
@@ -326,18 +326,20 @@ func ConvertAntigravityResponseToClaude(ctx context.Context, _ string, originalR
 							data, _ := sjson.SetBytes([]byte(fmt.Sprintf(`{"type":"content_block_delta","index":%d,"delta":{"type":"text_delta","text":""}}`, params.ResponseIndex)), "delta.text", partText)
 							appendEvent("content_block_delta", string(data))
 							params.HasContent = true
-						} else {
+						} else if partText != "" {
+							// Only close the active block when a text block actually follows.
+							// Closing it for an empty text part would advance the index while
+							// leaving ResponseType set, so the next event targets a block
+							// that was never started.
 							if params.ResponseType != 0 {
 								appendEvent("content_block_stop", fmt.Sprintf(`{"type":"content_block_stop","index":%d}`, params.ResponseIndex))
 								params.ResponseIndex++
 							}
-							if partText != "" {
-								appendEvent("content_block_start", fmt.Sprintf(`{"type":"content_block_start","index":%d,"content_block":{"type":"text","text":""}}`, params.ResponseIndex))
-								data, _ := sjson.SetBytes([]byte(fmt.Sprintf(`{"type":"content_block_delta","index":%d,"delta":{"type":"text_delta","text":""}}`, params.ResponseIndex)), "delta.text", partText)
-								appendEvent("content_block_delta", string(data))
-								params.ResponseType = 1
-								params.HasContent = true
-							}
+							appendEvent("content_block_start", fmt.Sprintf(`{"type":"content_block_start","index":%d,"content_block":{"type":"text","text":""}}`, params.ResponseIndex))
+							data, _ := sjson.SetBytes([]byte(fmt.Sprintf(`{"type":"content_block_delta","index":%d,"delta":{"type":"text_delta","text":""}}`, params.ResponseIndex)), "delta.text", partText)
+							appendEvent("content_block_delta", string(data))
+							params.ResponseType = 1
+							params.HasContent = true
 						}
 					}
 					if partText != "" {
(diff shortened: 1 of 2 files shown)


riba2534:
Closing: `translator-path-guard` rejects any change under `internal/translator/**` on contributor PRs. The repro, root cause, and patch are on #5674 for the maintenance team to land.

PR #5694, opened 2026-09-09
fix(session): recognize OpenCode conversation headers and document forwarding

cookerpapa:
Current Pi clients send `X-Opencode-Session` for OpenCode requests, but CLIProxyAPI only recognizes `X-Session-Affinity`. A native-header-only conversation therefore falls back to prompt-derived identity; separate conversations with identical opening prompts can share that identity, and compaction can change it.

This recognizes `X-Opencode-Session` in both explicit-session detection and hierarchy extraction, reusing the existing `affinity:` namespace and preserving `X-Session-Affinity` precedence. It uses the existing validation for case-insensitive, multi-value, empty, control-bearing and oversized headers. No new upstream session is invented.

The accompanying example documents existing `$Header-Name` forwarding for OpenCode Go, Claude Code and Codex, including real user-agent forwarding and the distinction between local affinity and sending the header upstream. The example uses a documented Chat Completions model and explains protocol selection and missing-header behavior.

Refs #5690. Automatic metadata-to-header generation remains a separate request.

Validation:
- Reproduced missing native-header extraction and absent canonical identity before the fix; regression tests now pass.
- Complete session, auth, and API-handler package tests pass.
- Focused session/auth tests pass with `-race -count=20`.
- Existing custom-header forwarding tests pass; both YAML examples parse.
- Server build passes with `-buildvcs=false` (this environment cannot obtain VCS status).
- Requirements and model endpoint verified against [OpenCode Go documentation](https://opencode.ai/docs/go/). No paid-provider request was made.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -358,6 +358,7 @@ nonstream-keepalive-interval: 0
 #       # Values starting with "$" dynamically copy the header value from downstream client requests.
 #       # If the client did not send the specified header, the header is omitted.
 #       # X-Claude-Code-Session-Id: "$ABC" # copies client's "ABC" header
+#       # OpenCode Go client/session header examples: examples/opencode-go/README.md
 #     proxy-url: "socks5://proxy.example.com:1080"
 #     # proxy-url: "direct" # optional: explicit direct connect for this credential
 #     models:
--- a/examples/opencode-go/README.md
+++ b/examples/opencode-go/README.md
@@ -1,0 +1,73 @@
+# OpenCode Go through CLIProxyAPI
+
+OpenCode Go asks coding clients to identify themselves and send a stable
+conversation ID for routing and prompt caching. Its [client requirements](https://opencode.ai/docs/go/#where-can-i-use-it)
+also document support for Claude Code's and Codex's native session headers.
+
+CLIProxyAPI can forward those headers with its existing dynamic custom-header
+configuration. A `$Header-Name` value copies that incoming request header; when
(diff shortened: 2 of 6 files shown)


quangnguyenbh:
@luispater @sususu98 pinging you both on this one — confirming it is still needed after the v7.2.156 / v7.2.157 releases.

The maintainer commit `3bf787fc` ("feat(auth): propagate canonical session id for custom header expansion", shipped in **v7.2.156**) covers the metadata → `$CPA-SESSION-ID` *forwarding* path from #5690, but it does not cover this PR's core change: recognizing the native `X-Opencode-Session` request header as an explicit local-affinity identity.

I checked the latest release **v7.2.157**, and `X-Opencode-Session` is still not handled anywhere:

- `sdk/cliproxy/session/info.go` (~L452): only `X-Session-Affinity` is read for the `affinity:` namespace
- `sdk/cliproxy/session/identity.go` (~L356): the explicit-session header list still contains `X-Session-Affinity` but not `X-Opencode-Session`

So clients that send only the native `X-Opencode-Session` header (current Pi / OpenCode Go) still fall back to prompt-derived identity, with the collision and compaction issues described in #5690.

This PR is therefore not superseded by `3bf787fc` — it is a small, self-contained change (2 logic hunks + regression tests), targets `dev`, and all three checks (`build`, `close-when-agents-md-changed`, `ensure-no-translator-changes`) are green. Could one of you take a look when you have a moment? I am happy to narrow the scope (e.g. drop the docs example and keep only the header recognition + tests) if that makes it easier to land and ship in an upcoming release.

PR #5701, opened 2026-09-09
feat(precompact): server-side pre-compaction of oversized requests

auroracapital:
## Summary

Adds server-side pre-compaction so a request routed to a context-capped provider (droid, cursor, via the `openai-compatibility` shim) that exceeds that provider's window gets its middle turns summarized by an aux model before being forwarded, instead of failing opaquely or timing out.

## Design

- Hooks into `RequestAfterAuthInterceptor`, which already fires after credential/executor/model selection and on the CLI-sub fallback path, so droid/cursor fallback attempts get the same pass automatically.
- New `internal/precompact` package: model-window lookup, token counting (reusing `helps.CountClaudeInputTokens` / `helps.CountOpenAIChatTokens`), turn-preserving split (system + tools + last N turns stay byte-identical for prompt-cache), aux-model summarization via the handler's own `ExecuteModel` (self-skipping via `InternalSource`), a per-session TTL cache so repeat turns don't re-summarize, and the interceptor that ties it together.
- New `pre-compact` config block (`enabled`, `aux-model`, `threshold`, `keep-recent-turns`, `cache-ttl`, `cache-max-sessions`), wired into hot-reload diffing.
- Aux-call failure never blocks the request: original body is forwarded and a warning is logged.
- One log line per compaction: `session, model, from_tokens, to_tokens, budget, window, aux_ms, cache_hit`.

## Testing

- `go build -o test-output ./cmd/server && rm test-output` — pass
- `go test ./internal/precompact/... ./sdk/api/handlers/... ./internal/config/...` — all pass
- `gofmt -w .` — clean

## Scope

Claude and OpenAI chat formats only; Gemini format and streaming mid-response compaction are out of scope for this change. `AGENTS.md` is intentionally left untouched per this repo's own guard.

### DIFF
--- a/internal/precompact/interceptor.go
+++ b/internal/precompact/interceptor.go
@@ -1,0 +1,141 @@
+package precompact
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/registry"
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/runtime/executor/helps"
+	cliproxyexecutor "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/executor"
+	log "github.com/sirupsen/logrus"
+)
+
+// CompactedHeader is set on the downstream response when compaction happened.
+const CompactedHeader = "X-Cliproxy-Compacted"
+
+// internalSourceValue mirrors handlers.modelExecutionInternalSource; requests
+// carrying it are auxiliary calls and are never compacted.
+const internalSourceValue = "plugin_host_model_callback"
+
+// Interceptor applies pre-compaction to selected-auth requests.
+type Interceptor struct {
+	reg        *registry.ModelRegistry
+	summarizer Summarizer
+	cache      *Cache
+}
+
+// New builds an interceptor. The cache is sized from cfg once; every other
+// setting is read per call from the config passed to Apply so hot reload applies.
+func New(cfg config.PreCompactConfig, reg *registry.ModelRegistry, s Summarizer) *Interceptor {
+	c := cfg.WithDefaults()
+	return &Interceptor{
+		reg:        reg,
+		summarizer: s,
+		cache:      NewCache(c.CacheMaxSessions, c.CacheTTLDuration()),
+	}
+}
+
+// Result reports what Apply did.
+type Result struct {
+	Compacted  bool
+	FromTokens int
+	ToTokens   int
+	CacheHit   bool
+	AuxMillis  int64
+}
+
+// Apply returns a replacement body when the request exceeds the model budget.
+// Any failure returns nil (forward the original) and is logged; it never blocks.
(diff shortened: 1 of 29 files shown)


sdk/api/handlers/handlers_interceptors.go
```diff
@@ -464,9 +464,14 @@ func (h *BaseAPIHandler) requestAfterAuthInterceptor(capture *requestAfterAuthCa
 }
 
 func (h *BaseAPIHandler) applyRequestInterceptorsAfterAuth(ctx context.Context, req coreexecutor.RequestAfterAuthInterceptRequest, requestID, skipPluginID string) coreexecutor.RequestAfterAuthInterceptResponse {
+	// Pre-compaction runs first so plugins see the body that will be sent upstream.
+	compacted := h.applyPreCompact(ctx, req)
```
cookerpapa, line 550: Keep token counting free of pre-compaction generation. This shared after-auth hook is also installed by `executeCountWithAuthManager` (used by Claude `/v1/messages/count_tokens`), so enabling pre-compaction can run `auxSummarizer.Summarize` during a count-only request and replace the body that gets counted. With a registered 1024-token model and `keep-recent-turns: 1`, my handler→manager fixture takes a 3513-token request, invokes the auxiliary executor once, and returns `{"input_tokens":44}` for the summary instead. The same fixture with pre-compaction disabled makes no generation call and counts the original body. Please distinguish count-only execution here and skip pre-compaction for it, while retaining the existing interceptor behavior.

cookerpapa (commented):
I reproduced two additional pre-compaction issues on 88713adbb3d3a747e2e1155547b369f4881a494f using local executor/summarizer fixtures, without live provider calls. The token-count endpoint case fails only when pre-compaction is enabled; the context-fit cases fail for an oversized summary and an oversized retained tail, while a short-summary control passes. Details and concrete measurements are inline. Review and reproductions prepared with Codex assistance.

internal/precompact/interceptor.go
```diff
@@ -0,0 +1,141 @@
...
+		i.cache.Put(session, Entry{Covered: len(sp.Middle), PrefixHash: HashMessages(sp.Middle), Summary: summary})
+	}
+	body, errBuild := Rebuild(req.SourceFormat, req.Body, sp, summary)
+	if errBuild != nil {
+		log.WithError(errBuild).Warn("precompact: rebuild failed; forwarding original")
+		return nil, res
+	}
+	to, _ := Count(req.SourceFormat, req.Model, body)
+	res.ToTokens = to
+	res.Compacted = true
```
cookerpapa, line 132: Check the rebuilt request against the remaining context budget before publishing compaction success. `Count` is currently used only for logging: an auxiliary answer of up to the fixed 4096-token limit, or a large preserved tail/system/tools block, can still leave the replacement too large. Against one registered 4096-token model with `max_tokens:512` and `keep-recent-turns:1`, my fixtures produce 3641 input tokens with a long summary, or 4042 with a large retained tail; both set `Compacted=true`, although input + reserved output exceeds 4096. A short-summary/short-tail control produces 43 input tokens. Compute the available summary budget from the retained content, constrain/retry summarization as appropriate, and validate the final count before replacing/caching it as a successful compaction; retain an explicit fallback when no fitting body is possible.

auroracapital:
Split into two PRs for easier review: routing/fallback in #5718, pre-compaction in #5719. Closing this one.

PR #5706, opened 2026-09-09
fix(translator): relay tool_result images as a user message for OpenAI chat completions

zhidacaishu:
## Problem

When a Claude-format request is translated to OpenAI Chat Completions, images returned inside a `tool_result` never reach the model. Anything a tool hands back as an image — a screenshot, a rendered chart, an image read from disk — is lost.

The failure is silent. No error is returned; the model simply behaves as if the tool produced nothing, and then either invents an answer or calls the same tool again. The same image sent in a plain `user` message is described correctly, which isolates the loss to the tool-role message.

## Cause

`convertClaudeToolResultContent` puts image parts inside the `role: "tool"` message:

```json
{"role":"tool","tool_call_id":"...","content":[{"type":"image_url","image_url":{"url":"data:image/png;base64,..."}}]}
```

The OpenAI Chat Completions schema only accepts text for tool-role messages. Upstreams discard the image parts instead of rejecting the request, so the content disappears without a trace.

## Reproduction

A solid red 64×64 PNG returned by a tool, with the model asked for its color. Three OpenAI-compatible upstream models, same request shape, only the translator differs:

| model | before | after |
|---|---|---|
| A | `Black` | `Red` |
| B | `Unknown` | `Red` |
| C | `White` | `Red` |

The "before" answers are pure guesses — the image never arrived.

## Change

`convertClaudeToolResultContent` now splits tool_result content into the text payload that stays on the tool message, and the image parts, which are returned to the caller. The caller replays those images in a `user` message emitted directly after the tool results, where every upstream accepts them.

- When the originating Claude user turn also carries its own text, the images are merged into that message instead, so the request keeps a single user turn rather than emitting two consecutive `user` messages.
- An image-only tool_result gets a short placeholder, so the tool message is never empty.
- A short notice labels the relayed images so the model knows they came from the preceding tool call.

This is strictly more compatible than the previous behaviour: an upstream that accepted images on tool messages accepts them on user messages too, so no configuration switch is needed.

## Tests

- Updated the two tests that asserted images stay inside tool content (`ToolResultTextAndImageContent`, `ToolResultURLImageOnly`).
- Added `TestConvertClaudeRequestToOpenAI_ToolResultImageMergesIntoUserText` for the merge path.
- `go test ./internal/translator/...` passes; `gofmt` and `go vet` are clean.

Also verified end to end against live OpenAI-compatible upstreams, including regression checks that text-only tool results, string tool results, plain chat, and user-message images all still behave as before.

https://claude.ai/code/session_01WarTu9ioEg6ucGDtnFbQkm

### DIFF
--- a/internal/translator/openai/claude/openai_claude_request.go
+++ b/internal/translator/openai/claude/openai_claude_request.go
@@ -178,7 +178,8 @@ func convertClaudeRequestToOpenAI(modelName string, inputRawJSON []byte, stream 
 				contentItems := make([][]byte, 0)
 				var reasoningParts []string // Accumulate thinking text for reasoning_content
 				var toolCalls []interface{}
-				toolResults := make([][]byte, 0) // Collect tool_result messages to emit after the main message
+				toolResults := make([][]byte, 0)       // Collect tool_result messages to emit after the main message
+				relayedToolImages := make([][]byte, 0) // Images pulled out of tool_result content for user-message relay
 
 				contentResult.ForEach(func(_, part gjson.Result) bool {
 					partType := part.Get("type").String()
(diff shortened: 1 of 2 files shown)


zhidacaishu:
Closing this: I opened it before noticing that `internal/translator/**` is guarded against pull requests by `translator-path-guard`. Refiled as #5707 with the diagnosis, the captured upstream payload and a reproduction, as the guard instructs.

The branch stays available at https://github.com/zhidacaishu/CLIProxyAPI/tree/fix/tool-result-image-relay if it is useful to the maintenance team — it is rebased on `dev`, and the `build` check passed here. Sorry for the noise.

PR #5723, opened 2026-09-10
fix(translator): surface audioTranscription parts in gemini-to-openai responses

### LINKED ISSUE #5722: bug(translator): Gemini 3.5 Transcribe transcripts (audioTranscription parts) are silently dropped in OpenAI chat-completions responses
[X] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.

**Describe the bug**

`gemini-3.5-transcribe` returns its transcript in a dedicated `audioTranscription` part instead of the regular `text` part. The Gemini → OpenAI chat-completions response converters (`internal/translator/gemini/openai/chat-completions/gemini_openai_response.go`) only handle `text`, `functionCall`, and `inlineData` parts, so the transcript is silently dropped and OpenAI-compatible clients receive an empty assistant message.

**Reproduction (upstream Gemini API, `v1beta` `:generateContent` with a WAV file as inlineData):**

```json
{
  "candidates": [
    {
      "content": {
        "parts": [
          { "text": "" },
          { "audioTranscription": { "text": "Testing one two three. The weather is sunny today." } }
        ],
        "role": "model"
      },
      "finishReason": "STOP"
    }
  ],
  "usageMetadata": { "promptTokenCount": 185, "totalTokenCount": 185 },
  "modelVersion": "gemini-3.5-transcribe"
}
```

After the Gemini → OpenAI conversion, `choices[0].message.content` / `choices[0].delta.content` is `""` — completion_tokens 0, transcript gone. Streaming (`:streamGenerateContent?alt=sse`) has the same shape and the same outcome.

**CLI Type**
gemini (both gemini-api-key and the aistudio websocket channel)

**Model Name**
gemini-3.5-transcribe (also affects any future speech-to-text model using the same part shape)

**LLM Client**
Any OpenAI-compatible client (verified via OpenWebUI chat completions and raw cURL against /v1/chat/completions)

**Request Information**

```bash
curl -X POST "https://generativelanguage.googleapis.com/v1beta/models/gemini-3.5-transcribe:generateContent?key=$KEY" \
  -H "Content-Type: application/json" \
  -d '{"contents":[{"role":"user","parts":[{"inlineData":{"mimeType":"audio/wav","data":"<base64>"}}]}]}'
```

Through CLIProxyAPI (OpenAI chat-completions endpoint, model `gemini-3.5-transcribe`), the same request returns 200 with an empty `message.content` and `usage.completion_tokens: 0`.

**Note**: the `text: ""` part present alongside `audioTranscription` makes the converter emit an empty-but-present content field, so clients see an empty reply rather than an error — easy to miss.

Related but orthogonal: #5278 registers the model for the interactions channel. This issue is about the response conversion on the generateContent surface (used by gemini-api-key and the aistudio websocket channel), where the request itself succeeds but the transcript is dropped.

The intended implementation code is in the first comment (per AGENTS.md policy for translator-only changes without push access). PR #5723 was opened as a demonstration before reading AGENTS.md and is closed.

xyonium:
## Problem

Fixes #5722.

`gemini-3.5-transcribe` returns its transcript in an `audioTranscription` part instead of the regular `text` part. The Gemini → OpenAI chat-completions response converters only handled `text`/`functionCall`/`inlineData` parts, so the transcript was silently dropped: OpenAI-compatible clients received an empty assistant message (`content: ""`, `completion_tokens: 0`) with HTTP 200 — easy to miss because nothing errors.

The empty `text: ""` part that ships alongside the transcript made the converter emit the empty content field, so the drop happened without any warning.

## Solution

Both converters (streaming and non-stream) fall back to `audioTranscription.text` when a part carries no plain `text`:

```go
// Speech-to-text models (gemini-3.5-transcribe) deliver the
// transcript in an audioTranscription part instead of text.
audioTranscriptionResult := partResult.Get("audioTranscription")
if audioTranscriptionResult.Exists() && !partTextResult.Exists() {
    partTextResult = audioTranscriptionResult.Get("text")
}
```

A part that has both `text` and `audioTranscription` keeps using `text`, matching the existing precedence between other part kinds. The transcript therefore lands in the normal `choices[0].delta.content` / `choices[0].message.content` field — no new response fields, no client changes.

## Scope

- `internal/translator/gemini/openai/chat-completions/gemini_openai_response.go`: two mirrored guard blocks (streaming + non-stream part loops).
- No request-side changes: `input_audio` → `inlineData` conversion already works.

## Verification

- Live upstream API (`v1beta`, same `glEndpoint` the aistudio executor uses): bare, `maxOutputTokens`-only, and SSE-streaming `generateContent` calls all return the transcript in `audioTranscription.text` (verbatim match against the source audio).
- New unit tests cover both converter paths with the exact upstream response shape (empty text part + audioTranscription part), asserting the transcript reaches `message.content` / `delta.content`.
- `go build ./...`, `gofmt`, and the full translator test suite are clean; the only failing test in `go test ./...` (`TestPionMediaRelayBridgesAudioAndDataChannel`, codex live module) fails identically on upstream/dev without this change.

## Notes

- Orthogonal to #5278 (interactions-channel registration): this PR fixes response conversion on the `generateContent` surface used by `gemini-api-key` and the aistudio websocket channel, where the request already succeeds today but returns an empty message.
- A companion models.json PR (router-for-me/models#55) adds `gemini-3.1-flash-lite-image` to the aistudio catalog; transcribe itself will follow the catalogs once this conversion path is merged.

### DIFF
--- a/internal/translator/gemini/openai/chat-completions/gemini_openai_response.go
+++ b/internal/translator/gemini/openai/chat-completions/gemini_openai_response.go
@@ -168,6 +168,12 @@ func ConvertGeminiResponseToOpenAI(_ context.Context, _ string, originalRequestR
 					if !inlineDataResult.Exists() {
 						inlineDataResult = partResult.Get("inline_data")
 					}
+					// Speech-to-text models (gemini-3.5-transcribe) deliver the
+					// transcript in an audioTranscription part instead of text.
+					audioTranscriptionResult := partResult.Get("audioTranscription")
+					if audioTranscriptionResult.Exists() && !partTextResult.Exists() {
+						partTextResult = audioTranscriptionResult.Get("text")
+					}
 					thoughtSignatureResult := partResult.Get("thoughtSignature")
 					if !thoughtSignatureResult.Exists() {
 						thoughtSignatureResult = partResult.Get("thought_signature")
@@ -367,6 +373,12 @@ func ConvertGeminiResponseToOpenAINonStream(_ context.Context, _ string, origina
 					inlineDataResult := partResult.Get("inlineData")
 					if !inlineDataResult.Exists() {
 						inlineDataResult = partResult.Get("inline_data")
+					}
+					// Speech-to-text models (gemini-3.5-transcribe) deliver the
+					// transcript in an audioTranscription part instead of text.
+					audioTranscriptionResult := partResult.Get("audioTranscription")
+					if audioTranscriptionResult.Exists() && !partTextResult.Exists() {
+						partTextResult = audioTranscriptionResult.Get("text")
 					}
 
 					if partTextResult.Exists() {
--- a/internal/translator/gemini/openai/chat-completions/gemini_openai_response_test.go
+++ b/internal/translator/gemini/openai/chat-completions/gemini_openai_response_test.go
@@ -77,3 +77,32 @@ func TestConvertGeminiResponseToOpenAINonStream_EmptyTextProducesEmptyString(t *
 		t.Fatalf("expected reasoning_content to be empty string \"\", got %v (type %v)", reasoning.Value(), reasoning.Type)
 	}
 }
+
+// Gemini 3.5 Transcribe returns its transcript in an audioTranscription part instead
+// of the regular text part. Without handling it, the transcript is silently dropped
+// and the OpenAI client sees an empty assistant message.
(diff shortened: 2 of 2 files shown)


xyonium:
Closing per AGENTS.md: translator-only changes from contributors without push access should be filed as an issue with the intended implementation code, which I've added to #5722. The branch stays available at xyonium:feat/gemini-transcript-part-openai if maintainers want to pull it directly.

PR #5731, opened 2026-09-11
fix(signature): preserve Gemini 3 server-side tool thought signatures (#5652)

### LINKED ISSUE #5652: fix(signature): server-side tool thoughtSignature dropped from Gemini requests (HTTP 400)
[  ] No, it's another issue.

Details, byte-level measurements and the failing/passing bodies are described below.

**Describe the bug**

When a Gemini 3 request carries both a server-side tool (e.g. `google_search`) and a custom tool (`functionCall`), replaying that model turn fails upstream:

```text
400 contents[1].parts[0]: Tool call part is missing thought_signature
```

CLIProxyAPI **deletes the `thoughtSignature` of `toolCall` / `toolResponse` parts** from the outgoing request. The same body sent straight to the upstream endpoint returns 200, and deleting that one signature by hand reproduces the identical 400.

The API contract asks for the echo — `Part.toolCall` in the v1beta discovery document reads *"The client is expected to echo this message back to the API."* So the signature must survive the proxy.

**Not a duplicate of #5602**

Both issues happen to cite `internal/signature/gemini_validation.go`, but they are different functions on different paths: #5602 is the Responses translator losing `functionResponse.id`, with CLIProxyAPI itself rejecting the request before forwarding (`ValidateGeminiFunctionCallPairing`, ~:393), while this one is the Gemini-native path where CLIProxyAPI strips a `thoughtSignature` and the *upstream* returns the 400 (`isLikelyGeminiOpaquePayload`, :515). `d0bb908`, which closed #5602, touches no file under `internal/signature/`, and this is still reproducible on current `main`.

**CLI Type**
gemini (`gemini-api-key` entry → Gemini-compatible upstream)

**Model Name**
`gemini-3.8-flash`

**LLM Client**
Kelivo (Flutter chat client), Gemini-native `streamGenerateContent`; CLIProxyAPI v7.2.154

**Request Information**

Reproduction: one turn with `google_search` enabled that emits both a server-side `toolCall` part and a custom `functionCall`, then replay that turn back to the proxy.

- body → upstream directly: **200**
- body → CLIProxyAPI → upstream: **400 `Tool call part is missing thought_signature`**
- body with that one `thoughtSignature` hand-deleted → upstream directly: **400, identical message**

**Expected behavior**

`thoughtSignature` on `toolCall` / `toolResponse` parts is forwarded unchanged, as the discovery document requires.

**Root Cause & Code Location**

`SanitizeGeminiRequestThoughtSignatures` (`internal/signature/gemini_sanitize.go`): a `toolCall` part is not a `functionCall`, so it reaches the third `switch` branch, which sets `SignatureActionDropSignature` and deletes the field.

The reason it gets there is the envelope whitelist. `isLikelyGeminiOpaquePayload` (`internal/signature/gemini_validation.go:515`) requires the envelope body to start with `0x01` (Tink prefix); server-side tool signatures carry a protobuf body instead. Measured with this repo's own helpers on real captures:

| signature | decoded | body first byte | `KnownEnvelope` | detected |
|---|---|---|---|---|
| model part | 105 B | `0x01` | true | gemini |
| `toolCall` | 650 B | `0x08` | false | unknown |
| `toolResponse` | 10077 B | `0x08` | false | unknown |

All three share the same outer envelope (`field 2 { field 1 }`); only the body differs.

**Suggested Fix**

Leave these parts untouched in `SanitizeGeminiRequestThoughtSignatures`:

```go
// Server-side tool blocks carry their own signatures; dropping one makes Gemini reject the replay.
if part.Get("toolCall").Exists() || part.Get("toolResponse").Exists() {
    partItems = append(partItems, partJSON)
    return true
}
```

Verified locally against v7.2.154: the previously failing body then returns 200, 3/3.

**OS Type**
 - OS: Linux (aarch64), Docker
 - Version: CLIProxyAPI v7.2.154

sususu98:
## Problem
Fixes #5652.

When a Gemini 3 request carries both a server-side tool (e.g. `googleSearch` / `GOOGLE_SEARCH_WEB`) and a custom client tool (`functionCall`), replaying that model turn fails upstream with:
```text
400 contents[1].parts[0]: Tool call part is missing thought_signature
```

### Root Cause
1. **Validation**: Server-side tool parts (`toolCall` / `toolResponse`) in Gemini 3 wrap the Tink AEAD ciphertext inside an inner Protobuf envelope (Field 1 varint `0x08` + Field 2 bytes `0x12` containing `0x01...`). `isLikelyGeminiOpaquePayload` only checked `payload[0] == 0x01`, which failed on the outer tag `0x08`, classifying the legitimate signature as an unknown envelope.
2. **Sanitization**: In `SanitizeGeminiRequestThoughtSignatures`, unrecognized signatures on non-functionCall parts fell through to `SignatureActionDropSignature`, dropping the `thoughtSignature` on `toolCall` and `toolResponse`.
3. **API Contract**: Unlike client-side `functionCall` where Google whitelists `"skip_thought_signature_validator"` for synthetic/migrated history, Google strictly validates server-side tools against genuine Tink ciphertexts and rejects `"skip_thought_signature_validator"` on `toolCall` with `400 Request contains an invalid argument`. Per Google's v1beta Discovery document contract (`Part.toolCall`: *"The client is expected to echo this message back to the API"*), server-side tool blocks must be echoed back verbatim.

## Solution
1. **Recognition (`internal/signature/gemini_validation.go`)**:
   - Implemented `isLikelyGeminiToolInvocationPayload` to parse inner Protobuf tags, verify wire format validity, and detect nested Tink ciphertexts starting with `0x01`.
   - Wired into `inspectGeminiField2Envelope` so server-side tool signatures are correctly classified as `SignatureProviderGemini`.
2. **Sanitization (`internal/signature/gemini_sanitize.go`)**:
   - Added early short-circuit skip in `SanitizeGeminiRequestThoughtSignatures` and `geminiContentsThoughtSignaturesNeedSanitize` for `toolCall`, `tool_call`, `toolResponse`, and `tool_response` parts.
   - Preserved existing sanitization rules: first `functionCall` gets the bypass sentinel when missing/incompatible, sibling calls remain unsigned, and unrecognized foreign signatures on normal `text` parts are dropped to prevent upstream `400 Corrupted thought signature` errors.
3. **Tests**:
   - `gemini_validation_test.go`: Added test for live-captured Gemini 3.8 Flash Google Search `toolCall` envelope; added negative tests rejecting malformed/truncated payloads; added `ValidateGeminiFunctionCallPairing` test accepting server-side tool blocks alongside function calls.
   - `gemini_sanitize_test.go`: Added tests verifying preservation of `toolCall` and `toolResponse` signatures (both camelCase and snake_case), mixed turn behavior (`toolCall` preserved + unsigned `functionCall` gets bypass sentinel), and foreign signature stripping on `text` parts.

## Verification
- Unit tests: `go test -race ./internal/signature/... ./internal/translator/gemini/...` (PASS)
- Build verification: `go build -ldflags="-s -w" -o /dev/null ./cmd/server` (clean)
- End-to-end: Replayed real Turn 2 request with `toolCall` (Google Search) + `toolResponse` + `functionCall` through CLIProxyAPI to Google AI Studio, successfully returning HTTP 200 OK.

### DIFF
--- a/internal/signature/gemini_sanitize.go
+++ b/internal/signature/gemini_sanitize.go
@@ -25,9 +25,11 @@ func GeminiReplaySignatureOrBypass(rawSignature string, blockKind SignatureBlock
 
 // SanitizeGeminiRequestThoughtSignatures applies Gemini replay policy to a
 // Gemini-shaped request. Existing provider signatures stay on their original
-// model parts. Only a missing or incompatible first functionCall gets the bypass
-// sentinel; unsigned sibling calls remain unsigned, matching native Gemini
-// parallel-call history. functionResponse parts never carry signatures.
+// model parts. Server-side tool blocks (toolCall and toolResponse) are echoed back
+// untouched per the Gemini API contract. Only a missing or incompatible first
+// functionCall gets the bypass sentinel; unsigned sibling calls remain unsigned,
+// matching native Gemini parallel-call history. functionResponse parts never carry
+// signatures.
 func SanitizeGeminiRequestThoughtSignatures(payload []byte, contentsPath string) []byte {
 	contentsPath = strings.TrimSpace(contentsPath)
 	if contentsPath == "" {
@@ -70,6 +72,10 @@ func SanitizeGeminiRequestThoughtSignatures(payload []byte, contentsPath string)
 				return true
 			}
 			if !isModelTurn {
+				partItems = append(partItems, partJSON)
+				return true
+			}
+			if part.Get("toolCall").Exists() || part.Get("tool_call").Exists() || part.Get("toolResponse").Exists() || part.Get("tool_response").Exists() {
 				partItems = append(partItems, partJSON)
 				return true
 			}
@@ -161,6 +167,9 @@ func geminiContentsThoughtSignaturesNeedSanitize(contents gjson.Result) bool {
 				return !needsSanitize
 			}
 			if !isModelTurn {
+				return true
+			}
+			if part.Get("toolCall").Exists() || part.Get("tool_call").Exists() || part.Get("toolResponse").Exists() || part.Get("tool_response").Exists() {
 				return true
 			}
 			hasFunctionCall := part.Get("functionCall").Exists()
--- a/internal/signature/gemini_sanitize_test.go
+++ b/internal/signature/gemini_sanitize_test.go
@@ -261,3 +261,91 @@ func TestSanitizeGeminiRequestThoughtSignaturesRemovesFunctionResponseSignature(
 		t.Fatalf("functionResponse nested thoughtSignature should be removed. Output: %s", string(out))
 	}
 }
+
+func TestSanitizeGeminiRequestThoughtSignatures_PreservesToolCallAndResponseSignatures(t *testing.T) {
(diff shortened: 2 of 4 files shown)


Eole7:
Is this fix only for AI Studio? I'm using Gemini models through Antigravity OAuth and I'm having this issue.

Issue #5738, opened 2026-09-11
Claude beta header is rewritten, dropping per-turn-control-2026-07-01 — 400s every per-message effort directive


maslyankov:
### Summary

CLIProxyAPI replaces the client's `Anthropic-Beta` header with its own assembled list and drops `per-turn-control-2026-07-01`. Recent Claude Code builds rely on that beta to send per-message effort directives, so every turn carrying one fails upstream with a hard 400:

```
messages.N.output_config: Extra inputs are not permitted
```

The request itself is valid — only the authorizing beta is missing by the time it reaches `api.anthropic.com`.

Affected: **7.2.147 and 7.2.157** (latest at time of writing). Verified on `darwin/aarch64`, Claude protocol, OAuth credentials.

### Reproduction

```bash
curl -s -w '\nHTTP %{http_code}\n' "http://127.0.0.1:8319/v1/messages?beta=true" \
  -H "x-api-key: $KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -H "anthropic-beta: per-turn-control-2026-07-01,mid-conversation-system-2026-04-07" \
  -d '{
        "model": "claude-fable-5-1",
        "max_tokens": 16,
        "messages": [
          {"role": "user",   "content": "hi"},
          {"role": "system", "content": [], "output_config": {"effort": "low"}},
          {"role": "user",   "content": "say ok"}
        ]
      }'
```

Result on both 7.2.147 and 7.2.157:

```json
{"type":"error","error":{"type":"invalid_request_error",
 "message":"messages.1.output_config: Extra inputs are not permitted"}}
```

The `{"role":"system","content":[],"output_config":{...}}` entry is the documented directive-only (per-message effort) form. Upstream confirms it is legal — this is the error text it returns for a *different* violation on the same endpoint:

```
messages.1: role 'system' must precede an 'assistant' message or end the array;
the directive-only form (content: [] with output_config) is accepted at any position
```

So the shape is accepted at any position; what is missing is the beta that enables it.

### Cause: the beta header is rewritten, not forwarded

Captured with `debug: true`, from one request's `error-v1-messages-*.log` — the two header sets in the same file:

**Downstream (what the client sent):**

```
claude-code-20250219, interleaved-thinking-2025-05-14, thinking-token-count-2026-05-13,
context-management-2025-06-27, prompt-caching-scope-2026-01-05,
[...]

PR #5745, opened 2026-09-11
fix(claude): forward caller betas the cloak does not manage

ggbdpq:
## Summary

The cloaked header path dropped every caller-supplied `Anthropic-Beta` on direct Anthropic when CLI-profile detection did not confirm the caller. A genuine Claude Code newer than the pinned 2.1.258 profile sends betas that profile predates (`per-turn-control-2026-07-01`, `mid-conversation-tool-changes-2026-07-01`); dropping them makes upstream reject per-turn effort directives with a hard 400 `messages.N.output_config: Extra inputs are not permitted` (verified upstream-side in the issue: the directive-only system turn is legal when the beta is present).

Caller betas the proxy itself assembles or gates stay governed by the measured baseline (Haiku effort gating, subagent cache-ttl omissions, etc. are unchanged) via a new `claudeManagedBetaSet`. Only unmanaged caller betas are forwarded, appended after the baseline — the shape a real newer client actually sends. Custom gateways keep all caller extensions as before.

## Verification

| Claim | Command | Result |
| --- | --- | --- |
| Unmanaged caller betas forwarded on direct Anthropic for an unconfirmed caller (fails before: absent from the assembled header, matching the issue's wire dump) | `go test -p 1 ./internal/runtime/executor/ -run TestApplyClaudeHeaders_ForwardsUnmanagedCallerBetas -v -count=1` | red before fix, green after |
| Managed betas keep their gating (effort stays off Haiku even when requested) | `go test -p 1 ./internal/runtime/executor/ -run TestApplyClaudeHeaders_StillGatesManagedCallerBetas -v -count=1` | pass |
| Existing beta/fingerprint contracts | `go test -p 1 ./internal/runtime/executor/ -count=1` | only the pre-existing xai Images/Videos combined-run flaky fails on this machine (single-run green, documented in prior PRs) |

## AI use

## Checklist

- [ ] Behavior change: unconfirmed callers' unmanaged betas are now forwarded on direct Anthropic (see Summary); complements #5444's version-floor approach and stays valid whether or not that lands.

### DIFF
--- a/internal/runtime/executor/claude_executor_beta_passthrough_test.go
+++ b/internal/runtime/executor/claude_executor_beta_passthrough_test.go
@@ -1,0 +1,65 @@
+package executor
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/router-for-me/CLIProxyAPI/v7/internal/config"
+	cliproxyauth "github.com/router-for-me/CLIProxyAPI/v7/sdk/cliproxy/auth"
+)
+
+// fixtureAuth builds a cloaked direct-Anthropic API-key auth for header tests.
+// The key value is a placeholder that never reaches a real upstream.
+func fixtureAuth() *cliproxyauth.Auth {
+	attrs := map[string]string{}
+	attrs[cliproxyauth.AttributeAPIKey] = strings.Join([]string{"test", "fixture", "key"}, "-")
+	attrs["fingerprint_profile"] = "claude-code-cli"
+	return &cliproxyauth.Auth{Attributes: attrs}
+}
+
+// A caller the cloak does not recognize (a Claude Code newer than the pinned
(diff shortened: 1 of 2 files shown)


ggbdpq:
Closing — superseded by `d9b8fdb7` on `dev` ("fix(claude): forward unmanaged caller betas on direct anthropic requests", Closes #5738), which landed the same design: a `claudeManagedBetaSet` distinguishing proxy-governed betas from caller extensions, with unmanaged caller betas forwarded verbatim on direct Anthropic endpoints and dropped only for unconfirmed CLI-profile shapes. The dev version additionally carries `claudeAFKModeBeta` and a larger passthrough test (99 lines vs 65 here); this branch has no unique coverage beyond it, so rather than resolving the now add/add conflicts I'm retiring the branch. Thanks to @idotorg (Luis Pater) for landing it.

Issue #5707, opened 2026-09-09
openai-compatibility: images inside tool_result are silently dropped (Claude -> OpenAI chat completions)


zhidacaishu:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.

(Strictly: there is no error. The request succeeds and the image is silently dropped, which is what makes this hard to notice.)

**Describe the bug**

When a Claude-format request is translated to OpenAI Chat Completions, images returned inside a `tool_result` never reach the model. Anything a tool hands back as an image — a screenshot, a rendered chart, an image read from disk — is lost.

The failure is silent. No error is returned; the model behaves as if the tool produced nothing and then either invents an answer or calls the same tool again. The identical image sent in a plain `user` message is described correctly, which isolates the loss to the tool-role message.

**CLI Type**

`openai-compatibility`

**Model Name**

Reproduced on three different vision-capable models behind `openai-compatibility` providers. The behaviour is format-level, not model-specific — all three fail identically, and all three succeed once the image is moved to a user message.

**LLM Client**

Claude Code (`POST /v1/messages`)

**Request Information**

Minimal reproduction. The image is a solid red 64×64 PNG, returned by a tool; the model is asked for its color.

```bash
curl -s http://127.0.0.1:8317/v1/messages \
  -H 'Content-Type: application/json' \
  -H 'x-api-key: YOUR_KEY' \
  -H 'anthropic-version: 2023-06-01' \
  -d '{
  "model": "YOUR_OPENAI_COMPAT_MODEL",
  "max_tokens": 80,
  "tools": [{"name":"screenshot","description":"Take a screenshot","input_schema":{"type":"object","properties":{}}}],
  "messages": [
    {"role":"user","content":"Take a screenshot then tell me its color."},
    {"role":"assistant","content":[{"type":"tool_use","id":"toolu_01","name":"screenshot","input":{}}]},
    {"role":"user","content":[
      {"type":"tool_result","tool_use_id":"toolu_01","content":[
[...]

Issue #4699, opened 2026-07-31
Responses → OpenAI-compatible upstream: stringified tool image output in function_call_output still billed as plain text (follow-up to #4458)


crebot51:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. The client (Codex CLI) sends a standard Responses API payload, but the proxy mistranslates it for the upstream, causing context-window overflow errors.
[ ] No, it's another issue.

**Describe the bug**

This is a follow-up to #4458 (fixed in the Chat Completions → Codex Responses direction). The same bug still exists in the **opposite direction**: Responses API (Codex CLI client) → OpenAI-compatible upstream (`openai-compatibility` provider).

When Codex CLI's `view_image` tool returns a screenshot, the session history contains a `function_call_output` item whose `output` is a **stringified JSON array** containing an `input_image` part with a `data:image/...;base64,...` URL. When translating `/v1/responses` requests to Chat Completions for an `openai-compatibility` upstream, the proxy keeps this base64 string as **plain text** instead of unwrapping it into a structured image content part.

The upstream (Kimi for Coding, `https://api.kimi.com/coding/v1`) then counts the base64 as text tokens. A single full-resolution simulator screenshot (~2.2 MB base64, 2,247,902 chars) costs ~1.25M tokens as text (vs ~2.6K tokens as an image), so a 2-message conversation explodes to 1,516,774 tokens and the upstream rejects the request:

```
Invalid request: Your request exceeded model token limit: 1048576 (requested: 1516774)
```

**Reproduction (verified against the proxy on 127.0.0.1:8317, v7.2.110)**

Send a Responses API request with a stringified image array in `function_call_output.output` (exactly what Codex CLI produces), routed to an `openai-compatibility` upstream:

```json
{
  "model": "k3",
  "input": [
    {"type": "message", "role": "user", "content": [{"type": "input_text", "text": "take a screenshot"}]},
    {"type": "function_call", "call_id": "c1", "name": "view_image", "arguments": "{}"},
    {"type": "function_call_output", "call_id": "c1",
     "output": "[{\"type\":\"input_image\",\"image_url\":\"data:image/png;base64,iVBORw0KGgo...\"}]"},
    {"type": "message", "role": "user", "content": [{"type": "input_text", "text": "what is the main color?"}]}
  ]
}
```

Measured results with a 942x2048 PNG (base64 length 11,392 chars):

| Scenario | Expected (image) | Actual |
|---|---|---|
[...]

PR #5754, opened 2026-09-12
fix(openai-compat): relay tool result images through user messages

dillonzq:
## Summary

OpenAI-compatible upstreams may reject or silently drop image parts inside `tool` messages even when they accept vision input in `user` messages. This PR keeps tool-result text on the `tool` message and relays explicitly typed images through the following `user` message, preserving parallel tool-call adjacency.

```json
{"role":"tool","tool_call_id":"call_1","content":"Read image file [image/jpeg]"}
{"role":"user","content":[
  {"type":"text","text":"Images returned by the preceding tool call(s):"},
  {"type":"image_url","image_url":{"url":"data:image/jpeg;base64,..."}}
]}
```

The relay:

- handles explicit `image_url`, `input_image`, and Claude `image` parts;
- merges images into an immediately following user message or appends one when needed;
- waits until consecutive tool results are complete;
- preserves unsupported image shapes and ordinary JSON objects containing image-related field names;
- returns the original payload without rebuilding text-only histories;
- replaces translated relay images with omission text for models configured with `input-modalities: [text]`.

The implementation stays in the OpenAI-compatible executor layer because pull requests may not modify `internal/translator/**`.

## Tests

Coverage includes mixed and image-only tool outputs, parallel tool results, Claude base64 images, detail normalization, unsupported shapes, non-image metadata fields, no-op payload identity, and streaming/non-streaming text-only behavior.

Validated with `go test ./...` and `go build -o test-output ./cmd/server`.

Related: #4699, #5707.

### DIFF
--- a/internal/runtime/executor/helps/openai_compat_tool_result_images.go
+++ b/internal/runtime/executor/helps/openai_compat_tool_result_images.go
@@ -1,0 +1,195 @@
+package helps
+
+import (
+	"strings"
+
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
+)
+
+const (
+	openAIToolResultImageRelayText   = "[Tool returned image content; the images follow in the next user message.]"
+	openAIToolResultImageRelayNotice = "Images returned by the preceding tool call(s):"
+)
+
+// RelayOpenAIToolResultImages keeps images out of Chat Completions tool
+// messages. Some OpenAI-compatible upstreams reject or drop multimodal tool
+// content even when they support image input in user messages.
+func RelayOpenAIToolResultImages(payload []byte) []byte {
+	messages := gjson.GetBytes(payload, "messages")
+	if !messages.IsArray() {
+		return payload
+	}
+
+	out := make([][]byte, 0, len(messages.Array())+1)
+	pendingImages := make([][]byte, 0, 2)
+	changed := false
+
+	flushImages := func() {
+		if len(pendingImages) == 0 {
+			return
+		}
+		out = append(out, prependOpenAIToolResultImages([]byte(`{"role":"user","content":[]}`), pendingImages))
+		pendingImages = pendingImages[:0]
+	}
+
+	for _, message := range messages.Array() {
+		messageJSON := []byte(message.Raw)
+		role := strings.ToLower(strings.TrimSpace(message.Get("role").String()))
+		if role == "tool" {
+			if text, images, ok := splitOpenAIToolResultImages(message.Get("content")); ok {
+				messageJSON, _ = sjson.SetBytes(messageJSON, "content", text)
+				pendingImages = append(pendingImages, images...)
+				changed = true
+			}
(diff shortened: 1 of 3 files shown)


dillonzq:
Addressed the two review notes in follow-up commits:

- added a lightweight pre-scan so requests with no relayable tool-result images return the original payload without rebuilding/copying the message history;
- expanded image conversion to cover Claude-style `type: "image"` base64/URL sources, while leaving unsupported image shapes unchanged instead of silently dropping them;
- added regression tests for both cases, including verifying the text-only no-op path returns the original byte slice.

CI has been retriggered on the updated head.

leaderlemon:
Two related observations on the text-only interaction.

**1. `RelayOpenAIToolResultImages` may not fire on the Claude path.**

The notice constant on line 13, `openAIToolResultImageRelayNotice = "Images returned by the preceding tool call(s):"`, is byte-identical to `toolResultImageRelayNotice` already emitted by `convertClaudeToolResultContent` in `internal/translator/openai/claude/openai_claude_request.go` (`4cd17293`).

On Claude -> OpenAI chat completions the images are lifted out of the `tool` message into a following `user` message *before* the executor runs. A relay that only extracts images from `role == "tool"` messages (your lines 43-44) therefore finds nothing to move on that path. It does fire for the Responses -> Chat Completions shape in your description — I'm only noting the Claude path looks already covered upstream, so this PR may not need to handle it.

**2. That is also why the text-only branch still leaks.**

`ShouldNormalizeOpenAIToolResultsForModel` -> `NormalizeOpenAIToolResultsTextOnly` only rewrites `role == "tool"` messages (the `message.Get("role").String() == "tool"` guard in `helps/openai_compat_tool_results.go`). By the time it runs on the Claude path, the image is already in the translator-created `user` message, out of its reach. For a model configured `input-modalities: [text]` the outgoing request still carries `image_url`.

Verified on v7.2.159 by asserting on the outgoing body at an httptest upstream with `input-modalities: ["text"]`: the relayed `image_url` is present. Same shape as the `non-stream_text-only` failure currently sitting in `openai_compat_executor_tool_results_test.go`.

So "Models explicitly configured as text-only keep the existing `NormalizeOpenAIToolResultsTextOnly` behavior" does hold — but that behavior no longer strips the image once the translator has relayed it. A strip-in-place pass over relayed `user` images on the text-only path would close it; adding a drop-variant to the relay here would not, since on the Claude path the relay does not run for that shape. `TestRelayOpenAIToolResultImagesLeavesTextOnlyPayloadUnchanged` covers a payload with no images, so it would not catch this.

Not blocking the vision path — the relay itself looks right for the Responses case.

dillonzq:
@leaderlemon 
Thanks for catching this. Fixed in `aa0078c0`: text-only normalization now also handles the translator-relayed `user` image block, replacing the image with the existing omission marker before the request reaches the upstream. I added streaming and non-streaming executor coverage for the Claude path; the full test suite, build, and PR CI pass.

PR #5793, opened 2026-09-13
store: keep PostgreSQL auth saves consistent

aikins01:
A PostgreSQL auth save can currently replace the local credential file before the database write succeeds. A database failure then leaves the local file ahead of durable state, and an unchanged-local retry can skip persistence.

This stages credential output and commits it to PostgreSQL before publishing the local file. If publication fails, it compensates the database write while retaining a per-credential session advisory lock. Save, deletion, and watcher persistence use the same lock, keyed by the resolved table identity, so another participating writer cannot complete a save during compensation. The existing database schema is unchanged.

This requires direct PostgreSQL connections or session pooling; transaction pooling is unsupported. Older instances must be stopped before upgrading writers that share an auth table. The connection and upgrade requirements are documented in `.env.example`.

The PostgreSQL integration tests exercise concurrent same-content writes, JSONB-equivalent content, deletion, watcher persistence, qualified table names, and cancellation during publication. Those nine cases passed ten consecutive runs with the race detector against PostgreSQL 16.


LTbinglingfeng:
we didn't want to let CPA use pgsql.

aikins01:
Ah, do you mean you’re planning to remove PostgreSQL support?

Issue #3358, opened 2026-05-12
DeepSeek prompt cache usage fields are not parsed


Shinokawa:
**Is it a request payload issue?**
[  ] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[x] No, it's another issue.

**If it's a request payload issue, you MUST know**
N/A. This is about usage parsing for an OpenAI-compatible provider response.

**Describe the bug**
DeepSeek OpenAI-compatible usage responses include prompt cache accounting fields named `prompt_cache_hit_tokens` and `prompt_cache_miss_tokens`.

CLIProxyAPI v7.0.3 currently parses the standard OpenAI fields (`prompt_tokens_details.cached_tokens` and `input_tokens_details.cached_tokens`) but does not map DeepSeek's cache hit/miss fields into `usage.Detail.CachedTokens`. As a result, downstream usage consumers can receive correct total/input token counts but lose the cached-token breakdown, which makes cache hit reporting and cost estimation inaccurate.

Example usage payload from a DeepSeek response:

```json
{
  "usage": {
    "prompt_tokens": 16075935,
    "completion_tokens": 44056,
    "total_tokens": 16119991,
    "prompt_cache_hit_tokens": 14771712,
    "prompt_cache_miss_tokens": 1260167
  }
}
```

Expected parsing:

- `InputTokens`: `16075935`
- `OutputTokens`: `44056`
- `TotalTokens`: `16119991`
- `CachedTokens`: `14771712`

I searched existing issues/PRs for `prompt_cache_hit_tokens` and `prompt_cache_miss_tokens` and did not find a duplicate. Existing `usage:null` PRs/issues appear to cover a separate stream-null-usage bug that is already addressed in v7.0.3.

**CLI Type**
openai-compatibility

**Model Name**
deepseek-v4-pro / deepseek-v4-flash

**LLM Client**
OpenAI-compatible clients using streaming and non-streaming chat completions.

**Request Information**
No secret-bearing request payload is needed to reproduce the parser issue. The minimal response usage payload above is sufficient for a unit test against `ParseOpenAIUsage` and `ParseOpenAIStreamUsage`.

**Expected behavior**
[...]

PR #5817, opened 2026-09-14
fix(usage): preserve DeepSeek prompt cache hits

ramapitecusment:
DeepSeek reports cached prompt tokens in `usage.prompt_cache_hit_tokens`. The OpenAI-compatible usage parser currently ignores this field, so a response with 100 input tokens, 80 cache hits and 10 output tokens is recorded as 100 uncached input tokens and zero cache reads. This loses the cache discount information used by downstream cost accounting.

Read the DeepSeek field as a fallback after the existing Chat Completions and Responses cache fields. The existing token breakdown then records 20 uncached input + 80 cached input + 10 output = 110 total. Explicit zero in a standard cache field remains authoritative.

The production change is three lines in the shared parser used by normal responses and SSE. It uses existing usage fields and requires no SDK, queue, journal or desktop changes. This covers documented responses with input/output counters; handling incomplete cache-only payloads remains outside this patch.

Validation:
- TDD: the new tests failed on unmodified `dev` for mixed/full cache in both response and streaming modes, then passed with the fallback.
- 16 regression subcases cover mixed/full/no cache, absent cache, both standard aliases, explicit-zero precedence, canonical bucket totals, and preservation through content/final-usage/`[DONE]` stream chunks.
- `go test ./internal/runtime/executor/helps ./sdk/cliproxy/usage -count=1` passed.
- `go build -o test-output ./cmd/server` passed (Go 1.26.5, macOS arm64).
- The wider executor suite encounters `TestOpenAICompatExecutorToolResultContentByInputModalities`; the same failure reproduces on clean upstream `bb20fa2d`, without this patch.
- Independent review found no actionable issues in this small diff.

References: [DeepSeek usage schema](https://api-docs.deepseek.com/api/create-chat-completion/) and [cache semantics](https://api-docs.deepseek.com/guides/kv_cache/).

Extracted from #5765 so this correction can be reviewed independently of persistent delivery and broader accounting changes. Related to the earlier #3802 proposal; this patch only maps the documented counter into the existing schema and does not add new cache-hit/cache-miss SDK fields.


sususu98:
Thanks for the PR!

Just a quick note from live verification: the official DeepSeek API (`api.deepseek.com`) currently already returns `usage.prompt_tokens_details.cached_tokens` alongside `prompt_cache_hit_tokens` (with identical values) in both streaming and non-streaming responses, matching their documented schema.

That said, adding `prompt_cache_hit_tokens` as a fallback is still very valuable for third-party gateways, self-hosted deployments (vLLM/SGLang), and custom proxies (such as reported in #3358) where the nested `prompt_tokens_details` object is omitted.

The fallback precedence and explicit-zero semantics look clean and well-tested. LGTM!

sususu98:
Closing as not planned.

After further architectural consideration, we have decided to keep the generic OpenAI-compatible usage parser strictly aligned with the canonical OpenAI specification (`prompt_tokens_details.cached_tokens`). As verified, official DeepSeek (`api.deepseek.com`) and modern open-source engines (e.g. llama.cpp, vLLM) already emit standard `prompt_tokens_details.cached_tokens` natively.

Introducing vendor-specific non-standard fields (`prompt_cache_hit_tokens`) into the shared parser creates dialect bloat and technical debt. Non-compliant intermediate proxies or gateways should align with the standard specification instead.

Thank you again for the effort and the well-crafted test suite!

PR #5834, opened 2026-09-14
fix(translator): coerce integral floats to integers for integer-typed tool-call arguments

bytemain:
## Problem

Some upstream models emit an integer-typed tool argument as a whole-number float, e.g. `"yield_time_ms": 14380.0`. Strict OpenAI-compatible clients reject the call. The Codex CLI's Rust tool router deserializes `i32` fields exactly and fails with:

```
ERROR codex_core::tools::router: error=failed to parse function arguments: invalid type: floating point `14380.0`, expected i32
```

(openai/codex#33452 — still open, not fixed as of 0.155.0-alpha.4.) The same shape hits `write_stdin.session_id`, `yield_time_ms`, `max_output_tokens`, `timeout_ms`, etc. whenever a non-OpenAI model is served through this proxy's OpenAI-compatible endpoint. Fixing it here benefits every client behind the proxy, not just Codex.

## Approach

Normalize at the single response choke point every provider already passes through — `Registry.TranslateStream` / `Registry.TranslateNonStream` in `sdk/translator/registry.go` — after the provider translator has produced the client-format output and **before** plugin `NormalizeResponseAfter` (which already wraps this exact spot). One hook therefore covers Claude/Gemini/Codex/Antigravity/Kimi/xAI/Devin **and** the OpenAI passthrough, with no changes to any individual translator under `internal/translator/` (per AGENTS.md).

The logic lives in a new package `internal/toolargs` and only runs when the client format is `openai` / `openai-response` **and** the original request declared `tools` (one `gjson` probe early-out for everything else, so the hot path is unaffected).

### Rules

- **Schema is the sole authority.** The parameters schema of the *same-named tool in the original client request* is used (both Chat Completions `tools[].function.parameters` and Responses `tools[].parameters` shapes). A missing, duplicated (ambiguous), or parameter-less tool, and any subschema that cannot be determined locally (`$ref`, `anyOf`/`oneOf`/`allOf`/`not`, or a type list that also admits `number`) → pass through untouched. Nothing is guessed.
- **Only proven targets are rewritten.** A JSON number at a position declared `integer` (optionally `["integer","null"]`), written with a fraction/exponent yet exactly integral: `14380.0` → `14380`, `1.438e4` → `14380`, `-5.0` → `-5`. `number`-typed params, strings, and non-integral floats (`14380.5`) are never changed and no value is ever invented — a genuinely fractional value for an integer param is left for the client to reject.
- **No whole-document float round-trip.** Arguments are tokenized with `json.Number`; only the target literals are spliced in place on the *original* bytes, so big integers, key order, whitespace, and unrelated fields survive byte-for-byte. Integrality is decided with `math/big.Rat` (exact).

### Coverage

| API | non-stream | stream |
|---|---|---|
| Chat Completions | `choices[].message.tool_calls[].function.arguments` | whole-value deltas only: a chunk that carries `function.name` **and** a complete JSON object (Claude/Gemini/Antigravity-style translators emit the assembled call in one delta) |
| Responses | `output[].type=="function_call"` → `arguments` | `response.function_call_arguments.done` (name resolved via the `item_id → name` announced by `response.output_item.added`, per stream) **and** `response.output_item.done` (stateless), so both agree |

Deltas are never rewritten. Custom (freeform) tool calls are skipped. Event order, `sequence_number`, item ids, `call_id`, and `finish_reason` are unchanged; SSE framing (`event:` line, `data:` prefix, both with and without an `event:` line) is reproduced verbatim.

**Known limit:** fragmented Chat Completions argument deltas (no name, partial JSON — e.g. the Codex delta path or an OpenAI-compatible passthrough that fragments) are passed through, because bytes already flushed to the client cannot be retracted. Buffering them would change wire timing and is deliberately left as a possible follow-up behind a flag.

### Observability

An in-process per `provider/model` counter (`toolargs.Stats()`; the repo has no metrics backend) plus a `Debug` log carrying only the tool name and the coerced parameter paths (e.g. `opts.retries`, `ids[2]`) — never argument values.

## Tests

- `internal/toolargs/coerce_test.go` (RED-first): the exact `14380.0` field sample; nested objects and (nested) integer arrays; a `number` param stays `14380.0`; `14380.5` passes through; big-int/key-order/whitespace byte preservation; `$ref`/`anyOf`/`oneOf`/`allOf`/root-`$ref`/`["integer","number"]`/missing-property/`null` schemas all pass through; `["integer","null"]`, exponent and negative forms; integer literal is a no-op; malformed arguments unchanged; tool lookup for both request shapes fails closed on duplicate/parameter-less/missing/empty name.
- `internal/toolargs/openai_test.go`: both APIs × both modes, the Responses event sequence (`added → delta → args.done → output_item.done → completed`) with stream memory released, `data:`-only framing, fragment passthrough, custom-tool skip, stats counting.
- `sdk/translator/registry_toolargs_test.go`: end-to-end through the registry for all four cases, ordering before `NormalizeResponseAfter`, no-op without tools and for non-OpenAI client formats. Mutation-verified: disabling the hook fails all four.

`go build ./...`, `go vet ./...` clean; `gofmt` clean on all touched files. Full `go test ./...` passes except `TestOpenAICompatExecutorToolResultContentByInputModalities` in `internal/runtime/executor`, which fails identically on clean `main` (7fa443dc) without this change (unrelated: tool-*result* image content).


bytemain:
Closing: on reflection the right layer for this is the client — Codex should tolerate integral floats for integer parameters (openai/codex#33452) — rather than the gateway rewriting model output. Leaving the branch up for reference in case a gateway-side option is ever wanted.

PR #5868, opened 2026-09-16
fix(devin): support swe-1-6-slow model variant

### LINKED ISSUE #5863: [Bug][Devin] Missing swe-1-6-slow Free-tier variant; bare swe-1-6 returns insufficient_quota
# Issue: Devin Free-tier `swe-1-6-slow` variant is not routable; bare `swe-1-6` returns `insufficient_quota`

## Environment

* CLIProxyAPI: v7.3.4 (Windows amd64)
* Devin auth: OAuth login via `-devin-login`
* Devin plan: Free

## Summary

On the same Devin Free account, the official Devin CLI exposes and successfully uses `swe-1-6-slow`.

CPA currently exposes `devin/swe-1-6`, resolves it to upstream `swe-1-6`, and Devin returns:

```text
permission_denied / insufficient_quota
```

CPA currently has no routable `swe-1-6-slow` variant, so a direct CPA A/B test using that exact upstream UID is not possible yet.

The model UID mismatch appears to be the likely cause of the Free-tier failure, and at minimum CPA currently cannot address the Free-tier SWE-1.6 variant exposed by the official Devin CLI.

### 1. Official Devin CLI Free catalog exposes `swe-1-6-slow`

On the same Free account, the official Devin CLI exposes:

```text
swe-1-6-slow
```

and successfully completes requests using that model.

### 2. CPA resolves `devin/swe-1-6` to bare `swe-1-6`

CPA logs confirm the upstream request uses:

```json
{
  "model": "swe-1-6",
  "session_id": "...",
  "cascade_id": "...",
  "max_tokens": 10,
  "prompts": ["..."]
}
```

### 3. Devin rejects the bare UID on the Free account

The upstream response is:

```json
{
  "error": {
    "message": "devin upstream error (permission_denied): an internal error occurred (trace ID: ...)",
    "type": "permission_error",
    "code": "insufficient_quota"
  }
}
```

### 4. Current CPA resolver has no `slow` variant

The current Devin model resolver explicitly maps:

```text
devin/swe-1-6
→ swe-1-6
```

and only handles `fast` as a special SWE-1.6 variant:

```go
case "swe-1-6":
    if effort == "fast" {
        return "swe-1-6-fast"
    }
    return "swe-1-6"
```

`-slow` is also absent from the recognized Devin suffix list.

The v7.3.4 binary likewise contains `swe-1-6` and `swe-1-6-fast`, but no `swe-1-6-slow`.

### 5. Direct `swe-1-6-slow` testing is currently blocked by CPA routing

Requests using:

```text
devin/swe-1-6-slow
```

are rejected by CPA before reaching Devin because that model is not registered/routable.

Therefore I cannot yet verify whether sending the exact upstream `swe-1-6-slow` UID through CPA succeeds, although the official Devin CLI does successfully use that UID on the same account.

### 6. Model aliases do not provide a workaround

I attempted both:

* `oauth-model-alias` in `config.yaml`
* `model_aliases` in the Devin auth JSON

Neither changes the final Devin upstream model UID selected by the resolver.

## Expected behavior

CPA should support the upstream:

```text
swe-1-6-slow
```

model UID so that it can be addressed directly.

A minimal non-breaking fix would be to add `swe-1-6-slow` as a supported Devin model/UID while preserving the existing behavior of `devin/swe-1-6`.

If plan-aware routing is desired, Free-tier requests to:

```text
devin/swe-1-6
```

could additionally resolve to:

```text
swe-1-6-slow
```

matching the official Devin CLI behavior, while preserving existing paid-tier behavior.

## Suggested validation

After adding `swe-1-6-slow` support:

1. Authenticate a Devin Free account.
2. Request `devin/swe-1-6-slow`.
3. Verify CPA sends upstream `chat_model_uid/model = swe-1-6-slow`.
4. Verify the request succeeds on the same account for which bare `swe-1-6` returns `insufficient_quota`.
5. Confirm existing `swe-1-6`, `swe-1-7`, and `swe-2` mappings remain unchanged.

## Additional context

Devin provider support is very recent, so this may be a Free-tier model-variant edge case.

Related: #5839 documents a separate SWE-2 compatibility difference where tool declarations are rejected while SWE-1.7 succeeds with the same payload.

avabbbb:
## Summary

Add support for Devin's `swe-1-6-slow` upstream model variant used by the current Free-tier SWE-1.6 path.

Fixes #5863.

## Problem

On the same Devin Free account:

- Official Devin CLI exposes and successfully uses `swe-1-6-slow`.
- CPA currently routes `devin/swe-1-6` to bare `swe-1-6`.
- Bare `swe-1-6` is rejected upstream with `permission_denied / insufficient_quota`.

CPA also did not recognize `-slow` as a Devin model UID suffix, so the Free-tier upstream variant could not be addressed directly.

## Fix

- Recognize the `-slow` Devin UID suffix.
- Add `devin/swe-1-6-slow` to the embedded Devin catalog.
- Add resolver/regression coverage for the slow variant.
- Preserve all existing mappings for `swe-1-6`, `swe-1-7`, and `swe-2`.

The remote catalog counterpart is added in: https://github.com/router-for-me/models/pull/58

## Validation

Same Devin Free account, same CPA build:

| Request              | Upstream UID   | Result                                      |
| -------------------- | -------------- | ------------------------------------------- |
| `devin/swe-1-6`      | `swe-1-6`      | `403 permission_denied: insufficient_quota` |
| `devin/swe-1-6-slow` | `swe-1-6-slow` | `200 OK`                                    |

Tests:

- `internal/runtime/executor/helps` — PASS
- `internal/registry` — PASS
- existing SWE-1.6 mapping regression — PASS
- SWE-1.7 regression — PASS
- SWE-2 regression — PASS

No existing model mapping is changed.


avabbbb:
Duplicate of #5867 (same commit dea4ce8, base=dev per repo retarget convention). Keeping #5867 as the canonical PR.

PR #5881, opened 2026-09-16
Fix proxied xAI (Grok) agent loop and tool calls

Arcadi4:


Arcadi4:
Apologies, I meant to open this in the fork.

PR #5901, opened 2026-09-17
feat(usage): observe and enqueue upstream_response_model

### LINKED ISSUE #5900: [Feature] Observe the upstream-declared model before force-mapping
## Problem

Usage records only store the model sent upstream (`model`) and the client-facing alias (`alias`). After a successful response, `force-mapping` rewrites the downstream model name back to the alias. The name the upstream actually declared in the raw HTTP/WebSocket body is discarded, so operators and [cpa-usage-keeper](https://github.com/Willxup/cpa-usage-keeper) cannot tell when the provider served a different model than the one CPA sent.

This matches the existing `model` / `alias` / response-tier split, but there is no field today for the pre-rewrite upstream declaration.

## Proposal

Observe the model name from the raw upstream payload **before** request-log gating and **before** force-mapping rewrite, then publish it on `usage.Record` and the Redis usage queue as `upstream_response_model`.

Intended behavior:

- Cover all protocols that already flow through `AppendAPIResponseChunk` / `AppendAPIWebsocketResponse` / `EmitWebSocketResponseEvent` (non-stream, SSE, WebSocket).
- Extraction families: Claude first-declaration, Gemini `modelVersion`, everything else OpenAI-shaped (`response.model` then `model`).
- Empty observations are omitted from the queue JSON (`omitempty`).
- Downstream client responses stay unchanged. Plugin API, TUI, Home, billing, and request-log structured fields are out of scope.

Companion change for persistence and UI: a matching issue on cpa-usage-keeper (linked from the follow-up PR). CPA should land first so the queue key exists before Keeper consumes it.

## Acceptance

- [ ] A force-mapping request that sends `sent-upstream-model` and receives a different raw `model` produces a usage-queue payload with `upstream_response_model` set to the raw value.
- [ ] A matching / empty observation does not write a misleading value (empty is omitted).
- [ ] Observation still runs when request-log is off / commercial mode is on.
- [ ] Downstream response `model` rewrite behavior is unchanged.

## Notes

Implemented locally against current `main` and verified with unit tests plus an isolated CPA + mock OpenAI + Keeper path. I will open a PR from a fork.

FlameMida:
Fixes #5900

Companion: https://github.com/Willxup/cpa-usage-keeper/issues/548 (Keeper PR follows; please land this CPA change first)

## Summary
- Observe the model name declared in the raw upstream HTTP/WebSocket payload before request-log gating and before force-mapping rewrite.
- Publish it on `usage.Record` and the Redis usage queue as `upstream_response_model` (omitted when empty).
- Cover Claude-first, Gemini `modelVersion`, and OpenAI-shaped extraction without changing downstream responses, plugin API, TUI, or billing.

## Test plan
- [x] `gofmt` on the touched packages
- [x] `go test ./sdk/cliproxy/usage/ ./internal/runtime/executor/helps/ ./internal/redisqueue/ -count=1`
- [x] Isolated force-mapping path: client alias `client-alias` sends `sent-upstream-model`, mock replies `upstream-reported-model`, usage-queue JSON contains `"upstream_response_model":"upstream-reported-model"`
- [ ] Reviewer: confirm plugin executors that skip `Append*` remain empty (accepted this round)


FlameMida:
Companion Keeper PR: https://github.com/Willxup/cpa-usage-keeper/pull/549

Issue #5838, opened 2026-09-15
xAI OAuth: POST /v0/management/api-call forwards literal $TOKEN$ (upstream 401)


anlostsheep:
**Is it a request payload issue?**
[x] Yes, this is a request payload issue. I am using a client/cURL to send a request payload, but I received an unexpected error.
[ ] No, it's another issue.

**Describe the bug**
On CLIProxyAPI **v7.3.3**, `POST /v0/management/api-call` does **not** substitute `$TOKEN$` for **xAI OAuth** auth files (`type=xai`, `auth_kind=oauth`).

The management API still returns HTTP 200 and forwards the upstream request. The upstream then sees a literal `Bearer $TOKEN$` and returns 401.

The same xAI OAuth credential works for normal inference (`POST /v1/responses` via the xAI executor / `cli-chat-proxy`). The same `api-call` succeeds if `Authorization` is set to the real `access_token` instead of `$TOKEN$`.

This started immediately after the automatic upgrade **v7.3.1 → v7.3.3** (v7.3.2 is in that jump; v7.3.2 changelog mentions APICall `$TOKEN$` body replacement / escaping). Quota probes that succeeded on v7.3.1 now persist empty results.

**CLI Type**
xAI / Grok Build OAuth (`type=xai`, `auth_kind=oauth`, auth JSON in `auth-dir`)

**Model Name**
`grok-4.6` (inference still works; this bug is on management `api-call`, not chat)

**LLM Client**
CPA Manager Plus quota refresh (it sends `Authorization: Bearer $TOKEN$` and `authIndex` from `GET /v0/management/auth-files`). Reproduced with a direct `api-call` as well.

**Request Information**

Auth file shape (secrets omitted):

```json
{
  "type": "xai",
  "auth_kind": "oauth",
  "base_url": "https://api.x.ai/v1",
  "access_token": "<oauth access token>",
  "refresh_token": "<refresh token>",
  "id_token": "<id token>",
  "email": "<redacted>",
  "disabled": false
}
```

`GET /v0/management/auth-files` lists the file as `type=xai`, `account_type=oauth`, with a non-empty `auth_index`.

Failing call (placeholder as sent by CMP / as reproduced):

```http
POST /v0/management/api-call
Content-Type: application/json

{
  "authIndex": "<auth_index from GET /v0/management/auth-files>",
  "method": "GET",
  "url": "https://cli-chat-proxy.grok.com/v1/billing?format=credits",
  "header": {
[...]

PR #5909, opened 2026-09-17
fix(management): reject unresolved token placeholders

camy-x:
`POST /v0/management/api-call` currently forwards requests containing a literal `$TOKEN$` when the supplied `auth_index` is missing or no longer resolves. This turns a local credential lookup failure into an upstream 401 while the management endpoint still returns an outer 200.

Reject unresolved placeholders with `400 auth token not found` before opening an upstream request. Add regression coverage for header and body placeholders, plus an end-to-end xAI OAuth test that reads the index from `/auth-files` and verifies the access token reaches the upstream request.

Addresses #5838.

Validation:
- `go test ./internal/api/handlers/management -count=1`
- `go build -o <temp> ./cmd/server`
- `go test ./... -count=1` passes all packages except the unrelated Windows timeout in `internal/runtime/executor`: `TestCodexExecutor_BootstrapBuffering_ByteCapReleasesStream` (also reproduces alone with a 45s timeout)


camy-x:
Closing as superseded by 0b55053944aa1547ba762c3fc519261cccc51d66. That commit covers this PR's unresolved-placeholder rejection and adds xAI token-storage fallback, access-token selection, refresh handling, and end-to-end coverage. There is no remaining functional change in this PR that needs merging. Thanks for the broader fix.

Issue #5911, opened 2026-09-17
devin: function_result conversion fails on structured, empty, and orphaned results


luyishui:
**Is it a request payload issue?**
[x] Yes. A `function_result` whose `function_call` was trimmed from the request
    history fails the whole upstream call.

**Describe the bug**

Three failure modes in `parseInteractionsPayload` when projecting tool results into
Devin prompts (source=4).

1. Structured results go up as raw JSON.

   When `result`/`output`/`content` is an array of content blocks or an object (the
   usual Claude `tool_result` shape), the raw JSON string becomes the prompt text, so
   the model reads `[{"type":"text","text":"..."}]` scaffolding instead of the tool
   output.

2. Empty results go up as an empty string.

   If `result`, `output`, and `content` are all absent or empty, the prompt text is
   `""`. The upstream rejects empty tool-result text.

3. Results with no matching call go up with a dangling `tool_call_id`.

   A `function_result` whose `call_id`/`id` has no pending emitted `function_call`
   (a compressed or truncated history) is emitted as source=4 with an empty or
   unresolvable `tool_call_id`, and the upstream rejects the whole request with
   invalid_argument. Pairing is positional upstream, so a result with nothing left to
   consume has to be sent as user text instead.

   The raw `messages` fallback has the same problem baked in: it ignores
   `assistant.tool_calls`, so every `role=tool` message on that path is an orphan by
   construction.

**Request Information**
A chat request carrying a `function_result` (or `role=tool` message) whose preceding
`function_call` was trimmed by history compaction reproduces bug 3: the upstream
answers invalid_argument.

**Expected behavior**
- structured results are projected to plain text
- empty results get a placeholder instead of `""`
- results without a pending call are sent as user text, not a dangling source=4 prompt

Issue #5893, opened 2026-09-17
[Bug] Built-in Devin provider drops images inside Claude tool_result before building the upstream request


Theigrams:
### Is it a request payload issue?

- [x] Yes — specific request content is silently lost during conversion; the request does not necessarily return an error.
- [ ] No

### Describe the bug

When using CPA's built-in Devin provider through `POST /v1/messages`, images nested inside Claude `tool_result.content` reach CPA but do not appear in the Devin request object. This affects screenshots returned by computer-use tools.

The tool-call ID survives, so CPA still constructs a tool-result prompt and continues the request. Its image payload is silently discarded, leaving the model without the screenshot needed for the next step.

### CLI Type

CPA's built-in Devin provider, using the Connect-RPC `GetChatMessage` endpoint at `server.codeium.com`.

### Model Name

Client model: `devin/swe-2`. The logged upstream selector is `swe-2-medium`; incoming `output_config.effort` is `medium`.

### LLM Client

ZCode. The request User-Agent identifies `ZCode/3.11.2`; the tool results come from its computer-use screenshot/zoom tools.

### OS Type / Environment

- Server: Linux, x86_64.
- Client: macOS.
- CLIProxyAPI: `v7.3.5`, commit `b681a1e0f7b89d26814f788b60bf84feaf72e912`.
- Endpoint: `POST /v1/messages`, streaming.
- Observation date: September 17, 2026, UTC+08:00.

### Request Information

Abbreviated request structure, with tool IDs, text and image data replaced:

```json
{
  "model": "devin/swe-2",
  "stream": true,
  "max_tokens": 128000,
  "thinking": {"type": "enabled", "budget_tokens": 8000},
  "output_config": {"effort": "medium"},
  "tool_choice": {"type": "auto"},
  "tools": [{
    "name": "screenshot",
    "description": "Capture a screenshot",
    "input_schema": {"type": "object", "properties": {}}
  }],
  "messages": [
    {"role": "user", "content": "Inspect the screenshot returned by the tool."},
    {"role": "assistant", "content": [
      {"type": "tool_use", "id": "tool_image_1", "name": "screenshot", "input": {}}
    ]},
    {"role": "user", "content": [
      {"type": "tool_result", "tool_use_id": "tool_image_1", "content": [
        {"type": "image", "source": {
          "type": "base64", "media_type": "image/png", "data": "<BASE64_PNG>"
        }}
      ]}
    ]}
  ]
}
```

[...]

PR #5913, opened 2026-09-17
fix(devin): harden function_result conversion for upstream pairing

luyishui:
## Summary

Fixes #5911.

`function_result` items in incoming interactions/messages still fail upstream in
two ways: empty results go up as an empty string, and results with no pending
call go up as a source=4 prompt with a dangling `tool_call_id`. (The structured
payload projection part of #5911 was covered by #5893's tool-result extraction;
this PR builds on top of it.)

## Problem

    function_result with empty result/output/content
        ↓
    prompt text "" -> upstream rejects empty tool-result text

    function_result whose call was trimmed by history compaction
        ↓
    source=4 prompt with empty/dangling tool_call_id
        ↓
    upstream invalid_argument -> the whole request fails

    role=tool on the raw `messages` fallback path
        ↓
    assistant.tool_calls never projected -> every result is an orphan there

## Fix

1. **Empty results**: a valid result with no text is emitted as `"[tool result]"`
   instead of `""`. Image-bearing results already carry their `[Image N]` header,
   so only fully empty payloads get the placeholder.

2. **Orphan results**: `appendToolResultPrompt` emits a result as source=4 only
   while a pending emitted call exists; pairing is positional upstream, so each
   kept result consumes one pending call. A result with nothing left to consume
   is folded into a user prompt (`"[tool result without matching call]"`) that
   still carries the text and any extracted images.

3. **`messages` fallback**: `assistant.tool_calls` are now projected onto the
   emitted prompt, so paired results stay native on that path instead of always
   folding.

4. **Image slot guard**: folded orphans share source=1 with real user prompts;
   `supplementImagesFromOriginal` skips marker-prefixed prompts so they neither
   absorb a user image slot nor shift the positional mapping for later prompts.

## What this does NOT change

- Results that do have a pending call still go up as source=4 with the same text
  and images.
- Structured/raw object payloads keep the current extraction behavior (raw
  objects are preserved by design).
- Request ordering is preserved.
- No translator code touched; no other provider affected.

## Tests

New cases in `devin_executor_test.go`:

| Case | Input | Expected |
|---|---|---|
| empty result | no result/output/content | `"[tool result]"` |
| image-only result | `[{type:image,...}]` | image header + attached image, no placeholder |
| missing id | result without call_id | folded to user prompt |
| duplicate result | two results, one call | second one folded |
| messages fallback | assistant.tool_calls + role=tool | native pairing kept |
| orphan before user image | folded orphan + later user message | user keeps its image slot |

## Before / After

| Case | Before | After |
|---|---|---|
| empty result | `""` sent upstream | `"[tool result]"` |
| orphan result | source=4 with dangling id, request rejected | user prompt, request succeeds |
| messages fallback | every tool result orphaned | paired results stay native |
| orphan ahead of user image | request rejected anyway | orphan folded, user image slot preserved |

## Verification

- `gofmt -l` clean
- `go build ./cmd/server` passed
- `go test ./internal/runtime/executor/...` passed


luyishui:
Superseded: these fixes landed upstream in b6fe4f20 (Closes #5911). Thanks! Remaining edge cases will follow in a separate PR.

PR #5914, opened 2026-09-17
fix(executor): normalize Codex multi-agent v2 agent_message items for Kimi Responses

pinguarmy:
## Problem

Codex sessions that used **multi-agent v2 subagents** inject `agent_message` input items into subsequent Responses API requests (subagent-to-leader handoff), whose content parts use `encrypted_content`. The Kimi upstream Responses endpoint rejects `agent_message` items outright:

```
HTTP 400 invalid_request_error
```

This broke **every** Codex session with subagent handoff history as soon as it was routed to a Kimi model — a 100% repro, independent of session size or content.

## Root cause

`KimiExecutor` forwards the translated request body to the upstream Kimi Responses URL without touching `agent_message` items, which Kimi does not recognize as a valid input item type.

## Fix

Add `normalizeKimiAgentMessages`, applied on both the streaming and non-streaming paths (next to the existing `normalizeKimiTools` / `normalizeKimiTemperature` hooks), which rewrites each top-level `agent_message` input item before forwarding:

- `type: "agent_message"` → `type: "message"`, `role: "user"`
- `encrypted_content` content parts → `input_text` with the ciphertext surfaced as text, so delegated task content is preserved instead of dropped (mirrors `normalizeCodexAgentMessages` in the Claude translator)

Properties:

- Only `agent_message` items are touched; neighbouring items, tool calls, and interleaved developer messages keep their exact positions (asserted in tests).
- Any `sjson` error falls back to the original body — the request is never left half-rewritten.
- Idempotent: rewritten items no longer match the rewrite predicate.

## Verification

- `go build ./...` clean.
- New unit tests: 4 subcases covering the rewrite, ciphertext surfacing, key removal, neighbour isolation, and pass-through for non-array input; plus an end-to-end non-streaming executor test.
- Existing executor and translator audit/compat suites still pass (interleaved developer-message / tool-output ordering assertions included).
- Live-verified against a production CPA instance running this patch: a request whose `input` contains an `agent_message` item with an `encrypted_content` part (the exact payload that previously returned HTTP 400) now returns HTTP 200 with a normal completion, on both the non-streaming and SSE streaming endpoints.

## Notes

- The plaintext handoff text is forwarded to the upstream model by design — the upstream cannot interpret encrypted parts; this matches the existing Claude-translator behaviour.


pinguarmy:
Note: this PR originally also carried two audit-only test files under `internal/translator/openai/openai/responses/` (sanitized session-structure and interleaved developer/tool ordering checks). The `ensure-no-translator-changes` policy blocks any `internal/translator` diff, so they were removed from this PR per that policy. They are preserved in a local audit branch; happy to hand them to the maintenance team via an issue if useful.

PR #5917, opened 2026-09-17
Fix: quota cooldown correctness for pooled credentials

auroracapital:
# Quota cooldown correctness for pooled credentials

## Summary

An exhausted credential could lose its open quota cooldown on every token
refresh, because registry reconciliation cleared the whole quota state and reset
the backoff exponent, so the credential rejoined rotation immediately and the
exponential ladder never escalated. This branch separates "reset because the
credential succeeded" from "reconcile transient error state", bounds every
cooldown that comes from an upstream reset hint, and stops the carried-over
recovery window from ratcheting forward indefinitely. It also exposes the
resulting per-credential quota state over the management API.

## What it fixes, and for whom

For anyone running more than one credential per provider behind the proxy:

- **Cooldowns survive a token refresh.** `resetModelState` still fully clears
  quota on a genuine success or an operator `ResetQuota`; reconciliation now uses
  `resetModelStateKeepingQuota`, which holds an unexpired cooldown open and
  carries the backoff exponent across an expired window.
- **An upstream reset can no longer park a credential for days.** Anthropic
  returns `anthropic-ratelimit-unified-*` headers on the fingerprinted
  `claude-cli` path that parse into resets of many hours to several days. Those
  were applied verbatim, bypassing the backoff ladder. `capQuotaCooldown` now
  bounds them at `quotaBackoffMax` at all three sites that consume a reset hint.
- **The recovery window is no longer a one-way ratchet.** The 429 branch of
  `applyAuthFailureState` carried a later `Quota.NextRecoverAt` forward without
  capping it, so a single weekly reset latched permanently and the credential
  could never shorten its own window.
- **Quota 429s without any reset signal get a sane floor.** These arrive with no
  `Retry-After`, no rate-limit headers, and the literal body `Error`, so a ladder
  starting at one second re-confirms known exhaustion repeatedly. The floor is
  provider-scoped, jittered, never shortens a longer cooldown, and is overridable
  through `SetAnthropicQuotaFloor`.
- **The state is observable.** A per-auth provider usage endpoint reports each
  credential's windows, including an estimated 5h and 7d reset derived from a new
  `Quota.FirstExceededAt`. The estimates are labelled as estimates, because the
  provider does not supply them.

## Deliberately part of this branch: replacing the existing jitter

The first commit, `fix(auth): match quota backoff to provider window and
desynchronise pool`, **replaces the current cooldown jitter rather than adding to
it.** Today `conductor_selection.go` spreads retries with `cooldownWaitJitterCap`
and a `wait/4` term applied at selection time. This branch instead jitters the
computed backoff by +/-20% at the point the cooldown is written, and raises
`quotaBackoffMax` from 30m to 5h to match the provider's rolling window.

The two approaches overlap, and a maintainer should choose one on purpose:

- Jittering at write time makes the stored `NextRetryAfter` itself
  desynchronised, so every reader of that field sees the spread, including the
  management API and any operator inspecting state.
- Jittering at selection time keeps the stored deadline exact and spreads only
  the moment of use.

This branch takes the first option because the rest of the commits reason about
the stored deadline directly. It is the base commit of the branch, and the later
commits assume the 5h ceiling, so it is not separable by simply dropping it; if
you prefer to keep the selection-time jitter, the cleanest path is to keep the
`quotaBackoffMax` change and drop the +/-20% term.

Provider-supplied `Retry-After` hints bypass jitter and are still honoured
exactly.

## Not included

- CLI-subscription fallback chains between providers. That is a deployment
  routing preference, not a correctness fix, and it does not belong in the
  cooldown path.
- Installing plugins from GitHub releases, which overlaps with the existing
  registry-driven plugin store.

## How to test

```
go build ./...
go test ./sdk/cliproxy/auth/...
go test ./internal/api/...
go test ./internal/runtime/executor/
```

Targeted regressions:

```
go test ./sdk/cliproxy/auth/ -run 'Cooldown|Quota|Backoff|RateLimit' -v
```

Notes:

- `TestOpenAICompatExecutorToolResultContentByInputModalities` and
  `TestMetaExecutor_ExecuteSuccessAndRateLimit` fail on the base tag as well.
  They are unrelated to this branch.
- `test(executor): make Claude fingerprint assertions host independent` fixes
  three assertions that called `helps.MapStainlessOS()`, which maps
  `runtime.GOOS`, while the production path deliberately writes the pinned
  `profile.OS` for unconfirmed clients. Those tests passed only on darwin. The
  expectation now derives from
  `ApplyClaudeDefaultDeviceProfileHeaders`, the same exported helper production
  uses, so they hold on any host. No production behaviour changes.
- The three `claude_ratelimit_cooldown` tests pinned the old policy of honouring
  a full multi-day reset. Their invariant, that a cooldown is never silently
  cleared by a concurrent success, an `Update`, or a restore, still holds; the
  magnitude assertions moved to the capped bound, and a regression test for a
  very long upstream reset was added.


0xble:
I tested head `75d2dcfd22482321f2435d10755700393902ae5e` against its merge base `8335eac731946bd4eff18f500653f93736df53d6` with a synthetic Codex credential-scoped 429. This helps with fresh cooldowns, but an already-recorded weekly deadline can still bypass the five-hour ceiling.

Results through `Manager.Register` → `Manager.MarkResult` → `GetByID`:

- Fresh auth, `CredentialScope: true`, HTTP 429, `RetryAfter: 168h`: base stores 168h; PR stores 5h for model quota, model retry and auth quota. This improvement passes.
- Same call with an existing 168h model/root quota state: PR still stores 168h in all three fields. This fails the proposed 5h invariant.
- A second, untouched healthy credential remains selectable on both revisions. This does **not** demonstrate cross-credential contamination.

The remaining path is in `MarkResult`: after `capQuotaCooldown`, the prior `state.Quota.NextRecoverAt` can replace the capped value. The credential root also preserves a later `auth.Quota.NextRecoverAt`, and the trailing `prevModelRetryAfter` preservation can restore a long retry deadline. In sibling propagation, `otherQuotaNext` is capped after `otherRetryAfter` was computed, so those fields can diverge too (the latter is a source-inspection finding, not a separate executed case).

Minimal reproducer, placed in package `auth` alongside existing tests (uses the existing `quotaResult` and `withQuotaCooldownEnabled` helpers):

```go
func TestReviewExistingCodexWeeklyDeadline(t *testing.T) {
    withQuotaCooldownEnabled(t)
    ctx := WithSkipPersist(context.Background())
    m := NewManager(nil, nil, nil)
    deadline := time.Now().Add(7 * 24 * time.Hour)
    a := &Auth{
        ID: "synthetic-codex", Provider: "codex", Status: StatusError,
        Unavailable: true, NextRetryAfter: deadline,
        Quota: QuotaState{Exceeded: true, Reason: "credential_quota", NextRecoverAt: deadline},
        ModelStates: map[string]*ModelState{
            "test-model": {
                Status: StatusError, Unavailable: true, NextRetryAfter: deadline,
                Quota: QuotaState{Exceeded: true, Reason: "quota", NextRecoverAt: deadline},
            },
        },
    }
    if _, err := m.Register(ctx, a); err != nil { t.Fatal(err) }
    delay := 7 * 24 * time.Hour
    r := quotaResult(a.ID, "test-model")
    r.CredentialScope, r.RetryAfter = true, &delay
    before := time.Now()
    m.MarkResult(ctx, r)
    got, _ := m.GetByID(a.ID)
    ceiling := before.Add(5*time.Hour + time.Second)
    if got.ModelStates["test-model"].NextRetryAfter.After(ceiling) || got.Quota.NextRecoverAt.After(ceiling) {
        t.Fatal("existing weekly quota deadline escaped the 5h ceiling")
    }
}
```

`go build -o /tmp/cliproxyapi-pr5917 ./cmd/server` passed, as did the existing auth and management package suites. The executor suite failed in `TestMetaExecutor_ExecuteSuccessAndRateLimit` and `TestOpenAICompatExecutorToolResultContentByInputModalities`; both failures also reproduce on the merge base.

Operationally, we observed healthy direct Codex calls while the proxy returned a long pool cooldown, followed by successful proxy requests after a restart. We did **not** preserve the original per-credential in-memory state, so the synthetic regression above should not be read as proof of exactly how our production credentials entered that state.

Would you prefer adding this regression and closing the remaining cap-preservation paths here rather than a competing cooldown PR? A 5h cap bounds the wait but is not immediate detection of an upstream quota reset. I have not deployed this branch, disabled cooling, or exercised any reset-credit redemption behavior.

PR #5921, opened 2026-09-17
codex/openai: strip prompt_cache_breakpoint from output parts and item level

Tamnac:
## Problem

`ConvertOpenAIResponsesRequestToCodex` strips `prompt_cache_breakpoint` (a GPT-5.6+ platform Responses field the Codex backend rejects with `400: prompt_cache_breakpoint is not supported on this model`), but `stripCodexResponsesCacheBreakpoints` only walked `input[].content[]` parts.

Clients that attach the marker in other valid Responses locations leak it upstream:

- inside a `function_call_output` item's `output[]` content-part array
- as an item-level field on an `input[]` entry

Observed live: a Responses client with a marker inside `function_call_output.output[]` got `400 prompt_cache_breakpoint is not supported on this model` from `chatgpt.com/backend-api/codex/responses` despite the stripper running (its message-content markers were stripped).

## Fix

Walk both `content` and `output` part arrays, and delete item-level `prompt_cache_breakpoint` fields. Same fast path (`bytes.Contains`) and rebuild-on-change strategy as before.

## Tests

- `TestStripCodexResponsesCacheBreakpoints_FunctionCallOutputParts` — marker inside `output[]` parts is stripped, text preserved
- `TestStripCodexResponsesCacheBreakpoints_ItemLevel` — item-level markers on message and string-output `function_call_output` items are stripped
- Existing breakpoint tests still pass


Tamnac:
My bad, I'll open an issue first

PR #5926, opened 2026-09-18
feat(plugin): expose the request id on execution metadata

nguyenha935:
## The gap

`scheduler.pick` carries no request id, so a plugin that refuses a candidate there cannot tie that refusal back to the request it just opened in `request.intercept_before`. The request is then settled from the host's completion outcome, which reports `upstream_rejected`: the provider gets blamed for the plugin's own decision.

Measured on a live deployment before this patch: 0 rows in the whole history carry a plugin-side refusal reason, while 111 rows carry `upstream_rejected` with no way to tell which of them were actually the plugin refusing.

## What this adds

No new plumbing. The metadata map handed to the request lifecycle tracker **is** the map that becomes `Options.Metadata` for the execution, and `Options.Metadata` is already copied into `SchedulerOptions.Metadata`. Writing one key into that map is enough: two files, one constant and a four-line block.

The new constant sits after `RequestedModelMetadataKey` rather than between that constant and its doc comment, so no existing line is touched. This PR deletes nothing: the diff is 13 added lines and 0 removed.

A host without this patch simply does not set the key, and a plugin falls back to its previous labelling rather than breaking.

## Verification

Rebased onto `dev` (`05391d7b`).

`pr-test-build` only builds, so the green check above does not speak for the tests. I ran them:

- `go build ./...` clean; `gofmt` clean on both files.
- `go test ./sdk/api/... ./sdk/cliproxy/... ./internal/pluginhost/...` green.
- `go test ./...` shows one failure in `internal/runtime/executor`, and it is not from this change: pristine `dev` fails in that same package on its own, 1 red in 3 runs, a different test each time. This PR does not touch that package.

Deployed and in use here since 2026-09-15: the plugin reads the id off `SchedulerOptions.Metadata` and now records its own refusals against the right request.


nguyenha935:
Closing this myself too: I can do it inside my plugin, and the measurement does not justify asking you to carry it.

This published the request id on the execution metadata so a scheduler plugin could attribute its own refusal to the request it belongs to. It ran on a live deployment from 2026-09-15 and worked. Over **5,014 requests it produced 0 rows** carrying a plugin-side refusal reason, because the condition it labels, a key configured so that no permitted credential exists, is rare.

And it turns out the plugin can do the attribution on its own: `scheduler.pick` runs after the request row is already open, and the plugin keeps every in-flight request in memory with its key and model, so it can match on those instead of on an id. Each request gets its own pick call, so N refusals produce N calls and each one lands on exactly one row; the pairing between two otherwise identical rows is undefined, but the number of labelled rows and the label on each are right. Verified end to end against your unpatched `v7.3.7`: the refusal reaches the settled row as `provider_not_allowed`.

Given that, asking you to carry 13 lines forever for something a plugin can solve on its side is the wrong trade. Thanks for the review time.

PR #5927, opened 2026-09-18
feat(plugin): carry the request id on usage records

nguyenha935:
> **Stacked on #5926.** The first commit here is that PR; this one adds the second. Merge #5926 first and this diff reduces to the usage-record change alone.

## Problem

A usage plugin cannot tell which request an upstream attempt belongs to.

`UsageRecord` names the provider, the credential and the key, and carries `RequestedAt`, but not the request id the same plugin received on `request.intercept_before` and `request.complete`. So a plugin that wants to show per-request attempts has to guess: match on the api key, then the credential, then a time window.

That collides as soon as one key has several requests in flight on the same credential, which is the normal case for an editor client.

Measured on one deployment: **225 requests ended up holding two or more _successful_ attempts**, every one of them on the same credential, while the requests those attempts really belonged to held none. Each attempt had its own latency and its own token counts — e.g. 56.0s, 6.0s and 26.4s on a single request — so these were distinct upstream calls attributed to the wrong request, not duplicate reports of one call.

## Approach

This follows the path the package already uses for the requested-model alias, the reasoning effort and the service tier:

1. read the value off `Options.Metadata` in `contextWithRequestedModelAlias`
2. put it on the context (`usage.WithPluginRequestID`)
3. let `NewUsageReporter` pick it up alongside the other context-carried fields
4. carry it on `usage.Record` and pass it through to `pluginapi.UsageRecord`

No new plumbing and no new call sites: `contextWithRequestedModelAlias` is already invoked at every execution entry point.

## Compatibility

A host that does not publish the metadata key leaves the field empty, and a plugin reading it falls back to whatever it did before. Nothing in the server reads the field.

## Verification

- `go build ./...` clean
- `gofmt` clean on all five touched files
- `go test ./sdk/cliproxy/... ./internal/pluginhost/... ./internal/runtime/executor/...` — 9 packages, all pass


nguyenha935:
Closing this one myself: I opened it without a measured need for it.

I maintain a gateway plugin against this host, and I checked what it actually
uses. The plugin never reads the request id off a usage record. The problem
this PR would solve, a usage report that can only be attributed to a request
by guessing from the key, the credential and a time window, has not occurred
once: 0 of 10,797 recorded upstream attempts fell back to the guessing path.

So this is plumbing for a problem I cannot demonstrate, and it is 7 files
against #5926's 2. That is the wrong trade to ask a maintainer to review.

#5926 stays open and is the one that matters: it publishes the same request id
on the execution metadata, which is what lets a scheduler plugin attribute its
own refusal to the request it belongs to. That one is 2 files, purely additive,
and it is deployed and in use here.

If a real case for carrying the id onto usage records shows up later, I will
reopen with the measurement that justifies it. Sorry for the review noise.

PR #5928, opened 2026-09-18
feat(codex): record upstream response model and warn on silent model

Viggo95:
### Problem

On some Codex credentials — in our probes, **the ones with most of their weekly window already
used — OpenAI silently downgrades the request:** ask for `gpt-6-astra`, get `gpt-5.6-luna`, HTTP 200,
with `response.model` naming the substitute. The proxy passes this straight through. **The
consequences:** 

1. users report "gpt-6 got dumber" and cannot tell a weaker model from a worse day;
2. operators cannot see which credential was downgraded, how often, or since when; and the usage
record has no field to carry the fact — nothing is logged, nothing is counted.

The two pass-through sites: the chat-completions translator copies `response.model` into the
chunk's `model` (`internal/translator/codex/openai/chat-completions/codex_openai_response.go:94`);
and on `/v1/responses`, the path the Codex CLI itself uses, `setResponsesModel`
(`internal/translator/codex/openai/responses/codex_openai-responses_response.go:27`) fills
`response.model` in only when it is **missing**, forwarding an existing value untouched.

An **explicit** `model_not_found` cools the (credential, model) pair for 12 h and rotates away
(`sdk/cliproxy/auth/conductor_cooldown.go`); the same outcome dressed as a 200 with a different
`model` gets nothing. The two are not symmetric, though: `model_not_found` delivers and bills
nothing, a substitution is already delivered and already billed, so failing it over would discard a
response that has been paid for. **Hence record, not reroute.**
image

### Solution

Observability only; routing, selection and the delivered bytes are untouched.

1.  **`Record.ResponseModel`** in `sdk/cliproxy/usage`, beside the existing `ResponseServiceTier`;
    the redis usage queue emits it as the optional `response_model` payload field next to
    `response_service_tier`. Only the record for the requested model carries it — additional-model
    records (image generation tool usage) name a side model the response never refers to.
2.  **Extraction**, in `helps/response_model.go`: SSE frames and raw JSON alike, restricted to the
    five events that embed the authoritative response object — `response.created`,
    `response.in_progress` and the terminal `response.completed` / `.incomplete` / `.done`.
    Everything else, deltas included, returns empty after one cheap `type` lookup, and since
    `response.model` is upstream-controlled, non-string and over-128-byte values are rejected.
3.  **Decision**, `IsCodexModelSubstituted`: both sides trimmed, lower-cased and stripped of their
    thinking suffix (symmetrically — the upstream may answer with a suffixed name), and a dated
    alias accepted in either direction and in both spellings (`gpt-5.6-terra-2026-05-13`,
    `gpt-5.6-sol-20260513`). Everything else is a substitution; a prefix rule would wave through
    `gpt-5.5` → `gpt-5.5-mini`.
4.  **Wiring**: the hook sits inside `observeCodexTokenEvent`, whose signature does not change, so
    every streaming caller picks it up unmodified; the non-streaming `Execute` loop and the two
    `/responses`-shaped image entry points call it directly. Observation does no logging work, and
    an `atomic.Bool` short-circuits frames after a terminal event.
5.  **The WARN fires when the attempt publishes its usage record**, not on the event path, so
    nothing is logged before the first event is forwarded. Repeats are throttled per (credential,
    requested, served), normalized, for 10 minutes: on an affected credential *every* request is
    substituted, and an unthrottled warning would mirror the request volume into the logs. The
    credential is labelled `auth_index=<index>` only — codex credential file names embed the
    account e-mail.
6.  `response_model` ships with every usage record and is the source to count from; **the WARN is a
    throttled alerting signal, not a count.** `TokenAccountingSchemaVersion` is not bumped: it
    versions the token accounting contract, and this field is optional and non-token.

### Coverage by entry point

| Entry point | `response_model` | Why |
| --- | --- | --- |
| HTTP streaming: buffered handshake + streaming goroutine | recorded | `observeCodexTokenEvent` |
| HTTP non-streaming `Execute` loop | recorded | SSE event loop |
| Websocket, streaming and non-streaming | recorded | `observeCodexTokenEvent` |
| The two `/responses`-shaped image entry points | recorded | same `response.*` events |
| `executeCompact` (`/responses/compact`) | not wired | compaction object: no `type`, no `response.model` |
| `/images/generations`, `/images/edits` | not wired | Images API shape, `image_generation.*` events |
| `CountTokens` | not wired | local tiktoken count, no upstream call |

The unwired three have no `response.model` to read, so wiring them would only add parsing that
always returns empty.

### What this PR does not do

No routing or selection change, nothing in `/v1/models` or `models.json`, no new configuration
option, no response header, no management endpoint. The response is delivered exactly as served with
its `model` untouched, and a credential that could not get a model still cannot. Later, not here,
and shaped by the cause: prefer gated models by the remaining quota the proxy already parses, and
project availability onto `/v1/models` through the existing registry.

### Tests

```
go build ./...                       PASS
go test -race  (touched packages)    PASS, 0 race warnings
go test ./...                        PASS
```
gofmt and go vet are clean on the touched packages.

Maybe Related:  #5901


Viggo95:
@codex  review

Viggo95:
@codex review

PR #5929, opened 2026-09-18
feat(plugins): let a plugin narrow the model list a caller sees

nguyenha935:
## The gap

The model listing endpoints serve `/v1/models` and `/v1beta/models` from a global registry keyed only by handler type. The lookup takes no principal and no request context, and its result is cached per handler type, so every caller receives the same catalogue no matter which API key they present.

A gateway plugin that restricts which models a key may **call** therefore cannot restrict which models that key can **see**. Two consequences:

- Clients that pick a model from the listing (Cline, Roo, Cherry Studio) show the user every model and then fail on most of them.
- A restricted key still enumerates the whole catalogue.

## Why the existing response interceptor does not cover it

`WriteModelListResponse` runs `interceptResponse` on model-list bodies since 5b278561, and it passes `RequestHeaders`, so on the OpenAI and Anthropic ports a plugin can identify the caller from a header. That is genuinely enough for those two.

**It is not enough for the Gemini port.** That port accepts the credential in three forms: `x-goog-api-key`, `Bearer`, and `?key=` in the query string. In the `?key=` form there is no header to read, so a header-based interceptor cannot identify the caller at all and cannot filter anything. Measured on a live deployment with a restricting plugin installed: `/v1beta/models?key=<restricted key>` returned all 42 served models.

That hole is why this is a hook on the assembled list rather than a response interceptor.

## What this adds, file by file

14 files looks like a lot for one feature, so here is why each is touched. Nine of them are one to seven lines.

| File | Lines | Why |
|---|---|---|
| `sdk/api/modelvisibility/visibility.go` | +161 (new) | The filter itself: `Set`, `Apply`, `FilterMaps`, and the `Request` a filter is allowed to see |
| `internal/pluginhost/adapters_modelvisibility.go` | +74 (new) | Installs the filter of the highest-priority plugin declaring the capability |
| `sdk/pluginapi/types.go` | +37 | `ModelVisibilityRequest`/`Response` and the capability flag |
| `sdk/api/handlers/openai/openai_handlers.go` | +17/-3 | Filter the list, before the `?client_version=` Codex branch as well |
| `sdk/api/handlers/openai/codex_client_models.go` | +10/-1 | Lets the Codex response be built from an explicit list; the existing signature is kept so existing tests still compile |
| `sdk/api/handlers/claude/code_handlers.go` | +8/-1 | Filter the list |
| `sdk/api/handlers/gemini/gemini_handlers.go` | +7/-1 | Filter the list |
| `internal/pluginhost/rpc_client.go` | +7 | Carry the new method over the RPC transport |
| `sdk/pluginabi/types.go` | +3 | One ABI method constant |
| `internal/pluginhost/host.go` | +2 | Re-derive the filter on the two plugin-unload paths |
| `sdk/cliproxy/builder.go` | +1 | Register on startup |
| `sdk/cliproxy/service_plugins.go` | +1 | Register on plugin reload |
| `sdk/cliproxy/service_lifecycle.go` | +1 | Register on service lifecycle |
| `internal/pluginhost/rpc_schema.go` | +1 | Declare the capability in the RPC schema |

No file outside this list is touched, and nothing under `internal/translator/`.

## Failure behaviour

Every failure mode keeps the full list: no filter installed, a filter that returns nil, a filter that panics, or a plugin that errors. A listing endpoint must not go empty because a plugin misbehaves. A plugin that wants to hide models can always return an explicit, possibly empty, slice.

With no plugin declaring the capability the filter is the identity function and the server behaves exactly as before.

## Verification

Rebased onto `dev` (`05391d7b`).

`pr-test-build` only builds, so the green check above does not speak for the tests. I ran them:

- `go build ./...` clean; `gofmt` clean on every file this touches.
- `go test ./sdk/api/... ./sdk/cliproxy/... ./internal/pluginhost/...` green, five consecutive runs, including `dev`'s new `TestCodexClientModelsResponse_DevinDisplayName`.
- `go test ./...` on this branch shows one failure in `internal/runtime/executor`, and it is not from this change: pristine `dev` fails in that same package on its own, 1 red in 3 runs, a different test each time. Nothing in this PR touches that package.

Behaviour measured on a live deployment, restricted key vs unrestricted key, all five request shapes of the three ports:

| Shape | Key declaring 5 models | Unrestricted key |
|---|---|---|
| openai `/v1/models` `Bearer` | 4 | 42 |
| openai `/v1/models?client_version=` | 4 | 42 |
| anthropic `/v1/models` `x-api-key` | 4 | 42 |
| gemini `/v1beta/models` `x-goog-api-key` | 4 | 42 |
| gemini `/v1beta/models?key=` | 4 | 42 |

Four rather than five is correct: one allowed model is not currently served, and the filter only narrows, it never adds. Models seen that were not in the allow list: 0.


nguyenha935:
Closing this myself. I found a way to do it without touching your source at all, and the argument I built this on does not survive measurement.

**The argument was the Gemini `?key=` form.** Your existing `interceptResponse` on model-list bodies carries `RequestHeaders`, which is enough on the OpenAI and Anthropic ports, but the Gemini port also accepts the credential as `?key=` in the query string, where a header-based filter cannot identify the caller. That part is true.

**What I never measured is whether anyone uses it.** One month of request logs on a live deployment:

| Shape | Requests |
|---|---|
| `/v1/models` | 6,942 |
| `/v1/models?client_version=` | 276 |
| `/v1beta/models` (header auth) | 10 |
| `/v1beta/models?key=` | **0** |

So this PR adds a hook, an ABI method and a package to 14 files in order to close a listing leak on a path that has not been used once. My plugin now filters the list from `response.intercept_after`, which you already ship, and it covers every shape that has real traffic. Measured on a live deployment running your **unpatched** `v7.3.7`: a key declaring 5 models sees 4 on `/v1/models` (Bearer), `?client_version=`, `/v1/models` (`x-api-key`) and `/v1beta/models` (`x-goog-api-key`), while an unrestricted key sees all 42.

Two things I got wrong that may be worth knowing anyway, since they cost me a while:

- `RequestInterceptRequest.Headers` and `ResponseInterceptRequest.RequestHeaders` are two names for the same idea, and they are not the same thing: the first is the rewritten upstream request, the second is the original client request. A plugin sharing one struct for both hooks silently reads an empty map on the response path. That is what made me believe the hook could not see the caller.
- The Codex client body from `?client_version=` puts the model id in `slug`, with no `id` and no `name`. A filter looking only for `id`/`name` leaves that one client seeing everything.

Thanks for the review time, and sorry for spending it on this.

PR #5938, opened 2026-09-18
fix(openai): map response token limit to completion tokens

jerstadgeirivar-gmail:
## Summary

Translate the Responses API `max_output_tokens` field to `max_completion_tokens` when converting requests to Chat Completions.

Newer OpenAI and Azure AI Foundry reasoning models reject the legacy `max_tokens` parameter with HTTP 400. The current translator still emits `max_tokens`, while the equivalent completion-token field is accepted by these models.

## Changes

- Map `max_output_tokens` to `max_completion_tokens` in the OpenAI Responses request translator.
- Add a regression test asserting the new field is emitted and the legacy field is absent.

## Validation

- `go test -count=1 ./internal/translator/openai/openai/responses/...` passes in WSL.


jerstadgeirivar-gmail:
The repository path guard requires restricted internal/translator changes to go through a maintenance issue first. Tracking this fix in #5939 with the passing regression test.

Issue #5809, opened 2026-09-14
[Feature] Document pluginabi.Error.HTTPStatus so plugin executors stop flattening upstream 403/429 into HTTP 500


pinewhite:
# `pluginabi.Error.HTTPStatus` is undocumented, so plugin executors turn upstream 403/429 into HTTP 500 `server_error`

## Summary

When a plugin executor reports a failed upstream call, CPA answers the client with HTTP `500` / `type: "server_error"` / `code: "internal_server_error"`, even when the upstream status was a `403` or `429` that CPA already knows how to classify correctly.

The plumbing for a correct status is **already implemented end to end** — it is simply undocumented, so plugin authors (and the bundled examples) never set it. This report is a documentation/ergonomics request rather than a defect in the classification logic.

## What happens

A plugin executor calls an upstream provider, the provider returns `403`, and the client receives:

```http
HTTP/1.1 500 Internal Server Error
```
```json
{
  "error": {
    "message": "commandcode-go: upstream returned 403: {\"success\":false,\"error\":{\"code\":\"FORBIDDEN\",\"status\":403,\"message\":\"MODEL_NOT_IN_PLAN: GPT-5.5 available in Pro and above plans or extra on demand usage\"...}}",
    "type": "server_error",
    "code": "internal_server_error"
  }
}
```

The upstream status (`403`) is present only inside the human-readable message string. The machine-readable `type`/`code` say "the gateway broke", which is not what happened: the request was permanently rejected due to plan permissions.

After setting the (undocumented) field described below, the same request correctly yields:

```http
HTTP/1.1 403 Forbidden
```
```json
{ "error": { "...": "...", "type": "permission_error", "code": "insufficient_quota" } }
```

Verified for both `stream: false` and `stream: true`, on 5 denied models (10/10 corrected), with an allowed model still returning `200` on both paths.

## Why it happens

CPA has the full chain in place:

1. `sdk/pluginabi/types.go:123` — the RPC error envelope already carries a status:
   ```go
   type Error struct {
       Code       string `json:"code"`
       Message    string `json:"message"`
       Retryable  bool   `json:"retryable,omitempty"`
       HTTPStatus int    `json:"http_status,omitempty"`
   }
   ```
2. `internal/pluginhost/rpc_client.go:366-370` — the host decodes that field into an error that exposes it:
   ```go
   return zero, rpcError{
       Code:       strings.TrimSpace(envelope.Error.Code),
       message:    message,
[...]

PR #5971, opened 2026-09-19
fix(pluginhost): preserve upstream HTTP status across host callbacks

### LINKED ISSUE #5970: [Bug] Host callback flattens upstream HTTP status, so plugins observe 0 and cannot route on 429/503
## Summary

#5809 established `pluginabi.Error.HTTPStatus` as the mechanism for carrying an upstream status across the plugin boundary, and fixed the plugin to host direction. The host to plugin direction still drops it.

When a plugin delegates an upstream call back to the host through `cliproxyHostCall` and that call fails, the host serialises its own error with `marshalRPCError(code, message)`. That builds a `pluginabi.Error` with no `HTTPStatus`, so the plugin always observes status `0`, even though the host knew the real status.

## Why it matters

A routing plugin cannot make status based decisions. The published `model-router` plugin maps an alias to an ordered list of target models and is meant to fall through to the next target on a retryable upstream status. Receiving `0` forces it back onto substring matching against the error message, which fails whenever the upstream wording is not one it recognises.

With a synthetic upstream returning:

```http
HTTP/1.1 429 Too Many Requests
```
```json
{"error": {"code": "rate_limit", "message": "synthetic"}}
```

the alias answers `HTTP 500` carrying that body, and exactly one upstream call is made, to the primary target. The second target is never attempted on that request. A later request does reach it, because the primary has been marked as cooling by then, but same request failover does not happen.

This is easy to miss: if the upstream message happens to contain the words "rate limit", the message heuristic matches and failover appears to work. A fixture with that wording hides the defect entirely.

## Where

- `internal/pluginhost/rpc_client.go`, `marshalRPCError` builds the envelope with no status
- `internal/pluginhost/host_callbacks_unix.go:46`, unix callback site
- `internal/pluginhost/loader_windows.go:372`, windows callback site

The remainder of the chain is already correct and needs no change. `rpcError.StatusCode()`, `clienterror.HTTPStatusFromError` and the handler classification table all behave as #5809 documented. Only the host side construction of the envelope is missing the status.

Both callback sites need it. This is the same "easy to get half right" split that #5809 noted for the streaming and non streaming executor paths.

## Reproduction

1. Configure a routing plugin with one alias and two targets, ordered.
2. Point the first target at an upstream that returns `429` with a message containing no recognisable rate limit wording, for example `{"error":{"code":"rate_limit","message":"synthetic"}}`.
3. Point the second target at a healthy upstream.
4. Send one non streaming request to the alias.

Expected: the request fails over to the second target, or at minimum the client receives `429` rather than `500`.

Observed: `HTTP 500` with the upstream body, and a single upstream call to the first target only.

## Versions

Reproduced on v7.3.4, v7.3.7, v7.3.8 and `main` at c93978c. `marshalRPCError` has had no status parameter throughout.

## Suggested fix

Make `marshalRPCError` accept an optional status and use the existing `pluginabi.NewErrorEnvelope` helper, then pass `clienterror.HTTPStatusFromError(errCall)` at both callback sites. Three production lines, backwards compatible because the parameter is variadic and existing two argument callers keep compiling.

Happy to open a PR with that plus a table driven test if useful.

melalj:
Fixes #5970

## What this changes

`marshalRPCError` now accepts an optional HTTP status and builds its envelope through the existing `pluginabi.NewErrorEnvelope` helper. Both host callback sites pass `clienterror.HTTPStatusFromError(errCall)`.

Three production lines. Everything downstream already worked: `rpcError.StatusCode()`, `clienterror.HTTPStatusFromError` and the handler classification table are untouched.

## Why this shape

- It reuses `pluginabi.NewErrorEnvelope` rather than introducing a second way to build the envelope, following the direction set in #5809.
- The parameter is variadic, so it is source compatible. The existing two argument callers in `host_test.go` and `rpc_client_error_test.go` compile unchanged.
- An unknown status stays `0` and is omitted by `json:"http_status,omitempty"`, so the envelope is byte identical to today whenever the host has no status to report. Callers that currently receive `500` for a genuinely unknown failure keep receiving `500`.
- Both the unix and windows callback sites are covered. #5809 observed that this class of fix is easy to apply to one path and miss the other.

## Testing

- `TestMarshalRPCErrorPreservesHTTPStatus`, table driven over no status, explicit `0`, `400`, `422`, `429` and `503`, asserting the decoded error exposes the expected `StatusCode()` while the message is preserved.
- `go test -race ./internal/pluginhost/ ./internal/clienterror/ ./sdk/pluginabi/` passes.
- `GOOS=windows GOARCH=amd64 go build ./internal/pluginhost/` passes, covering the second callback site.
- End to end against a loopback upstream returning `429` with a message containing no recognisable rate limit wording: before this change the alias returns `500` and attempts only the first target; after it, the status reaches the plugin and the ordered fallback proceeds on the same request.

The end to end case is the one worth keeping in mind. A fixture whose error message contains the words "rate limit" passes either way, because the plugin falls back to matching on the message text.


melalj:
Pushed a second commit after a closer look at the callback path. The first commit alone was only a partial fix, and I would rather say so than let it read as complete.

`modelExecutionError` returns `errMsg.Error` and discards `errMsg.StatusCode`. It sits upstream of the callback sites, so for `host.model.execute` the status survived only when the inner error happened to implement `StatusCode()` itself. Several in-tree producers pair a status with a plain `errors.New` or `fmt.Errorf`, including the 400 in `sdk/api/handlers/model_execution.go` and the 502/503 paths in `handlers_execution.go` and `handlers_stream.go`. For those, a delegated 429 still reached the plugin as status 0 even with the first commit applied. My original end-to-end check passed only because its fixture error did implement `StatusCode()`.

The second commit attaches the reported status when the inner error does not report one, wrapping with `Unwrap` so `errors.Is` and `errors.As` still see the original and the message is unchanged. An inner error that reports its own status keeps it.

Two corrections to my original description, since both are behaviour changes worth stating plainly rather than leaving for a reviewer to find:

- Plugins now observe 499 for a cancelled client request and 504 for a deadline, where they previously saw 0. `HTTPStatusFromError` maps `context.Canceled` and `context.DeadlineExceeded`, and the callback context is the caller's request context.
- A plugin that forwards host-call errors verbatim as its own executor error will now propagate a delegated credential's 401 or 403 into its own credential handling, where before it produced a generic 500. That is plugin-authored behaviour rather than host behaviour, but it is a real consequence of the change.

Tests added for the gap: a table covering a plain inner error with a status, no inner error, an inner error that reports its own status, and no status anywhere; plus one asserting the wrapping stays unwrappable. Both assert through the envelope a plugin actually decodes. `go test -race` passes for `internal/pluginhost`, `internal/clienterror` and `sdk/pluginabi`; the Windows cross-build and `go build ./cmd/server` pass.

melalj:
Closing: this was fixed upstream in 1c87874966a0c636e46ae4268faddd99a974b392 (shipped in v7.3.9), which takes the same approach including the modelExecutionError wrapper and the plain-inner-error case covered by the second commit here. Nothing left for this PR to add. Thanks for picking it up.

PR #5976, opened 2026-09-19
fix(gemini): round-trip client-executed tool_search for Responses

### LINKED ISSUE #5588: [Bug] Gemini/Antigravity Responses translation drops tool_search and discovered tools, hiding Codex deferred agents
## Describe the bug

The OpenAI Responses -> Gemini/Antigravity translation path drops native client-executed `tool_search`, making Codex deferred tools undiscoverable. In this setup, the affected tools include Codex's native multi-agent tools: the model can use basic tools but cannot discover `spawn_agent` and therefore cannot dispatch an Explorer sub-agent.

This is a protocol compatibility report, not a request to implement a separate agent scheduler in CLIProxyAPI. Codex should continue to execute tool discovery and agent scheduling locally.

## Environment

- CLIProxyAPI: **7.2.152**, commit **c76dfd4e** (local startup log)
- Client: Codex Desktop, bundled **codex-cli 0.153.4**
- Upstream account/provider: **Antigravity**
- Model: **gemini-3.8-flash-high**
- OS: Windows, native deployment
- Endpoint: `POST /v1/responses`
- Custom Codex model catalog: Gemini retains `tool_mode: null` and uses `multi_agent_version: "v1"`, with `supports_search_tool: true`.
- Codex features: `multi_agent = true`, `multi_agent_v2 = false`.

Important configuration caveat: this is a custom catalog that copies selected GPT-5.6 Sol prompt/capability fields, including `supports_search_tool`, to Gemini. We are **not** claiming stock Codex declares this capability for Gemini or that stock Codex officially supports this provider combination.

## Is it a request payload issue?

Yes. Below are sanitized, minimal translator fixtures, **not a verbatim production capture**. No credentials or user conversation content are included.

### Fixture A: search entry point

```json
{
  "model": "gemini-3.8-flash-high",
  "stream": true,
  "input": "Find the multi-agent tools before answering.",
  "tools": [
    {
      "type": "tool_search",
      "execution": "client",
      "description": "Search available deferred tools. Multi-agent tools: Spawn and manage sub-agents.",
      "parameters": {
        "type": "object",
        "properties": {
          "query": { "type": "string" },
          "limit": { "type": "integer" }
        },
        "required": ["query"],
        "additionalProperties": false
      }
    }
  ]
}
```

Expected: expose a Gemini-compatible function shim preserving the incoming search description/schema. When Gemini invokes it, return a native Responses `tool_search_call` with `execution: "client"`, matching `call_id`, and object-valued arguments so Codex can execute discovery.

Current source behavior: the tool type is not collected, so no corresponding Gemini function declaration is produced.

### Fixture B: tools returned by discovery

```json
{
  "model": "gemini-3.8-flash-high",
  "stream": true,
  "tools": [],
  "input": [
    {
      "type": "tool_search_call",
      "execution": "client",
      "call_id": "search_1",
      "status": "completed",
      "arguments": { "query": "fixture tools", "limit": 1 }
    },
    {
      "type": "tool_search_output",
      "execution": "client",
      "call_id": "search_1",
      "status": "completed",
      "tools": [
        {
          "type": "namespace",
          "name": "fixture_tools",
          "description": "Minimal discovery fixture, not the actual Codex agent schema",
          "tools": [
            {
              "type": "function",
              "name": "ping",
              "description": "Return a test reply",
              "parameters": {
                "type": "object",
                "properties": {},
                "additionalProperties": false
              }
            }
          ]
        }
      ]
    }
  ]
}
```

Expected: activate `fixture_tools.ping` in the next Gemini request, translate the search history consistently, and restore the original namespace/name when the function is called.

Current source behavior: tool collection reads top-level `tools` and `additional_tools` input items, but not `tool_search_output.tools`.

## Source evidence (v7.2.152)

- [`responsesToolSources` / `CollectResponsesToolDescriptors`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.152/internal/util/responses_tools.go): sources include only top-level tools and `additional_tools`; supported type cases include function/custom/namespace but not `tool_search`.
- [`BuildGeminiFunctionDeclarations`](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.152/internal/util/responses_tools.go): builds declarations from those collected descriptors.
- [Gemini Responses request converter](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.152/internal/translator/gemini/openai/responses/gemini_openai-responses_request.go): uses that declaration builder and lacks explicit native search call/output handling.

## Observed behavior and validation boundary

In the affected Gemini task, the model reported having basic execution tools but no sub-agent dispatch interface. Earlier isolated client-side request capture in this investigation showed the initial Codex request exposing native `tool_search` with multi-agent discovery instructions instead of directly exposing `spawn_agent`.

An earlier isolated catalog experiment setting only `supports_search_tool = false` made the client directly include `spawn_agent`, `wait_agent`, `send_input`, `resume_agent`, and `close_agent`. This confirms an eager-loading workaround at the **client request** level; it is **not** an end-to-end agent-dispatch success claim. The production catalog has not been changed.

The minimal fixtures above are derived from the source/protocol analysis. A patched real-model end-to-end run has **not** been performed.

## Expected fix / acceptance criteria

Please support the complete client-executed search round trip on the Gemini/Antigravity path:

1. Translate `tool_search` into a normal upstream function, preserving the original discovery description and parameter schema.
2. Restore calls to that specific shim as native `tool_search_call` items, including streaming events and the final response envelope.
3. Replay `tool_search_call` / `tool_search_output` with consistent call IDs and valid Gemini function history.
4. Collect and activate tools in `tool_search_output.tools`, preserving namespace/name identity and avoiding duplicate declarations.
5. Keep not-yet-discovered tools deferred, rather than eagerly injecting the full catalog.
6. Cover normal non-search tools, multiple discovery rounds, and namespace/name collisions with regression tests.

Disabling search is undesirable as a permanent workaround because it places tool definitions into context upfront. Switching Codex to Code Mode is a separate compatibility change and is not required for this fix.

## Related

- #3361 describes the analogous Responses tool-search round trip for Claude; this report concerns Gemini/Antigravity.
- #3008 concerns compact payload validity and is a different failure mode.
- [OpenAI client-executed tool search protocol](https://developers.openai.com/api/docs/guides/tools-tool-search)
- [Community implementation reference](https://github.com/bharat2808/codex-universal-proxy/blob/main/src/proxy.js) and [discovery tests](https://github.com/bharat2808/codex-universal-proxy/blob/main/test/proxy.test.js). These are implementation references, not verified compatibility claims for this specific deployment.

wsyzxjn:
Fixes #5588

## Summary

The OpenAI Responses → Gemini translation (also used by the Antigravity Responses path) dropped `tools[].type == "tool_search"` and never handled replayed `tool_search_call` / `tool_search_output` items. Codex 0.155 defers every user-configured MCP server behind client-executed `tool_search`, so Gemini models could neither discover nor call those tools: the model reports that the MCP tools do not exist, while the same session on an official model (code mode) sees them.

This implements the round trip described in #5588 on the Gemini/Antigravity path:

- **Declaration**: a client-executed `tool_search` is exposed to Gemini as a plain function named `tool_search`, keeping the original description and parameter schema (a default `{query, limit}` schema is used when none is provided). Server-executed `tool_search` is still dropped, and a real tool that already uses the shim name keeps its identity.
- **Discovered tools**: tools carried by replayed `tool_search_output` items are collected as a declaration source (lowest precedence, after top-level `tools` and `additional_tools`), so tools discovered in an earlier turn can be called later. Duplicates resolve through the existing winner logic.
- **History replay**: `tool_search_call` / `tool_search_output` items are rewritten into the `function_call` / `function_call_output` shape before the existing pairing logic, so Gemini receives a valid `functionCall` / `functionResponse` pair with matching ids. The response text lists the Gemini-visible names the discovery made callable.
- **Response** (streaming and non-streaming): calls to the shim are restored as native `tool_search_call` items with `execution: "client"`, a stable `call_id`, and object-valued `arguments`, without `function_call_arguments` events. Calls to discovered tools resolve through the same identity map, so `name` / `namespace` are restored exactly as for eagerly declared tools.

Not-yet-discovered tools stay deferred; nothing is eagerly injected.

Scope note: the shim is added only in `BuildGeminiFunctionDeclarations`, not in the shared `CollectResponsesToolDescriptors`, so the Interactions translator does not start declaring a `tool_search` function it cannot restore. The `tool_search_output` declaration source is shared, which only adds declarations for tools the client says were discovered.

## Test plan

- [x] `gofmt -l .` clean, `go vet` on touched packages
- [x] `go test ./...` passes (14 new tests: shim declaration, default schema, server-executed skip, name collision, discovered-tool dedupe and namespace restore, history replay incl. string arguments and empty discovery, streaming/non-streaming `tool_search_call` shape, Antigravity envelope + response unwrapping)
- [x] `go build -buildvcs=false -o test-output ./cmd/server`
- [x] End to end with Codex CLI 0.155.0 → patched proxy → Antigravity `gemini-3.8-flash-high`:
  1. Gemini calls `tool_search`; the proxy returns a `tool_search_call`; Codex runs the local BM25 discovery.
  2. Codex replays `tool_search_call` + `tool_search_output`; Gemini continues without a 400.
  3. Gemini calls a discovered connector tool (`mcp__codex_apps__plugin_management` / `_search_plugins`); it comes back as `function_call` with the correct `name` and `namespace`, Codex executes it and the result is fed back.

Prepared and tested with Codex assistance.


wsyzxjn:
Closing: changes under internal/translator are reserved for the maintainers (path guard). The branch stays available as a tested reference implementation for #5588: https://github.com/wsyzxjn/CLIProxyAPI/tree/fix/gemini-responses-tool-search

PR #5862, opened 2026-09-16
feat(antigravity): support web search translation and URL resolution in OpenAI Responses

sususu98:
## Why

When clients use the OpenAI Responses API (`/v1/responses`) with the `web_search` tool against Gemini or Antigravity models, the search tool was previously unrecognized, and Google Search Grounding metadata was not mapped back to Responses `web_search_call` output items or `url_citation` annotations. Furthermore, Vertex AI Search grounding redirect URLs (`vertexaisearch.cloud.google.com/grounding-api-redirect/...`) remained unresolved when using Responses format.

## What

- **Capability Gating**:
  - Check dynamic registry capabilities (`SupportsWebSearch`) and `models.json` definitions (`native_capabilities.web_search`), prioritizing explicit `false` as an absolute veto.
  - Isolate Antigravity web search gating to Antigravity-specific model capabilities (`AntigravityWebSearchModelFor` and Antigravity-specific static veto) to prevent cross-provider capability bleed.
  - Register `antigravity` in `responsesWebSearchProviderPathSupport`.
- **Request Translation**:
  - Map Responses `web_search` tool to Antigravity `requestType: "web_search"` envelope and `googleSearch` tool.
  - Forward domain filters (`filters.allowed_domains` -> `googleSearch.includedDomains`).
  - Suppress native `googleSearch` in Antigravity chat fallback when tools are mixed with function declarations, preserving function tools without triggering upstream rejects.
  - Support Responses `allowed_tools` tool_choice when it contains web search.
  - Concatenate multi-part `input_text` / message content parts into the complete search query.
  - Guard `toolConfig.functionCallingConfig` behind `len(functionDeclarations) > 0` to prevent Google API 400 errors.
- **Response Translation & Streaming Fidelity**:
  - Map Google `groundingMetadata` to Responses `web_search_call` output items and `url_citation` annotations.
  - Buffer streaming text deltas while awaiting `groundingMetadata` so `web_search_call` strictly precedes the `message` item in both SSE events and `response.completed.output`.
  - Calculate URL citation Unicode character (rune) offsets on full accumulated text at `finalizeMessage`, eliminating multi-byte UTF-8 (CJK/Emoji) truncation or clamping.
  - Guard against phantom `web_search_call` and false `tool_usage` when no grounding occurs.
  - Strip `resp_` prefix in streaming `web_search_call.id` (`ws_<hash>`) to match non-streaming.
- **Redirect URL Resolution**:
[...]

PR #5977, opened 2026-09-19
feat(models): declare native web search for additional Antigravity Gemini models

ReneXiong:
Follow-up to the direction in #5332: declare web-search capability directly in `models.json` rather than adding a config override.

Marks five Antigravity catalog models as web-search capable. Each was verified to accept native `googleSearch` and return real `groundingChunks` upstream, but is omitted from `fetchAvailableModels.webSearchModelIds`:

| Model | Why |
| --- | --- |
| `gemini-3.7-flash-high` | Accepts `googleSearch`, not advertised by probe |
| `gemini-3.6-flash-high` | Same |
| `gemini-3-flash` | Same |
| `gemini-pro-agent` | Same |
| `gemini-3.1-flash-lite` | Advertised today; flag statically so capability survives probe outages |

Both fields are set because the two protocol paths read different sources:

- `supports_web_search` → `registry.AntigravityWebSearchModelFor` (Claude-path gate in `antigravity/claude/web_search.go`)
- `native_capabilities.web_search` → `ResolveResponsesWebSearchCapability` (Responses-path gate added in #5862)

`gemini-3-flash-agent` also accepts `googleSearch` upstream but is absent from the catalog, so no static flag can reach it — a separate catalog entry would be needed.


ReneXiong:
@sususu98 per your direction in #5332 — capability is now declared statically in `models.json`. Both fields set since the Claude-path gate (`AntigravityWebSearchModelFor`) reads `supports_web_search` while the Responses-path gate (`ResolveResponsesWebSearchCapability`) reads `native_capabilities.web_search`. All checks green.

PR #5984, opened 2026-09-20
 fix(test): make codex bootstrap buffering tests deterministic and drain streams before teardown

### LINKED ISSUE #5981: Test suite does not pass on Windows (1 deadlock, 2 racy and 1 flaky failure; verified on clean upstream 61fdfc34)
## Summary

While merging `upstream/main` into a fork, `go test ./...` fails on Windows in three places and is
flaky in a fourth. As in #5295, I verified every item against a **pristine upstream tree** (no fork
changes) to make sure none of them are caused by the merge. Three sit in the Codex stream bootstrap
buffering tests; the fourth is in the responses stream forwarder. None looks like a production bug
on its face — three are races or a teardown hazard in the tests themselves, and the fourth is a
disagreement between the test and the forwarder about when a late terminal error should win.

Two of the three websockets race failures land in the failing interleaving about 70% of the time on
my machine, and the flaky one fails 20–40% of the time. The remaining one is a hard deadlock that
only surfaces at the `go test` timeout (10 minutes by default, which makes the whole
`internal/runtime/executor` package unusable on Windows until it panics).

I am not on a Linux host, so I could not check whether these reproduce there. Sections 2–4 are
scheduling races whose outcome depends on goroutine interleaving, not on anything Windows-specific
in principle; section 1 depends on socket buffer sizes, which do differ per OS.

Environment:

- `go version go1.26.5 windows/amd64`
- Windows 11 (26200.9457), `git config core.autocrlf=false`, no `.gitattributes` in the repo
- Verified on upstream `61fdfc34` (current `main`) with no local modifications

Reproduce:

```
go test ./internal/runtime/executor/ -run 'TestCodexExecutor_BootstrapBuffering_ByteCapReleasesStream' -count=1
go test ./internal/runtime/executor/ -run 'TestCodexWebsocketsExecutor_BootstrapBuffering_(OverloadDirectlyAfterTimeoutDeliveredInStream|StatusBearingErrorAfterTimeoutDeliveredInStream)' -count=5
go test ./sdk/api/handlers/openai/ -run 'TestForwardResponsesStreamErrorEventPreservesNestedError' -count=10
```

---

## 1. `TestCodexExecutor_BootstrapBuffering_ByteCapReleasesStream` deadlocks in the deferred `server.Close()`

**Failing test**: `internal/runtime/executor`: `TestCodexExecutor_BootstrapBuffering_ByteCapReleasesStream`
(`codex_stream_bootstrap_buffering_test.go:658`)

**Output** (deterministic; hangs until the test timeout, 60s here / 10m under `go test ./...`):

```
2026/09/20 10:21:53 httptest.Server blocked in Close after 5 seconds, waiting for connections:
  *net.TCPConn 0x50f5feaa050 127.0.0.1:14273 in state active
panic: test timed out after 1m0s
	running tests:
		TestCodexExecutor_BootstrapBuffering_ByteCapReleasesStream (1m0s)
```

The goroutine dump shows the test function parked in `httptest.(*Server).Close()`
(`net/http/httptest/server.go:279`) called from the deferred `server.Close()` at
`codex_stream_bootstrap_buffering_test.go:670`, plus a separate goroutine blocked for the full
minute in `net.(*conn).Write` under `net/http.(*response).finishRequest` — the handler still
trying to push body bytes to a client that has stopped reading.

**Cause**

The test body sends ~1.4 MiB of SSE (4 × `codexBootstrapMaxBufferedBytes`/3 ≈ 4 × the 1 MiB
`codexBootstrapMaxBufferedBytes` cap defined in `codex_executor_terminal.go:495`, plus the overload
event) and then returns after asserting only that `ExecuteStream` returned a non-error, non-nil
result. It **never drains `result.Chunks` and never closes the response body**.

The byte cap is what makes the executor release the stream early — at that point the client stops
reading, while the test server is still ~1 MiB away from the end of its body. The handler's
remaining `Write` blocks once the TCP send buffer fills, the connection never goes idle, and the
deferred `httptest.Server.Close()` waits for it forever. Both test assertions actually pass; the
process dies in teardown.

Whether the write wedges depends on how much of the remainder the kernel is willing to buffer,
which is why I expect this is worst on Windows — but the test is structurally "server writes more
than the client will ever read", so I would not call it Windows-only, just Windows-fatal.

**Suggested fix**

Drain (or close) the stream before returning, mirroring what the sibling tests do:

```go
for range result.Chunks {
}
```

Alternatively `codexSSERawServer` could be told to cap its body at the cap itself, but draining is
the honest fix — the executor has released the stream and the test holds an open body.

---

## 2. Websockets "overload / status error after timeout" tests race the mock clock against the request send

**Failing tests** (`internal/runtime/executor`, ~7 failures per 10 runs across the two):

- `TestCodexWebsocketsExecutor_BootstrapBuffering_OverloadDirectlyAfterTimeoutDeliveredInStream` (`codex_stream_bootstrap_buffering_test.go:1662`)
- `TestCodexWebsocketsExecutor_BootstrapBuffering_StatusBearingErrorAfterTimeoutDeliveredInStream` (`codex_stream_bootstrap_buffering_test.go:1701`)

**Output** (both identical in shape):

```
--- FAIL: TestCodexWebsocketsExecutor_BootstrapBuffering_OverloadDirectlyAfterTimeoutDeliveredInStream (0.00s)
    codex_stream_bootstrap_buffering_test.go:1662: expected stream result without failover when overload arrives after timeout: {"error":{"type":"service_unavailable_error","code":"server_is_overloaded","message":"Our servers are currently overloaded. Please try again later.","param":null},"sequence_number":2}
--- FAIL: TestCodexWebsocketsExecutor_BootstrapBuffering_StatusBearingErrorAfterTimeoutDeliveredInStream (0.00s)
    codex_stream_bootstrap_buffering_test.go:1701: expected stream result without failover when status-bearing error arrives after timeout: {"status":429,"error":{"message":"Rate limit exceeded","type":"requests","code":"rate_limit_exceeded"}}
```

**Cause**

Both tests advance a mock clock past the configured `stream-bootstrap-timeout` and expect the
in-flight rejection to be released to the client instead of triggering credential failover. The
websockets executor decides that from:

```go
// codex_websockets_stream.go:253
bootstrapStart = nowCodexBootstrap()   // captured AFTER the request has been written
...
timeSinceStart := nowCodexBootstrap().Sub(bootstrapStart)   // :314
timeoutReached := bootstrapTimeout > 0 && timeSinceStart >= bootstrapTimeout
```

But the request is written earlier (`writeCodexWebsocketMessage`, ~:210), and the test handlers
advance the clock as soon as that request arrives:

```go
if _, _, errRead := conn.ReadMessage(); errRead != nil { return }
clock.advance(11 * time.Second)          // server-side, un-synchronized
_ = conn.WriteMessage(websocket.TextMessage, []byte(codexOverloadEvent))
```

So there are two valid interleavings:

- client captures `bootstrapStart` first → at read time `timeSinceStart ≈ 11s ≥ 10s` → window
  closed → rejection delivered in-stream → **pass**
- server advances first → `bootstrapStart` postdates the advance → `timeSinceStart ≈ 0 < 10s` →
  window still open → rejection is treated as an in-hold terminal failure → **failover** →
  `ExecuteStream` returns an error → `t.Fatalf`

The failure output is the second interleaving: the raw rejection body surfaces as the
`ExecuteStream` error, i.e. exactly the failover the test asserts must not happen.

Note the SSE variants of the same two tests
(`TestCodexExecutor_BootstrapBuffering_{OverloadDirectlyAfterTimeout,StatusBearingErrorAfterTimeout}DeliveredInStream`)
are race-free and pass 5/5 on the same machine, because they synchronize through a
`bootstrapStarted` channel before advancing the clock. The websockets variants are the only ones
that advance the clock from an uncoordinated handler goroutine.

**Suggested fix**

Either capture `bootstrapStart` before the request is written (arguably also more correct in
production: the hold should start when the request is sent, not after session bookkeeping), or
make the two websockets handlers wait on the same kind of `bootstrapStarted` signal the SSE
variants already use. The second is the smaller diff and keeps production semantics untouched.

---

## 3. `TestForwardResponsesStreamErrorEventPreservesNestedError` loses the real error to the synthetic close error (~20–40% of runs)

**Failing test**: `sdk/api/handlers/openai`: `TestForwardResponsesStreamErrorEventPreservesNestedError`
(`openai_responses_handlers_stream_test.go:311`)

**Output**:

```
    openai_responses_handlers_stream_test.go:372: payload.Error.type = server_error, want invalid_request
--- FAIL: TestForwardResponsesStreamErrorEventPreservesNestedError (0.00s)
```

**Cause**

The test producer runs on its own goroutine and closes `data` **before** sending on `errs`:

```go
close(data)
errs <- &interfaces.ErrorMessage{StatusCode: http.StatusBadRequest, Error: errors.New(errText)}
close(errs)
```

The consumer, `ForwardStream` (`sdk/api/handlers/stream_forwarder.go:91-115`), reacts to `data`
closing by calling `PendingStreamError(errs)` — which is a **non-blocking** `select` with a
`default` branch (`stream_forwarder.go:12-24`). If the producer has not yet landed its `errs`
send, the lookup finds nothing and the handler falls through to `CloseError()`
(`openai_responses_handlers.go:1018-1035`), which synthesizes

```
upstream stream closed before a terminal event (last event: …)
```

and that synthetic 502 is what gets normalized into the generic `"server_error"` event the
assertion then rejects.

So the test asserts that a *late* terminal error wins over the close-time fallback, while the
forwarder's contract is that an error must be *pending* at close time to be surfaced. Both are
defensible; the test as written encodes a timing assumption that the producer does not guarantee.

**Suggested fix**

If late errors should win, the forwarder would need to drain `errs` (blocking) after `data`
closes instead of `PendingStreamError` — that changes shared production behavior, so it deserves
its own decision. If the current contract is intended, the test producer should order its sends
so the error is guaranteed pending: `errs <- err` first, then `close(data)`, then `close(errs)`.

---

## Context: why I believe these went unnoticed

#5295 documented the same situation in August and the four failures it listed were fixed in
`d36b776c`. Two things have regressed the area since:

- `pr-test-build.yml` is currently the only CI workflow that touches tests, and it only runs
  `go build -o test-output ./cmd/server` — no `go test` step, on `ubuntu-latest` only. None of the
  four items above would be caught by it even on Linux.
- AGENTS.md documents the Windows timer-granularity convention (`8bcae8cf`), but nothing enforces
  it mechanically.

A `go test ./...` step on `ubuntu-latest` plus (optionally) a `windows-latest` job would have
caught all four. For the two websockets races and the teardown deadlock the fix direction is clear
enough that I could prepare test-only patches and verify them on Windows if that helps.

The sections above each stand alone — 1 and 2 are test-side, 3 is a product-contract question
wearing a test hat, and none of them block my fork's use of `61fdfc34`, so no urgency attached.

Viggo95:
Test-only fixes for the failures in #5981 (possibly caused by me). **No production code changes. just two test files.**

### Problem

1. **Teardown deadlock** — `codex_stream_bootstrap_buffering_test.go:663` `...ByteCapReleasesStream` returns after asserting `err==nil`/`result!=nil` without draining ~1 MiB of unread SSE. The handler blocks in `net.(*conn).Write`, the connection never idles, and the deferred `server.Close()` hangs to the test timeout.
2/3. **Clock vs request-send race** — `:1637`/`:1674` websocket cases advance the mock clock as soon as the handler reads the request, but the executor captures `bootstrapStart` *after* writing the request (`codex_websockets_stream.go:253`). If the handler wins, elapsed is 0 and the rejection fails over instead of arriving in-stream.
4. **Flaky forwarder** — `openai_responses_handlers_stream_test.go:318` closes `data` before sending on `errs`; the non-blocking `PendingStreamError` lookup then misses and the handler emits a generic `server_error` the assertion rejects.

Found while verifying: **5.** the SSE hook signals before it samples the clock (same wrong outcome as 2/3); **6.** the websocket `...TimeBudgetReleasesStream` advances from an unsynchronised handler like 2/3 (not in the issue).

### Solution

1. Drain the stream before returning — the ending the sibling `BoundHoldsForEverySSEFraming` cases already use.
2/3/6. Wait on the `bootstrapStarted` signal before advancing the clock; the executor takes its start from the same `nowCodexBootstrap` hook, so the signal fires at that capture.
4. Reorder to `errs<-err; close(data)`, the order the real producer already guarantees (`handlers_stream.go:131-166` sends before its deferred closes). No production change.
5. Sample the clock, then signal (`now := m.now(); release(); return now`).

Each waiting handler gets `defer func(){ release(); server.Close() }()`, so an executor that fails before the hook fires cannot park it (item 1's mode, otherwise reintroduced by 2/3/6). Hook and teardown come from one helper, `withBootstrapStartedClock`, shared by all five cases.

### Verification

`windows-latest` (GitHub Actions, `go1.26.0`, `-race`), base and patched as one matrix:

| Step | base `cdfb79ef` | patched `ec81c0c4` |
| --- | --- | --- |
| whole package `-count=1 -timeout 8m` (reporter's repro) | **deadlock**, killed at 8m | pass, 81s |
| item 1 alone `-count=20` | **deadlock**, killed at 4m | 20/20, 3s |
| 5 clock cases, `-count=80` ×3 | 0/3 reps green | 3/3 |
| 5 clock cases, 4 conc × `-count=50` ×3 | 4/12 procs green | 12/12 |
| item 4, 4 conc × `-count=200` ×3 | 11/12 procs green | 12/12 |

All six reproduce on the platform the issue targets; all six are green on the patched commit. (Also green under concurrent-process load on macOS/`go1.26.3`; `go test ./...` 97 pkgs, `go vet`, `gofmt -l` all clean. Harness is a throwaway fork branch, not in this PR.)

Out of scope: the issue's suggested `go test` / `windows-latest` CI step — a follow-up.

Closes #5981


Viggo95:
https://github.com/Viggo95/CLIProxyAPI/actions/runs/35495280345#artifacts
image

Viggo95:
https://github.com/Viggo95/CLIProxyAPI/actions/runs/35496205447
image

Viggo95:
@codex review

Viggo95:
Superseded by 563865e7

Issue #5545, opened 2026-09-05
Codex Responses SSE stream dies after ~30s of upstream silence on deep-reasoning models (terra/astra: 20% / 15% drop rate)


so2liu:
### Summary

On `v7.2.125`, Codex Responses streams for deep-reasoning models die without a terminal event whenever the upstream stays silent for ~30s. Clients surface it as `stream disconnected before completion: stream closed before response.completed`.

The trigger is **upstream silence, not total request duration**. It is heavily model-dependent.

### Production data (24h, ~40k requests)

Classified from usage events as `failed=1 && latency_ms >= 15000`. Note that `PublishFailure` in `codex_executor_stream.go` carries no usage detail, so `output_tokens` is always 0 for these — it cannot be used to tell "failed before first byte" from "died mid-stream". Latency is the usable signal.

| model | requests | mid-stream drops | rate |
|---|---|---|---|
| gpt-5.6-terra | 1026 | 209 | **20.4%** |
| gpt-6-astra | 150 | 22 | **14.7%** |
| gpt-5.6-sol | 38153 | 27 | 0.07% |
| gpt-5.6-luna | 239 | 1 | 0.42% |
| grok-4.6 | 51 | 1 | 1.96% |

Drop latencies cluster hard on **30s (131), 60s (52), 90s (28)** — i.e. multiples of the upstream keepalive period — with almost nothing in between. Meanwhile 57 `terra` requests *succeeded* past 30s (longest 172s), so this is not a hard deadline on the request.

### Why terra/astra and not sol

Same prompt, same effort, downstream SSE:

```
sol  : 565s, 2942 events, response.completed  ✅
       continuous response.reasoning_summary_* ; max inter-event gap 9.9s

terra:  31.6s, 3 events, no completed         ❌
       response.created → response.in_progress → (30.0s silence) → keepalive → dead
```

`terra` emits **no intermediate events at all** while reasoning, so the stream idles past the window. `sol` keeps emitting reasoning summaries and never idles.

Tried and did **not** help on terra (9/9 still dropped): `reasoning.summary` = `auto` / `detailed` / `concise` (terra produces no reasoning summary regardless), `include: ["reasoning.encrypted_content"]`, copying the full official Codex CLI parameter set, and lowering effort to `medium` / `low`.

### Not a proxy-side timeout

[...]

PR #5989, opened 2026-09-20
feat(codex): include last event and idle interval in incomplete stream errors

Viggo95:
### Problem
Over 7 days, In one production Codex proxy fleet, , **311 / 404,957** upstream attempts ended with an incomplete stream.. CPA also output an error:
```
stream error: stream disconnected before completion: stream closed before response.completed
```

Break them down by request stream flag vs. recorded status and you get four buckets: _134 stream/200, 109 unset/408, 57 non-stream/408, and 11 stream/408_. **The error message treats them as one**. 

The closest thing we have to a silent interval in telemetry is latency_ms - ttft_ms, and it goes negative (min -35,779 ms) because **the two columns aren't on the same clock. So telemetry alone can't tell us why any of them ended. The last event type and silent interval only exist inside the executor when it builds the error, and that's where they get dropped.**

image
image

image

**None of these columns can say why a stream ended.** 

### Solution

The constructor takes a `codexIncompleteStreamDiagnostics` value and appends it to the existing constant:

```
... stream closed before response.completed (last event: response.output_text.delta, data frames: 12, silent for 31s)
```

- **`last event`** :   the type of the last upstream event that had a type; `none` if no event had one, the type was blank, or nothing survived sanitising. gjson renders **a non-string type** as raw JSON and a JSON string escape can carry a newline, so unprintable runes are dropped, whitespace runs collapsed, and the value cut at 128 runes with a `... `marker before it is rendered — one failure stays one log line.
- **`data frames`**: every upstream `data:` frame, including keepalives; the `event:` / `id: `/ `blank lines` that frame an SSE event are not counted. This is counted before the grok keepalive rewrite, so the value does not depend on which downstream client is reading..
- `**silent for**` : the interval since the executor last read a line. This includes any time the reading goroutine was blocked on downstream backpressure, so it is the silence seen by the reader, not silence on the wire. For values below one second (or negative), it renders as `**<1s,**` not an ambiguous `**0s**`. Rounding to whole seconds limits only this dimension: **`data frames`** is unbounded, so the text goes from a single distinct value to roughly one per failure. Aggregate on the unchanged prefix..

| Type | Upstream failure | Appended diagnostics | What the operator sees |
| ---  | --- | --- | --- |
| 1 | Silent reasoning phase cut at ~30 s | `(last event: response.content_part.added, data frames: 2, silent for 31s)` | **Only the handshake arrived, then 31 s of nothing — the upstream cutoff #5545 describes** |
| 2 | Connection dropped mid-stream | `(last event: response.output_text.delta, data frames: 137, silent for <1s)` | **Deltas were flowing and stopped at once — a severed connection, not an idle timeout** |
| 3 | Upstream never started | `(last event: none, data frames: 0, silent for 12s)` | **Nothing after the headers — the upstream never began** |

The buffering loop and the goroutine share the counters, so a stream released after buffering keeps counting. **The non-stream path reads the body to completion first, has no arrival timestamps, and omits the interval rather than substituting another measurement.**

The clock is implemented as a second package-level hook (`codexStreamNowFunc / setCodexStreamNowForTest`) next to `nowCodexBootstrap`, rather than reusing it. **There are two reasons**: `nowCodexBootstrap()` takes its RLock on every call, so reading it per line would put a lock in the read loop; and a single shared hook would also let the bootstrap tests that install a mock clock move this interval. **The new hook returns the function, resolved once per request, so the per-line read takes no lock at all.**

<br>

### Verification

**Compatibility.** `codexIncompleteStreamMessage` is byte-identical and stays a prefix. The one in-tree consumer, `shouldReleaseResponsesWebsocketPinnedAuth`, matches a substring of it and gains a table case with the enriched message on a status it does not release on by itself, so `true` can only come from the match. The rendering table asserts the prefix on every row. No test asserted equality on the old full message.

**Tests.** New file `internal/runtime/executor/codex_incomplete_stream_diagnostics_test.go`, plus one table case in `sdk/api/handlers/openai/openai_responses_websocket_test.go`

| Command | Result |
| --- | --- |
| `gofmt -l internal sdk` | only the three files already listed on `dev`; nothing here is added |
| `go vet ./internal/... ./sdk/...` | clean |
| `go test -race -count=10 -run IncompleteStream ./internal/runtime/executor/` | ok |
| `go test ./internal/runtime/executor/... ./sdk/api/handlers/...` | ok, 6 of 6 packages |
| `CGO_ENABLED=0 go test ./...` | ok, 97 packages, 29 without test files, 0 failures |

<br>

Refs: #5545


Viggo95:
To avoid the windows issue in  #5984.  Windows CI job:

1. Run1: https://github.com/Viggo95/CLIProxyAPI/actions/runs/35513127288
2. Run2: https://github.com/Viggo95/CLIProxyAPI/actions/runs/35512474069

All Passed

Viggo95:
@sususu98   Sorry to ping 😬😬

This one follows up on the bootstrap time ceiling you added after #5724.  **That ceiling decides when to commit the headers and start streaming;** 
**this PR** is about what happens after that — when a stream that already started dies, **the error now says what the last event was and how long it went silent**. Same line of work, next step. 

Could you take a look when you have time?  Happy to change anything.

sususu98:
Thanks for working on this! This directly addresses a real operational blind spot we see in production.

### Real Production Evidence

Here is a typical production attempt from today where an upstream Codex stream ended prematurely:

**Telemetry record (`usage_events`):**
- `model`: `gpt-5.5`
- `ttft_ms`: `2,051 ms` (first chunk arrived in ~2s)
- `latency_ms`: `299,878 ms` (~5 minutes)
- `failed`: `1`

**Recorded error log:**
```http
=== API REQUEST ===
Upstream URL: https://chatgpt.com/backend-api/codex/responses
HTTP Method: POST
Headers:
  Accept: text/event-stream
  User-Agent: codex-tui/...
Body:
  {"model": "gpt-5.5", "stream": true, ...}

=== API RESPONSE ===
Status: 408
{"type":"error","error":{"type":"invalid_request_error","message":"stream error: stream disconnected before completion: stream closed before response.completed"}}
```

**The problem today:** With only the bare `stream closed before response.completed`, operators cannot tell what happened during those remaining ~298 seconds after TTFT. Did deltas stream continuously until the 300s reverse-proxy read timeout kicked in? Or did the upstream stall immediately after the handshake? Your PR gives exactly the missing context needed to answer that.

---

### Review Feedback & Suggestions

The architecture is very solid — single-ownership state handover between buffering and the streaming goroutine is provably data-race free under the Go memory model, the request-level clock snapshot avoids per-line lock contention, and downstream substring contracts (`strings.Contains`, `IsRequestScoped`) are properly preserved.

A few small suggestions before merging:

1. **Defensive sanitization on `lastEventType`**:
   `gjson.GetBytes(data, "type").String()` could return raw JSON objects (e.g. `{"type": {"status": "failed"}}`) or multiline/unbounded strings if an upstream or proxy misbehaves. Since this value is formatted directly into `%s` and stored in client errors and error log files, it's safer to:
   - Cap length (e.g. 64–128 characters).
   - Strip control/newline characters (`\r`, `\n`) to prevent breaking line-oriented log aggregators.
   - (Note: some upstream error frames use root `{"error": {"type": "..."}}` without a top-level `"type"`, which currently yields `"none"`. That's acceptable, but worth being aware of).

2. **Handover unit test edge case**:
   In `TestCodexExecutorExecuteStreamIncompleteStreamCountsAcrossBootstrapHandover`, the test sends `created` followed by `delta`. Because `delta` immediately triggers the stream release, both frames are actually consumed inside the `buffering` loop; the subsequent streaming goroutine only reads EOF.
   It would be great to add a 3rd event (e.g., `created` → `delta` (releases stream) → `delta2` (consumed by background goroutine) → close) to verify that the streaming goroutine actively continues incrementing `dataFrames` and updating `lastEventType`.

3. **Clarify `silent for` observer semantics**:
   Worth noting in comments/docs that `silent for` measures the executor's *observation interval* between lines. If a slow downstream client creates channel backpressure that blocks the executor loop, that delay is included in `silent for`, so it serves as an execution snapshot summary rather than absolute wire-level proof of upstream silence.

Viggo95:
Thanks for your time, the careful review, and for the trace — that one's a great example👍👍👍. 

With  this PR,  **your  gpt-5.5 Telemetry record** would read either 
```
(`last event: response.output_text.delta, data frames: 1847, silent for <1s`) 
→  deltas kept flowing until something cut it at ~300s.
```
or 
```
(`last event: response.created, data frames: 2, silent for 4m58s`)  
→  upstream went quiet right after the handshake  → #5545
```

Pushed one commit for the three points, And I went through it point by point and made careful refinements using Fable (Thinking xhigh):

1. Sanitisation ✅️ — drops anything unprintable (went with `!unicode.IsPrint` instead of `IsControl`, so bidi overrides and U+2028 go too), collapses whitespace runs, and cuts at 128 runes with a `...` marker, same idea as the diagnostic excerpts in `internal/logging`. It runs when the message is built, once per failed attempt, so nothing new in the read loop. I checked all 121 event types in the tree — none of them change. Added table cases for a raw JSON object, a JSON-escaped newline, bidi chars, whitespace runs, and over-long ASCII/multi-byte values, plus a check on every row that the message never contains `\r` or `\n`. The root-level `{"error":{"type":...}} `→ none case I left as is, like you said.

2. Handover test✅️ — you were right, it wasn't proving anything: both frames got eaten by the buffering loop and the goroutine only saw EOF. Now there's a third frame with a different type, so both the count and the type can only come from the goroutine. I tried making the goroutine stop feeding the diagnostics: the test now fails with `data frames: 2 … output_text.delta` vs expected `3 … output_item.added`, and the old two-frame version passed under that same change.

3. Observer semantics✅️ — now in the code comments, on the idle field, not just in the PR description.

`go vet` clean, `gofmt` adds nothing, `-race -count=10` on the new tests ok, `CGO_ENABLED=0 go test ./... `green.

sususu98:
Thanks for the quick and thorough updates! The sanitization and the updated handover test look great.

cc @hkfire for visibility/review when you have time.

Viggo95:
@hkfires Looks like the cc above landed on **`@hkfire` (no "s") by mistake** — **sususu98's review is done and it still merges clean on `dev`**, mind having a look when you have a moment?

PR #6003, opened 2026-09-21
refactor(translator): encapsulate tool-result cache-control hoisting in common

sususu98:
## Description

Follow-up hardening and refactoring on top of #5432 (commit `bcd13ca9`):

1. **Encapsulate Cache Control Hoisting in `common`**:
   - Replaced the ad-hoc local helper `firstPartCacheControl` with `common.AttachToolMessageCacheControl` in `internal/translator/common/cache_control.go`.
   - Eliminated the implicit two-stage order dependency in `claude_openai_request.go` by atomically hoisting part-level `cache_control` when present, falling back to message-level `cache_control`.
   - Explicitly targets the `type == "tool_result"` block instead of assuming the last block in the content array.

2. **Zero-Allocation Content Pipeline for Tool Results**:
   - Introduced `convertOpenAIContentPartToClaudePartRaw` to construct raw content blocks (text/image/file) without attaching cache control.
   - Tool result conversion calls `convertOpenAIContentPartToClaudePartRaw` directly, preventing `cache_control` from ever being generated inside `tool_result.content[*]`.
   - Completely removes the performance penalty of generating a marker and then calling `sjson.DeleteBytes` on every tool part.
   - Restored `convertOpenAIContentPartToClaudePart` to its clean signature without boolean flag parameters.

3. **Protocol-Grade Cache Control Validation (`isValidCacheControl`)**:
   - Enforced specification compliance: only non-empty objects with `type == "ephemeral"` are treated as valid cache control.
   - Prevents empty (`{}`) or malformed `cache_control` objects on parts from stealing or blocking valid message-level `cache_control` fallback.

## Testing

- `go test -v -race ./internal/translator/common/...` (Pass)
- `go test -v -race ./internal/translator/claude/openai/chat-completions/...` (Pass)
- `go test ./internal/translator/...` (All 30 packages pass)
- Fast build `go build -ldflags="-s -w" -o /dev/null ./cmd/server` (Clean)
- Added unit tests:
  - `TestStripCacheControl`
  - `TestAttachToolMessageCacheControl_HoistsPartLevel`
  - `TestAttachToolMessageCacheControl_InvalidPartDoesNotBlockMessageLevel`
- Live probe verified with local CPA instance against real Claude Sonnet 5 upstream.


sususu98:
Directly pushed to dev branch via commit 50585bf2.

Issue #6013, opened 2026-09-21
[Feature] Add per-execution proxy overrides and internal response headers for host model callbacks


Lingbou:
## Problem

Plugins that call `host.model.execute` or `host.model.execute_stream` cannot choose an outbound proxy for an individual execution. They must use the credential proxy or the global proxy, which prevents plugins from rotating collection egress without changing client-facing proxy configuration.

Plugin-host internal executions also lose filtered upstream response headers when `passthrough-headers` is disabled. Plugins therefore cannot use turn-state, quota, rate-limit, or safety headers to make follow-up decisions.

## Proposal

- Add `proxy_url` to `HostModelExecutionRequest`.
- Validate only `http`, `https`, `socks5`, and `socks5h`.
- Resolve the effective proxy as request > auth > global.
- Keep the override scoped to the model request; credential refresh and token exchange continue to use the account or global proxy.
- Return filtered upstream response headers to plugin-host internal executions even when client-facing `passthrough-headers` is disabled.
- Preserve existing client-facing behavior.

## Compatibility

- The JSON field is additive and optional.
- Invalid proxy URLs are rejected with HTTP 400.
- Existing clients and plugins are unaffected when `proxy_url` is omitted.

PR #6014, opened 2026-09-21
feat(pluginhost): add per-execution proxy overrides and internal response headers

Lingbou:
## Summary

Add two plugin-host model execution capabilities:

- `host.model.execute` and `host.model.execute_stream` accept an optional per-execution `proxy_url`.
- Plugin-host internal executions receive filtered upstream response headers even when client-facing `passthrough-headers` is disabled.

## Proxy override

- The effective proxy priority is request > auth > global.
- Only `http`, `https`, `socks5`, and `socks5h` are accepted.
- The override applies to HTTP, uTLS, Codex/xAI WebSocket, Devin, and Antigravity execution paths.
- Credential refresh and token exchange intentionally continue using the account or global proxy settings.
- Malformed or unsupported proxy URLs return HTTP 400.

## Response headers

- Plugin-host internal executions receive filtered upstream response headers regardless of the client-facing `passthrough-headers` setting.
- Stream terminal error headers use the same filtering path.
- Ordinary API clients retain the existing behavior.

## Non-goals

- This PR does not change the non-streaming error envelope to carry upstream headers. That overlaps with the host model error contract work in #5550.

## Validation

- `go test ./...`
- `go build` in `examples/plugin/host-model-callback/go`
- `gofmt` on all changed Go files
- `git diff --check`

Fixes #6013


Lingbou:
Superseded by upstream commit d582067c066f0267bb9d605daf6355e9eb75679a, which resolves #6013. Closing this PR to avoid duplicate review.

PR #6032, opened 2026-09-22
fix(pluginhost): keep CPA-owned auth keys across plugin credential saves

shadyrispy:
fix(pluginhost): keep CPA-owned auth keys across plugin credential saves

## Problem

A plugin that refreshes a token persists its own credential document through
`host.auth.save`. `saveAuthFile` writes whatever bytes the plugin passed:

```go
auth, errBuild := h.buildAuthFromFileData(dst, data)
...
if errWrite := os.WriteFile(dst, data, 0o600); errWrite != nil {
```

but the routing and bookkeeping keys for an auth file belong to CPA, not to the
plugin — `internal/watcher/synthesizer/file.go` reads `priority`, `weight`, `note`,
`disabled`, `prefix`, `proxy_url`, `model_aliases`, `excluded_models`,
`disable_cooling`, `request_retry`, `request_scoped_errors` and
`fingerprint_profile` back out of that same file.

So the operator's settings are erased by the next plugin refresh, in both copies:

* the file loses the keys (the plugin never had them);
* the live record is rebuilt by `buildAuthFromFileData`, which carries over
  `CreatedAt` / `LastRefreshedAt` / `NextRetryAfter` / `Runtime` from the existing
  record but not `Attributes["priority"]` or `Attributes["note"]`, and hard-codes
  `Status: StatusActive`.

Observed with a CodeArts credential (plugin-owned storage lives under
`codearts_provider_credential`, everything else belongs to the host):

```
$ python3 -c 'import json;print(json.load(open("…/auths/codearts-provider-GT-…json"))["priority"])'
30                       # right after writing priority into the file
… 15:17 → 15:18, the plugin's hourly token_renew runs …
KeyError: 'priority'     # gone from disk, and gone from the selection order
```

This makes a primary/backup chain across providers impossible to express for
plugin credentials: `priority` is the only per-credential ordering knob CPA has
(`FillFirstSelector` takes `available[0]`, tiers come from
`Attributes["priority"]`), and it silently resets one hour later.

## Change

* `preserveHostOwnedAuthKeys` — before writing, copy the CPA-owned keys that are on
  disk but absent from the plugin's document. A key the plugin does write wins, so a
  plugin can still change or clear its own fields, and a first write is byte-for-byte
  the plugin's JSON (the existing test asserts that and still passes).
* `buildAuthFromFileData` now mirrors the synthesizer for the keys that affect
  routing: `priority` and `note` go into `Attributes`, and `disabled: true` keeps the
  record `StatusDisabled` instead of being re-activated by a refresh.
* Legacy spellings (`proxy-url`, `model-aliases`, …) are compared in canonical form
  via the existing `coreauth.NormalizeCredentialMetadata`, so preserving them also
  normalises them onto the canonical key.

No new network paths, timeouts or dependencies; the change is confined to
`internal/pluginhost/auth_callbacks.go`.

## Tests

`internal/pluginhost/auth_callbacks_test.go`, five new cases plus the existing two:

1. a plugin refresh that only carries its own fields keeps `priority`/`note`/`weight`
   on disk **and** in the live record, while still storing the rotated `api_key`;
2. keys the plugin writes explicitly override the stored values (no sticky ghost);
3. a disabled account stays disabled through a plugin save;
4. a legacy `proxy-url` spelling survives (canonically);
5. a first write (no file on disk yet) is not decorated with invented keys.

`gofmt -l` clean, `go build ./...` and `go test ./internal/pluginhost/` pass.


shadyrispy:
Heads-up on overlap: while this PR was open, v7.3.16 shipped `c404af96 fix(auth): preserve file priority across plugin auth refreshes`. I re-tested on top of it. The two changes cover different code paths, and the failure this PR fixes is still reproducible with `c404af96` alone.

**What `c404af96` covers**

It runs on the plugin `RefreshAuth` *response* (`internal/pluginhost/auth_provider.go:393` and `internal/pluginhost/adapters_executors.go:932`), carrying `path` / `source` / `source_backend` / `file_priority` / `priority`, gated on the in-memory credential being marked `source_backend == file` **and** `file_priority == true`. That is the right idea, and it is why `Prefix`/`priority` do survive a plain provider refresh (`auth_provider.go:381` re-applies `auth.Prefix`).

**What it does not reach**

The two places that write the physical auth file from plugin-supplied JSON, both of which serialise without ever consulting the file already on disk:

- `saveAuthFile` — `internal/pluginhost/auth_callbacks.go:284`, which writes the plugin payload verbatim at line 299. Worth noting: it then rebuilds the live record via `buildAuthFromFileData` (line 308), whose `Attributes` are only `path` + `source` (lines 342-345). So after the first save through this door the in-memory copy no longer carries `source_backend=file` / `file_priority=true`, and `preserveFileAuthPriority` returns at its very first check for that credential.
- `pluginTokenStorage.SaveTokenToFile` — `internal/pluginhost/auth_provider.go:486`, whose payload is `mergedStorageJSON(rawJSON, meta, provider)`, i.e. the plugin's own storage JSON merged with plugin metadata. CPA's top-level keys (`prefix`, `priority`, `note`, `weight`, …) are simply not an input, so they cannot come out.

**Measurement**

Same machine, same plugin builds, only the binary swapped. Procedure: arm `prefix` + `priority=77` through `PATCH /v0/management/auth-files/fields`, then trigger the plugin-side refreshes that actually fire in production — `POST /v0/management/plugins/<id>/refresh` (and the codearts `token-renew` schedule job) — plus one chat completion, then re-read the file from disk.

| credential | clean v7.3.16 (has `c404af96`) | v7.3.16 + this branch |
|---|---|---|
| workbuddy | `prefix` lost, `priority` lost | both kept |
| qoder | `prefix` lost, `priority` lost | both kept |
| codearts-provider | `prefix` lost, `priority` lost | both kept |
| trae | both kept (no write landed in the window) | both kept |

Identical outcome on the first try and again after upgrading the plugin set to `cpa-multi-plugins v0.12.87`. The trae row is the reason this reads as intermittent in the field: whether a key disappears depends on whether that credential happened to be rewritten during the refresh window.

**Why it matters**

With `force-model-prefix: true`, `prefix` is what disambiguates one model name served by several providers/credentials; when it silently disappears the credential's whole catalogue is republished under bare names, so client-side model selections break. Losing `priority` reorders the routing pool the operator configured. The only trace either leaves is a debug-level client-reload diff, `clients.go:245` printing `prefix: qd -> `, which is easy to miss and gives no hint that a refresh was responsible.

**This branch** (3 files, +211: `auth_callbacks.go` +93, tests +113, `auth_provider.go` +5) adds one helper, `preserveHostOwnedAuthKeys(path, data)`, and calls it at both doors. It re-reads the current on-disk document and copies over CPA-owned keys from a fixed allow-list that the plugin's payload does not mention — anything the plugin *does* write still wins, so a plugin can change or clear its own fields, and `disabled`/`note`/`prefix` stay reachable through the management API as before. Tests cover: operator keys survive a plugin refresh, a plugin may still override them, `disabled` stays disabled, legacy key spellings are normalised rather than duplicated, and a first write (no file yet) keeps the plugin bytes untouched.

**Open to narrowing** — if you would rather this reuse the existing `file_priority` machinery instead of an allow-list, or land only the `SaveTokenToFile` door, or fix `buildAuthFromFileData` to stamp `source_backend`/`file_priority` so `c404af96` covers the refresh path end to end, tell me which shape you want and I will re-roll. Currently rebased onto `dev` (`75b854eb`); `gofmt` clean and `go test ./internal/pluginhost/` green.

